PR #522, opened 2015-01-08
H5o

chinasaur:


chinasaur:
I don't have a good build environment, so all these changes are quick and dirty; I patched the generated .c in my env and confirmed that it fixed compiler warnings, then made equivalent changes in the .pyx without testing.  Hope it works out!

PR #523, opened 2015-01-08
Incompatible pointer fixes

chinasaur:
I don't have a good build environment, so all these changes are quick and dirty; I patched the generated .c in my env and confirmed that it fixed compiler warnings, then made equivalent changes in the .pyx without testing. Hope it works out!

### DIFF
--- a/api_gen.py
+++ b/api_gen.py
@@ -64,11 +64,11 @@ class Line(object):
                             """, re.VERBOSE)
 
     SIG_PATTERN = re.compile("""
-                            (unsigned[ ]+)?
-                            (?:[a-zA-Z_]+[a-zA-Z0-9_]*\**)
-                            [ ]+[ *]*
-                            (?P<param>[a-zA-Z_]+[a-zA-Z0-9_]*)
-                            """, re.VERBOSE)
+                             (?:unsigned[ ]+)?
+                             (?:[a-zA-Z_]+[a-zA-Z0-9_]*\**)
+                             [ ]+[ *]*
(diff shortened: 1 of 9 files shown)


chinasaur:
Yeah, that wasn't a good idea...  I reverted the api_functions.txt changes for now.

chinasaur:
Okay, I think this is ready to pull.  Apologies again for the poor dev strategy.  Next time I have patches I'll be sure to have a reasonable build environment first.

api_gen.py
```diff
@@ -91,10 +91,11 @@ def __init__(self, text):
         self.fname = parts['fname']
         self.sig = parts['sig']
 
-        self.args = self.SIG_PATTERN.findall(self.sig)
+        sig_const_stripped = self.sig.replace('const', '')
+        self.args = self.SIG_PATTERN.findall(sig_const_stripped)
         if self.args is None:
```
chinasaur: I'm not sure None is an adequate check, but out of scope for this change.

tacaswell: `self.args` will only be `None` if the regex fails to match, if it is non-None, then it found one or more matches.

andrewcollette:
Tests pass on Travis and Windows.  Good catch with the rdcc variable too... we'd been using the wrong type for that.

Before I merge, please create a new PR and use rebase to squash things down a bit... right now we have 22 log entries for a relatively minor set of changes.

tacaswell:
For future reference, you don't have to create a new PR, you can just squash and then force-push to the current branch (before you deleted it).

andrewcollette:
@tacaswell... thanks!  Will use this in future.

PR #524, opened 2015-01-09
Const char defs

chinasaur:
Sorry about this, but I'll take one more stab in the dark using travis as my build env...  Presumably will still barf on const * const *, but hopefullly gets closer.


chinasaur:
Okay, that looks about as expected.  Let me know if you have a suggestion for handling const * const *.  Adding a special (* const) match is pretty ugly...

PR #526, opened 2015-01-10
Fixing incompatible pointer types.

chinasaur:
This is mostly const correctness, but there are a few more subtle ones
(int*->size_t*, hsize_t**->hsize_t*).

The longer story is that GCC4.9 now warns for these, but doesn't have a -Wno,
so these mostly cosmetic bugs were preventing h5py from building in GCC with
-Werror.


andrewcollette:
Thanks!

PR #527, opened 2015-01-12
Subclass from abcs

aragilar:
This changes the `DictCompat` class to be two classes `MappingWithLock` and `MutableMappingWithLock`, which inherit from `abc.Mapping` and `abc.MutableMapping`. This means that some functions which `DictCompat`, and hence Group etc., did not have (e.g. update), are now added.

### DIFF
--- a/h5py/_hl/attrs.py
+++ b/h5py/_hl/attrs.py
@@ -19,7 +19,7 @@ from .dataset import readtime_dtype
 from .datatype import Datatype
 
 
-class AttributeManager(base.DictCompat, base.CommonStateObject):
+class AttributeManager(base.MutableMappingWithLock, base.CommonStateObject):
 
     """
         Allows dictionary-style access to an HDF5 object's attributes.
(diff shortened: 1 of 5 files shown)


tacaswell:
:+1: from me.

andrewcollette:
It's an interesting idea.  I would be concerned about advertising too much compatibility with mappings... for example, the values containership test here will load every single item in the group, one by one, to check for equality.

The inheritance graph is also very complex; I count 6 new classes, each using multiple inheritance.  For example, looking at base.py I can't figure out how len() is actually computed for groups or attributes... where does this happen?

aragilar:
The six new classes replace 4 classes. Each of the new classes corresponds to a ABC in `collections`, with locks added where necessary (i.e. if locks weren't added, they'd be exactly the same). Rather than using `collections.<ABC>.register` for groups and attributes, they're explicitly subclassed.

In the case of `len()`, each class which subclasses `(Mutable)MappingWithLock` needs to provide __len__() (as documented in `collections`, and which they currently do).

What's gained from these changes is that more things that were implicit are now explicit (subclass of ABC rather than register, classes are explicitly mutable or immutable rather than one superclass representing both), and that expected dict methods (e.g. update, pop) are added.

andrewcollette:
Sounds good!  Let's merge it and see what feedback we get.

PR #533, opened 2015-01-15
More robust configuration system

behrmann:
Hey,

recently I tried to install h5py on a Debian system (a current testing) via pip and failed, since the hardwired paths and names for HDF5 just don't work for Debian, due to ist multiarch approach. 
Also the recent changes using pkgconfig that could in principle solve this problem don't work, because Debian packages HDF5 both in a serial and a mpi-enabled version (using openmpi) and pkgconfig.parse only show the version that was installed first, thus allowing only to build a serial or an mpi-enabled version but not both.
I'encountered similar problems on a current Fedora rawhide.

I've thus broken apart all configuration steps and put them in this changeset. More information can be found in the commit message of the first commit, that is basically a squashed version of this branch https://github.com/behrmann/h5py/tree/debian_install 

I have tested this changeset on a current Debian testing (with python and python3, both serial and mpi-enabled), a Fedora rawhide (both python and python3, only serial due to mpi4py apparently being broken in Fedora) and Arch (both python and python2, both serial and mpi-enabled). All build and pass the test.

I'm sorry for my commits being so noisy, since I strip whitespace...

### DIFF
--- a/docs/build.rst
+++ b/docs/build.rst
@@ -49,7 +49,7 @@ Install with pip
 Simply run::
 
     $ pip install h5py
-    
+
 All dependencies are installed automatically.
 
 Via setup.py
@@ -88,12 +88,21 @@ Custom installation
 You can specify build options for h5py with the ``configure`` option to
 setup.py.  Options may be given together or separately::
 
-    $ python setup.py configure --hdf5=/path/to/hdf5
+    $ python setup.py configure --hdf5=/path/to/hdf5/{lib/include}
+    $ python setup.py configure --hdf5-libdir=/path/to/libhdf5/lib
+    $ python setup.py configure --hdf5-includedir=/path/to/hdf5/include
+    $ python setup.py configure --hdf5-libname=customname-libhdf5
     $ python setup.py configure --hdf5-version=X.Y.Z
     $ python setup.py configure --mpi
-    
-Note the ``--hdf5-version`` option is generally not needed, as h5py 
+
+Note the ``--hdf5-version`` option is generally not needed, as h5py
 auto-detects the installed version of HDF5 (even for custom locations).
+
+The custom library name may be needed when multiple versions of hdf5 are
+installed under different names.
+
+Furthermore note, that ``--hdf5-libdir`` and ``--hdf5-includedir`` take
+precedence over ``--hdf5``
 
 Once set, build options apply to all future builds in the source directory.
 You can reset to the defaults with the ``--reset`` option::
@@ -104,17 +113,24 @@ You can also configure h5py using environment variables.  This is handy
 when installing via ``pip``, as you don't have direct access to setup.py::
 
     $ HDF5_DIR=/path/to/hdf5 pip install h5py
+    $ HDF5_LIB=/path/to/hdf5/lib pip install h5py
+    $ HDF5_INCLUDE=/path/to/hdf5/include pip install h5py
+    $ HDF5_LIBNAME=customname-libhdf5 pip install h5py
     $ HDF5_VERSION=X.Y.Z pip install h5py
-    
+    $ HDF5_MPI=1 pip install h5py
+
 Here's a list of all the configure options currently supported:
 
-======================= =========================== ===========================
-Option                  Via setup.py                Via environment variable
-======================= =========================== ===========================
-Custom path to HDF5     ``--hdf5=/path/to/hdf5``    ``HDF5_DIR=/path/to/hdf5``
-Force HDF5 version      ``--hdf5-version=X.Y.Z``    ``HDF5_VERSION=X.Y.Z``
-Enable MPI mode         ``--mpi``                   (none)
-======================= =========================== ===========================
+===============================  ===========================================  ===================================
+Option                           Via setup.py                                 Via environment variable
+===============================  ===========================================  ===================================
+Custom path to HDF5              ``--hdf5=/path/to/hdf5``                     ``HDF5_DIR=/path/to/hdf5``
+Custom path to HDF5 lib dir      ``--hdf5-libdir=/path/to/hdf5/lib``          ``HDF5_DIR=/path/to/hdf5``
+Custom path to HDF5 include dir  ``--hdf5-includedir=/path/to/hdf5/include``  ``HDF5_DIR=/path/to/hdf5``
+Custom name for HDF5             ``--hdf5-libname=customname-libhdf5``        ``HDF5_LIBNAME=customname-libhdf5``
+Force HDF5 version               ``--hdf5-version=X.Y.Z``                     ``HDF5_VERSION=X.Y.Z``
+Enable MPI mode                  ``--mpi``                                    ``HDF5_MPI=1``
+===============================  ===========================================  ===================================
 
 
 Building against Parallel HDF5
--- a/docs/build.rst
+++ b/docs/build.rst
@@ -125,8 +125,8 @@ Here's a list of all the configure options currently supported:
 Option                           Via setup.py                                 Via environment variable
 ===============================  ===========================================  ===================================
 Custom path to HDF5              ``--hdf5=/path/to/hdf5``                     ``HDF5_DIR=/path/to/hdf5``
-Custom path to HDF5 lib dir      ``--hdf5-libdir=/path/to/hdf5/lib``          ``HDF5_DIR=/path/to/hdf5``
-Custom path to HDF5 include dir  ``--hdf5-includedir=/path/to/hdf5/include``  ``HDF5_DIR=/path/to/hdf5``
+Custom path to HDF5 lib dir      ``--hdf5-libdir=/path/to/hdf5/lib``          ``HDF5_LIB=/path/to/hdf5``
+Custom path to HDF5 include dir  ``--hdf5-includedir=/path/to/hdf5/include``  ``HDF5_INCLUDE=/path/to/hdf5``
 Custom name for HDF5             ``--hdf5-libname=customname-libhdf5``        ``HDF5_LIBNAME=customname-libhdf5``
 Force HDF5 version               ``--hdf5-version=X.Y.Z``                     ``HDF5_VERSION=X.Y.Z``
 Enable MPI mode                  ``--mpi``                                    ``HDF5_MPI=1``
--- a/setup_configure.py
+++ b/setup_configure.py
@@ -265,7 +265,7 @@ def autodetect_version(libdirs, libnameregexp, mpi=False, hdf5_version=None):
 
     version = "{0}.{1}.{2}".format(int(major.value), int(minor.value), int(release.value))
 
-    if hdf5_version is not None:
+    if hdf5_version is not None and not sys.platform.startswith('win'):
         assert hdf5_version == version
 
     return version
--- a/setup_configure.py
+++ b/setup_configure.py
@@ -2,16 +2,16 @@
 """
     Implements a new custom Distutils command for handling library
     configuration.
-    
+
     The "configure" command here doesn't directly affect things like
     config.pxi; rather, it exists to provide a set of attributes that are
     used by the build_ext replacement in setup_build.py.
-    
+
     Options from the command line and environment variables are stored
     between invocations in a pickle file.  This allows configuring the library
     once and e.g. calling "build" and "test" without recompiling everything
     or explicitly providing the same options every time.
-    
+
     This module also contains the auto-detection logic for figuring out
     the currently installed HDF5 version.
 """
@@ -21,6 +21,9 @@ import os
 import os.path as op
 import sys
 import pickle
+import re
+import platform
+
 
 def loadpickle():
     """ Load settings dict from the pickle file """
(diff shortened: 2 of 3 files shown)


andrewcollette:
Looks like an issue when autodetect fails (and we should probably figure out why it fails in this case, too).

behrmann:
Could you elaborate on what you think fails? 

Just to be on the same page, on Debian and Fedora the build simply fails if `pkgconfig` is not installed. 
If it is installed all thinks might happen. On Debian you just get either the pkgconfig info of the serial oder the openmpi HDF5 library, whoseever -dev package was installed first. On Arch it just silently fails and says that HDF5 is not installed, even if it is.
So I unfortunately think that pkgconfig is just to unreliable to use it.

`autodetect_version` from setup.py is not sufficient either, because it only looks in very nonstandard places (for most distros), if pkgconfig fails, and on top of that requires the HDF5 library to be named libhdf5.so, which at least on Debian it's not. If autodetect_version screws up, you also get the very confusing output that the version is `'1.8.4'`

The `--hdf5` option is insufficient, too, since for one it is very confusing (`--hdf5-prefix` would have been a better name), but necessarily fails if the library and the header files are not directly in `$(--hdf5)/{lib,include}`, which again on Debian they're not.

All this info is then added into a long list of more hardcoded paths, which are then used to build h5py, which more often than not failed for me for the above reasons.

I'm not saying the current setup does not work, only that it is very fragile, which i noticed after moving to using virtualenvs exclusively and trying to install it from pip and not via apt.
h5py does build on Debian if pkgconfig is installed and one tries to build a current master, but I tried building tag `2.4.0`, which got me sucked into that problem. Fortunately for me, pkgconfig did not solve my problems, since I wanted to build a specific version. ;)

andrewcollette:
Sorry, I was referring to the Travis-CI test case: it looks like a number of variables are getting set to "???" when the autodetection fails, and the "???" is making its way into places it shouldn't.

The pkgconfig contributions themselves were an attempt to fix exactly that same hard-coding problem, so I'm certainly on board with changes that make the setup.py process more robust.

behrmann:
Ah, sorry about that. That was mostly for testing purposes to have a
clear output if anything failed, a value of None should be just as fine.
I only found it confusing that the configuration always show a version
of `1.8.4` if it fails.

andrewcollette:
HDF5 1.8.4 is the oldest version in common use (by one of the Ubuntu LTS distributions), so the autodetection would fall back to that if it failed.

behrmann:
I see, but wouldn't it be more natural to assume something is wrong and there is no libhdf5, if it fails to be autodetected?

andrewcollette:
The expectation was that in a substantial number of cases, HDF5 would be present somewhere the compiler could see it, but the autodetection (based on ctypes) wouldn't.

behrmann:
I am not sure I share this interpretation, but I guess it might be dropped now
(or somehow relaxed, coming from unixoid platforms I tend to forget the Windows
world), since in this pull request, the path for the library is search by
greping the output of `ldconfig`. If it is not in there, it will not be in the
system.

andrewcollette:
Good point.  Btw, on Windows, we require that ``--hdf5`` and ``--hdf5-version`` are supplied as part of the build process.

behrmann:
Ah, good point. That needs to be changed in my version, because it sanity checks
the supplied version against the autodetected one, which would now necessarily fail
on Windows then. I will change that.

aragilar:
I just noticed the pull request, and have a few suggestions.
1. Try pkgconfig before ldconfig. The problem with parsing the output of ldconfig is you'll have to worry about what arch your compiling with (and check you use the right one), and you'll ignore libraries that are not included in ldconfig (e.g. user version of hdf5 with pkgconfig info).
2. Check "hdf5-serial" before "hdf5" for serial build (I would avoid doing the same for mpi, as at least in Debian's case, there are two possible versions of mpi (openmpi and mpich), and you would need to work out which one you want to compile against)

Incidentally, the reason the order of installation of the different hdf5 libraries makes a difference is because the priority of the alternatives is equal (50). I'll file a bug against Debian, it would make more sense if the serial version had a higher priority than the mpi version.

The new configuration system is definitely an improvement though.

docs/build.rst
```diff
@@ -104,17 +113,24 @@ You can also configure h5py using environment variables.  This is handy
...
-======================= =========================== ===========================
-Custom path to HDF5     ``--hdf5=/path/to/hdf5``    ``HDF5_DIR=/path/to/hdf5``
-Force HDF5 version      ``--hdf5-version=X.Y.Z``    ``HDF5_VERSION=X.Y.Z``
-Enable MPI mode         ``--mpi``                   (none)
-======================= =========================== ===========================
+===============================  ===========================================  ===================================
+Option                           Via setup.py                                 Via environment variable
+===============================  ===========================================  ===================================
+Custom path to HDF5              ``--hdf5=/path/to/hdf5``                     ``HDF5_DIR=/path/to/hdf5``
+Custom path to HDF5 lib dir      ``--hdf5-libdir=/path/to/hdf5/lib``          ``HDF5_DIR=/path/to/hdf5``
```
aragilar: Typo here and line below (s/HDF5_DIR/HDF5_LIB and s/HDF5_DIR/HDF5_INCLUDE)

behrmann:
@andrewcollette: I added a check to skip the sanity check on windows platforms in 20a44ce4eacd600212fff5a95d96697d6f0b5b9d .

@aragilar: I fixed the typo in 9dd67d81f129e24956d48667edd7582ca2b1a4ea.
I am not convinced, that looking at the pkgconfig-output first would be the best solution. 
1. `ldconfig` can find user-installed libraries, since these can be set with the environment variables `$LD_RUN_PATH` and `$LD_LIBRARY_PATH`. 
2. I am not sure that pkgconfig solves the architecure issue.
3. If pkgconfig only provides a single entry for HDF5, even if multiple versions are installed, the autodetection would still be defeated if we gave precedence to pkgconfig data.

Regarding the second point, I just noticed that at least Debian does provide provide access all pkgconfig data, since one can simply use `pkgconfig.parse('hdf5-serial')`  and `pkgconfig.parse('hdf5-openmpi')`.
I will make appropriate changes and also add some fixes for different mpi implementations and if you guys think it makes more sense to use pkgconfig first and fall back to other stuff later, I am fine with that.

aragilar:
`ldconfig` manages a cache (recreated typically when post-install scripts are run by package managers), not the actual linking. It does not include anything outside of `/lib` and `/usr/lib` (default config files typically add `/usr/local/lib` and multiarch paths). It won't find user libraries, or possibly even admin installed libraries (unless added to the config files). An easy test for this is to add some path with shared library files in it to your $LD_LIBRARY_PATH, then run `ldconfig`. You won't find those shared library files in the output.

If `pkg-config` gives the wrong output, then it's a bug with the distro packaging of hdf5 (which if they have h5py in the distro would result in a FTBFS, which at least in Debian are rated critical, which should mean a fix is quickly produced). If the `ldconfig` method is first, then any issues with that will be fixed in the same way Debian has fixed building h5py currently (https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=756668, i.e. setting CPPFLAGS and LDFLAGS), which means `pip install h5py` won't work.

andrewcollette:
@aragilar, @behrmann... I wonder if we are going about this the wrong way.  Since the beginning of the project, there's been an assumption that h5py should build from source "out of the box" in a single step by end users.  

But the native Python distutils toolchain really isn't there yet for packages which need Cython and a bunch of C libraries; in other words, we are trying to write "configure" because distutils doesn't provide it.  That adds an enormous amount of complexity to setup.py and related facilities, and incidentally makes the h5py project responsible for build failures on every platform, many of which we're not in a good position to debug.

Many Linux distributions now ship pre-built versions of h5py, and with wheels it's increasingly possible for volunteers to build pip-installable packages which don't rely on a compiler toolchain on the destination computer.  Maybe a better approach here would be to figure out what the minimum set of configuration options is (path to HDF5, HDF5 version, etc.), and just make those options easy to set on the command line or setup.cfg.  In other words, make life easier for the downstream maintainers and emphasize their role in distributing h5py.

The package would of course still be buildable by end users, in many cases out of the box.  But like packagers, they would occasionally be responsible for supplying ``--hdf5``, ``--hdf5-version`` and so forth, either on the command line or in setup.cfg.  We wouldn't poke around the system and try to figure things out.

Thoughts?

behrmann:
@aragilar: You are right. My assumption that `ldconfig` would check `LD_RUN_PATH` or `LD_LIBRARY_PATH` was wrong, but a minimally invasive patch to fix that would be to look through these paths, too. 
If one looks through all currently loaded libraries in `ldconfig` and the contents of 
I'm happy to do the work, `LD_{RUN,LIBRARY}_PATH`, this should cover all reasonable use cases or at least the same set as pkgconfig.
I am not against using pkgconfig, I honestly didn't know better, but it took me a while to figure out, that on Debian I can look for diffent pkgconfigs (hdf5, hdf5_serial, hdf5_openmpi, and I assume hdf5_mpich) and I don't know how this changes for different distros (there is no pkgconfi on Arch and I don't know the status for Fedora), but having multiple pkgconfigs to choose from makes the pkgconfig version more complicated, too.

@andrewcollette: Yeah, I see your point, but I don't fully share it. :) I actually did my work to scratch my itch of not being able to install h5py in a virtualenv. Nevertheless, the work done so far already leads all necessary switches to the outside, so that packagers should now be able to do their work more easily, since I've introduced seperate switches for library and include directories as well as library name on top of the existing switches.

As I see it, the work that still needs to be done is to accomodate other MPI implementations than OpenMPI in the search for everything.
If you wish I will make the changes to give pkgconfig precedence over my current solution. Just point the direction where this should go.

aragilar:
I looked at how openSUSE, Fedora, Debian and Arch did hdf5/mpi. Debian (and hence its derivatives) appears to be the only ones to use pkg-config for hdf5. Arch doesn't allow for different versions of hdf5. Fedora and openSUSE put the serial versions in the default library path, and have separate directories for the different mpi versions. Fedora (and presumably openSUSE) use the module system (Fedora's MPI plan is at https://fedoraproject.org/wiki/Packaging:MPI, I haven't found openSUSE's), whereas Debian uses pkg-config, the alternatives system, and namespacing commands (e.g. mpirun.<mpi version>).

Fedora's MPI plan implies the mpi versions are not in the output of `ldconfig` so we'll need to consider LD_LIBRARY_PATH.

In terms of what we should do in terms of poking, I think we should at least try to support running "pip install h5py" on Debian and Fedora (and linking to the serial version), as if we can get that working, then that should mean that it'll work for most users (especially the less-technically inclined, who would be using Debian/Fedora derivatives). Also, given two different people (@behrmann and I) separately tried to get this to work says something about the need to have some level of poking :wink:.

I also had a quick look at OSX (Fink, MacPorts and homebrew), and it looks like it's similar to Arch, in that there's no special support for parallel mpi and serial.

andrewcollette:
Let's use pkgconfig where possible, and resort to asking ldconfig on platforms where pkgconfig is unavailable or clearly broken.

behrmann:
Ok, I'm good with that and will get to work on that hopefully some time this weekend.

ghisvail:
Sorry for the duplicate in #536, although there are 2 separate issues I see here.

You guys already nailed the problem and pkgconfig will sure be a more comprehensive solution than the multiple hooks solution I proposed #536.

ghisvail:
Any progress on this issue ? What is left to be done to close this ?

behrmann:
Things that still need doing:
1) `pkgconfig` needs to be the first thing we try (it is only a fallback in what
I proposed)
2) The fall-back search needs ti work with mpich, too. So far it only looks for
OpenMPI, since I wasn't thinking about anything else.

I'm currently refactoring to suit these needs, but the current state works and
builds on all the platforms I could test it on.

ghisvail:
I have tried your branch on current Debian Jessie and autodetection now works as expected.

Well done !

behrmann:
Thanks. It would have been bad if it wouldn't work on Jessie, since my machine
at work runs it, too. ;)
Ok, back to further changes...

ghisvail:
@behrmann 
I have used your first commit of this PR and modified it a bit to generate a patch to be applied on top of vanilla h5py v2.4. It should hopefully be enough to fix the Debian package for now, until a proper fix lands in h5py v2.5+.

@andrewcollette 
Would you consider applying the fix of the build system to v2.4.1 as well ?

andrewcollette:
@ghisvail, the proposed changes are too large for a bugfix release.

ghisvail:
@andrewcollette 
> the proposed changes are too large for a bugfix release.

FYI, the patch I have only affects the setup*.py scripts and nothing in the module. Your call.

behrmann:
Hi, here a current status report on where I am. My current work can be found in my branch debian_install (https://github.com/behrmann/h5py/tree/debian_install). I thin I squashed all the bugs I've been introducing and I can currently build h5py with and without MPI support with either OpenMPI or MPICH. Disclaimer: The MPICH version doesn't pass the tests, since my mpi4py is not built against MPICH, so I still have to test that, but the other stuff works as expected.
`pkgconfig` is used first, but the stuff I hacked together in the initial commit of this PR is available as fallback and finds everything to succesfully find everything that would also be provided by `pkgconfig`.

The next thing for me to do, besides building mpi4py against MPICH an testing that, to test my current state on Fedora and Arch.

ghisvail:
@behrmann 
This is brilliant work.

I am actually interested in getting an updated Debian package for v2.4+ to build. For now, the Debian package does not consider the MPI component of h5py, but that could be added later. I extracted a patch from your debian_install branch and successfully updated the current Debian package with v2.4. Since this patch is unlikely to land on a v2.4.x bugfix release according to @andrewcollette, it would need to be applied downstream at the packaging level until v2.5 eventually comes out. Is that ok with you @behrmann ?

behrmann:
Well, I wouldn't go that far. It is somewhat working now. If it helps you build
the Debian package, excellent, but I am far from done; I have an idea for a more
elegant solution.
The current state only works for Debian, if I remember correctly, I was still
working to build on Fedora, though.

ghisvail:
@behrmann 
It truns out that the old trick of overriding CFLAGS and LDFLAGS in the Debian packaging still works for v2.4.0, so I left it that way for now. I will push a new version out in Debian experimental soon. I hope your stuff gets merged in v2.5 so we can get rid off this hack and use a proper solution based on python-pkgconfig. Thanks for working on this.

andrewcollette:
@behrmann, any progress?

behrmann:
Not really, too much other work. I realized though, that the proper approach
would be to factor out my work here into a separate small module. Since this is
a rather generic problem, it would be best to make it transparent to h5py and
just have a module generate a pkg-config file to a temporary directory, then in
the installer prepend that path to `$PKG_CONFIG_PATH` and use `pkgconfig`
without any change from there.

ghisvail:
I second @behrmann on this.

Besides the busted version autodetection feature, h5py does build properly if you have a valid pc file describing your HDF5 install using the facilities provided by python-pkgconfig. It would be better and cleaner to capitalize on that and refactor everything else (default or use-provided arguments) to produce a pc file.

setup_configure.py
```diff
@@ -103,134 +379,135 @@ def reset_rebuild(self):
...
+                self.hdf5_define_macros = versioninfo[4]
                 print("Autodetected HDF5 %s" % self.hdf5_version)
             except Exception as e:
                 sys.stderr.write("Autodetection skipped [%s]\n" % e)
-                self.hdf5_version = '1.8.4'
-                
-        if self.mpi is None:
-            self.mpi = oldsettings.get('cmd_mpi')
-                
+                self.hdf5_libdir = '???'
```
ax3l: better: `= None`

ax3l:
It's too bad this branch is very much stalled.

> libdir, ... flags

I like the new flags and the controllability via env vars (for external control, e.g. in pip scenarios).

> multiarch & multiple installs

don't try to find them, use the first one you find or you are ending up re-inventing cmake modules.

take the first you find from `pkgconfig` or `h5cc -show` and if that is not what the user wants (even it is configured as such, since tools like `h5cc` will be configured to that lib) then a user has to point via the flags or env vars.

ax3l:
btw: #707 is updated and is a quick solution in case this PR stays abandoned.
It would also allow to merge my PR in #721 ;)

tacaswell:
This is 3+ years out of date and has merge conflicts, closing due to inactivity.

@behrmann If you want to pick this up again please re-open (or make a new PR, not sure how bad the merge-conflicts are)!

behrmann:
Thanks for closing this! I actually had forgotten about it and guess it would
needed to be done anew, mostly to get a grip on what the current state of
building h5py is.

Issue #537, opened 2015-01-27
h5py/defs.c: No such file or directory


LanceWesterhoff:
Hello-
I went through and tried to build h5py on both CentOS6 and CentOS5. I end up with the following error each time. Since it's a fresh download, I wonder if there is something odd about the way I am installing the package that the current version doesn't support (h5py-2.1.0 installed this same way without difficulty). I'm not seeing anything in the logs that would indicate the problem. One thought is that I'm installing h5py in a separate directory path (e.g. "tools") from my python install, but I think this would be a supported install. Any suggestions? Thanks!

```javascript

[lance@hostname h5py-2.4.0]$ uname -a
Linux hostname 2.6.32-431.5.1.el6.centos.plus.x86_64 #1 SMP Tue Feb 11 21:21:23 UTC 2014 x86_64 x86_64 x86_64 GNU/Linux
[lance@hostname h5py-2.4.0]$ cat /etc/redhat-release 
CentOS release 6.5 (Final)
[lance@hostname h5py-2.4.0]$ export LD_LIBRARY_PATH=/opt/python/Python-2.7.9-01262015/lib:${LD_LIBRARY_PATH}
[lance@hostname h5py-2.4.0]$ export LD_LIBRARY_PATH=/opt/python/Python-2.7.9-tools-01262015/lib:${LD_LIBRARY_PATH}
[lance@hostname h5py-2.4.0]$ export PATH=/opt/python/Python-2.7.9-01262015/bin:${PATH}
[lance@hostname h5py-2.4.0]$ export PATH=/opt/python/Python-2.7.9-tools-01262015/bin:$PATH
[lance@hostname h5py-2.4.0]$ export PYTHONPATH=/opt/python/Python-2.7.9-tools-01262015/lib/python2.7/site-packages
[lance@hostname h5py-2.4.0]$ gcc --version
gcc (GCC) 4.9.0
Copyright (C) 2014 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

[lance@hostname h5py-2.4.0]$ /opt/python/Python-2.7.9-01262015/bin/python setup.py configure --hdf5=/opt/python/Python-2.7.9-tools-01262015/
running configure
Autodetected HDF5 1.8.14
********************************************************************************
                       Summary of the h5py configuration

[...]

PR #541, opened 2015-02-10
Cythonize system fallback

ulrikpedersen:
This patch relates to #537 "h5py/defs.c: No such file or directory".

For old versions of setuptools (probably <0.7), the Cython.Build.cythonize() function can fail silently to process .pyx files into .c files. Updating setuptools fixes this problem, but this update can have more complex implications on a python site installation - and this may not be desired in a production environment.

This patch adds a fallback solution which first checks whether the expected .c files has been produced by cythonize(). If the expected .c files are not found, the system cython is executed on the appropriate set of .pyx files (only for the missing .c files).

The use of the cython from the system environment is similar to the approach of the [numpy build system](https://github.com/numpy/numpy/blob/master/tools/cythonize.py)

### DIFF
--- a/setup_build.py
+++ b/setup_build.py
@@ -109,14 +109,13 @@ class h5py_build_ext(build_ext):
     def run_system_cython(pyx_files):
         try:
             retcode = subprocess.call(['cython', '--fast-fail', '--verbose'] + pyx_files)
-            print "Retcode: ", str(retcode)
             if not retcode == 0:
                 raise Exception('ERROR: Cython failed')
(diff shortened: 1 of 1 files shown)


andrewcollette:
Looks like the build failure is Python 3 related.

ulrikpedersen:
Ah, yes; I'm stuck in the old days with python2.7 and forgot about the python3 print function... It builds on travis now.

andrewcollette:
Thanks!

PR #544, opened 2015-02-26
Incompatible function pointer fix

chinasaur:
Another incompatible pointer type warning fix.  See https://github.com/h5py/h5py/pull/526.


andrewcollette:
Thanks!

PR #551, opened 2015-03-17
Single Writer Multiple Reader SWMR

ulrikpedersen:
This is a changeset to support the new [SWMR feature in the HDF5 library](http://www.hdfgroup.org/HDF5/docNewFeatures/NewFeaturesSwmrDocs.html). 

The SWMR feature is scheduled for inclusion in the HDF5 1.10 release - but is already implemented in the development version from 1.9.178. The HDF Group provide a development snapshot, currently with this version number - and of course it can be found in the [HDF Group subversion repository](http://svn.hdfgroup.uiuc.edu/hdf5/branches/revise_chunks).

SWMR development snapshots can be found on the ftp server: ftp://ftp.hdfgroup.uiuc.edu/pub/outgoing/SWMR

This is a new (upcoming) feature in the HDF5 library, and is providing a few new function calls in the API. Because of that I have taken care to support the feature only when the HDF5 library is >= 1.9.178. Previous versions will still work - but not have these new functions, related to SWMR. This is following the model in h5py of supporting version-specific features.

Unittests are included for both cases: HDF5 library with and without the SWMR feature.

The documentation and examples are hopefully clear enough to explain the feature, how to use it and it's limitations.

Please review and let me know if this is useful :smiley:


andrewcollette:
Yes, it is certainly useful.  I would be OK with releasing this as part of the next minor version, since it doesn't affect builds against public versions of HDF5.

Some nitpicks:

* Could you make the intro to the documentation section even stronger in saying that the feature is experimental (and that HDF5 1.10 is not publicly ready).
* For high-level methods like refresh and flush, rather than raising AttributeError I'd prefer if they were conditionally defined based on the version.  This is because we encourage people to load the package in ipython and bang on the Tab key to look around... non-working methods may be confusing.
* Conditional definition of the ``swmr`` keyword to ``File`` would be a giant pain in the ass, so I'm OK with just adding it to the signature.  But I'd like it to raise ValueError if people pass in True and the swmr features aren't available.

Overall, this is a great and much-requested feature.  Hopefully we can hit the ground running when 1.10 is released.

ulrikpedersen:
> * For high-level methods like refresh and flush, rather than raising AttributeError I'd prefer if they were conditionally defined based on the version. This is because we encourage people to load the package in ipython and bang on the Tab key to look around... non-working methods may be confusing.
> * Conditional definition of the swmr keyword to File would be a giant pain in the ass, so I'm OK with just adding it to the signature. But I'd like it to raise ValueError if people pass in True and the swmr features aren't available.

OK, I have had a go at doing those tweaks. Is that in line with what you had in mind @andrewcollette?

> * Could you make the intro to the documentation section even stronger in saying that the feature is experimental (and that HDF5 1.10 is not publicly ready).

Still need to update the docs...

andrewcollette:
Looks good so far!

ulrikpedersen:
> * Could you make the intro to the documentation section even stronger in saying that the feature is experimental (and that HDF5 1.10 is not publicly ready).

No problem. I've done that with a nice big "warning" box now:

h5pyswmrwarning

Note that the first line of the swmr.rst doc need to have the h5py version number N.N.N substituted before a release is made...

andrewcollette:
Looks great... thanks again.  We may ask for your help again once 1.10 is released, in case the SWMR support in h5py needs a little tweaking.

ulrikpedersen:
That's cool - just drop me an email if or when. In the next few months my colleagues will be getting our hands dirty with a bit more swmr work so I ex

PR #555, opened 2015-04-01
add user and api docs to release tarball

ghisvail:
Dear Andrew,

It would be nice if both the user and developer documentation were provided as part of the release tarball in PyPI so distributions can provide them as part of the packaging process. This PR addresses this by appending the relevant documentation folders in the MANIFEST.in.

Thanks for considering this,

Ghis

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -12,3 +12,5 @@ include setup_build.py
 include setup_configure.py
 include ANN.rst
 include README.rst
+recursive-include docs *
+recursive-include docs_api *


andrewcollette:
I appreciate the idea... unfortunately this will add the Sphinx source docs to the tarball, not the docs themselves.  To build the API docs, h5py must also be on os.path somewhere.

Also, why can't distributions simply pull the code from GitHub?  We tag all the releases...

ghisvail:
> This will add the Sphinx source docs to the tarball, not the docs themselves.

Which is what I need. Debian, and other licensing cautious distros, do not allow binaries (doc pdf for instance) without their corresponding source anyway. With that PR, we have all the stuff in place to produce the HTML and PDF output from the source sphinx docs.

> Also, why can't distributions simply pull the code from GitHub? We tag all the releases...

1) Not everyone does that in a consistent manner, (i.e. format of the tag, etc...)
2) Different hosting = different URL pattern for each of them (Github, Bitbucket, SourceForge...)
3) PyPI = our one stop shop for everything, whenever possible

andrewcollette:
Sounds OK to me.  I don't think having the Sphinx source in the tarball can possibly hurt anything.

Thanks!

PR #560, opened 2015-04-09
remove extra DS_Store files from release tarball

ghisvail:
Dear Andrew,

I have noticed that, since version v2.4, an extra DS_Store file gets into the release tarball in the lzf/ folder. This PR makes sure it will be excluded from future releases.

Cheers,
Ghis


tacaswell:
Should probably add it to the '.gitingore' as well?

ghisvail:
It's already in .gitignore I believe.

andrewcollette:
Thanks!

PR #561, opened 2015-04-14
do not ship doc build in release tarball

ghisvail:
Hi Andrew,

I noticed that the latest v2.5.0 release tarball have a few issues with regards to the documentation. I will make 2 different PRs for these, since both issues relate to different things.

This PR prevents any future temporary sphinx build directory being shipped in the release tarball. Release v2.5.0 has one of these temporary build directory in docs/ but not in docs_api/ which tells me this was not intended.

Cheers,


andrewcollette:
Good catch!

PR #562, opened 2015-04-14
Bug/fix apidoc build

ghisvail:
This PR addresses some issues with the build process of the apidoc.

1) The version string used to be imported from h5py which fails. This PR uses hardcoded values like in the user doc. This could be refactored later.

2) The automod.py now works with Python 3.


andrewcollette:
Importing the version string here is actually intentional... since the low-level docs use autodoc/automod, h5py has to be importable when sphinx is run.

ghisvail:
> Importing the version string here is actually intentional...

Which fails on me with:

```
ImportError("You cannot import h5py from inside the install directory.\nChange to another directory first.")
```

After running:

```
PYTHONPATH=. sphinx-build -N -bhtml docs_api/ docs_api/_build
```

Not quite sure about the rationales behind why you should be outside the source tree to run this.

This is not a deal breaker for me though, since only the user docs will be packaged in the future. Feel free to reject the PR if you feel no further action should be taken on this.

andrewcollette:
Because there's a directory named "h5py" in the source tree (but which doesn't contain the built Cython modules), Python's import machinery gets confused.  You will have to install h5py somewhere, or build it and set PYTHONPATH to the appropriate build directory.  I suppose an inplace build would work, but I haven't tried it.

You can probably see why the docs/ and docs_api/ stuff is not for public consumption. :) Your Py3 patch for automod is welcome, though.  Would you like to amend this PR or make a new one?

ghisvail:
I have reverted the version string commit, leaving only the fix for Python 3.

andrewcollette:
Thanks!

PR #576, opened 2015-05-22
Requirements for setup (numpy, Cython) added

jlombacher:
This is needed if no numpy, Cython is installed, e.g. in Virtualenvs

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -136,6 +136,6 @@ setup(
   ext_modules = [Extension('h5py.x',['x.c'])],  # To trick build into running build_ext
   requires = ['numpy (>=1.6.1)', 'Cython (>=0.17)'],
   install_requires = ['numpy>=1.6.1', 'Cython>=0.17', 'six'],
-  setup_requires = ['pkgconfig', 'six'],
+  setup_requires = ['numpy>=1.6.1', 'Cython>=0.17', 'pkgconfig', 'six'],
   cmdclass = CMDCLASS,
 )
--- a/setup_build.py
+++ b/setup_build.py
@@ -83,7 +83,14 @@ class h5py_build_ext(build_ext):
         except EnvironmentError:
             pass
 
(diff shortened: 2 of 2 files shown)


andrewcollette:
If I recall correctly, what this does is install temporary copies alongside the h5py build directory.  That isn't very helpful, since NumPy has to be available at runtime.

Can you test this in a new virtualenv, and see if it actually installs NumPy to the equivalent of site-packages?

jlombacher:
It's not installed when it's only in setup_requires, but because it's also included in install_requires.
If you don't put it in setup_requires the setup won't work, there is probably a way to fix it, that  numpy is not needed for the setup ... but I think Cython is.
I'll try to look into it in more detail.

andrewcollette:
The root of the problem is that setup.py eventually calls numpy.get_include(), which doesn't work right if NumPy isn't installed in the correct place before setup.py begins executing.  I don't think there's a way around this, but let me know what you find out.

jlombacher:
I think this way it should work. If numpy is not available it's installed for the setup in the egg directory and the includes will be pulled from there. 
I'm not sure if it's a bug of numpy so it should be fixed there.

andrewcollette:
The include files have to be pulled from the actual installed version of NumPy to avoid a possible version mismatch.

Frankly I don't think there's any way to solve this with pip and setup.py, since pip doesn't have a real dependency resolver.  Have you tried using conda?

jlombacher:
As fare as I understand it, it won't be a problem because:
- If numpy is not installed, it's first installs in .egg, for building, and than the same numpy version is installed on the system
- If numpy is installed, no further numpy version will be installed, and for both installation and runtime the system numpy will be used
So always the equal header files are used.

Now my test is finished and it revealed the same result. 

But thanks for the hint with conda, I'll have a look at it.

andrewcollette:
Good point.  I'm going to merge this into master and we'll see what the feedback is.

PR #579, opened 2015-05-31
Large number of fixes and cleanups triggered by pylint

andrewcollette:
Fixes an enormous number of issues identified through static analysis, including:

- Rarely-used code paths broken by mistyped variable names
- Duplicated code which had been moved from function to module level and not cleaned up
- Whole classes which were unused
- Undefined format-string arguments to rarely-triggered exceptions
- Bad exception handlers which eat SystemExit, etc., by mistake
- Missing public close() methods on nearly all low-level identifiers
- Unneeded Mapping classes in h5py._hl.base
- Pointless one-line functions used as iteration callbacks when list.append would suffice
- Piles of undocumented modules, functions, and methods

Also switches on pylint as part of the Travis build for pull requests.

### DIFF
--- a/h5py/_hl/selections2.py
+++ b/h5py/_hl/selections2.py
@@ -76,7 +76,7 @@ def read_selections_scalar(dsid, args):
 class ScalarReadSelection(object):
 
     """
-        Undocumented.
+        Implements slicing for scalar datasets.
     """
     
     def __init__(self, fspace, args):
@@ -95,7 +95,10 @@ class ScalarReadSelection(object):
         yield self.fspace, self.mspace        
 
 def select_read(fspace, args):
-    """ Undocumented. """
+    """ Top-level dispatch function for reading.
+    
+    At the moment, only supports reading from scalar datasets.
+    """
     if fspace.shape == ():
         return ScalarReadSelection(fspace, args)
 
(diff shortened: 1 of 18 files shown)


h5py/_hl/selections2.py
```diff
@@ -71,6 +75,10 @@ def read_selections_scalar(dsid, args):
 
 class ScalarReadSelection(object):
 
+    """
+        Undocumented.
```
tacaswell: This seems like cheating!

h5py/h5s.pyx
```diff
@@ -47,7 +47,7 @@ UNLIMITED = H5S_UNLIMITED
 NO_CLASS = H5S_NO_CLASS
 SCALAR   = H5S_SCALAR
 SIMPLE   = H5S_SIMPLE
-_NULL = H5S_NULL
+globals()["NULL"] = H5S_NULL  # "NULL" is reserved in Cython
```
tacaswell: Did you intend to drop the leading `_` here?  I don't really understand what is going on here, just aggressively pattern matching.

h5py/_hl/selections2.py
andrewcollette: Yes, I suppose I should set a good example...

h5py/h5s.pyx
andrewcollette: This is intentional... Cython doesn't allow "NULL = H5S_NULL" since "NULL" is a reserved word.  We used to patch this up in __init__.py.

PR #597, opened 2015-07-20
Corrected return types on attribute open low-level api functions

ulrikpedersen:
Correcting a few H5Aopen* functions to return hid_t instead of herr_t.

For some reason (unknown to me) returning the herr_t doesn't break anything noticeable when using the current released HDF5 libraries. However, this caused a whole bunch of unittest failures when using the latest SWMR development snapshot: v1.9.222 on ftp://ftp.hdfgroup.uiuc.edu/pub/outgoing/SWMR/src


andrewcollette:
Thanks for catching this!

PR #604, opened 2015-08-05
Update mpi.rst

JensTimmerman:
python setup.py build --mpi results in 'error: option --mpi not recognized' but using this option for configure does seem to work.

The old option seems to work in 2.2, but I need the new version in 2.5


andrewcollette:
Thanks!

PR #606, opened 2015-08-14
Vector assigment

rainwoodman:
To my surprise, it wasn't possible to assign a vector to a component of a compound dataset. I am not aware of any other alternatives to write to a vector component. Hence this patch.


andrewcollette:
Thanks!

PR #608, opened 2015-08-22
update faq section on variable length arrays

dguest:
The FAQ seems to be out of date with respect to variable length arrays. The table of supported types was updated recently, but I found one section that seemed to contradict this update.


andrewcollette:
Thanks!  Looks like we missed this when the feature was added.

Issue #620, opened 2015-09-14
enumeration within variable-length array not maintained across file save/load


smason:
I've just started playing with `h5py` (version 2.5.0) and wanted to use an array consisting of a variable length enumeration, I can create it and load it back in without error, but the datatype from the loaded data seems to have lost the details of the enumeration.  For example:

    with h5py.File('myfile.hdf5','w') as f:
        dt1 = h5py.special_dtype(vlen=h5py.special_dtype(enum=(np.int32, dict(foo=1, bar=2))))
        df = f.create_dataset('test', (2,), dtype=dt1)
        df[:] = [[1],[2,1]]

    with h5py.File('myfile.hdf5','r') as f:
        dt2 = f['test'].dtype

    print(h5py.check_dtype(enum=h5py.check_dtype(vlen=dt1)))
    print(h5py.check_dtype(enum=h5py.check_dtype(vlen=dt2)))

I get back the expected `{'foo': 1, 'bar': 2}` for the first print while `None` is printed for `dt2`—I'm expecting the output to be the same.  The output of `h5dump` contains:

    DATATYPE  H5T_VLEN { H5T_STD_I32LE}

So it seems the enum is never being written.  If I don't wrap the enum in a variable length array everything seems OK in Python and `h5dump`.

PR #621, opened 2015-09-14
fix for issue #620

smason:
These patches add a test and a fix for my recently reported issue #620


andrewcollette:
Great!  Looks like that "logical" keyword is showing up in a lot of places.

smason:
Thanks, that was quick!
Was amazed that was all it took.

Issue #599, opened 2015-07-21
Non string attributes broken in latest master (after #597)


gpwright:
Latest version of h5py built against hdf5 swmr no longer supports creation of non string attributes for example ints or floats.  String attributes work without error.

My configuration:

OS: Ubuntu 14.04. 
hdf5 version: hdf5-1.9.222-swmr0
h5py version: ea4e2d6bd978b226e9223053592b52dbfb5a2f2b

E.g.

```python
import h5py

f = h5py.File("test.h5", "w")
g = f.create_group("group")
g.attrs["A"] = 2
```

Result:

```python
gpwright@gpwright-desktop:~$ python test.py
Traceback (most recent call last):
  File "test.py", line 5, in <module>
    g.attrs["A"] = 2
  File "_objects.pyx", line 54, in h5py._objects.with_phil.wrapper (/home/gpwright/Sources/h5py/h5py/_objects.c:2511)
  File "_objects.pyx", line 55, in h5py._objects.with_phil.wrapper (/home/gpwright/Sources/h5py/h5py/_objects.c:2464)
  File "/usr/local/lib/python2.7/dist-packages/h5py-2.5.0-py2.7-linux-x86_64.egg/h5py/_hl/attrs.py", line 93, in __setitem__
    self.create(name, data=value, dtype=base.guess_dtype(value))
  File "/usr/local/lib/python2.7/dist-packages/h5py-2.5.0-py2.7-linux-x86_64.egg/h5py/_hl/attrs.py", line 169, in create
    htype = h5t.py_create(original_dtype, logical=True)
  File "h5t.pyx", line 1448, in h5py.h5t.py_create (/home/gpwright/Sources/h5py/h5py/h5t.c:14960)
  File "h5t.pyx", line 1468, in h5py.h5t.py_create (/home/gpwright/Sources/h5py/h5py/h5t.c:14791)
  File "h5t.pyx", line 1482, in h5py.h5t.py_create (/home/gpwright/Sources/h5py/h5py/h5t.c:14336)
  File "h5t.pyx", line 1302, in h5py.h5t._c_int (/home/gpwright/Sources/h5py/h5py/h5t.c:12338)
ValueError: Not a datatype or dataset (Not a datatype or dataset)
```

gpwright:
Hi,

Does anyone have any insights about this?  I just tested this again on latest master and it's still a problem: 441ba1ca8658e20e335aacec043afd771ea9508d

[...]

Issue #616, opened 2015-09-09
SWMR has some problems:


NAThompson:
Not unexpected as it's experimental, but:

1) If we use `hdf5-1.9.222-swmr0.tar.gz` ftp'd from ftp://ftp.hdfgroup.uiuc.edu/pub/outgoing/SWMR/src (which is linked from <a href="http://docs.h5py.org/en/latest/swmr.html">the documentation</a>) and the following dockerfile:

```<bash>
FROM ubuntu:15.04
ENV HOME /root
RUN apt-get update
RUN apt-get -yq install gcc build-essential subversion zlib1g-dev
ADD hdf5-1.9.222-swmr0.tar.gz /root
RUN cd ; cd hdf5-1.9.222-swmr0; ./configure --prefix=/usr/
RUN cd ; cd hdf5-1.9.222-swmr0; make && make install

RUN apt-get install -y python3-dev
RUN apt-get install -y python3-pip
RUN pip3 install Cython
RUN pip3 install numpy
RUN pip3 install six
RUN pip3 install h5py
CMD python3 -c "import h5py"
```

we get the error:

```<bash>
docker build -t hdf5v19 .
docker run hdf5v19
ValueError: Not property objects (Not property objects)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/local/lib/python3.4/dist-packages/h5py/__init__.py", line 31, in <module>
    from .highlevel import *
  File "/usr/local/lib/python3.4/dist-packages/h5py/highlevel.py", line 13, in <module>
    from ._hl.base import is_hdf5, HLObject
  File "/usr/local/lib/python3.4/dist-packages/h5py/_hl/base.py", line 78, in <module>
    dlapl = default_lapl()
  File "/usr/local/lib/python3.4/dist-packages/h5py/_hl/base.py", line 65, in default_lapl
    lapl = h5p.create(h5p.LINK_ACCESS)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper (/tmp/pip-build-wf8df8jq/h5py/h5py/_objects.c:2574)
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper (/tmp/pip-build-wf8df8jq/h5py/h5py/_objects.c:2533)
[...]

PR #625, opened 2015-09-30
Longint fixes

gpwright:
Between version 1.9.178 and 1.9.222 of HDF5 the type hid_t was changed from int to long int.  h5py now handles this appropriately by modifying it's own hid_t cython declaration based on the HDF5 version number.

Note: this changeset fixes issue #599 and issue #616

### DIFF
--- a/h5py/api_types_hdf5.pxd
+++ b/h5py/api_types_hdf5.pxd
@@ -13,10 +13,7 @@ include "config.pxi"
 
 cdef extern from "hdf5.h":
   # Basic types
-  IF HDF5_VERSION >= (1, 9, 222):
-    ctypedef long int hid_t
-  ELSE:
-    ctypedef int hid_t
+  ctypedef long int hid_t
   ctypedef int hbool_t
   ctypedef int herr_t
   ctypedef int htri_t
--- a/h5py/api_types_hdf5.pxd
+++ b/h5py/api_types_hdf5.pxd
@@ -213,6 +213,7 @@ cdef extern from "hdf5.h":
 # === H5I - Identifier and reflection interface ===============================
 
   ctypedef enum H5I_type_t:
+    H5I_UNINIT       = -2,  # uninitialized Group
     H5I_BADID        = -1,  # invalid Group
     H5I_FILE        = 1,    # group ID for File objects
     H5I_GROUP,              # group ID for Group objects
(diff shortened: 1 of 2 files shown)


gpwright:
Btw I'm pushing this to master but can we cherry-pick it onto the 2.5 branch since that's what I'm actually developing on?  Should I create a separate pull request or can it be done as part of this one?

andrewcollette:
I have some lingering concerns about this.  Specifically, as I recall we never relied on the exact type involved in the "ctypedef int hid_t" declaration; in fact, this bit from the Cython docs (number 3 in the second list) specifically states that the exact type should *not* matter:

http://docs.cython.org/src/userguide/external_C_code.html#referencing-c-header-files

In other words, "hid_t" should always be what shows up in the Cython output so the "int" or "long int" should not be visible.  Then it will map to whatever int/long int is declared by HDF5.

Any idea of how to address this?  One approach might be to run Cython with the "int" definition, and then the "long int" definition, and simply diff the output files.  Then we could see where the exact type was "leaking" into the Cython output.

gpwright:
I understand your point, particularly the section in the cython docs that says "will work okay whatever the actual size of a word is (provided the header file defines it correctly)".

Moreover I just ran the diff as you suggested on builds with and without the typedef change, and the diff showed no changes to the builds except for the typedef and some unknown changes to the binary data inside the egg file.

I just verified for my own sanity that by reverting back to typedef int and trying to "import h5py" I am back to the original problem "Value Error: Not a property list class".

I have little or no knowledge of cython but my explanation would be that at runtime, python is instantiating int objects rather than long objects based on the cython type.  The python int object is then assigned a long and overflows.

andrewcollette:
Interesting!  So what happens if we simply use "ctypedef long int" on both HDF5 1.8 and 1.10?

andrewcollette:
(ctypedef long int hid_t, that is)

gpwright:
I built h5py against HDF5 1.8 with ctypedef long int, and it seems fine.  I can import h5py, and run some simple tests of my own, and all the build-in tests are passing.

Version 1.10 I couldn't find online... has this been released yet?

So are you thinking to skip the version check and just always use long?

andrewcollette:
Sorry, 1.9 is the branch that will become 1.10.

Since the choice of type evidently doesn't affect the Cython-generated C files, I would suggest removing the version check and just using long int everywhere.

gpwright:
I removed the version check

andrewcollette:
Looks good.  Thanks again for tracking this down!

PR #631, opened 2015-10-08
Read fixed-length unicode

xen-0:
Change set to allow h5py to read fixed-length unicode string datasets.
We generate a lot of hdf5 files with fixed-length unicode attributes so this is a convenient change.
Scalar datasets should behave in the same way as variable length datasets.

I haven't added a test for this change as there isn't a means to write fixed-length unicode with h5py, but if you're happy for me to include a pre-written hdf5 file to use in a test case, I will.

### DIFF
--- a/h5py/_hl/attrs.py
+++ b/h5py/_hl/attrs.py
@@ -80,7 +80,7 @@ class AttributeManager(base.MutableMappingHDF5, base.CommonStateObject):
         is_fl_unicode = False
         try:
             is_fl_unicode = attr.get_type().get_cset() == 1 and not attr.get_type().is_variable_str()
-        except AttributeError, e:
+        except AttributeError:
             pass
 
         if is_fl_unicode:
--- a/h5py/_hl/attrs.py
+++ b/h5py/_hl/attrs.py
@@ -88,12 +88,15 @@ class AttributeManager(base.MutableMappingHDF5, base.CommonStateObject):
 
         attr.read(arr, mtype=htype)
 
-        if is_fl_unicode:
-            arr = convert_utf8_array(arr)
 
         if len(arr.shape) == 0:
+            if is_fl_unicode:
+                arr = numpy.array(arr.tobytes().decode('utf-8'))
             return arr[()]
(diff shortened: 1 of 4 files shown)


andrewcollette:
I agree wholeheartedly that we need to figure out what to do about such strings.  But I'm not loving this Python-side conversion approach, particularly since the types don't round trip and it's difficult to test.

One the Unicode front I actually think it's too early for pull requests.  It's clear h5py needs some kind of explicit, documented approach to handling Unicode so that the fallout of such decisions is understood.  For example, this PR will return NumPy Unicode strings ("U"), which then can't be stored in the file, because HDF5 doesn't support UTF-32.  Likewise, because this PR bypasses the type system in h5t.pyx, the dtype of the dataset won't match the dtype of the returned data, which is one of the invariants h5py tries to maintain.

I would suggest that the next step, for those who are interested in getting Unicode working with h5py in a sane fashion, would be to generate a PEP-style document that we can discuss on the mailing list.  We have been going in circles around Unicode for years and I think having a concrete proposal that everyone can discuss and understand, as opposed to individual short-term fixes, will be the best way to get this functionality into h5py.  This would also give us an opportunity to coordinate with PyTables, who have their own Unicode approach that we'll eventually have to harmonize when the projects merge.

andreabedini:
I agree with what @andrewcollette says. Also "fixed-length unicode" makes me cringe as it's poorly worded (unicode is not an encoding).

@xen-0 a solution would be the following: since you are the one choosing the encoding, you can save the encoded strings for what they are, sequences of bytes (once encoded they are not strings anymore). The "string" data type in HDF5 has to be considered more like metadata describing about how to interpret the sequence of bytes it contains. Given it's not expressive enough for you, just avoid it. The only thing you loose if the ability of other applications to interpret automatically your data (e.g. h5dump will show numbers).

I hope it makes sense.

xen-0:
Fixed length UTF-8 would be more accurate.

We'll stick with ASCII for the moment - the files we create may be opened with tools outside of our control.

PR #635, opened 2015-10-21
Add support for client classes to automatically yield h5py datasets.

bsmithyman:
Implement a new case in `Group.__setitem__`, which allows instances of custom classes to automatically convert themselves to h5py datasets. If the right-hand-side of the setitem call has a method `__toh5ds__`, call that method and expect it to return a new, unlinked dataset.

Example code:

````
import h5py
import numpy

class RHS(object):
    
    def __init__(self, arr, attrs):
        
        self.arr = arr
        self.attrs = attrs
    
    def __toh5ds__(self, group, name):
        
        ds = group.create_dataset(None, data=self.arr)
        ds.attrs['name'] = name
        for attr in self.attrs:
            ds.attrs[attr] = self.attrs[attr]
        
        return ds

h5file = h5py.File('test.h5', 'w')
g = h5file.create_group('test')

instance = RHS(numpy.array([1, 2, 3, 4, 5]), {'colour': 'red', 'shape': 'pointy'})

g['a'] = instance
````

### DIFF
--- a/h5py/_hl/group.py
+++ b/h5py/_hl/group.py
@@ -266,6 +266,12 @@ class Group(HLObject, MutableMappingHDF5):
         Numpy dtype
             Commit a copy of the datatype as a named datatype in the file.
 
+        Class that implements __toh5ds__
+            Call obj.__toh5ds__(self, name), to allow client code to support
+            automatic coercion to a dataset. The name is provided to the
+            client for metadata generation, but the client is not permitted
+            to change the name used for the assignment.
+
         Anything else
             Attempt to convert it to an ndarray and store it.  Scalar
             values are stored as scalar datasets. Raise ValueError if we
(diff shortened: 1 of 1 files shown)


andrewcollette:
@bsmithyman, this is an interesting idea but it would have to be discussed on the mailing list before we get to a PR.

bsmithyman:
Thanks @andrewcollette; my apologies for the gaffe then. I've written a post to the mailing list, and referenced both the PR and my original commit (in case you want to close this).

bsmithyman:
After discussion on the h5py mailing list, it was decided that this makes more sense at a higher level. Objects do not "round-trip", and implementation of this would require complicating the h5py stack. In keeping with h5py's design goal to remain small and efficient, this isn't a direction that makes sense at the core h5py level. Implementation as a userspace wrapper (similar to the `hickle` project) may make sense in future.

PR #639, opened 2015-10-29
construct python dtype with correct offsets of compound types

nevion:
There was an assumption made about offsets in the dtype generated from h5 types.  Structures don't have to be packed yet were assumed to be with the dtype constructor used.  The constructor swapped to supports the offsets information that makes sure the dtype accurately describes the h5 compound type.

### DIFF
--- a/h5py/h5t.pyx
+++ b/h5py/h5t.pyx
@@ -1078,6 +1078,7 @@ cdef class TypeCompoundID(TypeCompositeID):
         cdef int nfields
         field_names = []
         field_types = []
+        field_offsets = []
         nfields = self.get_nmembers()
 
         # First step: read field names and their Numpy dtypes into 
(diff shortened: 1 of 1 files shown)


andrewcollette:
Thanks!  This does seem like a more correct approach.  Can I ask, though, if this was intended to fix a specific problem you encountered?  The py_dtype() methods generally don't return byte-for-byte representations of their HDF5 types.  For example, compound types can map to complex numbers if the field names are correct, and vlens map to Python objects even though the C representations are totally different.

What happens at the moment is that Python will use whatever default offsets come from the dtype constructor (presumably it will be fully packed), and the HDF5 type conversion will put the fields in the correct place.  Likewise when writing back to the file.  Is there some reason this behavior should be avoided?

nevion:
I rely on accurate dtype descriptions for memory and file types.

Packet tables do not write packed structs for instance and I ran into this issue reading them.  I'm not sure about normal tables converting to packed structs but I don't think they do either.  And if you specified the dataset type to be an unpacked record, well, again it might be the same case.  I also read in, for some data, what I call posix packet tables (a binary record file) where I write an h5 file which describes the structures to.

Btw packing will lead to slower memory accesses and may break SSE2/vectorized processing alignment requirements.

In short you aren't being as compatible as you should be, for no discernible reason/gains.

andrewcollette:
All great points.  I agree it's valuable to have the dtype reflect the actual offsets chosen by the HDF5 file author.  Thanks!

PR #640, opened 2015-11-03
setup: Support building with current mpi4py version

wence-:
mpi4py migrated the mpi4py.mpi_c module to mpi4py.libmpi in April 2014.
After the release of v1.3.1 but before v2.  Sniff this in setup
configure and import MPI types from the appropriate module in h5p.pyx,
to allow building with MPI on and modern mpi4py versions.


wence-:
I'd like to be able to build h5py against modern mpi4py.  But currently can't, because mpi4py.mpi_c is removed since April 2014 (between 1.3.1 and 2.0).  This fix changes the setup and build to deal with both versions.

andrewcollette:
@wence-, sounds good to me!  As you know we don't have MPI-specific automated tests at the moment.  Can I ask you to be available if issues related to this patch come up after it's released?

wence-:
Certainly.

I intend to use this feature and will be build-testing with MPI enabled, so I, at least, have a vested interest in it working.  Do you intend a point release any time soon?  I'd like to be able to point users to a released version, rather than the bleeding edge.

andrewcollette:
Sounds great.  I will actually have some time in the next few days so I can roll this into 2.6.0.

wence-:
Great, thanks.

PR #642, opened 2015-11-07
Use item value in KeyError instead of error message

mgeier:
This is not mentioned in [the docs](https://docs.python.org/3/library/exceptions.html#KeyError), but I think a `KeyError` is supposed to be constructed with the actual key, whose `repr()` is then used in the error message.

If it's constructed with an error message string, this message will be displayed with quotes, which looks strange, especially if the key is a string, and even worse if the key is an empty string, e.g.:

    >>> data['']
    KeyError                                  Traceback (most recent call last)
    ...
    KeyError: ' not found'

See also http://stackoverflow.com/a/24999035/500098.

### DIFF
--- a/h5py/_hl/dims.py
+++ b/h5py/_hl/dims.py
@@ -81,7 +81,7 @@ class DimensionProxy(base.CommonStateObject):
                     
             res = h5ds.iterate(self._id, self._dimension, f, 0)
             if res is None:
-                raise KeyError('%s not found' % item)
+                raise KeyError(item)
             return Dataset(res)
 
     def attach_scale(self, dset):


andrewcollette:
Yes... it's only a pity we can't do the same with lookups on the Group objects.

mgeier:
Thanks for merging!

I don't understand all this Cython code, but why is it not possible to construct the `KeyError` with the given Python object instead of the error message string (which I didn't find anywhere in the code ...)?

When I try accessing a non-existing dataset within a group, I get this:

    KeyError: "Unable to open object (Object 'x' doesn't exist)"

It would be more Pythonic to get this:

    KeyError: 'x'

But I guess all this isn't really that important anyway ...

andrewcollette:
Unfortunately those KeyErrors come directly from HDF5, which means we don't have access to the original object name.

mgeier:
OK, I see.

I'm not necessarily saying that h5py should change this, I'm asking mainly out of curiosity:

In whatever context this error is raised, shouldn't the original key given by the user be available?
What about ignoring the message from HDF5 and just using the original key from the user?

I guess one answer to my own question would be that combined keys like `'/group2/subgroup7/data5'` wouldn't give any information which exact part of the query is invalid, but I saw that an error in one of the group names just raises this very generic error:

```
KeyError                                  Traceback (most recent call last)
...
KeyError: 'Unable to open object (Component not found)'
```

The more explicit error is only raised if all the names except the last one in the path are correct.

I think this would be more helpful in the general case:

```
KeyError: '/group2/subgroup7/data5'
```

PR #648, opened 2015-11-17
Added MPI collective I/O support

valiantljk:
Added collective I/O in h5py2.5.0. The changes basically followed Andrew's previous branch 'h5py-mpi_collective_24', which were developed in 2013-2014, and updated in May 2015, but then not maintained, nor merged to h5py master. 
The changes modified the related files to provide the collective io property, and fixed the function name errors and also conducted tests with 3200 processes.
To use the collective I/O: 
```
with dset.collective:
                dset[start:end,:] = temp
```
A test code is available in examples/collective_io.py, 

A sample test result on [Cori](http://www.nersc.gov/users/computational-systems/cori/):
* Data size x: 6553600 y: 1024
* File size: ~50 GB
* Number of processes: 3200
* Independent write time: 10.32 s
* Collective write time: 4.49 s


andrewcollette:
This feature would be great.  Looks like you've got some Travis failures, though.

valiantljk:
Hi Andrew, 
The numpy is supposed to be skipped in the Travis test, as set in the pylintrc, 'extension-pkg-whitelist=numpy,h5py', but I still get these errors 'Module 'numpy' has no 'asscalar' member (no-member)', as shown in 436.6 build. Any ideas?

Best,
Jialin

jakirkham:
@valiantljk @andrewcollette 

The pylint noise appears to be caused by some changes in pylint 1.5.x and astroid 1.4.x. This happens on master too. See issue here ( https://github.com/h5py/h5py/issues/655 ).

I propose that these pinned at present as I am doing in this PR ( https://github.com/h5py/h5py/pull/656 ).

jakirkham:
@valiantljk, could you try merging or rebasing on top of ( https://github.com/h5py/h5py/pull/656 )? I expect this will fix the `pylint` errors.

valiantljk:
@jakirkham , I added the two lines as shown in #656 , but still couldn't pass the pylint tests.

jakirkham:
@valiantljk, while it is true that is pylint is failing, the reason it is still failing is not the same. Instead, it appears to be complaining about two "private" function call that are in your PR. Whether there is a better way to do this or not, I will let you discuss this with @andrewcollette.

Error: ( https://travis-ci.org/h5py/h5py/jobs/94718698#L127-L128 )
Code: ( https://github.com/h5py/h5py/pull/648/files#diff-51d64f4574399e0fc3ed97ee67713edeR166 )
Code: ( https://github.com/h5py/h5py/pull/648/files#diff-51d64f4574399e0fc3ed97ee67713edeR169 )

andrewcollette:
@valiantljk, in this case I think it is safe to disable that PyLint warning.  We don't want to make `_dxpl` part of the public API and your approach here is perfectly fine.  See the AstypeContext class in dataset.py for an example of how to do this.

Also, thanks for taking the time to get this feature working... many folks from the HPC community have been asking about it.

valiantljk:
@andrewcollette, sounds great. Let me fix it shortly.

andrewcollette:
Looks good... thanks again.

Issue #638, opened 2015-10-23
ENH: Provide an `ndim` attribute for `Dataset`


jakirkham:
Ideally, this could be implemented by just finding the length of `.shape`. It would be nice to be a little more similar to `numpy` in this way.

jakirkham:
I might be willing to give this a shot with some guidance.

PR #649, opened 2015-11-20
Add `ndim` to Dataset

jakirkham:
Fixes https://github.com/h5py/h5py/issues/638

Requires https://github.com/h5py/h5py/pull/656
Requires https://github.com/h5py/h5py/pull/653

Simply finds the length of shape.

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -175,6 +175,11 @@ class Dataset(HLObject):
         return DimensionManager(self)
 
     @property
+    def ndim(self):
+        """Numpy-style attribute giving the number of dimensions"""
+        return self.id.rank
+
+    @property
     @with_phil
     def shape(self):
         """Numpy-style shape tuple giving dataset dimensions"""
--- a/h5py/tests/hl/test_dataset_getitem.py
+++ b/h5py/tests/hl/test_dataset_getitem.py
@@ -58,6 +58,10 @@ class TestEmpty(TestCase):
         dsid = h5py.h5d.create(self.f.id, b'x', tid, sid)
         self.dset = h5py.Dataset(dsid)
         
+    def test_ndim(self):
+        """ Verify number of dimensions """
+        self.assertEquals(self.dset.ndim, 0)
+        
     def test_shape(self):
         """ Verify shape """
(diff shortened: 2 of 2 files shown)


jakirkham:
I think can and maybe should be cached, but I am not sure how that is being done currently. So, could use a few pointers.

jakirkham:
Added some method of caching. There is probably a better one.

jakirkham:
Could use some pointers on how to test this.

andreabedini:
isn't this the same as [Dataset.rank](https://github.com/jakirkham/h5py/blob/add_ndim/h5py/h5d.pyx#L140)?

jakirkham:
Really? I don't seem to be able to use `rank` on `Dataset`s. It looks like this commit is really old and I am on `2.5.0` so I am not sure why this is.

In any event, the point of the addition of `ndim` here is to be more duck type friendly with regards to `numpy.ndarray`s. There is no `rank` for them and `ndim` is heavily used.

andrewcollette:
@jakirkham, it looks like you have accidentally added .ndim to the low-level DatasetID in h5d.pyx.  I think the class you are looking for is in h5py/_hl; you could do something like the following:

```
@property
def ndim(self):
    return self.id.rank
```

jakirkham:
Thanks for the pointer, @andrewcollette. I have done exactly that and tried to add some tests. They seem ok, but I would like to see the tests pass. Let me know if you would have them changed and/or moved somewhere else.

jakirkham:
Alright, looks like the tests are passing. Ready for more feedback, if needed.

jakirkham:
Had added some tests for `Dataset.shape`, but have broken them out in this PR ( https://github.com/h5py/h5py/pull/653 ) for separate consumption.

jakirkham:
`pylint` was giving me some grief as it was rules in `pylintrc` so I had to pin the version to something more reasonable.

jakirkham:
How does this look so far?

andrewcollette:
Looks good... thanks.

jakirkham:
Think I forgot a `with_phil` decorator. At least, I think it is needed here. So, I put it in this PR ( https://github.com/h5py/h5py/pull/660 ).

Issue #587, opened 2015-06-17
Minimum Cython version in documentation incorrect


jjhelmus:
The minimum version of Cython required to build h5py from source is incorrect in the documentation and setup.py file.  These both state Cython version 0.17 or later yet using a version of Cython lower than 0.19 cannot cythonize the `/h5py/_conv.pyx` file.  Specifying Cython version 0.19 or later seems reasonable.

Partial log when building with Cython 0.18:

```
$ python setup.py build
zip_safe flag not set; analyzing archive contents...
...
Autodetected HDF5 1.8.5
********************************************************************************
                       Summary of the h5py configuration

    Path to HDF5: None
    HDF5 Version: '1.8.5'
     MPI Enabled: False
Rebuild Required: False

********************************************************************************
Executing api_gen rebuild of defs
Executing cythonize()
Compiling /home/jhelmus/python/h5py-2.5.0/h5py/defs.pyx because it changed.
Compiling /home/jhelmus/python/h5py-2.5.0/h5py/_errors.pyx because it changed.
Compiling /home/jhelmus/python/h5py-2.5.0/h5py/_objects.pyx because it changed.
Compiling /home/jhelmus/python/h5py-2.5.0/h5py/_proxy.pyx because it changed.
Compiling /home/jhelmus/python/h5py-2.5.0/h5py/h5fd.pyx because it changed.
Compiling /home/jhelmus/python/h5py-2.5.0/h5py/h5z.pyx because it changed.
Compiling /home/jhelmus/python/h5py-2.5.0/h5py/h5.pyx because it changed.
Compiling /home/jhelmus/python/h5py-2.5.0/h5py/h5i.pyx because it changed.
Compiling /home/jhelmus/python/h5py-2.5.0/h5py/h5r.pyx because it changed.
Compiling /home/jhelmus/python/h5py-2.5.0/h5py/utils.pyx because it changed.
Compiling /home/jhelmus/python/h5py-2.5.0/h5py/_conv.pyx because it changed.
Compiling /home/jhelmus/python/h5py-2.5.0/h5py/h5t.pyx because it changed.
Compiling /home/jhelmus/python/h5py-2.5.0/h5py/h5s.pyx because it changed.
Compiling /home/jhelmus/python/h5py-2.5.0/h5py/h5p.pyx because it changed.
Compiling /home/jhelmus/python/h5py-2.5.0/h5py/h5d.pyx because it changed.
[...]

Issue #577, opened 2015-05-23
Incorrect dependency on Cython?


QuLogic:
The readme states that Cython is required for building. However, `setup.py` includes `Cython` in `requires` which is a _runtime_ requirement.

If for example, you install `h5py` in `conda` (which is pre-built), then run `python setup.py install` for a package that depends on `h5py`, it will attempt to install `Cython` as a dependency of `h5py`.

So it seems that either the readme is wrong, or `setup.py` is wrong.

andrewcollette:
As I recall this was a workaround for other broken behavior with setuptools/pip.  It's mostly working now, and I'm not eager to change anything because the whole process is so fragile.

QuLogic:
I don't see how that could be the case. Entries in `requires` are installed _after_ the requested package, so they couldn't have any effect on the installation process.

The log shows it was added in 0fad9f00f70d256125b9940a7a5738ae20fe60c6 as part of a refactoring, not in response to any bugs. So as long as removing it works for you, I don't think it would break anything.

andrewcollette:
Hmmm.... maybe I was too hasty on this one.  Btw we just merged a PR which adds NumPy & Cython to setup_requires.

aragilar:
This was fixed by #650.

PR #650, opened 2015-11-23
Clean up versions and testing

aragilar:
This is the result of https://groups.google.com/forum/#!topic/h5py/CRi4pEeVpdU.
It fixes incorrect versions in setup.py, tox.ini, and in a number of places in the documentation (as per #587 and #577).

It also improves the tox.ini (avoids setup.py test), and uses tox for testing on travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,22 +3,27 @@ language: python
 notifications:
   email: false
 
-python:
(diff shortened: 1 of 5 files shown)


andrewcollette:
Looks great.  Thanks!

PR #651, opened 2015-11-24
Move to new travis container infrastructure

aragilar:
Travis has a new way of running tests, using containers instead of virtual machines. Details can be found at https://docs.travis-ci.com/user/migrating-from-legacy/, but for us it should result in faster testing, as we can keep pip wheel cache (and I plan on testing if we can use it to store the eggs from setup_requires), and it's claimed that the container will start faster than the vm.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,17 @@ language: python
 notifications:
   email: false
 
+addons:
+  apt:
+    packages:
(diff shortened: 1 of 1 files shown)


andrewcollette:
Anything that makes the tests faster is fine with me.

Issue #655, opened 2015-12-02
pylint very noisy


jakirkham:
So, it doesn't appear that `pylint` makes any noise on `master`. ( https://travis-ci.org/h5py/h5py/jobs/92838612 ) However, when adding some relatively simple tests, I am seeing a ton of noise from `pylint` related to code I'm not touching. ( https://travis-ci.org/h5py/h5py/jobs/94427849 )

PR #656, opened 2015-12-02
tox - pin pylint and astroid to last working versions 

jakirkham:
Fixes https://github.com/h5py/h5py/issues/655

Appears the new 1.5.x version `pylint` ignores or badly misinterprets the settings in `pylintrc` and that `astroid` 1.4.x also has some bugs that generate a bunch of false positives. Until this can be addressed, simply use the last known working versions of `pylint` (i.e. 1.4.x) and `astroid` (i.e. 1.3.x), which do not have this problem.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -5,7 +5,8 @@ envlist = {py26,py27,py32,py33,py34}-{test,pylint}-{nodeps,deps}
 deps =
     deps: numpy>=1.6.1
     deps: cython>=0.19
-    pylint: pylint
+    pylint: astroid>=1.4
+    pylint: pylint>=1.5
     py26: unittest2
 commands =
     test: python -c "import h5py; h5py.run_tests()"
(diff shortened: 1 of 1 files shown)


jakirkham:
Cache was tricking it into installing the pinned version when they weren't pinned. This should correct that problem as it requires the problematic versions now. Then it pins them one after another to demonstrate the issues seen until all pins are applied and the problem is eliminated.

jakirkham:
Not ready to merge yet.

jakirkham:
Alright, this checks out. Ready to merge.

andrewcollette:
@jakirkham, thanks for debugging this!

jakirkham:
Of course.

jakirkham:
@andrewcollette, looks like some of the builds have hung probably due to the issues Travis CI was having yesterday ( https://www.traviscistatus.com/incidents/mdpq4zcsmvmr ). Maybe it is worth canceling or restarting these so the build queue can maintain its normal bandwidth. Especially as this problem is now resolved.

PR #658, opened 2015-12-05
Fix incorrect tox test command

### LINKED ISSUE #654: Testing with tox not returning non-zero exit code on failure
Here is an example where it is doing this ( https://travis-ci.org/h5py/h5py/jobs/94427845 ). Is there some reason that `tox` is not returning a non-zero exit code on test failure.

aragilar:
Fixes #654, but it's probably worth checking the output from travis (since the tests should pass).

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -9,7 +9,7 @@ deps =
     pylint: pylint>=1.4,<1.5
     py26: unittest2
 commands =
-    test: python -c "import h5py; h5py.run_tests()"
+    test: python -c "from sys import exit; import h5py; exit(0) if h5py.run_tests().wasSuccessful() else exit(1)"
     pylint: pylint h5py
 changedir =
     test: {toxworkdir}


jakirkham:
Needs this ( https://github.com/h5py/h5py/pull/656 ) as `pylint` and `astroid` got updated and seem to not be respecting `pylintrc` and may have other bugs.

tox.ini
```diff
@@ -9,7 +9,7 @@ deps =
     pylint: pylint>=1.4,<1.5
     py26: unittest2
 commands =
-    test: python -c "import h5py; h5py.run_tests()"
+    test: python -c "from sys import exit; import h5py; exit(0) if h5py.run_tests().wasSuccessful() else exit(1)"
```
jakirkham: I think this `exit(0) if h5py.run_tests().wasSuccessful() else exit(1)`
could be simplified to `exit(not h5py.run_tests().wasSuccessful())`.
Probably could use a relative import to make this shorter still.

andrewcollette:
Thanks!

PR #660, opened 2015-12-12
Add `with_phil` decorator to `ndim` property

jakirkham:
Follow up to https://github.com/h5py/h5py/pull/649

Not sure if this was needed, but seems to be everywhere else. So, added it here.

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -201,6 +201,7 @@ class Dataset(HLObject):
         return DimensionManager(self)
 
     @property
+    @with_phil
     def ndim(self):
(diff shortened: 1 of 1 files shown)


jakirkham:
Is this ok, @andrewcollette?

andrewcollette:
Yes, it's good practice... thanks.

jakirkham:
Ok, thanks for clarifying. What is this lock protecting against? Just other `h5py` operations?

andrewcollette:
Yes, every part of the API in h5py now uses explicit thread locking.  You can check out the comments in `_objects.pyx` for an overview.

PR #664, opened 2016-01-05
Initial version of null dataspace code

### LINKED ISSUE #279: Properly support empty attributes and datasets
HDF5 has the concept the use of an "empty" or "null" dataspace for attribute and dataset objects, having a data type but neither shape nor data.  No such construct exists in NumPy.

These objects are showing up in the real world, e.g. in netcdf4 files.  An interim solution (issue #277) raises IOError on read.

The following mechanisms have been discussed to support this HDF5 feature.

## Low level interface

Will be handled with the native HDF5 machinery.  No additional C types are needed. AttrID and DatasetID objects with H5S_NULL dataspace will have object.shape == None.

## High level interface: exception approach

It is attractive to have reads & writes to these objects fail with an exception, as obviously there's no data to handle.  However, this has the following unintended side effects:

1. If an exception is strictly raised, it breaks attrs.values(), itervalues() and items() for any object which happens to be "contaminated" by an empty attribute.
2. If empty attributes are instead eliminated from the output of values() and friends, we now have a situation where len(keys()) != len(values()).
3. If we eliminate the attributes from keys() as well, each attribute must be opened and inspected, not only in keys() but also for containership testing.
4. There is no way to get the type of an empty attribute from the high level interface.

We could create empty datasets by allowing shape=None.  No mechanism is proposed for creating empty attributes.

## High level interface: "Empty" marker approach.

A new class, h5py.Empty, could be created and used to fill in as the "value" for empty attributes and datasets.  It would have a .dtype attribute.  Reading an empty float32 attribute would return an Empty instance with obj.dtype == dtype('float32').  Empty attributes could be created by assigning an Empty instance to a name.  

Likewise, empty datasets could be created either by calling create_dataset with shape=None or by assigning data=Empty(dtype), or by assigning an Empty instance to a name in the file.

This has the following drawbacks:

1. It introduces a new type used for a very specific purpose, which is alien to both NumPy and HDF5.
2. People may become confused by slicing into a dataset and getting neither a NumPy array nor an array scalar in return.

## High level interface: Zero-length NumPy array

The same basic approach as above, only instead of an Empty instance, a NumPy array is used with shape (0,).  The array provides a .dtype attribute intrinsically and can so be used to create empty attributes or datasets by simple assignment.  There is precedent for this in the NumPy world; zero-length selections already return such an object (e.g. array[0:0]).

Assigning this object to an attribute or object name creates an object with H5S_NULL dataspace, instead of the present H5S_SIMPLE dataspace with shape (0,) (is this used by anyone?).  Datasets can still be created with shape (0,) via create_dataset.

Disadvantages:

1. Overlaps with shape (0,), which is permitted for datasets and attributes.
2. Technically has rank-1, which is mildly annoying from a consistency standpoint.
3. We would trade the ability to create shape-(0,) attributes for the ability to create H5S_NULL attributes.

## Status

We propose to implement the third solution: zero-length NumPy array.  Community feedback is welcome.

aragilar:
Closes #279.

### DIFF
--- a/h5py/_hl/base.py
+++ b/h5py/_hl/base.py
@@ -20,7 +20,7 @@ import six
 from collections import (Mapping, MutableMapping, KeysView, 
                          ValuesView, ItemsView)
 
-from .. import h5d, h5i, h5r, h5p, h5f, h5t
+from .. import h5d, h5i, h5r, h5p, h5f, h5t, h5s
 
 # The high-level interface is serialized; every public API function & method
 # is wrapped in a lock.  We re-use the low-level lock because (1) it's fast, 
@@ -78,6 +78,13 @@ def default_lcpl():
 
 dlapl = default_lapl()
 dlcpl = default_lcpl()
+
+
+def is_empty_dataspace(obj):
+    """ Check if an object's dataspace is empty """
+    if obj.get_space().get_simple_extent_type() == h5s.NULL:
+        return True
+    return False
 
 
 class CommonStateObject(object):
@@ -397,4 +404,25 @@ class MutableMappingHDF5(MappingHDF5, MutableMapping):
     """
 
     pass
-    
+
+
+class Empty(object):
+
+    """
+        Proxy object to represent empty/null dataspaces (a.k.a H5S_NULL).
+
+        This can have an associated dtype, but has no shape or data. This is not
+        the same as an array with shape (0,).
+    """
+    shape = None
+
+    def __init__(self, dtype):
+        self.dtype = dtype
+
+    def __eq__(self, other):
+        if isinstance(other, Empty) and self.dtype == other.dtype:
+            return True
+        return False
+
+    def __repr__(self):
+        return "Empty(dtype={0!r})".format(self.dtype)
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -22,7 +22,7 @@ from six.moves import xrange    # pylint: disable=redefined-builtin
 import numpy
 
 from .. import h5, h5s, h5t, h5r, h5d, h5p, h5fd
-from .base import HLObject, phil, with_phil
+from .base import HLObject, phil, with_phil, Empty, is_empty_dataspace
 from . import filters
 from . import selections as sel
 from . import selections2 as sel2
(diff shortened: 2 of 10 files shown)


h5py/_hl/base.py
```diff
@@ -397,4 +404,25 @@ class MutableMappingHDF5(MappingHDF5, MutableMapping):
...
+    def __init__(self, dtype):
+        self.dtype = dtype
+
+    def __eq__(self, other):
+        if isinstance(other, Empty) and self.dtype == other.dtype:
+            return True
+        return False
+
+    def __repr__(self):
+        return "Empty(dtype={})".format(repr(self.dtype))
```
andrewcollette: I think `{r}` can be used for this.

andrewcollette:
A couple of things come to mind:

1. #279 suggests that accessing the contents of an empty dataset would produce an Empty object... this PR simply raises an exception.  Thoughts?
2. What does Dataset.shape report for empty datasets (ideally it should be None)?
3. There will need to be a section added to the documentation describing how Empty works
4. How can I create empty datasets and attributes using Empty?  #279 suggests this should be handled by assigning to a name in the file (or a new entry in .atts), or by using shape=None when calling create_dataset.

aragilar:
Updated the code (I thought I'd written both the read and write parts, but was evidently imaging things...).

The code should act as expected, but I do have one question: should h5py.Empty have a default? It'd be easier to use. On the other hand if the default ever changes, it'd be less obvious than if the default dtype for create_dataset changed.

aragilar:
@andrewcollette Is there anything that's still missing?

h5py/_hl/dataset.py
```diff
@@ -399,6 +405,10 @@ def __getitem__(self, args):
         * Boolean "mask" array indexing
         """
         args = args if isinstance(args, tuple) else (args,)
+        if is_empty_dataspace(self.id):
+            if args != tuple():
```
andrewcollette: I think we should also accept Ellipsis.  For example, scalar datasets can't be sliced with dset[:] but can with dset[...].

andrewcollette:
@aragilar, thanks for pinging me on this; it had slipped off my radar.  This will be an excellent feature!  Just add the ability to index with Ellipsis and I'll merge it.

aragilar:
@andrewcollette Done

PR #675, opened 2016-02-02
Drop 3.2 support from tests

andrewcollette:
Python 3.2 is not supported by our automatic test suite and dependencies.  Drop from CI testing.


tacaswell:
:+1:

PR #676, opened 2016-02-09
support for local python installations

andreh7:
I have a local python installation on a RHEL6 system where I don't have sudo rights.

This patch puts the value of distutils.sysconfig.get_config_var("LIBDIR") (before /opt/local/lib and /usr/local/lib) which will give the directory where libpythonX.Y.so is installed. 

This was only added on the non-win platform code part in setup_build.py as I don't have the means to test it on Windows.

### DIFF
--- a/setup_build.py
+++ b/setup_build.py
@@ -10,6 +10,7 @@ try:
 except ImportError:
     from distutils.extension import Extension
 from distutils.command.build_ext import build_ext
+import distutils.sysconfig
 import sys
 import os
 import os.path as op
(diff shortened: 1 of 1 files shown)


andrewcollette:
Hmmm... this doesn't seem very h5py-specific.  We'd like to help but I'm not eager to add code the h5py to work around generic distutils issues.  Are there any options to `build_ext`, environment variables, etc., that might help in your case? It's also hard to predict what effect this might have on other build setups.

andreh7:
thanks for the fast reply !

In fact, I tried a few options like setting LDFLAGS before running setup.py and using the --library-dirs option to setup.py configure and build but in the end I got the impression that this is ignored because standard library paths like /usr/local/lib and /opt/local/lib for the location of libpython2.7 are hardwired into setup_build.py .

----
Trying again, it looks like installing with pip seems to work without the modification (with a locally unpacked directory as well as from the remote repository) but not with running setup.py . On the other hand, the same happens for numpy (pip working, setup.py not working with the local python installation), so I guess it's ok...

Issue #552, opened 2015-03-17
support CORE initialization from in-memory images


dnelson86:
Hi,

A feature request to support operations on in-memory HDF5 image files.
In particular, to be able to initialize the CORE driver with an in-memory buffer

```
H5Pset_file_image(fapl_id, buf, buf_len);
```

These are 1.8.9+ features.

See page 24 of: http://www.hdfgroup.org/HDF5/doc/Advanced/FileImageOperations/HDF5FileImageOperations.pdf

Thanks!

andrewcollette:
No objection from me!  A community-supplied PR would be necessary, though.

dnelson86:
Relevant:

https://groups.google.com/forum/#!topic/h5py/W25rGn4msmM

gheber:
+1 from me. I just thought about that very function the other day playing w/ PySpark.

shoyer:
I am also very interested in this feature, but I'm not familiar enough with C to write the PR myself.

Note that pytables has support for this, which might might be a helpful starting point: 
https://pytables.github.io/cookbook/inmemory_hdf5_files.html#memory-images-of-hdf5-files

That final example seems to be missing the argument `driver_core_image=image`:
https://github.com/xray/xray/issues/23#issuecomment-36185024

nevion:
I had this ready some time ago, but haven't had time to address what we wanted to do for the future:
https://github.com/h5py/h5py/pull/460

Been using it roughly every day since so it's definitely working for me.  

What we wanted to do was proper buffer handling though which required some changes to the callbacks the HDF group uses.  In my experience to date, it is quite the crap shoot getting them to accept patches and I haven't had time to even make them.

gheber:
What kind of changes would you like to see? I'd be happy to shepherd things on this end.

nevion:
@gheber I can expand more concretely if needed but basically we need to be able to control the lifetime management of a python c buffer object and have that be part of the H5LT functions/structs.  It is not sufficient to assume we can just pass the file image alone since we need to reference count on the python object to make sure the pointer the buffer/file image is working with stays valid.  Copying the image works for some people but given people using HDF5 are often using significant chunks (close to 100%) of ram, copying is not an option that works for many quite often, myself included.

[...]

PR #680, opened 2016-02-24
Minimally implement file image operations

andreabedini:
This adds low-level wrappers for two HDF5 calls:
  - `H5Fget_file_image`
  - `H5Pset_file_image`

The first can be used to retrive the image of an open file, while the second
to provide an image for a file driver to use as the initial content of the
file.

In its current implementation `H5Pset_file_image` simply copies the image obtained from Python. A more complex mechanism allows to share a file image with HDF5, avoiding copies.

Addresses #552, with my apologies to @nevion for taking a different approach :-)

@andrewcollette please give feedback, especially about the following two points:

- Extensions `h5py/h5*.pyx` seem all to be very minimal, almost entirely relying on Cython auto-generated code to call CPython APIs. In both `h5f.pyx` and `h5p.pyx` I call CPython APIs, I hope this doesn't go against a design decision.
- I didn't implement any high-level interface for this because I don't know how you would design it (and because for PyTables I only need to call the low-level API). It's still possible to call this with the low-level interface.

```
# image is a Python object supporting the buffer interface (bytes, numpy array, etc)

fapl = h5py.h5p.create(h5py.h5p.FILE_ACCESS)
fapl.set_fapl_core()
fapl.set_file_image(image)
fid = h5py.h5f.open(b'/tmp/somewhere.h5', h5py.h5f.ACC_RDONLY, fapl=fapl)
f = h5py.File(fid)
```

Thanks.

Edit: I just found a related discussion i #460. Perhaps I should read that too!


nevion:
This directly works with the H5F api, which has getters/setters for images which is fine most of the time but has inefficiencies because of the copying it uses when using large files.  Addressing this should not close #552 as that uses a different backend HDF5 includes (the memory based driver included with H5LT api) which is made and optimized for ram based usages (no actual files involved).  With that API, while you can completely avoid copies and files,  it needs some small extensions to support what boils down to external reference counts on userdata (=python objects) for real safely.  At current, the user has to maintain the python objects to keep validity or risk segfaults.  This gets awkward fast and users of SWIG will probably know what I'm talking about.

Remember, large datasets do happen in this domain of users and people do design system resources with their dataset sizes in mind.  Not addressing this breaks such designs where it shouldn't as you can easily end up with 2 or 3 copies of stuff sitting around.

So +1 for sure, but make sure to know the distinction of this issue with #552, this way isn't going to scale to large datasets and just wraps the simple builtin copy based setter/getter for file images.

andreabedini:
Thanks for the feedback @nevion! Yes, this is just a little step forward. I am aware of the overhead of unnecessary copies but I haven't found a way around just yet (I am still figuring out the API). It wasn't my intention suggesting this replace #552, far from that.

h5py/h5p.pyx
```diff
@@ -1161,6 +1165,30 @@ cdef class PropFAID(PropInstanceID):
...
+        @with_phil
+        def set_file_image(self, image):
+            """ (BUFFER image)
+            Sets an initial file image in a memory buffer.
+            """
+
+            cdef Py_buffer buf
+
+            if image is None:
+                # Release file image buffer if set
```
andrewcollette: This behavior should be documented in the docstring.

h5py/h5p.pyx
```diff
@@ -1161,6 +1165,30 @@ cdef class PropFAID(PropInstanceID):
...
+                H5Pset_file_image(self.id, NULL, 0)
+
+            if not PyObject_CheckBuffer(image):
```
andrewcollette: This seems to conflict with `image` possibly being None.

andrewcollette:
I think this is a good intermediate step on the way to #552, and should be helpful in the context of the PyTables project.  A high-level interface isn't needed right now.  I had some minor line notes.

It would also be nice if you could add a minimal test suite; anywhere under h5py.tests is fine, it's a bit of a mess in there at the moment.  An empty HDF5 file is about 800 bytes; you could simply embed a string literal in the test file.

andreabedini:
Thank you for the feedback @andrewcollette. I hope I incorporated your suggestions (perhaps I should add even more tests).

As a side note, I started working on implementing the rest of the file image API and it's a bit of a nightmare :-) but I hope I will be able make progress.

nevion:
@andrewcollette  I don't see how this is an intermediate step to #552, but it is a usable holdover most of the time.  And your comment about a high level  interface not being needed right now - well, I'm not sure exactly what you mean but  #552 really would be a tiny constructor detail to the user.  It's needed as soon as you need large files in memory (with large being relative to the machine you have to work on).   Particularly annoying on budget commodity machines many labs and institutions have as workstations.

Anywho, I think it's in mergable state already so while I'm not an h5py dev, what are you waiting for - merge it!  I hate seeing simple and useful stuff like this bitrott and miss releases... seen it too many times.

andrewcollette:
@andreabedini, thanks!  Looks great.

@nevion, the high-level interface changes in #552 would be fine as they mainly extend the existing CORE functionality.

gretahuang:
Hi, I'm new to h5py but have been trying to use the above functionality with the added step of decompressing a gzipped .h5 file, but I keep running into this error: `OSError: Unable to open file (Truncated file: eof = 8722055168, sblock->base_addr = 0, stored_eoa = 25232587836)`.

Code is as follows
```output = (read in example.h5.gz as bytes string)
decompressed_output = gzip.decompress(output)     # returns Python bytes object
fapl.set_fapl_core()
fapl.set_file_image(decompressed_output)
fid = h5py.h5f.open(b'/tmp/tmp.h5', flags=h5py.h5f.ACC_RDONLY, fapl=fapl)
store = h5py.File(fid)```

gretahuang:
Hi, I'm new to h5py but have been trying to use the above functionality with the added step of decompressing a gzipped .h5 file, but I keep running into this error: `OSError: Unable to open file (Truncated file: eof = 8722055168, sblock->base_addr = 0, stored_eoa = 25232587836)`.

Code is as follows
```output = (read in example.h5.gz as bytes string)
decompressed_output = gzip.decompress(output)     # returns Python bytes object
fapl.set_fapl_core()
fapl.set_file_image(decompressed_output)
fid = h5py.h5f.open(b'/tmp/tmp.h5', flags=h5py.h5f.ACC_RDONLY, fapl=fapl)
store = h5py.File(fid)```

PR #683, opened 2016-03-04
Fix deprecation of - for numpy boolean arrays

### LINKED ISSUE #681: h5py/_hl/dataset.py:70:DepreciationWarning
I just started using h5py recently, and just noticed that I was getting the following warning when I write NetCDF files:

h5py/_hl/dataset.py:70:DeprecationWarning: numpy boolean negative, the `-` operator, is   deprecated, use the `~` operator or the logical_not function instead.

This is the line in question:

     if isinstance(chunks, tuple) and (-numpy.array([ i>=j for i,j in zip(tmp_shape,chunks) if i is not None])).any():

I changed the `-` to a `~ `as the warning suggested and the warning has gone away:

    if isinstance(chunks, tuple) and (~numpy.array([ i>=j for i,j in zip(tmp_shape,chunks) if i is not None])).any():
 
Maybe that could be changed in the next release.

aragilar:
Fixes #681 by removing the negation of numpy boolean array. Instead replaced by a more readable version.

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -73,7 +73,9 @@ def make_new_dset(parent, shape=None, dtype=None, data=None,
 
     tmp_shape = maxshape if maxshape is not None else shape
     # Validate chunk shape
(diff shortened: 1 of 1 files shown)


andreabedini:
much more readable :+1:

Issue #690, opened 2016-03-16
HDF5 Direct Chunk Writing not supported


simongregorebner:
Currently HDF5 direct chunk writing (https://www.hdfgroup.org/HDF5/doc/Advanced/DirectChunkWrite/) is not supported by h5py.

Direct chunk writing is required to be able to write pre-compressed data directly into the HDF5 file without going through the filter-pipeline of HDF5. As modern high throughput detectors like the Dectris Eiger (https://www.dectris.com/EIGER_X_Detectors.html), and/or several other detectors at Synchrotron and FEL facilities, provide data in such a speed and volume that the HDF5 library is not able to keep up while going through its filter pipeline, this functionality is necessary.

aragilar:
Fixed by #691.

PR #691, opened 2016-03-16
Direct Chunk Write Functionality Added

simongregorebner:
This pull request adds direct chunk writing to h5py. It is fixing issue #690 

To address the issue that direct chunk writing is potentially dangerous for non experts, I did not made this functionality directly available through the dataset api but only if going through dataset.id .

Example of directly writing pre-compressed data using bitshuffle (https://github.com/kiyo-masui/bitshuffle):

```
import h5py
import numpy
import bitshuffle
import bitshuffle.h5
import struct

filename = "direct_chunk.hdf"
filehandle = h5py.File(filename, "w")

block_size = 2048
dataset = filehandle.create_dataset("data", (100, 100, 100), maxshape=(None, 100, 100), compression=32008, compression_opts=(block_size, bitshuffle.h5.H5_COMPRESS_LZ4), chunks=(1,100,100), dtype='float32')

array = numpy.random.rand(100, 100)
array = array.astype('float32')

compressed = bitshuffle.compress_lz4(array, block_size)

# Adding header information according to HDF5 plugin specification
bytes_number_of_elements = struct.pack('>q', (100*100*4))
bytes_block_size = struct.pack('>i', block_size*4)
byte_array = bytes_number_of_elements + bytes_block_size + compressed.tobytes()

index = 0
dataset.id.write_direct_chunk((index, 0, 0), byte_array)

filehandle.close()
```

### DIFF
--- a/h5py/h5d.pyx
+++ b/h5py/h5d.pyx
@@ -398,40 +398,35 @@ cdef class DatasetID(ObjectID):
 
     IF HDF5_VERSION >= (1, 8, 11):
 
-        def write_direct_chunk(self, offsets, data, filter_mask=0, PropID dxpl=None):
-            """
-            Direct chunk write to dataset
+        def write_direct_chunk(self, offsets, bytes data, H5Z_filter_t filter_mask=H5Z_FILTER_NONE, PropID dxpl=None):
+            """ no return
+
+            Writes data from a bytes array (as provided e.g. by struct.pack) directly
+            to a chunk at position specified by the offsets argument.
+
+            Feature requires: 1.8.11 HDF5
             """
 
             cdef hid_t dset_id
             cdef hid_t dxpl_id
             cdef hid_t space_id = 0
-            cdef uint32_t filters
-            cdef hsize_t *offset
+            cdef hsize_t *offset = NULL
             cdef size_t data_size
-            cdef char *buf
             cdef int rank
 
+            dset_id = self.id
+            dxpl_id = pdefault(dxpl)
+            space_id = H5Dget_space(self.id)
+            rank = H5Sget_simple_extent_ndims(space_id)
+
+            if len(offsets) != rank:
+                raise TypeError("offset length (%d) must match dataset rank (%d)" % (len(offsets), rank))
+
             try:
-                dset_id = self.id
-                dxpl_id = pdefault(dxpl)
-                filters = filter_mask
-                space_id = H5Dget_space(self.id)
-                rank = H5Sget_simple_extent_ndims(space_id)
-
-                if len(offsets) != rank:
-                    raise TypeError("offset length (%d) must match dataset rank (%d)" % (len(offsets), rank))
-
                 offset = <hsize_t*>emalloc(sizeof(hsize_t)*rank)
                 convert_tuple(offsets, offset, rank)
-                buf = data
-                data_size = len(data)
-
-                H5DOwrite_chunk(dset_id, dxpl_id, filters, offset, data_size, buf)
-            except BaseException as e:
-                print e
+                H5DOwrite_chunk(dset_id, dxpl_id, filter_mask, offset, len(data), <char *> data)
             finally:
                 efree(offset)
                 if space_id:
                     H5Sclose(space_id)
-            pass
--- a/h5py/h5d.pyx
+++ b/h5py/h5d.pyx
@@ -399,7 +399,7 @@ cdef class DatasetID(ObjectID):
     IF HDF5_VERSION >= (1, 8, 11):
 
         def write_direct_chunk(self, offsets, bytes data, H5Z_filter_t filter_mask=H5Z_FILTER_NONE, PropID dxpl=None):
-            """ no return
+            """ (offsets, bytes data, H5Z_filter_t filter_mask=H5Z_FILTER_NONE, PropID dxpl=None)
 
             Writes data from a bytes array (as provided e.g. by struct.pack) directly
             to a chunk at position specified by the offsets argument.
--- a/h5py/h5d.pyx
+++ b/h5py/h5d.pyx
@@ -395,3 +395,43 @@ cdef class DatasetID(ObjectID):
             """
             H5Drefresh(self.id)
 
+
+    IF HDF5_VERSION >= (1, 8, 11):
+
+        def write_direct_chunk(self, offsets, data, filter_mask=0, PropID dxpl=None):
+            """
+            Direct chunk write to dataset
+            """
+
+            cdef hid_t dset_id
+            cdef hid_t dxpl_id
+            cdef hid_t space_id = 0
+            cdef uint32_t filters
+            cdef hsize_t *offset
+            cdef size_t data_size
+            cdef char *buf
+            cdef int rank
+
+            try:
+                dset_id = self.id
+                dxpl_id = pdefault(dxpl)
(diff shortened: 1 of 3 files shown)


h5py/h5d.pyx
```diff
@@ -395,3 +395,43 @@ cdef class DatasetID(ObjectID):
             """
             H5Drefresh(self.id)
 
+
+    IF HDF5_VERSION >= (1, 8, 11):
+
+        def write_direct_chunk(self, offsets, data, filter_mask=0, PropID dxpl=None):
```
andrewcollette: Please use the Cython types where appropriate (for e.g. data, filter_mask).  This will save some work in the body of the function; for example, you won't have to cast to `filters`.

h5py/h5d.pyx
```diff
@@ -395,3 +395,43 @@ cdef class DatasetID(ObjectID):
...
+            """
+            Direct chunk write to dataset
```
andrewcollette: See other docstrings in the low-level interface for an example of what's expected here.

h5py/h5d.pyx
```diff
@@ -395,3 +395,43 @@ cdef class DatasetID(ObjectID):
...
+            """
+
+            cdef hid_t dset_id
+            cdef hid_t dxpl_id
+            cdef hid_t space_id = 0
+            cdef uint32_t filters
+            cdef hsize_t *offset
```
andrewcollette: All pointers should be explicitly initialized to NULL, especially if you'll be freeing them later.

h5py/h5d.pyx
```diff
@@ -395,3 +395,43 @@ cdef class DatasetID(ObjectID):
...
+                if len(offsets) != rank:
+                    raise TypeError("offset length (%d) must match dataset rank (%d)" % (len(offsets), rank))
+
+                offset = <hsize_t*>emalloc(sizeof(hsize_t)*rank)
+                convert_tuple(offsets, offset, rank)
+                buf = data
+                data_size = len(data)
+
+                H5DOwrite_chunk(dset_id, dxpl_id, filters, offset, data_size, buf)
+            except BaseException as e:
```
andrewcollette: No need to trap exceptions here; the `except` block should go.

h5py/h5d.pyx
```diff
@@ -395,3 +395,43 @@ cdef class DatasetID(ObjectID):
...
+                print e
+            finally:
+                efree(offset)
+                if space_id:
+                    H5Sclose(space_id)
+            pass
```
andrewcollette: As should this.

andrewcollette:
This will be a useful feature.  Please also add some unit tests; see the h5py/tests/old subpackage for examples of low-level tests.

andreabedini:
big :+1:, also it would be great to refactor a bit the tests for low-level api, for #680 I had no idea where to put them.

simongregorebner:
I updated the pull request based on your feedback ...

h5py/h5d.pyx
```diff
@@ -395,3 +395,38 @@ cdef class DatasetID(ObjectID):
             """
             H5Drefresh(self.id)
 
+
+    IF HDF5_VERSION >= (1, 8, 11):
+
+        def write_direct_chunk(self, offsets, bytes data, H5Z_filter_t filter_mask=H5Z_FILTER_NONE, PropID dxpl=None):
+            """ no return
```
andrewcollette: Please reproduce the full signature on the first line... see other functions in this module for reference.  This is needed as Cython produces a generic `f(*args, **kwds)` signature.

Also, it seems nitpicky but please break lines at 79 characters for PEP8 compliance.

h5py/tests/old/test_h5d_direct_chunk_write.py
```diff
@@ -0,0 +1,31 @@
...
+        filename = "direct_chunk.hdf"
+        filehandle = h5py.File(filename, "w")
+
+        dataset = filehandle.create_dataset("data", (100, 100, 100), maxshape=(None, 100, 100), chunks=(1, 100, 100), dtype='float32')
+
+        # writing
+        array = numpy.zeros((10, 100, 100))
+        for index in range(10):
+            a = numpy.random.rand(100, 100).astype('float32')
+            dataset.id.write_direct_chunk((index, 0, 0), a.tobytes(), filter_mask=1)
```
andrewcollette: Double-check this on Python 2... I couldn't find a tobytes() method on the ndarray object under Python 2.7 with NumPy 1.8.2.

h5py/tests/old/test_h5d_direct_chunk_write.py
```diff
@@ -0,0 +1,31 @@
...
+
+from .common import ut, TestCase
+
+
+@ut.skipUnless(h5py.version.hdf5_version_tuple >= (1, 8, 11), 'Direct Chunk Writing requires HDF5 >= 1.8.11')
+class TestWriteDirectChunk(TestCase):
+    def test_write_direct_chunk(self):
+
+        filename = "direct_chunk.hdf"
+        filehandle = h5py.File(filename, "w")
```
andrewcollette: There should be facilities in the test suite to create and clean up temporary files (I think in the "common" module); this will dump a file into the user's current working directory.  Other tests should provide an example.

h5py/tests/old/test_h5d_direct_chunk_write.py
```diff
@@ -0,0 +1,31 @@
...
+            dataset.id.write_direct_chunk((index, 0, 0), a.tobytes(), filter_mask=1)
+            array[index] = a
+
+        filehandle.close()
+
+        # checking
+        filehandle = h5py.File(filename, "r")
+        for i in range(10):
+            read_data = filehandle["data"][i]
+            assert(array[i] == read_data).all()
```
andrewcollette: Please use the TestCase.assertX methods as a bare "assert" will be optimized out when Python is run with the `-O` flag.

andrewcollette:
Overall looks good.  Some additional comments before we can merge.  Keep in mind the automatic pull request testing is evidently checking HDF5 < 1.8.11, so we'll have to manually make sure this works on Python 2 and 3 before merging.

h5py/h5d.pyx
simongregorebner: Sorry, I'm not 100% sure what you are looking for. The first line of the comment should state the signature of the function again? 

Something like this?
```
 """ (offsets, bytes data, H5Z_filter_t filter_mask=H5Z_FILTER_NONE, PropID dxpl=None)
...
```

andrewcollette: Here's a good example from the existing low-level interface:

https://github.com/h5py/h5py/blob/master/h5py/h5d.pyx#L84

h5py/tests/old/test_h5d_direct_chunk_write.py
leosala: tobytes() has been introduced in numpy 1.9.0 (http://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.tobytes.html). We can change it to tostring(), it will be the same (and from numpy 1.9.0 is an alias for tobytes() )

andrewcollette:
Looks good.  Thanks again for contributing this feature.

ulrikpedersen:
:+1:  Many thanks @simongregorebner for this one!

PR #700, opened 2016-04-15
Try to prevent numpy's dropping of py2.6 and py3.3 from breaking tests

aragilar:
Based on https://github.com/numpy/numpy/blob/master/doc/release/1.11.0-notes.rst, numpy is dropping support for python 2.6, 3.2 and 3.3 in 1.12. This tells tox to use earlier versions for testing.


andrewcollette:
Thanks!

PR #701, opened 2016-04-15
Fix offsets in aligned compound datatypes

### LINKED ISSUE #698: Offsets in aligned compound datatypes
It seems h5py does not correctly compute offsets in _aligned_ compound datatypes; namely, it neglects padding bytes. Failing test:
```python
data = np.ones(10, dtype=np.dtype('i4,i8', align=True))

with h5py.File('aligned.h5', 'w') as f:
    f['data'] = data

with h5py.File('aligned.h5', 'r') as f:
    h5data = np.array(f['data'])

assert np.all(h5data == data)
```

I believe I know how to fix this. Function `_c_compound()` in the `H5T` wrappers determines offsets by accumulating sub-datatype sizes. Instead, it could e.g. read offsets from the `dtype.fields` dict. Let me know if you want a PR.

jbernhard:
Fixes #698.

This turned out to be incredibly easy.  I also added some relevant tests.

Note that the precise calculation of offsets depends on the machine's architecture / memory model.  My 64-bit linux box aligns to 8 bytes, so both these test datatypes have size == 16 even though the underlying data only need 10.  A 32-bit machine should pad to 12 bytes.

### DIFF
--- a/h5py/h5t.pyx
+++ b/h5py/h5t.pyx
@@ -1414,14 +1414,13 @@ cdef TypeCompoundID _c_compound(dtype dt, int logical):
     # Initial size MUST be 1 to avoid segfaults (issue 166)
     tid = H5Tcreate(H5T_COMPOUND, 1)  
 
-    offset = 0
     for name in names:
         ename = name.encode('utf8') if isinstance(name, unicode) else name
         dt_tmp = dt.fields[name][0]
+        offset = dt.fields[name][1]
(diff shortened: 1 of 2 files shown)


andrewcollette:
Thanks!  This is a welcome fix.  The failed Travis-CI test is noise related on ongoing issues with NumPy on Python 2.6, so I'll go ahead and merge it.

PR #702, opened 2016-04-19
Base compound type size off of numpy itemsize

cmorse:
This was causing some interoperability issues with some C hdf5 code that expects the sizeof datatypes to be a multiple of 8 (The same value that numpy's itemsize uses).

### DIFF
--- a/h5py/h5t.pyx
+++ b/h5py/h5t.pyx
@@ -1411,15 +1411,19 @@ cdef TypeCompoundID _c_compound(dtype dt, int logical):
     cdef dict fields = dt.fields
     cdef tuple names = dt.names
 
-    # Initial size MUST be 1 to avoid segfaults (issue 166)
-    tid = H5Tcreate(H5T_COMPOUND, 1)  
+    # Set initial size to itemsize
+    tid = H5Tcreate(H5T_COMPOUND, dt.itemsize)
 
     for name in names:
         ename = name.encode('utf8') if isinstance(name, unicode) else name
         dt_tmp = dt.fields[name][0]
         offset = dt.fields[name][1]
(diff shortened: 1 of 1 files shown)


lnunno:
I would like to see this merged in as well.

cmorse:
Not sure why github closed this for me. @andreabedini, @andrewcollette Would appreciate some feedback on this PR.

AcidLeroy:
It would be great to have this.

andreabedini:
@cmorse first of all sorry for the delay in getting back to you. I had a look and I am not sure that problem this is supposed to fix (honest question).

Doing `tid = H5Tcreate(H5T_COMPOUND, dt.itemsize)` does seem more sensible but looking at HDF5 code, it doesn't look like the size is really used anywhere in `H5T__create` and also `H5T__insert` seems to have logic to keep track of the total size of the members.

andreabedini:
oh! I see. Calling `H5Tget_size` on the type created by h5py will return just 1, is this what you are experiencing? (HDF5 API is a bit weird here)

cmorse:
@andreabedini No worries.

The problem this pull request is trying to fix is that in the project I am working on, we are intentionally putting additional padding at the end of datatypes to allow for future expansion. The additional padding is used so that additional fields can be added in the future. When the datatype defined lacks the padding, it's size doesn't line up with what's expected elsewhere in our code, which causes problems with reading the data.

aragilar:
@cmorse I combined this and #738 in #793 (plus some additional compound datatype work), can you see if the latest master fixes your problem?

cmorse:
@aragilar Sorry for the late reply on this. A coworker has verified that #738 fixes the issue we were having.

PR #703, opened 2016-04-20
Test minimum required versions

aragilar:
It appears #701 added a test which required numpy 1.11 (which is why some tests are failing). This wraps that test in a check against numpy 1.11, and tests the minimum version of numpy and cython required (numpy 1.7 is the earliest version to support python 3.3, numpy 1.9 is the earliest version to support 3.4).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,10 @@ env:
   - TOXENV=py27-test-deps
   - TOXENV=py33-test-deps
   - TOXENV=py34-test-deps
+  - TOXENV=py26-test-mindeps
+  - TOXENV=py27-test-mindeps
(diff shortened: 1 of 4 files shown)


aragilar:
@andrewcollette Any changes need to be made before this can be merged?

andrewcollette:
@aragilar, thanks for pinging me on this... I have been running around giving software demos for the past month.  It looks good, except I am not wild about disabling the alignment test on NumPy < 1.11.  Is there an easy way to modify the test instead?  If not we can go ahead and merge it to fix the errors.

aragilar:
Done

andrewcollette:
Thanks!

PR #707, opened 2016-05-13
Fix pkgconfig handling in setup_configure autodetect_version.

asford:
`autodetect_version` in `setup_configure` is currently broken in cases requiring the use of pkgconfig to resolve the correct libhdf5 libdir. This breaks HDF5 autodetection in the default installation scheme in ubuntu/debian, causing h5py to revert back to '1.8.4' compatibility mode. This prevents the inclusion of the newly introduced `set_file_image` & `get_file_image` apis. This causes a test failure when running `setup.py test` on debian setups.

While #533 may resolve this issue, it appears that this request is currently stalled. This minor fix resolves the default case in ubuntu/debian.

### DIFF
--- a/setup_configure.py
+++ b/setup_configure.py
@@ -204,7 +204,11 @@ def autodetect_version(hdf5_dir=None):
     libdirs = ['/usr/local/lib', '/opt/local/lib']
     try:
         if pkgconfig.exists("hdf5"):
-            libdirs.append(pkgconfig.parse("hdf5")['library_dirs'])
+            hdf5_library_dirs = pkgconfig.parse("hdf5")['library_dirs']
+            if isinstance(hdf5_library_dirs, basestring):
+                libdirs.append(hdf5_library_dirs)
+            else:
+                libdirs.extend(hdf5_library_dirs)
     except EnvironmentError:
         pass
     if hdf5_dir is not None:
--- a/setup_configure.py
+++ b/setup_configure.py
@@ -1,4 +1,3 @@
-
 """
     Implements a new custom Distutils command for handling library
     configuration.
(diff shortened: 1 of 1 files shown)


andrewcollette:
Thanks!  There is a minor issue in that `basestring` is not defined for Python 3.  We can't rely on `six` being available when setup.py is run, so you might try one of the usual workarounds.  One slightly- embarrassing example I've seen is `isinstance(x, (type(b""), type(u"")))`.

ax3l:
@andrewcollette this PR is updated and helps already a lot of people.

andreabedini:
hey @ax3l I am not sure what's going on here. `pkgconfig.parse` seems to always return a dictionary of sets, see https://github.com/matze/pkgconfig/blob/master/pkgconfig/pkgconfig.py#L167. I checked this is what happens on Debian

```
>> pkgconfig.parse("hdf5")
defaultdict(<class 'set'>, {'libraries': {'hdf5'}, 'library_dirs': {'/usr/lib/x86_64-linux-gnu/hdf5/serial'}, 'include_dirs': {'/usr/include/hdf5/serial'}, 'define_macros': set()})
```

On what system do you see it returning a single string?

ax3l:
@andreabedini I am just experiencing the same problem on Ubuntu (travis) and Debian (my dev) systems. But following the signature you linked, it should simply read
```python
if pkgconfig.exists("hdf5"):
    libdirs.extend(hdf5_library_dirs['library_dirs'])
```
and nothing else.

ax3l:
Ah I see the bug. This should be the diff in the current master:
```diff
diff --git a/setup_configure.py b/setup_configure.py
index fc7fb08..c670ff1 100644
--- a/setup_configure.py
+++ b/setup_configure.py
@@ -204,7 +204,7 @@ def autodetect_version(hdf5_dir=None):
     libdirs = ['/usr/local/lib', '/opt/local/lib']
     try:
         if pkgconfig.exists("hdf5"):
-            libdirs.append(pkgconfig.parse("hdf5")['library_dirs'])
+            libdirs.extend(pkgconfig.parse("hdf5")['library_dirs'])
     except EnvironmentError:
         pass
     if hdf5_dir is not None
```

append is adding the `list` as an object inside the dictionary key `'library_dirs'`. But of course, it should *extend* the list by adding each element in the iterable. will open a quick PR.

andreabedini:
Superseded by #734, closing.

PR #708, opened 2016-05-15
updating docs/quick.rst file

infinite-Joy:
The place showing the value for dset[10] seems to be wrong
Have corrected it.
Great documentation btw.


tacaswell:
:+1:

andrewcollette:
Good catch... thanks!

PR #716, opened 2016-06-04
Add support for pathlib/PEP519

aragilar:
Currently h5py only support strings as paths, this add support for pathlib/PEP519 and simplifies some path related code.

### DIFF
--- a/h5py/_hl/compat.py
+++ b/h5py/_hl/compat.py
@@ -1,0 +1,79 @@
+"""
+Compatibility module for high-level h5py
+"""
+import sys
+import six
+
+
+def fspath(path):
+    """
+    Return the string representation of the path.
+    If str or bytes is passed in, it is returned unchanged.
+    This code comes from PEP 519, modified to support earlier versions of
+    python.
+
+    This is required for python < 3.6.
+    """
+    if isinstance(path, (six.text_type, six.binary_type)):
+        return path
+
+    # Work from the object's type to match method resolution of other magic
+    # methods.
+    path_type = type(path)
+    try:
+        return path_type.__fspath__(path)
+    except AttributeError:
+        if hasattr(path_type, '__fspath__'):
+            raise
+        try:
+            import pathlib
+        except ImportError:
+            pass
+        else:
+            if isinstance(path, pathlib.PurePath):
+                return six.text_type(path)
+
+        raise TypeError("expected str, bytes or os.PathLike object, not "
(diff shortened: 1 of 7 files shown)


aragilar:
Uh, broken by https://github.com/numpy/numpy/issues/7498, it's probably easier to just not have python 2.6 tested with numpy 1.11...

aragilar:
I'll rebase this once #733 get merged.

h5py/_hl/files.py
```diff
@@ -16,6 +16,19 @@
 import sys
 import os
 
+try:
```
andreabedini: is there a way to hide this somewhere? perhaps in `compat.py` itself? also it's duplicated in `group.py`.

h5py/_hl/compat.py
```diff
@@ -0,0 +1,79 @@
+"""
+Compatibility module for high-level h5py
+"""
+import sys
+import six
+
+
+def fspath(path):
```
andreabedini: see my other comment in `files.py`. perhaps we can do: if python version < 3.6 `def fspath ...` else `from os import fspath`? analogously for fsdecode an fsencode (for python < 3.2)? then we just do a single `from .compat import fspath`. hope it make sense

andreabedini:
lgtm, I left a couple of comments on how to organise the compat module.

aragilar:
The reason behind the try/except rather than if is that PEP 519 is kinda vague where it's getting implemented, as pathlib is a provisional module (so changes get backported). Moving the try/except into compat makes sense, I'll do that after I rebase.

andreabedini:
@aragilar I merged #733. Yeah I was just wondering if there's a way of neatly hiding the module shim in the `compat.py` module. Actually perhaps it's better to include `pathlib` in the module as a reference of what the module is shimming. Anyway none of this is a big deal, your contribution is precious and I don't want to waste it on these trivialities :) Just let me know when you think you have done.

aragilar:
It's only replacing os, not pathlib, and pathlib has a existing backport. Also, there are other path handling libraries (e.g. py.path used in pytest), which are supported automatically by using PEP 519. We could use pathlib internally, and depend on pathlib, but that doesn't get us much.

andreabedini:
great work @aragilar. looks good to me, @andrewcollette merge ? :)

andrewcollette:
Looks great to me.  Thanks again for putting the time in to do this!

PR #718, opened 2016-06-12
Add split file driver

bplotnick:
This pull request adds support for the "split" file driver, which writes to a metadata file and a data file.


andreabedini:
Thanks @bplotnick, sorry things are slow ... :)

h5py/_hl/files.py
```diff
@@ -60,6 +60,19 @@ def make_fapl(driver, libver, **kwds):
...
+    elif driver == 'split':
+        if 'metafile' in kwds:
+            kwds['metafile'] = os.path.join(kwds['metafile'],r'%s')
+        kwds['metafile'] = kwds.get('metafile','')+'-m.h5'
+        if 'rawfile' in kwds:
+            kwds['rawfile'] = os.path.join(kwds['rawfile'],r'%s')
+        kwds['rawfile'] = kwds.get('rawfile','')+'-r.h5'
+        try:
+           kwds['rawfile'] = kwds['rawfile'].encode(sys.getfilesystemencoding())
+           kwds['metafile'] = kwds['metafile'].encode(sys.getfilesystemencoding())
```
aragilar: Rather than using `.encode`, can you use `encode_filename` from compat.py. It avoids a number of other problems.

aragilar:
Can you add some tests to make sure everything works. Some things to test would be how different types are handled (e.g. bytes, strings, paths), and that defaults work correctly?

takluyver:
Dropping the milestone because this hasn't moved in a couple of years, but if someone does want to pick it up again, we can always reassess it for a future release.

aragilar:
#1468 appears to have merged this, closing.

PR #721, opened 2016-06-17
Close #542 Introduce `HDF5_MPI`

### LINKED ISSUE #542: Environment Flag for MPI Setup
Since `pip` does [not (yet) support a `--build-option=`](https://github.com/pypa/pip/issues/18) flag it would be nice if `--mpi` could be controlled [via an environment flag, too](http://docs.h5py.org/en/latest/build.html#custom-installation) - currently: *(none)*.

ax3l:
Introduces an environment control variable `HDF5_MPI` to complement the `--mpi` `configure` option as described in #542.

Implemented the same way as the other HDF5 controls to allow easier installs via pip where direct access to `configure` is not possible.

A travis-ci test was added to keep the MPI features stable in the CI.

### DIFF
--- a/docs/build.rst
+++ b/docs/build.rst
@@ -77,6 +77,7 @@ when installing via ``pip``, as you don't have direct access to setup.py::
 
     $ HDF5_DIR=/path/to/hdf5 pip install h5py
     $ HDF5_VERSION=X.Y.Z pip install h5py
+    $ CC="mpicc" HDF5_PARALLEL="ON" HDF5_DIR=/path/to/parallel-hdf5 pip install h5py
     
 Here's a list of all the configure options currently supported:
 
@@ -85,7 +86,7 @@ Option                  Via setup.py                Via environment variable
 ======================= =========================== ===========================
 Custom path to HDF5     ``--hdf5=/path/to/hdf5``    ``HDF5_DIR=/path/to/hdf5``
 Force HDF5 version      ``--hdf5-version=X.Y.Z``    ``HDF5_VERSION=X.Y.Z``
-Enable MPI mode         ``--mpi``                   (none)
+Enable MPI mode         ``--mpi``                   ``HDF5_PARALLEL=ON``
 ======================= =========================== ===========================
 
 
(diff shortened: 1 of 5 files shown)


ax3l:
possible collision with #533 (which is a more intrusive change and stalled since several months...)

setup_configure.py
```diff
@@ -58,6 +58,7 @@ class EnvironmentOptions(object):
     def __init__(self):
         self.hdf5 = os.environ.get('HDF5_DIR')
         self.hdf5_version = os.environ.get('HDF5_VERSION')
+        self.mpi = os.environ.get('HDF5_PARALLEL') == "ON"
```
tacaswell: Could this be done in a 'truthy' way that does not privilege the string "ON"?  I would expect `HDF5_PARALLEL=1` to also work.

docs/build.rst
```diff
@@ -85,7 +86,7 @@ Option                  Via setup.py                Via environment variable
 ======================= =========================== ===========================
 Custom path to HDF5     ``--hdf5=/path/to/hdf5``    ``HDF5_DIR=/path/to/hdf5``
 Force HDF5 version      ``--hdf5-version=X.Y.Z``    ``HDF5_VERSION=X.Y.Z``
-Enable MPI mode         ``--mpi``                   (none)
+Enable MPI mode         ``--mpi``                   ``HDF5_PARALLEL=ON``
```
tacaswell: the env `HDF5_MPI` fits the pattern a bit better (and is a bit more guessable).

setup_configure.py
ax3l: I thought about this first, too.
Unfortunately one would then expect that =FALSE,0,OFF should work too and I was not sure where to end...

Just "setting" a var in bash to smth. is therefore no great option.

docs/build.rst
ax3l: Actually "parallel" is correct in hdf5 notation (see hdf5 configure and h5p namings). The --mpi option is actually not precise enough when talking about mpi io based file operations (one can e.g. also do mpi based, independent posix operations to individual files).

ax3l: updated the PR: renamed for consistency with `--mpi`

setup_configure.py
ax3l: found [this way](http://stackoverflow.com/questions/715417/converting-from-a-string-to-boolean-in-python/715455#715455) but would depend on a pre-defined list of "what shall be true" strings.

do you suggest
- s.toLower() in ["on", "1", "true"] as `True`
- everything else as false

?

setup_configure.py
```diff
@@ -58,6 +58,7 @@ class EnvironmentOptions(object):
     def __init__(self):
         self.hdf5 = os.environ.get('HDF5_DIR')
         self.hdf5_version = os.environ.get('HDF5_VERSION')
+        self.mpi = os.environ.get('HDF5_MPI') == "ON"
```
ax3l: @tacaswell 

found [this way](http://stackoverflow.com/questions/715417/converting-from-a-string-to-boolean-in-python/715455#715455) but would depend on a pre-defined list of "what shall be true" strings.

do you suggest
- s.toLower() in ["on", "1", "true"] as `True`
- everything else as false

?

tacaswell: For mpl we use every value as `True`, and the env not being defined as `False`, but we may not be following conventions well enough.

For simplicity I agree just one 'true' value is better than a mid-sized set.

ax3l: thank you for the feedback.

yes, I explicitly wanted an way to disable it again. e.g. right now the `--mpi` flag can only be deactivated again by removing the `h5config.pkl` pickle file which is a bit odd.

> For simplicity I agree just one 'true' value is better than a mid-sized set. 

ok then we leave it like it is right now. "ON" is true, everything else is false.

ax3l:
@tacaswell I updated the PR with your suggestions :)

tacaswell:
Seems reasonable to merge to me.

attn @andrewcollette

ax3l:
rebased against latest updates in `master`

ax3l:
ping @ andreabedini :)

andreabedini:
great work @ax3l

ax3l:
thank you! :)

PR #724, opened 2016-06-23
TST: adapt test to run as root user

matthew-brett:
When root user opens a read-only file, default mode is 'r+' rather than
'r', as expected by the current tests.

See: https://github.com/h5py/h5py/issues/696


andrewcollette:
Thanks!  This has been an annoyance for a while now.

PR #733, opened 2016-07-12
Don't try to test with numpy 1.11.X on py26, it keeps breaking

aragilar:
As per https://github.com/numpy/numpy/issues/7498, numpy 1.11.0 and 1.11.1 are broken on python 2.6. Given that there's probably no easy way for numpy to fix this, not testing 1.11.X on python 2.6 is probably the best solution, in that if a user has a working 1.11.X numpy on python 2.6, then h5py will probably work fine (given we do test 1.11.X with other versions, so numpy changes should be picked up).

@andrewcollette Would be able to merge this quickly, so pull requests pass/fail correctly again?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,11 @@ matrix:
     - python: 3.5
       env:
       - TOXENV=py35-test-deps
+matrix:
+  include:
+    - python: 3.5
+      env:
+      - TOXENV=py35-test-mindeps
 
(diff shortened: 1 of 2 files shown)


andreabedini:
:+1: are you ok for me to merge it @andrewcollette ?

andrewcollette:
Yes please do!

PR #734, opened 2016-07-13
Package Config: Fix Detection

ax3l:
Fix the detection of `pkgconfig` to avoid fallbacks to HDF `1.8.4` or non-detection on default installs on Ubuntu/Debian based systems.

This is a fixed version of PR #707 by @asford with the feedback provided by @andreabedini.

The main problem is the mismatch of `list` [append vs. extend](http://stackoverflow.com/questions/252703/append-vs-extend):
`pkgconfig.parse("hdf5")` returns a dictionary of sets of which we take the set of `library_dirs` which should be extending the libdirs list element-wise (and *not* "adding a `set` object to a `list`" via `append`).

(note: it's actually done correctly in `setup_build.py` already.)

Tested on Debian 8.5 "jessie" with default hdf5 libraries (`1.8.13`) installed.

### DIFF
--- a/setup_configure.py
+++ b/setup_configure.py
@@ -204,7 +204,7 @@ def autodetect_version(hdf5_dir=None):
     libdirs = ['/usr/local/lib', '/opt/local/lib']
     try:
         if pkgconfig.exists("hdf5"):
-            libdirs.append(pkgconfig.parse("hdf5")['library_dirs'])
+            libdirs.extend(pkgconfig.parse("hdf5")['library_dirs'])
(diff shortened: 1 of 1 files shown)


andreabedini:
LGTM, @andrewcollette merge? (and let me know if you're ok with me going ahead without asking first)

andrewcollette:
@andreabedini, I have very little time these days to keep up with PRs, so please merge this (and others) at your discretion.  I'm always happy to chime in if asked for advice, though!

Issue #8, opened 2012-12-27
Advanced indexing


andrewcollette:
_Original author: bioinformed@gmail.com (November 19, 2008 16:34:35)_

Is it possible to enable advanced indexing to extract an array of row 
indexes or a boolean array?  There seems to be support for extracting 
slices and coordinates, but not arbitrary indexes.

_Original issue: http://code.google.com/p/h5py/issues/detail?id=8_

johntyree:
What do you mean by duplicate coordinates here?

The docs describe it as ["duplicate selections"](http://docs.h5py.org/en/latest/high/dataset.html#fancy-indexing), but either way I'm not sure what it's referring to. Not repeated indices in a fancy indexing selection, it seems.

```python
>>> f['test'] = np.arange(5)
>>> f['test'][[0,0]]        
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper (/home/vagrant/pisi/tmp/h5py-2.5.0-3/work/h5py-2.5.0/h5py/_objects.c:2458)
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper (/home/vagrant/pisi/tmp/h5py-2.5.0-3/work/h5py-2.5.0/h5py/_objects.c:2415)
  File "/home/john/projects/genia/genia-dev/lib/python2.7/site-packages/h5py/_hl/dataset.py", line 451, in __getitem__
    self.id.read(mspace, fspace, arr, mtype)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper (/home/vagrant/pisi/tmp/h5py-2.5.0-3/work/h5py-2.5.0/h5py/_objects.c:2458)
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper (/home/vagrant/pisi/tmp/h5py-2.5.0-3/work/h5py-2.5.0/h5py/_objects.c:2415)
  File "h5py/h5d.pyx", line 177, in h5py.h5d.DatasetID.read (/home/vagrant/pisi/tmp/h5py-2.5.0-3/work/h5py-2.5.0/h5py/h5d.c:3058)
  File "h5py/_proxy.pyx", line 130, in h5py._proxy.dset_rw (/home/vagrant/pisi/tmp/h5py-2.5.0-3/work/h5py-2.5.0/h5py/_proxy.c:1670)
[...]

PR #739, opened 2016-07-18
Check for duplicates in fancy index validation

qxcv:
Currently, a duplicate selection like `dataset[[1, 2, 2, 3]]` yields a cryptic internal error (see discussion in #8). This PR makes h5py raise an error at validation time when duplicate indices are passed instead of waiting until the underlying library chokes on the indices.

Note that I chose `TypeError` instead of `IndexError` for consistency with the existing implementation. `IndexError` would make more sense, but changing from `TypeError` would probably break a lot of existing user code.

### DIFF
--- a/h5py/_hl/selections.py
+++ b/h5py/_hl/selections.py
@@ -353,7 +353,9 @@ class FancySelection(Selection):
                 except TypeError:
                     pass
                 else:
-                    if sorted(arg) != list(arg):
+                    list_arg = list(arg)
+                    adjacent = zip(list_arg[:-1], list_arg[1:])
+                    if any(fst >= snd for fst, snd in adjacent):
                         raise TypeError("Indexing elements must be in increasing order")
 
         if len(sequenceargs) > 1:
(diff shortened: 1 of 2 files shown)


tacaswell:
This may have conflicts with #778 

I think this one should go in first.

laliberte:
I agreed. This seems higher up the list.

laliberte:
I've pushed an update to #778 and checked for potential conflicts with current PR and there seems to be none.

tacaswell:
@laliberte Moving the discussion to here from the other PR, I am not sure how to remove the limitation as in https://github.com/h5py/h5py/issues/8#issuecomment-142655958 @andrewcollette says there needs to a 1:1 mapping.  Even if you can get around this limitation, than should be put off till after 2.7, the question now is should we merge _this_ PR as it stands?

laliberte:
Yes!

I've just checked and this limitation can be substantially narrowed (but not eliminated). My implementation has no conflict with this PR and therefore it should be merged.

laliberte:
Also, I think we should definitely postpone opening vector indexing of #778 until everything is sure to work. It opens up a lot of new edge cases.
With all this said, +1 for merge.

PR #749, opened 2016-08-11
ENH: add long double datatype to lookup

tacaswell:


cgohlke:
`npy_complex256` is not available on all platforms. This PR breaks msvc builds: `error C2065: 'npy_complex256' : undeclared identifier`

andreabedini:
we should test stuff on appveyor :-/ @cgohlke do you know on what platform is broken?

cgohlke:
It seems `npy_complex256` is only available in numpy if either NPY_BITSOF_FLOAT, NPY_BITSOF_DOUBLE, or [`NPY_BITSOF_LONGDOUBLE == 128`](https://github.com/numpy/numpy/blob/master/numpy/core/include/numpy/npy_common.h#L1035).

Numpy builds for the official CPython distributions on Windows are using `NPY_BITSOF_LONGDOUBLE == NPY_BITSOF_DOUBLE == 64` to be compatible with the MS C runtime. 

Other platforms might use `NPY_BITSOF_LONGDOUBLE == 96` or `256`.

tacaswell:
Sorry @cgohlke I seem to cause you a fair amount of work :disappointed:

andreabedini:
This is a tricky issue because _long double_ may map to a bit anything. I fought with this a long writing https://github.com/PyTables/PyTables/pull/214. I'd suggest we remove all mentions of _(double|complex)256_ an we relay only on _long(dobule)_. This is what both HDF5 and numpy do. (i.e. numpy has `NPY_CLONGDOUBLE` and `NPY_COMPLEX128` but no `NPY_COMPLEX256` as it wouldn't exist on all platforms). If everybody agrees, I'll try to find the time to fix this.

PR #752, opened 2016-08-27
Make fapl kwds

andreabedini:
This is helpful to the development of PyTables 4 where we risk passing to the backend keywords that are meant for us.


andreabedini:
Tests pass, so I guess I'll merge ...

PR #754, opened 2016-09-08
Add pypy possible support by avoiding potential race condition

### LINKED ISSUE #720: Making h5py PyPy compatible
To make h5py more PyPy compatible the following bugreport to PyPy will be of interest.

https://bitbucket.org/pypy/pypy/issues/2328/osx-crash-in-h5py-while-using-brainstorm

aragilar:
This probably fixes #720 (I don't have pypy around and I'm not sure what I need to do to make it numpy-compatible, so the tests aren't working at the moment).
Also, I was running into some kind of race condition on cpython 3.5 due to problems in __dealloc__ which seemed to be identical to the problems on pypy, so this isn't just a pypy fix.
I've added pypy to travis, but it's not required to pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,6 +49,11 @@ matrix:
             - openmpi-bin         # 1.4.3
             - libopenmpi-dev
             - libhdf5-openmpi-dev # 1.8.4
+    - python: pypy
+      env:
+      - TOXENV=pypy-test-deps
+  allow_failures:
+    - python: pypy
 
 install:
   - pip install tox
--- a/h5py/_objects.pxd
+++ b/h5py/_objects.pxd
@@ -15,9 +15,11 @@ cdef class ObjectID:
     cdef readonly hid_t id
     cdef public int locked              # Cannot be closed, explicitly or auto
     cdef object _hash
+    cdef unsigned long _pyid
 
 # Convenience functions
 cdef hid_t pdefault(ObjectID pid)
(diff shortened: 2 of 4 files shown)


andreabedini:
Thanks @aragilar, this looks good to me. I want to run some tests before merging it.

andreabedini:
Looks good to me, and Travis tests all pass (pypy gets stuffed pip installing numpy). I will go ahead and merge. @cgohlke open a bug if it breaks for you.

cgohlke:
This does not look safe to use on 64 bit CPython for Windows. The 64 bit memory address of a Python object is stored in an 32 bit `unsigned long`: [`self._pyid = id(self)`](https://github.com/h5py/h5py/blob/master/h5py/_objects.pyx#L169).

andreabedini:
@cgohlke thanks, I'm trying to figure out the right type for `_pyid`. For CPython it would be essentially a pointer (`void *`)? but I have no idea if that would that work for pypy too?

aragilar:
Would Py_ssize_t work, or is there a way of getting cython to convert c99 sized types to the correct type?

cgohlke:
I think the unsigned `Py_uintptr_t` might be better.

amoose136:
I'm glad other people have noticed this. Currently h5py compatibility with pypy is the only thing stopping me from running my code with pypy instead of the default interpreter.

hyperkang:
Does h5py work with pypy now?

PR #758, opened 2016-10-02
Initial support for appveyor

aragilar:
This adds some initial support for testing h5py on appveyor (I've squashed the commits so we don't have all my attempts making a really messy history). I haven't got it working, so maybe someone who knows windows/appveyor better than me will be able to identify the problem.

Currently:
- HDF5 builds from source (using cmake, 1.8.17 seems to be the first version where this is properly supported)
- h5py kinda finds HDF5 (it knows that 1.8.17 is installed, but fails to find the header files)
- Support for tox is done (assuming that we can get h5py working)
- h5py wheels are generated (but don't include HDF5, I don't know how linking works on windows)
- There's some caching going on, currently just wheels via pip, but we should also cache HDF5 (especially as appveyor is much slower than travis).

UPDATE:
h5py now finds HDF5, but there's symbols missing, I'm not sure what's going on there...

UPDATE:
h5py now installs, but tests are failing

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,78 @@
+# from https://packaging.python.org/en/latest/appveyor/
+environment:
+  HDF5_VERSION: "1.8.17"
+  HDF5_DIR: "C:\\hdf5-%HDF5_VERSION%"
+  TOX_TESTENV_PASSENV: "HDF5_DIR"
+
+  matrix:
+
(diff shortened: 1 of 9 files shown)


aragilar:
I'm going to close this for now, as this still needs a fair bit of work, and the branch is a mess.

PR #738, opened 2016-07-15
Correctly compute size of compound dtype for non-monotonically increasing offsets

smutch:
Hopefully fixes #737.

If a dtype contains offsets which are not monotonically increasing then the resulting size computed in `h5t._c_compound` is incorrect.  This PR attempts to fix this issue and add a relevant test case.

PR #765, opened 2016-10-19
construct dtype using types sorted by offset for compound dtypes. Fixes #764

### LINKED ISSUE #764: Garbage returned from reading hdf5 vlen dataset
It seems that there has been a regression in the handling of vlen datasets. This can be easily reproduced using an HDF5 file produced by an old version of the FLASH simulation code. The easiest way to get it is to use the yt curldrop:

```
curl -JO http://use.yt/upload/1106d6c4
```

This will create a file named `DensTurbMag_hdf5_plt_cnt_0015` in the current directory.

To reproduce the issue, run the following test script:

```
import h5py
f = h5py.File('DensTurbMag_hdf5_plt_cnt_0015')
print(f['simulation parameters'][:])
```

Under h5py 2.6.0 this prints:

```
array([ (73, 4.6683907401e-313, 1.69759663317e-313, -4e-323, -759169024, 1113971554, -402653184, 1108822134)],
      dtype={'names':['total blocks','time','timestep','redshift','number of steps','nxb','nyb','nzb'], 'formats':['<i4','<f8','<f8','<f8','<i4','<i4','<i4','<i4'], 'offsets':[0,24,32,40,4,8,12,16], 'itemsize':48})
```

Whereas under h5py 2.5.0 it prints:

```
array([(73, 751000000000.0, 25000000000.0, 0.0, 22, 8, 8, 8)],
      dtype=[('total blocks', '<i4'), ('time', '<f8'), ('timestep', '<f8'), ('redshift', '<f8'), ('number of steps', '<i4'), ('nxb', '<i4'), ('nyb', '<i4'), ('nzb', '<i4')])
```

This was originally reported by a user on the yt mailing list. See https://bitbucket.org/yt_analysis/yt/issues/1291/simulation-parameters-in-flash-25-datasets.

ngoldbaum:
This fixes the issue reported in #764. Thanks to @nevion for the suggested solution.

### DIFF
--- a/h5py/h5t.pyx
+++ b/h5py/h5t.pyx
@@ -1111,7 +1111,7 @@ cdef class TypeCompoundID(TypeCompositeID):
                 ordered_fields = sorted(
                     collated_fields, key=operator.itemgetter(2))
                 field_names, field_types, field_offsets = \
-                    [list(t) for t in zip(*ordered_fields)]
+                    map(list, zip(*ordered_fields))
             typeobj = dtype({
                 'names': field_names,
                 'formats': field_types,
--- a/h5py/h5t.pyx
+++ b/h5py/h5t.pyx
@@ -26,6 +26,7 @@ from utils cimport  emalloc, efree, \
 
 # Runtime imports
 import sys
+import operator
 from h5 import get_config
 import numpy as np
 from ._objects import phil, with_phil
@@ -1006,7 +1007,6 @@ cdef class TypeCompositeID(TypeID):
         """
         return H5Tget_member_index(self.id, name)
 
-
 cdef class TypeCompoundID(TypeCompositeID):
 
     """
@@ -1106,7 +1106,17 @@ cdef class TypeCompoundID(TypeCompositeID):
         else:
             if sys.version[0] == '3':
                 field_names = [x.decode('utf8') for x in field_names]
-            typeobj = dtype({'names': field_names, 'formats': field_types, 'offsets': field_offsets})
+            if len(field_names) > 0:
+                collated_fields = zip(field_names, field_types, field_offsets)
+                ordered_fields = sorted(
+                    collated_fields, key=operator.itemgetter(2))
+                field_names, field_types, field_offsets = \
+                    [list(t) for t in zip(*ordered_fields)]
+            typeobj = dtype({
+                'names': field_names,
+                'formats': field_types,
+                'offsets': field_offsets
+            })
 
         return typeobj
 


h5py/h5t.pyx
```diff
@@ -1006,6 +1006,8 @@ cdef class TypeCompositeID(TypeID):
         """
         return H5Tget_member_index(self.id, name)
 
+def get_third(iterable):
```
ngoldbaum: Dunno if there's a better place to put this. Unfortunately it's needed because cython 0.19 doesn't support lambdas inside cdef function definitions.

pitrou: Perhaps use `key=operator.itemgetter(2)`?

ngoldbaum: Thank you for the suggestion, I didn't know that was in the `operator` module.

nevion: I figured it didn't save or help anything and perhaps even made it slower, so I didn't mention this previously.  Not like I benchmarked it tho.

ngoldbaum:
@andreabedini I see you've been merging pull requests recently. Is there any chance I can get you to review this pull request?

h5py/h5t.pyx
```diff
@@ -1106,7 +1106,17 @@ cdef class TypeCompoundID(TypeCompositeID):
         else:
             if sys.version[0] == '3':
                 field_names = [x.decode('utf8') for x in field_names]
-            typeobj = dtype({'names': field_names, 'formats': field_types, 'offsets': field_offsets})
+            if len(field_names) > 0:
+                collated_fields = zip(field_names, field_types, field_offsets)
+                ordered_fields = sorted(
+                    collated_fields, key=operator.itemgetter(2))
+                field_names, field_types, field_offsets = \
+                    [list(t) for t in zip(*ordered_fields)]
```
andreabedini: isn't this just `field_names, field_types, field_offsets = zip(*ordered_fields)` ?

ngoldbaum:
I'm actually not sure whether #738 is the same issue. I tried adding the test from that PR to this one and this PR does not fix that test. I think it might be the case that something wrong is happening in two different parts of the code.

What this *does* do is restore the ability to correctly read one of the datasets in the HDF5 file I link to in #764, as h5py 2.5.0 and older was able to do. I wasn't able to come up with a test for this functionality besides reading that one particular dataset, otherwise I would have added a unit test here.

h5py/h5t.pyx
ngoldbaum: Almost, it would even work in Python I think, but not in Cython. Since `field_names`, `field_types`, and `field_offsets` are declared to be `list`, not `tuple`, this assignment will fail at runtime unless we explicitly convert the result of `zip(*ordered_fields)` to `list`. Alternatively we could have made them tuples instead of lists. I didn't end up doing that to avoid invasively changing code elsewhere.

ngoldbaum:
I believe I also checked and #738 doesn't read in the data correctly from my file either.

nevion:
If binding code, and numpy handle out of order offsets correctly, everything would work fine but #738 finds one spot where it didn't, perhaps there are others either inside H5py bindings or even numpy itself.   I somewhat expected this patch unnecessary after reading #738, but the code here guards/worksarounds problems being in potentially more than one place, as @ngoldbaum kinda surmises.

I think the test cases would just be creating an out of order compound type using the lower level H5T api (which h5py exposes), then converting to a dtype, and examining the dtype and whether or not the size is the true size of the compound type it was constructed from, as well as all fields being the same offset.

all for this patch being axed if we get the compound type construction/mapping bulletproof and correct.

h5py/h5t.pyx
andreabedini: ah ok, so it's a Cython thing. In Python `a, b, c = rhs` only requires `rhs` to be iterable. If `field_names, field_types, field_offsets = list(zip(*ordered_fields))` works in Cython I'd still prefer that.

ngoldbaum: That won't work, but `map(list, zip(*ordered_fields))` will. I'll replace it with the more concise version...

andreabedini:
@ngoldbaum @nevion how about this plan. We merge this PR because it fixes a real problem (or at least part of it). Then we continue the discussion in an umbrella issue for out of order compound types. Actually I might ask @ngoldbaum to create the new issue, since I am sure he's done some fresh research on the subject. (I have been away from h5py for a while and can't guarantee any time to do real work on it) CC our godfather @andrewcollette ;-)

PR #773, opened 2016-11-08
Use size_t to store Python object id

cgohlke:
`unsigned long` is not large enough to store the address of a Python object on Windows 64-bit.

See also https://github.com/h5py/h5py/blob/master/h5py/_objects.pyx#L169 and https://docs.python.org/3/library/functions.html#id


andreabedini:
Thanks @cgohlke

Issue #584, opened 2015-06-10
bug when indexing with numpy integer array of size 1


SuperElectric:
Indexing a multidimensional Dataset with an numpy array of length 1 is treated the same as indexing with the integer it contains. This is inconsistent with numpy.ndarray's behavior, and with the behavior of Dataset when indexing it with a list or tuple of integers.

I am using h5py 2.2.1, the version used by Ubuntu 14.04 (the current long-term support version of Ubuntu).

Example, where `dset` is a Dataset of shape `(3, 4, 5, 6)`:

```
>> dset[[1], ...].shape  # indexing with a size-1 list produces expected shape
>> (1, 4, 5, 6)
>> dset[numpy.arange(1), ...].shape  # indexing with a size-1 ndarray does not
>> (4, 5, 6)
>> dset[numpy.arange(2), ...].shape  # ndarrays are ok as long as their size > 1
>> (2, 4, 5, 6)
```

jakirkham:
Also ran into this surprising bug recently. Would love to resolve this somehow.

jakirkham:
Looking at the code, I think the issue has to do with selection. In particular, this [line]( https://github.com/h5py/h5py/blob/2.6.0/h5py/_hl/selections.py#L85 ) causes the problem. This can be demonstrated simply as follows.

```python
In [1]: import numpy

In [2]: int([1, 2])
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-18ff4f9b52da> in <module>()
----> 1 int([1, 2])

TypeError: int() argument must be a string or a number, not 'list'

In [3]: int([1])
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-7c9c44330c2c> in <module>()
----> 1 int([1])

TypeError: int() argument must be a string or a number, not 'list'

In [4]: int(numpy.array([1, 2]))
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-4-9da7ac2589b6> in <module>()
----> 1 int(numpy.array([1, 2]))

TypeError: only length-1 arrays can be converted to Python scalars

In [5]: int(numpy.array([1]))
Out[5]: 1
```

The last case avoids the `FancySelection` case (potentially relying on this special case behavior). As a result, it goes through a code path that allows this entry to be treated as a simple integer index.

[...]

PR #774, opened 2016-11-09
WIP: Workaround NumPy index array bug

jakirkham:
Fixes https://github.com/h5py/h5py/issues/584

Fixes a deviation from NumPy behavior when using NumPy index arrays. Namely indexing by a singleton NumPy index arrays result in the index dimension being dropped when it should remain as a singleton dimension. We add a workaround and some tests to ensure this behavior is fixed.


jakirkham:
Apparently the bug is the lack of error message for this case. Going to go ahead and close.

Issue #493, opened 2014-10-22
More precise exceptions for failed openings


anntzer:
In Python 3.3+, it would be nice if h5py.File passed through the OSError subclass raised when failing to open file, e.g. FileNotFoundError when the file doesn't exist, PermissionError in we don't have the rights, etc., rather than just an OSError.

PR #780, opened 2016-11-24
fix append mode file create exception swallowing

coderfi:
For some reason, on my MacOSX, I get an error when trying to `h5py.File(..., mode='a')`.
The error seemed non-sensical, until I dug into the code and saw that the root cause error was getting swallowed.
This patch reraises the original exception on the append mode create.

FYI, this is the error I get (I haven't yet figured out why I get this, but that is a different problem).

IOError: Unable to open file (Unable to flock file, errno = 35, error message = 'resource temporarily unavailable')

### DIFF
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -112,7 +112,12 @@ def make_fid(name, mode, userblock_size, fapl, fcpl=None, swmr=False):
         try:
             fid = h5f.open(name, h5f.ACC_RDWR, fapl=fapl)
         except IOError:
-            fid = h5f.create(name, h5f.ACC_EXCL, fapl=fapl, fcpl=fcpl)
+            exc_info = sys.exc_info()  # save the root cause
+            try:
+                fid = h5f.create(name, h5f.ACC_EXCL, fapl=fapl, fcpl=fcpl)
+            except:
+                # reraise the original root cause
+                raise exc_info[0], exc_info[1], exc_info[2]
     elif mode is None:
         # Try to open in append mode (read/write).
         # If that fails, try readonly, and finally create a new file only


aragilar:
@coderfi Can you post the error you first encountered?
Do you know which version of HDF5 you're using, https://lists.hdfgroup.org/pipermail/hdf-forum_lists.hdfgroup.org/2016-May/009480.html suggests this may be something to do with SWMR (Single Writer Multiple Reader) support?

tacaswell:
Be nice to be python 3 only in this case https://docs.python.org/3/reference/simple_stmts.html#raise

tacaswell:
power-cycled to run CI against current master.

takluyver:
I think the `raise etype, value, traceback` syntax doesn't work in Python 3, you'd need something like [six.reraise()](https://six.readthedocs.io/#six.reraise).

I'm not sure that this fix is right overall. If there's an error creating the file (e.g. you don't have write permissions to that directory), this change would silence the real error and raise the file not found error instead.

Maybe what we need is to check more precisely whether the first error is the obvious one (that file doesn't exist). If it is, try to create the file, but if it's anything else, just raise it directly.

takluyver:
I dug into this a bit more. Unfortunately, it looks like HDF5 doesn't give us any way to access the POSIX errno, which would let us reliably distinguish between different error causes. #493 was also asking for something similar

I've asked on the HDF Forum for a way to access the POSIX errno when relevant: https://forum.hdfgroup.org/t/get-posix-errno-from-hdf5-errors/5003

takluyver:
Closing: I don't think this is the right fix, as described above. The root cause is that we don't have a good way to distinguish different sorts of error opening the file, to separate the error we expect from those we don't.

PR #788, opened 2016-12-05
Use unicode prefix to enforce checking encoding

aragilar:
Currently this passes utf8-encoded bytes on python2, which may or may not be valid for a filename.


andreabedini:
Yes, @aragilar, unicode literals should be marked with `u`.

PR #790, opened 2016-12-06
Update windows type headers

aragilar:
This updates the windows interger headers for newer VS support.

### DIFF
--- a/windows/inttypes.h
+++ b/windows/inttypes.h
@@ -1,0 +1,306 @@
+// ISO C9x  compliant inttypes.h for Microsoft Visual Studio
+// Based on ISO/IEC 9899:TC2 Committee draft (May 6, 2005) WG14/N1124 
+// 
+//  Copyright (c) 2006-2013 Alexander Chemeris
+// 
+// Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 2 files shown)


andreabedini:
Is this only precautionary or it fixes some problem with building on windows? (Just to understand)

aragilar:
I needed this to get h5py to build on appveyor with python3, so this should fix problems, but it can wait until the appveyor work is ready to go (unless someone else has found this necessary to build on windows). I'll open an issue to track progress on getting appveyor to work.

aragilar:
I've rebased this on top of the current master, to see if this gains us anything, and to make sure that this doesn't break anything.

tacaswell:
should this be merged for #784 ?

aragilar:
It doesn't need to be, it can wait till after we do the release. When I was experimenting on getting appveyor to work, I thought I did need this, but that may have been due to something else at the time.

tacaswell:
Do we still need this, appevyor seems to be working with python3

aragilar:
I'm happy to close this, there could have been a bug in something else which got fixed at some point, or something changed on appveyor, but given everything seems to work without these changes, I think closing this makes sense, we can always come back to this if need be.

PR #794, opened 2016-12-08
Fix pathlib tests on windows

aragilar:
Windows does not like repeatedly opening an already open tempfile, this works around this.

### DIFF
--- a/h5py/tests/common.py
+++ b/h5py/tests/common.py
@@ -13,6 +13,7 @@ import sys
 import os
 import shutil
 import tempfile
+from contextlib import contextmanager
 
 from six import unichr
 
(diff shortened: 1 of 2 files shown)


andreabedini:
I am not sure I understand. Aren't the temporary files getting closed by the context manager?

andreabedini:
ok, had ☕️ now. So we use `tempfile.NamedTemporaryFile` only to have a name? we don't want it open? Then perhaps we should use `tempfile.TemporaryDirectory` and use a fixed name.

aragilar:
We need a named temporary file, but it can't be open on Windows (see the Windows NT comment at https://docs.python.org/2/library/tempfile.html#tempfile.NamedTemporaryFile). The temporary directory with a specific name would work, but the result is no different than the current code (instead of using `tempfile.NamedTemporaryFile` I'd need to use `tempfile.mkdtemp` as `tempfile.TemporaryDirectory` is python 3 only unfortunately) . I've chose this method as it was the recommended method for pathlib support for matplotlib which I'd previously written, but I can change the closed_tempfile context manager.

andreabedini:
ah, I forgot `tempfile.TemporaryDirectory` is not available on Py2 (:-1:). It's all good, I'll merge.

PR #804, opened 2016-12-14
Work around Python 2 not having surrogateescape handler

### LINKED ISSUE #803: Unable to import h5py in py2.7 when filesystem encoding is not utf-8
Merging #800 broke importing h5py on python 2.7 when the filesystem encoding is not utf-8.  Although I suspect that the broken commit was fc9b8133.  The particular problem is the following.

If os.fsencode/decode do not exist, then _hl/compat.py defines:

```python
    encoding = sys.getfilesystemencoding()
    if encoding == 'mbcs':
        errors = 'strict'
    else:
        errors = 'surrogateescape'    
    def fsencode(filename):
        """
        Encode filename to the filesystem encoding with 'surrogateescape' error
        handler, return bytes unchanged. On Windows, use 'strict' error handler if
        the file system encoding is 'mbcs' (which is the default encoding).
        """
        if isinstance(filename, six.binary_type):
            return filename
        elif isinstance(filename, six.text_type):
            return filename.encode(encoding, errors)
        else:
            raise TypeError("expect bytes or str, not %s" % type(filename).__name__)
```

Py2.7 doesn't define the "surrogateescape" error handler:

```python
u'\u03b1'.encode("utf-8", "surrogateescape") => '\xce\xb1'
u'\u03b1'.encode("ANSI_X3.4-1968", "surrogateescape") => LookupError: unknown error handler name 'surrogateescape'
```

The python-future package does, but it's broken (PythonCharmers/python-future#256).  Assuming that's fixed, I guess one could do:

```python
from future.utils.surrogateescape import register_surrogateescape
register_surrogateescape()
```

In py2 mode in compat.py.

aragilar:
This should fix #803. I've also added additional checks to travis to test for issues when LC_ALL=C/LANG=C, which should prevent breaking on systems without unicode support.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,22 @@ env:
 #  - TOXENV=py35-test-deps
 #  - TOXENV=py35-test-mindeps
 # commented out because of https://github.com/travis-ci/travis-ci/issues/4794
+  - TOXENV=py26-test-deps
+    TOX_TESTENV_PASSENV=LANG LC_ALL
+    LANG=C
+    LC_ALL=C
+  - TOXENV=py27-test-deps
+    TOX_TESTENV_PASSENV=LANG LC_ALL
+    LANG=C
+    LC_ALL=C
+  - TOXENV=py33-test-deps
+    TOX_TESTENV_PASSENV=LANG LC_ALL
+    LANG=C
+    LC_ALL=C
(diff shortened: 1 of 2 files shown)


wence-:
Seems to work for me, thanks.

Issue #732, opened 2016-07-08
Error while creating external links with unicode names


kormitigrov:
We are unable to create an ExternalLink to the file with the unicode (namely cyrilic) name. A simple example follows:

```python
# -*- coding: utf-8 -*-

import h5py

# creating and opening file
h5file = h5py.File('d:/temp/test.h5', 'w')

# creating the link to the file with unicode name (existing)
# h5file['link'] = h5py.ExternalLink('d:/work/grp_results.h5', '/')
h5file['link'] = h5py.ExternalLink('d:/Работа/grp_results.h5', '/')

# accessing the link - does not work
print(h5file['link'])
```

The traceback is rather odd:

```
Traceback (most recent call last):
  File "C:\Program Files (x86)\JetBrains\PyCharm Community Edition 5.0.3\helpers\pydev\pydevd.py", line 2407, in <module>
    globals = debugger.run(setup['file'], None, None, is_module)
  File "C:\Program Files (x86)\JetBrains\PyCharm Community Edition 5.0.3\helpers\pydev\pydevd.py", line 1798, in run
    launch(file, globals, locals)  # execute the script
  File "C:\Program Files (x86)\JetBrains\PyCharm Community Edition 5.0.3\helpers\pydev\_pydev_imps\_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc) 
  File "D:/REXLAB/test5.py", line 13, in <module>
    print(h5file['link'])
  File "h5py\_objects.pyx", line 54, in h5py._objects.with_phil.wrapper (D:\Build\h5py\h5py-2.5.x\h5py\_objects.c:2467)
  File "h5py\_objects.pyx", line 55, in h5py._objects.with_phil.wrapper (D:\Build\h5py\h5py-2.5.x\h5py\_objects.c:2424)
  File "C:\Program Files\WinPython-64bit-3.4.4.2Qt5\python-3.4.4.amd64\lib\site-packages\h5py\_hl\group.py", line 164, in __getitem__
    oid = h5o.open(self.id, self._e(name), lapl=self._lapl)
  File "h5py\_objects.pyx", line 54, in h5py._objects.with_phil.wrapper (D:\Build\h5py\h5py-2.5.x\h5py\_objects.c:2467)
  File "h5py\_objects.pyx", line 55, in h5py._objects.with_phil.wrapper (D:\Build\h5py\h5py-2.5.x\h5py\_objects.c:2424)
[...]

PR #809, opened 2016-12-21
Proposed high level interface for VDS

aaron-parsons:
Hi,

So this is a proposed high level interface for the VDS implementation. I'm looking for some feedback on whether you guys think it's any good/ what you are after.

The idea is to create at least two sets of objects that look like datasets, but aren't. One set is the source dataset, and the other is the virtual dataset. You can then specify mapping between the two using fancy indexing. A list of these maps is then provided into the "create" method.

I've tested it on the Excalibur, Percival, Edge and a couple of other Diamond use cases, and it seems to work fine.

I'm working on some proper unit tests now, and will also perform some tidying, but wanted an opinion before moving too far forward. 

Aaron


aragilar:
@aaron-parsons Can you add an example of how this would be used (e.g. a short program which reads one of the example datasets, plus a link to the example dataset and maybe even a file?)? Currently, I don't have a good understanding the reasoning behind specific choices made (I'll add specific comments inline), and currently, the code reads unpythonically (the getters/setters, the type checks, etc.), which to me confuses the issue more.

h5py/_hl/group.py
```diff
@@ -111,6 +113,32 @@ def create_dataset(self, name, shape=None, dtype=None, data=None, **kwds):
                 self[name] = dset
             return dset
 
+    def create_virtual_dataset(self,VMlist=[], fill_value=0x1):
```
aragilar: Why is the `VMlist` default an empty list? Are you aware of http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments? The commonly understood idiom is to set to `None, and create a new list inside the function, see the link.

aragilar: Why is the `fill_value` default `0x1`?

h5py/_hl/group.py
```diff
@@ -111,6 +113,32 @@ def create_dataset(self, name, shape=None, dtype=None, data=None, **kwds):
...
+        '''
+        Creates the virtual dataset from a list of virtual maps
+        '''
+        if type(VMlist) is not type([]):
+            VMlist = [VMlist]
+        
+        if len(VMlist)<1:
+            print "you have not given me any VirtualMaps!"
+            return 0
```
aragilar: I'm confused here, is this an error? A warning? What is the significance of returning 0?

h5py/_hl/group.py
```diff
@@ -530,3 +558,94 @@ def __init__(self, filename, path):
 
     def __repr__(self):
         return '<ExternalLink to "%s" in file "%s"' % (self.path, self.filename)
+
+
+class DatasetContainer(object):
```
aragilar: Is this supposed to match the API of the normal dataset?

h5py/_hl/group.py
```diff
@@ -530,3 +558,94 @@ def __init__(self, filename, path):
...
+        This is an object that looks like a dataset, but it not. It allows the user to specify the maps based on lazy indexing, 
+        which is natural, but without needing to load the data. If we can specify the shape, it's faster, if not, it gets pull from the actual source data.
+        '''
+        self.path = path
+        self.key = key
+        self.shape = shape
+        self.slice_list = [slice(None, None, None)] * len(self.shape)
+        
+        
+    def get_path(self):
```
aragilar: Not needed, `self.path` works.

h5py/_hl/group.py
```diff
@@ -530,3 +558,94 @@ def __init__(self, filename, path):
...
+        return self.path
+    
+    def get_key(self):
```
aragilar: Similarly not needed.

h5py/_hl/group.py
```diff
@@ -530,3 +558,94 @@ def __init__(self, filename, path):
...
+        return self.key
+    
+    def __getitem__(self, val):
+        if type(val) is not type(()):# make sure it's always a tuple
```
aragilar: Why does it need to be a tuple? Why not any sequence type?

h5py/_hl/group.py
```diff
@@ -530,3 +558,94 @@ def __init__(self, filename, path):
...
+            val = (val,)
+        self.slice_list = list(val) + [slice(None, None, None)]*(len(self.shape)-len(val)) # generate the right slice
+        return copy(self)
+
+    @property
+    def block_shape(self): # block shape is the size of the "SLICED dataset
+        outshape = ()
```
aragilar: `outshape` should be a list

h5py/_hl/group.py
```diff
@@ -530,3 +558,94 @@ def __init__(self, filename, path):
...
+        for i in range(len(self.shape)):
+            sanitized = self.get_slice_list()[i].indices(self.shape[i])
+            outshape += tuple([(sanitized[1]-sanitized[0])/sanitized[2]])
+        return outshape
+
+    def get_slice_list(self):
+        return self.slice_list
+    
+    def set_block_shape(self,shape):
```
aragilar: This appears to be useless, `the_block_shape` appear nowhere.

h5py/_hl/group.py
```diff
@@ -530,3 +558,94 @@ def __init__(self, filename, path):
...
+
+    @property
+    def block_shape(self): # block shape is the size of the "SLICED dataset
+        outshape = ()
+        for i in range(len(self.shape)):
+            sanitized = self.get_slice_list()[i].indices(self.shape[i])
+            outshape += tuple([(sanitized[1]-sanitized[0])/sanitized[2]])
+        return outshape
+
+    def get_slice_list(self):
```
aragilar: Similarly, not needed.

h5py/_hl/group.py
```diff
@@ -530,3 +558,94 @@ def __init__(self, filename, path):
...
+    '''
+    Virtual source and VirtualTarget are just easy identifiers for the user. They do the same thing.
+    '''
+    pass
+
+class VirtualTarget(DatasetContainer):
+    '''
+    Virtual source and VirtualTarget are just easy identifiers for the user. They do the same thing.
+    '''
+    pass
```
aragilar: These two classes seem extraneous, are they really needed?

h5py/_hl/group.py
aragilar: Would a better name be `VirtualDataset`?

h5py/_hl/group.py
```diff
@@ -530,3 +558,94 @@ def __init__(self, filename, path):
...
+        self.create_src_dpsace()
+
+    def get_block_shape(self):
+        '''
+        the two could be different ranks causing a rank deficiency in the map
+        '''
+        rank_def = len(self.target.shape) - len(self.src.shape)
+        return (1,)*rank_def + self.src.shape
+
+    def create_src_dpsace(self):
```
aragilar: Why is this not part of `__init__`?

h5py/_hl/group.py
```diff
@@ -530,3 +558,94 @@ def __init__(self, filename, path):
...
+        self.src_dspace = h5s.create_simple(self.src.shape)
+#         print self.src.get_slice_list()
+        # if the slice is None then we must want to take the first position
+        start_idx = tuple([0 if ix.start is None else ix.start for ix in self.src.get_slice_list()])
+        # if the count idx is None then we must want all the elements
+        count_idx = tuple([1 if ix.step is None else ix.step for ix in self.src.get_slice_list()])
+#         print count_idx
+        self.src_dspace.select_hyperslab(start=start_idx, count=count_idx, block=self.src.shape)
+    
+    def get_src_dpsace(self):
```
aragilar: This is also unnecessary.

aaron-parsons:
Hi James, Thanks for the quick feedback! The getters and setters are a hangovers from working on larger projects, where these are typically used to aid bug shooting. I'll change them to be class methods for this case, so they appear like bound variables.

Happy to provide example data+scripts, but wasn't sure where to put them/send them. What's best?

tacaswell:
I suggest adding a new page to https://github.com/h5py/h5py/tree/master/docs with some prose + examples.

ulrikpedersen:
Thanks for doing this @aaron-parsons :+1: 

> I suggest adding a new page to https://github.com/h5py/h5py/tree/master/docs with some prose + examples.

I agree with @tacaswell that docs are required. 

Also there is an "examples" folder where you could dump some stand-alone example scripts. I suggest you reproduce the examples/use-cases that come with the HDF5 library.

CC: @epourmal

epourmal:
Happy New Year to everyone!

The HDF Group will be more than happy to host those examples at our support Web site. Please send email to help@hdfgroup.org when they are available.

Unfortunately, I am behind on the VDS User's Guide. It is long overdue... It would be great to have pointers to the examples in that document too.

aragilar:
I've looked at this again (with the help of the example @aaron-parsons provided, thanks), and the information on VDS on the HDF5 website, and there's three questions that any high-level interface needs to answer:

1. How filenames are encoded. There appears to be no specification for how filenames are to be encoded inside the HDF5 file for VDS (like External Links). This creates problems when dealing with non-ASCII filenames on Windows (see for example https://github.com/h5py/h5py/issues/732, https://github.com/h5py/h5py/pull/800 and https://github.com/h5py/h5py/issues/818). While it's probably too late to change the API, we need to handle filename encodings correctly (and document what users need to do on their part).

2. Can virtual datasets be modified. This doesn't appear to be the case (based on the low-level and HDF5 APIs), but there doesn't appear to be an explicit statement on this (or maybe I just can't read and completely missed it). We should be explicit about this, within the h5py docs and with error messages.

3. How conflicting types are handled. I don't see where and whose responsibility it is to deal with conflicting types. What happens when I try reading a virtual dataset which has the dtype set to float, but the underlying dataset is made of strings (I don't see any code to prevent this, especially as one of the VDS examples explicitly mentions that the underlying datasets do not need to exist when the virtual dataset is created)? If it's HDF5 who handles this, we need to make sure h5py passes the error on correctly. If HDF5 doesn't handle this, we need to work out how h5py handles this (and what the user needs to do in this case)? This also needs to be documented in the h5py docs.

While the answers are probably simple and I missed some information (treat VDS filenames the same as other filenames, virtual datasets cannot be modified, HDF5 raises an error), they are things we need to handle correctly.

aragilar:
On the current implementation, the example, to me, shows the API is too complex. I'd suggest having two new classes, a `VirtualDataset` class which subclasses from `h5py.Dataset`, overriding methods where necessary, and raising errors on unsupported operations, and what would effectively be a namedtuple (which I've called `h5py.VirtualReference` in my example below, so as not to be confused with `VirtualMap` and its API, which is similar), which would hold the needed information. I would recommend allowing for the selections anything which can be used on normal datasets (integers, numpy-like indexing etc.). I'd also recommend allowing passing actual HDF5 files to the `source_file` argument, and passing h5py datasets to the `source_dataset` argument, which will allow shapes and dtypes to be validated before writing the virtual dataset to the file. My modified version of the example is as follows:
```
import h5py

FILE_NAMES_TO_CONCATENATE = ['1.h5', '2.h5', '3.h5', '4.h5', '5.h5']
SOURCE_DATASET_PATH = '/data'

with h5py.File("VDS.h5", 'w', libver='latest') as f:
    with h5py.File(file_names_to_concatenate[0],'r') as h:
        inner_shape = h[SOURCE_DATASET_PATH].shape
        dtype = h[SOURCE_DATASET_PATH].dtype
    total_shape = tuple(len(FILE_NAMES_TO_CONCATENATE)) + inner_shape

    virtual_dataset = f.create_virtual_dataset(
        "virtual_dataset",
        [
            h5py.VirtualReference(
                target_section=i,
                source_file=filename,
                source_dataset=SOURCE_DATASET_PATH,
                source_selection=None,
                source_dtype=dtype,
            ) for i, filename in enumerate(FILE_NAMES_TO_CONCATENATE)
        ], shape=total_shape, dtype=dtype, fillvalue=0
    )
    print(virtual_dataset)
```

aaron-parsons:
Dear folks,

Thanks for the support on this pull request, and sorry it has taken so long to come back with a response. I've also managed to tidy up the request considerably.

Our focus for the use of virtual dataset at Diamond is primarily focussed on stitching together SWMR datasets in various (sometimes quite complicated) configurations. Some of these are detailed in: https://support.hdfgroup.org/HDF5/docNewFeatures/VDS/HDF5-VDS-requirements-use-cases-2014-12-10.pdf We also intend to use it for manipulation of big datasets via processing tools, predominately savu (https://github.com/DiamondLightSource/Savu). For this reason the main facility we require is access to a certain level of fancy indexing, in a simple to use interface.

This is the reason for over-riding the get_item and not just using the example that @aragilar demonstrated. It would be nice to eventually move this in to the dataset api to take advantage of all the nice work done by other people, but this requires a dataset id. Since one of the conditions for VDS is that we can write the new dataset without the sources existing (this is actually very important), this is not possible for us at the moment.

In reponse to the comments by @aragilar :

1. The same conditions apply here as for the underlying C code (and errors are handled by the C code), I won't address them in this higher level interface, but happy to do so if there are some guidelines?
2.Unfortunately they can't be modified except on creation. This is why we need the virtual_map interface. Otherwise we could add references as we go and could use the dataset api.
3. Conflicting types are handled by the underlying C. If it can cast to the new type then it will (in fact this was a pre-requisite for VDS), if not it will fail with the same error as it would normally.

The majority of our use cases I have outlined both in the examples directory, and also with unit tests for the low and high level interfaces. These should all pass in Travis. I'm sure as we begin to use it in savu over the coming months, there will be further commits (+unit tests) from our side to shoot bugs, but for now I think this a reasonable first start.

aaron-parsons:
Hi @tacaswell and @aragilar,

Do you have any more comments for this pull request?
I added support + test for ellipsis, which seems to break some of the other tests you have.
Otherwise, we are currently using this without problems.

What would be needed to complete the PR?

Aaron

aragilar:
Sorry, I been spending all my free time fixing my desktop, I'll try to look at this over the weekend.

aaron-parsons:
No worries. Hope it get fixed!

aragilar:
I haven't had a chance to play with the code, so I'll have more feedback later, but three things:
1. Can you fix the failing tests on Python 3, the ellipse code doesn't work on Python 3?
2. The vds tests aren't running (you probably need to add an __init__.py), and aren't Python 3 compatible, can you fix those?
3. Can you redo the history so that there isn't a merge from master, it makes the history messy (you should just be able to rebase onto master)? If you're not comfortable with that, I can do that.

aaron-parsons:
Hi,

I've done all of these. I can't actually see what tests are running on travis, it just provides me a summary, They all run locally now when I py.test the tests folder, so I guess that's ok.

Ellipsis, was just a try/catch.

Let me know what you think of the code, it seems to work pretty well from our end.

Aaron

tacaswell:
I addressed most of my own comments in #894

ulrikpedersen:
> I addressed most of my own comments in #894

Hi @tacaswell - is this PR effectively closed in favour of #894 which seems more recent? If yes, can #894 replace this PR on milestone ["2.8"](https://github.com/h5py/h5py/milestone/11)?

tacaswell:
I am not going to have time to come back to this for a while.  I would be happy if someone from Diamond took over finishing up the last couple of to-dos on #894

takluyver:
This evolved into #1060, which was just merged.

tacaswell:
Closing now that #1060 is in.

PR #811, opened 2016-12-26
Update classifiers to include supported python version and interpreters

### LINKED ISSUE #810: PyPi entry doesn't include supported python versions
The PyPi entry for h5py doesn't include the version support (e.g. python3 support) - leading it to be included in lists of packages incompatible with python3 (e.g. the one linked here, https://medium.com/broken-window/python-3-support-for-third-party-libraries-dcd7a156e5bd#.22slwvwof , though that may be corrected soon)

aragilar:
Fixes #810, doesn't include python 3.2 (as we dropped it) or python 3.6 (as we haven't started testing on it), but includes pythons 2.6, 2.7, 3.3-3.5.

Also adds Cython classifier (as other Cython-using projects do, e.g. pandas), and CPython interpreter classifier.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -104,7 +104,16 @@ Intended Audience :: Developers
 Intended Audience :: Information Technology
 Intended Audience :: Science/Research
 License :: OSI Approved :: BSD License
+Programming Language :: Cython
 Programming Language :: Python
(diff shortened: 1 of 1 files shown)


tacaswell:
:+1: Thanks!

Issue #641, opened 2015-11-05
Interpret H5T_NATIVE_B8 as numpy.bool


llchan:
I'm migrating from PyTables, which uses `H5T_NATIVE_B8` as the bool representation.  In h5py, this HDF5 type currently raises `TypeError: No NumPy equivalent for TypeBitfieldID exists`.  Since it's currently unusable anyways, would anyone be opposed to having h5py interpret those columns as bool so the files are compatible?  I also have some existing C code that writes PyTables-readable data, and it would make my life a lot easier to write one format instead instead of toggling between two modes.

If we're worried about breaking existing stuff, we could make it an opt-in feature, but that seems unnecessary since the existing behavior is that it cant read the data at all.

andrewcollette:
I don't have any objection to reading or interoperating with such data, as long as we continue to create files using the now-established h5py types.  You could look in h5t.pyx to get started.

llchan:
Cool, I'll put together a PR

achilleas-k:
Is there a workaround for this issue? Is there any way to read `H5T_NATIVE_B8` type data from an existing HDF5 file using h5py?

llchan:
Apologies for letting this slip. I had a workaround a local branch, but hadnt put together a PR because the use case went away (wound up just converting types upstream of h5py). Let me take a look when I get back to my other computer.

will133:
I have recently encountered this issue as well as I believe this is the only showstopper when I try to migrate our codebase to use h5py only instead of using both pytables and h5py.  I'm very interested in testing this out if you have a workaround for this issue.

will133:
I'm looking at the h5py source a bit but I don't know what's a proper way to implement this.  I basically added these in h5t.pyx:

```
diff --git a/h5py/h5t.pxd b/h5py/h5t.pxd
index 13e9c8c..676da9c 100644
--- a/h5py/h5t.pxd
+++ b/h5py/h5t.pxd
@@ -34,9 +34,6 @@ cdef class TypeVlenID(TypeID):
 cdef class TypeTimeID(TypeID):
     pass
 
-cdef class TypeBitfieldID(TypeID):
-    pass
-
 cdef class TypeReferenceID(TypeID):
     pass
 
@@ -51,6 +48,9 @@ cdef class TypeIntegerID(TypeAtomicID):
 cdef class TypeFloatID(TypeAtomicID):
     pass
 
+cdef class TypeBitfieldID(TypeAtomicID):
+    pass
+
[...]

PR #821, opened 2017-01-11
Add H5T_NATIVE_B8 -> bool support

llchan:
This adds `H5T_NATIVE_B8` -> `bool` support. The primary motivation for this is PyTables interop (they encode bools as 0/1 in a `H5T_NATIVE_B8` field).

A new `b8_to_bool` flag is added to the `H5PYConfig` object to enable this feature. It's off by default to preserve the old exception-raising behavior in case anyone relies on that. If we don't care about preserving that we can set it to True by default.

Comments are welcome. This is my first time working with the h5py source so it's possible I'm doing something unconventional.

Addresses #641

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -149,9 +149,6 @@ matrix:
             - libhdf5-openmpi-dev # 1.8.11 based on trusty
 
     # PyTables compatibility tests
-    - python: 2.7
-      env:
-      - TOXENV=py27-test-mindeps-tables
     - python: 3.7
       env:
       - TOXENV=py37-test-mindeps-tables
--- a/.travis.yml
+++ b/.travis.yml
@@ -148,6 +148,14 @@ matrix:
             - libopenmpi-dev
             - libhdf5-openmpi-dev # 1.8.11 based on trusty
 
+    # PyTables compatibility tests
+    - python: 2.7
+      env:
+      - TOXENV=py27-test-mindeps-tables
+    - python: 3.7
+      env:
+      - TOXENV=py37-test-mindeps-tables
+
     # Additional python versions which are not officially supported
     - python: "nightly"
       env:
--- a/h5py/_conv.pyx
+++ b/h5py/_conv.pyx
@@ -13,10 +13,13 @@
     Low-level type-conversion routines.
 """
 
+from h5 import get_config
 from h5r cimport Reference, RegionReference, hobj_ref_t, hdset_reg_ref_t
 from h5t cimport H5PY_OBJ, typewrap, py_create, TypeID
 cimport numpy as np
 from libc.stdlib cimport realloc
+
+cfg = get_config()
 
 # Initialization
 np.import_array()
@@ -805,6 +808,19 @@ cdef int conv_ndarray2vlen(void* ipt, void* opt,
     return 0
 
 # =============================================================================
+# B8 to enum bool routines
+
+cdef herr_t b82boolenum(hid_t src_id, hid_t dst_id, H5T_cdata_t *cdata,
+                        size_t nl, size_t buf_stride, size_t bkg_stride, void *buf_i,
+                        void *bkg_i, hid_t dxpl) except -1:
+    return 0
+
+cdef herr_t boolenum2b8(hid_t src_id, hid_t dst_id, H5T_cdata_t *cdata,
+                        size_t nl, size_t buf_stride, size_t bkg_stride, void *buf_i,
+                        void *bkg_i, hid_t dxpl) except -1:
+    return 0
+
+# =============================================================================
 
 cpdef int register_converters() except -1:
 
@@ -812,6 +828,9 @@ cpdef int register_converters() except -1:
     cdef hid_t vlentype
     cdef hid_t pyobj
     cdef hid_t enum
+    cdef hid_t boolenum = -1
+    cdef int8_t f_value = 0
+    cdef int8_t t_value = 1
 
     vlstring = H5Tcopy(H5T_C_S1)
     H5Tset_size(vlstring, H5T_VARIABLE)
@@ -821,6 +840,10 @@ cpdef int register_converters() except -1:
     vlentype = H5Tvlen_create(H5T_STD_I32LE)
 
     pyobj = H5PY_OBJ
+
+    boolenum = H5Tenum_create(H5T_NATIVE_INT8)
+    H5Tenum_insert(boolenum, cfg._f_name, &f_value)
+    H5Tenum_insert(boolenum, cfg._t_name, &t_value)
 
     H5Tregister(H5T_PERS_HARD, "vlen2str", vlstring, pyobj, vlen2str)
     H5Tregister(H5T_PERS_HARD, "str2vlen", pyobj, vlstring, str2vlen)
@@ -840,9 +863,13 @@ cpdef int register_converters() except -1:
     H5Tregister(H5T_PERS_SOFT, "vlen2ndarray", vlentype, pyobj, vlen2ndarray)
     H5Tregister(H5T_PERS_SOFT, "ndarray2vlen", pyobj, vlentype, ndarray2vlen)
 
+    H5Tregister(H5T_PERS_HARD, "boolenum2b8", boolenum, H5T_NATIVE_B8, boolenum2b8)
+    H5Tregister(H5T_PERS_HARD, "b82boolenum", H5T_NATIVE_B8, boolenum, b82boolenum)
+
     H5Tclose(vlstring)
     H5Tclose(vlentype)
     H5Tclose(enum)
+    H5Tclose(boolenum)
 
     return 0
 
@@ -865,5 +892,8 @@ cpdef int unregister_converters() except -1:
 
     H5Tunregister(H5T_PERS_SOFT, "vlen2ndarray", -1, -1, vlen2ndarray)
     H5Tunregister(H5T_PERS_SOFT, "ndarray2vlen", -1, -1, ndarray2vlen)
+
+    H5Tunregister(H5T_PERS_HARD, "boolenum2b8", -1, -1, boolenum2b8)
+    H5Tunregister(H5T_PERS_HARD, "b82boolenum", -1, -1, b82boolenum)
 
     return 0
--- a/h5py/_conv.pyx
+++ b/h5py/_conv.pyx
@@ -821,6 +821,19 @@ cdef herr_t boolenum2b8(hid_t src_id, hid_t dst_id, H5T_cdata_t *cdata,
     return 0
 
 # =============================================================================
+# B8 to UINT8 routines
+
+cdef herr_t b82uint8(hid_t src_id, hid_t dst_id, H5T_cdata_t *cdata,
+                     size_t nl, size_t buf_stride, size_t bkg_stride, void *buf_i,
+                     void *bkg_i, hid_t dxpl) except -1:
+    return 0
+
+cdef herr_t uint82b8(hid_t src_id, hid_t dst_id, H5T_cdata_t *cdata,
+                     size_t nl, size_t buf_stride, size_t bkg_stride, void *buf_i,
+                     void *bkg_i, hid_t dxpl) except -1:
+    return 0
+
+# =============================================================================
 
 cpdef int register_converters() except -1:
 
@@ -866,6 +879,9 @@ cpdef int register_converters() except -1:
     H5Tregister(H5T_PERS_HARD, "boolenum2b8", boolenum, H5T_NATIVE_B8, boolenum2b8)
     H5Tregister(H5T_PERS_HARD, "b82boolenum", H5T_NATIVE_B8, boolenum, b82boolenum)
 
+    H5Tregister(H5T_PERS_HARD, "uint82b8", H5T_NATIVE_UINT8, H5T_NATIVE_B8, uint82b8)
+    H5Tregister(H5T_PERS_HARD, "b82uint8", H5T_NATIVE_B8, H5T_NATIVE_UINT8, b82uint8)
+
     H5Tclose(vlstring)
     H5Tclose(vlentype)
     H5Tclose(enum)
@@ -895,5 +911,8 @@ cpdef int unregister_converters() except -1:
 
     H5Tunregister(H5T_PERS_HARD, "boolenum2b8", -1, -1, boolenum2b8)
     H5Tunregister(H5T_PERS_HARD, "b82boolenum", -1, -1, b82boolenum)
+
+    H5Tunregister(H5T_PERS_HARD, "uint82b8", -1, -1, uint82b8)
+    H5Tunregister(H5T_PERS_HARD, "b82uint8", -1, -1, b82uint8)
 
     return 0
--- a/h5py/h5t.pyx
+++ b/h5py/h5t.pyx
@@ -784,19 +784,12 @@ cdef class TypeTimeID(TypeID):
     """
     pass
 
-
 cdef class TypeBitfieldID(TypeID):
 
     """
         HDF5 bitfield type
     """
-
-    cdef object py_dtype(self):
-        if H5Tequal(self.id, H5T_NATIVE_B8):
-            return dtype('?')
-        else:
-            raise TypeError("No NumPy equivalent for %s exists" % self.__class__.__name__)
-
+    pass
 
 cdef class TypeReferenceID(TypeID):
 
--- a/h5py/h5t.pyx
+++ b/h5py/h5t.pyx
@@ -784,12 +784,19 @@ cdef class TypeTimeID(TypeID):
     """
     pass
 
+
 cdef class TypeBitfieldID(TypeID):
 
     """
         HDF5 bitfield type
     """
-    pass
+
+    cdef object py_dtype(self):
+        if H5Tequal(self.id, H5T_NATIVE_B8):
+            return dtype('?')
+        else:
+            raise TypeError("No NumPy equivalent for %s exists" % self.__class__.__name__)
+
 
 cdef class TypeReferenceID(TypeID):
 
--- a/h5py/tests/hl/test_datatype.py
+++ b/h5py/tests/hl/test_datatype.py
@@ -308,7 +308,7 @@ class TestB8Bool(TestCase):
 
     def _test_b8(self, arr1):
         path = self.mktemp()
-        with tables.open_file(path, 'a') as f:
+        with tables.open_file(path, 'w') as f:
             if arr1.dtype.names:
                 f.create_table('/', 'test', obj=arr1)
             else:
--- a/h5py/tests/hl/test_datatype.py
+++ b/h5py/tests/hl/test_datatype.py
@@ -8,6 +8,10 @@ from itertools import count
 import numpy as np
 import six
 import h5py
+try:
+    import tables
+except ImportError:
+    tables = None
 
 from ..common import ut, TestCase
 
@@ -271,3 +275,61 @@ class TestStrings(TestCase):
         assert string_info.encoding == 'ascii'
         assert string_info.length == 10
         assert h5py.check_vlen_dtype(dt) is None
+
+
+@ut.skipUnless(tables is not None, 'tables is required')
+class TestB8Bool(TestCase):
+
+    """
+    Test loading of H5T_NATIVE_B8 as numpy.bool
+    """
+
+    def test_b8_bool(self):
+        arr1 = np.array([False, True], dtype=np.bool)
+        self._test_b8(arr1)
+
+    def test_b8_bool_compound(self):
+        arr1 = np.array([(False,), (True,)], dtype=np.dtype([('x', '?')]))
+        self._test_b8(arr1)
+
+    def test_b8_bool_compound_nested(self):
+        arr1 = np.array(
+            [(True, (True, False)), (True, (False, True))],
+            dtype=np.dtype([('x', '?'), ('y', [('a', '?'), ('b', '?')])]),
+        )
+        self._test_b8(arr1)
+
+    def test_b8_bool_array(self):
+        arr1 = np.array(
+            [((True, True, False),), ((True, False, True),)],
+            dtype=np.dtype([('x', ('?', (3,)))]),
+        )
+        self._test_b8(arr1)
+
+    def _test_b8(self, arr1):
+        path = self.mktemp()
+        with tables.open_file(path, 'a') as f:
+            if arr1.dtype.names:
+                f.create_table('/', 'test', obj=arr1)
+            else:
+                f.create_array('/', 'test', obj=arr1)
+
+        with h5py.File(path, 'r') as f:
+            dset = f['test']
+
+            # read uncast dset to make sure it raises as before
+            with self.assertRaises(
+                TypeError, msg='No NumPy equivalent for TypeBitfieldID exists'
+            ):
+                dset[:]
+
+            # read cast dset and make sure it's equal
+            with dset.astype(arr1.dtype):
+                arr2 = dset[:]
+            self.assertArrayEqual(arr1, arr2)
+
+            # read uncast dataset again to ensure nothing changed permanantly
+            with self.assertRaises(
+                TypeError, msg='No NumPy equivalent for TypeBitfieldID exists'
+            ):
+                dset[:]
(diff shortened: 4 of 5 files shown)


llchan:
I should add that the tests use `tables` to generate the H5T_NATIVE_B8 columns. If `h5py` has a better way to generate the test files we should update the tests to use that instead. If not, we may want to add `tables` to the CI setups.

llchan:
Did some performance testing. Whatever mechanism this is invoking in the HDF5 code is quite slow, despite the converter being a no-op, and I'm seeing a `H5T_NATIVE_B8` read taking 10x longer than the same read in PyTables. If I modify the h5py code to represent bools as `H5T_NATIVE_B8`, the read times are identical. I'll poke around the HDF5 source and see if there's any obvious reasons why a no-op conversion should affect read times that much.

In any case, this PR is still valid, but we may be able to implement it more efficiently by giving HDF5 the "best" memory type to read into and calling `numpy.ndarray.view` on it after the fact. Would likely make the change more invasive though.

llchan:
Profiling is showing a majority of the time spent in `H5T__conv_struct_opt`, which appears to be a function for converting between compound datatypes. It's doing a ton of unnecessary work, e.g. gazillions of `memmove`s when the types I'm converting between should be converted in-place. So the best way to fix this would be to add a fast path to that function. However, in my experience, HDF5 is a very slow moving project, and even if we improve the implementation it'll take a while for it to be released, and I don't know if they are willing to backport the changes to 1.8.x (I can't upgrade to 1.10.x until PyTables supports it). We may want to more strongly consider the idea of reading into a dtype that matches the HDF5 dtype, and casting to the final view in h5py.

Btw, I am on HDF5 1.8.17, but as far as I can tell the latest 1.10.x source has the same code for this function.

epourmal:
Please submit the patch. We will try to move quickly and we can out it in both 1.8 and 1.10

1.10.1 is coming in late February.

Thank you!

Elena

Sent from my iPhone

On Jan 12, 2017, at 1:04 PM, Lawrence Chan <notifications@github.com<mailto:notifications@github.com>> wrote:


Profiling is showing a majority of the time spent in H5T__conv_struct_opt, which appears to be a function for converting between compound datatypes. It's doing a ton of unnecessary work, e.g. gazillions of memmoves when the types I'm converting between should be converted in-place. So the best way to fix this would be to add a fast path to that function. However, in my experience, HDF5 is a very slow moving project, and even if we improve the implementation it'll take a while for it to be released, and I don't know if they are willing to backport the changes to 1.8.x (I can't upgrade to 1.10.x until PyTables supports it). We may want to more strongly consider the idea of reading into a dtype that matches the HDF5 dtype, and casting to the final view in h5py.

Btw, I am on HDF5 1.8.17, but as far as I can tell the latest 1.10.x source has the same code for this function.

-

aragilar:
It should be possible to create the datasets via the h5py low-level interface, rather than using pytables. I'd move the tests which test for pytables compatibility (i.e. explicitly depending on pytables, rather than just reading hdf5 datatypes) into a separate directory (so h5py/tests/pytables/), which makes it easy to find and add these tests (I presume there's going to be quite a few, so I'll look at adding pytables to the tox/Travis/appveyor configs).

I wouldn't bother about the `b8_to_bool` flag, it'll only complicate things (as long as we don't change how we write, and only increase the types we can read, I don't see there being an issue).

llchan:
@epourmal I havent looked into a HDF5 patch yet, but will take a look when I have a chance. Realistically speaking, it probably won't happen fast enough to make a late February release. If you guys have the resources and could take a look as well, that would be helpful. I think it should be easy to reproduce, just a no-op conversion within a compound type. There should be no memory operations and the performance should be reasonably close to a read without a conversion.

@aragilar sounds good to me. I only guarded it with the `b8_to_bool` flag in case anyone relied on the exception to switch to an alternate code path. I can remove all that stuff and make it always-on.

I'll also look into the low-level interface for the tests.

llchan:
@aragilar Actually, upon further consideration, I'm not sure if we want it always-on. What if someone writes a `H5T_NATIVE_B8` thinking they will get a `int8` or `uint8`? As it stands now, they will see the exception and look for alternatives (maybe the low-level interface?). If we read them as bool by default, it could lead to unexpected behavior. If anything, a 1-byte bitfield would be more appropriately read as a `uint8`.

Perhaps the best way to make this automagically do the right thing is to have h5py detect that the file was written by PyTables and enable the compatibility mode. PyTables writes attributes to the HDF5 file, so it should be easy to detect. I'd have to do some testing to see what happens when a C program writes an HDF5 file and PyTables appends something to that file. Off the top of my head I think each dataset has its own PyTables-specific attributes, so we can do this on a per-dataset basis.

aragilar:
If they want to read it as `uint8`, then `Dataset.astype` (http://docs.h5py.org/en/latest/high/dataset.html#Dataset.astype) should work fine, we can even add a test for this to be sure. Guessing based on heuristics seems like a bad idea, further complicating things.

llchan:
Okay, maybe automatically detecting PyTables isnt the best idea, could lead to needless maintenance on someone else's schedule. Scrap that. The rest of this assumes that there shall be no automated detection based on HDF5 attributes/metadata.

---

I think that enabling _any_ bitfield conversion by default is a bad idea, because the interpretation of the bits is application-specific, and automatic conversion can have unexpected pitfalls:

###### Automatic `bool` conversion can be bad
Suppose `h5[key] = data` contains a B8 bitfield intended as a real bitfield, and `data[0] = 0xA` and `data[1] = 0xF`, etc. If we enable automatic bool conversion, a lot of things can behave unexpectedly:
```python
data = h5[key][:]
same = data[0] == data[1]  # broken (silently)
changed = data[:-1] == data[1:]  # broken (silently)
n_unique = len(np.unique(data))  # broken (silently)
bits = numpy.unpackbits(data)  # broken (noisily)
```

###### Automatic `uint8` conversion can be bad
Suppose `h5[key] = mask` contains a B8 bitfield intended as a bool, and `mask[0] = True` and `mask[1] = False`, etc. When used as a array slice, this mask will now index into the array (choosing only elements 0 and 1) instead of masking elements.
```python
mask = h5[key][:]
masked = data[mask]  # broken (silently)
```

Yes, if you control both the writing process and the reading process and know how h5py handles bitfields, you can use astype and make it all work just fine. But if someone is sent a file and needs to poke around without knowing much about it, the default behavior of h5py should have as few silent gotchas as possible.

###### Proposal
We include all the necessary HDF5-level converters, but it's up to the user to cast the dataset to a readable type. We can add helper functions to simplify some of these for simple cases like "make all B8 fields bool" or "make all bitfields unsigned integer fields of the same width". An uncast dataset will raise the same exception that it used to, indicating to the user that an explicit cast is required.

Thoughts?

aragilar:
Seems good, and we should document that casting is required in the dataset documentation (either before or after http://docs.h5py.org/en/latest/high/dataset.html#creating-and-reading-empty-or-null-datasets-and-attributes).

tacaswell:
@llchan What is the state of this?

khornlund:
Any update on this getting merged in?

taldcroft:
:+1: on getting reviving this!  https://github.com/astropy/astropy/issues/6473

llchan:
Sorry for letting this sit for so long.  I wound up changing my (C/C++) writer to write h5py-style bools directly, so this became a non-blocking issue for me and fell by the wayside.  It's been 2+ years since I looked at this, so I suspect the PR would need some updating, and to be quite honest I don't quite have the bandwidth to push on this at the moment.  Does someone else want to carry on with it?

danwil02:
Any progress on getting this merged in? This is a blocking issue for my project and we've had difficulty building this PR on windows. It would be fantastic if this could get merged so that we can just grab the new version of the package from the anaconda repo.

llchan:
Okay since there seems to be more interest, I'll see what I can do to resurrect this PR. I think the proposal in the comments was not implemented so it will take some code rework.

danwil02:
Thanks mate, I appreciate your effort on this. Let us know how you go

llchan:
Pushed up a new version of the PR that favors explicit type casting, e.g.
```python
with dset.astype(np.bool):
    data = dset[:]
```

Not sure why the CI tests are failing, works for me under pytest locally.  Can someone familiar with tox/travis take a look?

llchan:
Alright after some fiddling it turns out the CI issues were unrelated to the PR.  py37-test-mindeps was broken all along but it was not actually tested in travis so adding a py37-test-mindeps-tables surfaced some pre-existing issues.  I fixed those for the sake of getting this PR in the green.  If a maintainer prefers to fix those CI issues in a separate commit/PR I can rebase this one off master when that's ready.

danwil02:
Great work mate! Glad to see this moving forward :)
When do you think we could expect an update to the windows h5py package on the anaconda repo with this change?

llchan:
That will be up to the maintainers, I'm not involved in the release process, and I imagine it depends a lot on the other changes that are in progress.

If you're blocked on this, you could always pull down my branch or apply the diff to your own checkout and install that.  I'm not familiar with anaconda so I can't give you more concrete steps on how to do that, but I'm sure it's possible and supported.

MikeScavell:
I'm also at a point where this functionality would be useful to my project. I managed to build the version with the `b8_to_bool` flag before you did the force-push, but I only got it building for Linux. It worked exactly as I needed for my project but I was really hoping to get it built for Windows.

I managed to get your latest branch built for Windows (using AppVeyor) but it seems to be missing some of the functionality that was in the `b8_to_bool` flag version (the main issue being it won't correctly read BitFields). Is there any chance you can reintegrate the `b8_to_bool` flag and/or ability to read BitField data in the latest version? I had a shot at it myself but I'm not well experienced in constructing my own Python packages.

llchan:
Some of the discussion above led to one of the other contributors suggesting we stick to explicit casts rather than having it in a global config flag, so I removed the flag and you should read your datasets like
```python
with dset.astype(np.bool):
    data = dset[:]
```
(also works for compound dtypes)

If you have an example that doesnt work with that, paste it over and I'll be happy to address.

MikeScavell:
Thanks for getting back to me. I made sure to try your recommendation (which didn't quite work for me) before posting my last comment. Here is a code snippet I'm using and the errors I get.

If I use this on my file:
`import h5py
import numpy as np

filename = "log.h5"

f = h5py.File(filename, 'r')

dset = f["Baseline/SimulationTime"]

with dset.astype(np.bool):
    data = dset[:]`
I get the error "Can't read data (no appropriate function for conversion path)".

If I try to type "f["Baseline/SimulationTime"]" in the IPython console I get "No NumPy equivalent for TypeBitfieldID exists".

This was definitely working with your version that used the `b8-to-bool` flag as I still have that package built on my Linux VM and I can still run it successfully on the same file. Let me know if you have any ideas on how I can get past this issue. If needed, I can probably send you a copy of a HDF5 file that works on one build but not the other.

Thanks for your time!

llchan:
I should have been more clear, the `np.bool` is just an example for a bool array.  Most likely your dataset has a compound dtype, so you would need to do something more like
```python
dtype = np.dtype([('foo', 'i8'), ('bar', '?')])
with dset.astype(dtype):
    data = dset[:]
```

I still have my original PR in a branch, so if we want the old method of a global flag to toggle all B8 fields to bools, I could always re-add that stuff.  Would like to hear opinions from other contributors/maintainers, because I think that while this explicit casting is clear and non-magical, it's rather cumbersome and doesnt help much when you don't know the dtype up front.

llchan:
Okay actually I re-read some of the discussion above (sorry, this is from a long time ago so not really fresh in my memory), and I think it would be acceptable to make this always-on and unconfigurable, and if anyone needs B8 fields to behave in some way _other_ than a bool, they will need to explicitly cast.  Not 100% safe for everyone, but will prob cover 99% of use cases and make things Just Work out of the box.

Hang tight while I flip it on.

MikeScavell:
I agree with your first comment, part of the issue I also had was I was unsure what data was in any given field (could be bool, string, int, double, or any combination of these or more). I think the always on idea is good since it will work for most people, and anyone who knows they have something different can explicitly cast it.

Thanks so much for your efforts llchan, I've got the package installed on Windows and it is working perfectly. This is extremely helpful to the project I'm working on at the moment and you've made a few people very happy! 😄

takluyver:
I may be missing things, but from my rough understanding of these concepts, I'd prefer to keep the explicit cast to bool.

Pytables is a tool which uses HDF5 as a storage format for its own concepts, so it can make a decision that it's going to store booleans with the B8 bitfield type. By contrast, h5py aims to expose HDF5 functionality as faithfully and idiomatically as possible for use in Python, and I don't think automatically casting 8-bit bitfields to booleans does that, especially because it loses data.

As you said previously, by all means have convenient ways to get the data as booleans, and document them for compatibility with Pytables. But I don't think it should do that by default. If it does anything by default, I'd prefer preserving all the data (presumably as uint8), but I'd also be happy if you need to explicitly specify either type, since AFAIK numpy doesn't have equivalent bitfield types.

llchan:
Does reading as uint8 work for everyone?  It won't round-trip through pytables+h5py anymore, but maybe that's not a big deal.  Like you said, uint8 would preserve all the bits in case someone is actually using it as a bitfield, and presumably that's the type they would want anyways. Requiring the cast makes h5py unusable for datasets like this unless the dtype is known exactly ahead of time, which imho hurts usability significantly.

At any rate, I don't have a strong preference, but given that I don't use pytables-encoded bools anymore, I don't really want to spend too much more time on this PR, and I'd like to either merge in something simple, or pass on the torch to someone else.

takluyver:
Thanks for the work you've put in on this even when it's no longer something you need. :-)

Is it easy for you to roll it back to the variant which required specifying the target type explicitly? I think we could merge it like that, because it's a clear improvement over the status quo where you can't read that data at all. Then we can open a new issue to decide whether there should be a default type.

takluyver:
Am I right in thinking that rolling back to requiring an explicit cast would just involve dropping commit b01fe389cdbe393a6e569b25ac5d25397517f8a9?

llchan:
That's correct, let me push that up. I also have an old version of this work floating around somewhere that includes a global h5py config var to enable auto-convert. If I find that, should we include that with this PR?

llchan:
To elaborate on the above, we could do something like
```python
h5py.get_config().b8_dtype = np.dtype('?')
```
and that would case all B8s to be interpreted as bools.  That should give users the flexibility to use e.g. u1 instead if they want a bitfield, or dont do any configuration and get the current type exception.

takluyver:
I prefer having it decided locally by the code accessing it, rather than a global config variable. I generally like libraries to avoid config options unless necessary, because it makes it harder to see what one piece of code will do if it can be influenced by a config option set somewhere else.

If the others are happy with this, I propose that we make it possible to load B8 bitfield data, and then open a new issue to discuss whether & how to make it more convenient.

Does `with dset.astype(np.uint8):` also work with this PR?

llchan:
Sounds good, I can keep this PR straightforward and require the cast and we can discuss alternatives in another issue/PR.

And yeah, normally I wouldn't be a fan of global vars either, but there's already an existing global config so I thought that might be a reasonable place to kick h5py into "pytables-compat-mode".  Perhaps there's enough stuff in the low-level h5py API to discover the underlying dtype but I'm not familiar with it.  I think with the high-level API you won't be able to get the dtype of the dataset.  Anyways that stuff can be discussed later.

Also, I should reiterate what I mentioned in a comment closer to the top: even though these are no-op conversions, hdf5 does still need to make copies of the data (with the no-frills way I've set up the converters, at least).  Haven't tested this in 2+ years but I imagine the overhead is still present and significant.

takluyver:
OK, thanks for the heads up.

Am I right in thinking this PR currently only allows conversion to bool, not to uint8 even if you specify that with `.astype()`?

llchan:
Pushed another commit with the analogous code path for B8 -> UINT8.

takluyver:
Thanks. I'm happy with this PR now, but I'd like to let one of the other maintainers take a quick look before merging it (@tacaswell @aragilar @aparamon?).

The test failures on Appveyor look like a random network error rather than an actual problem.

takluyver:
OK, let's get this merged. Thanks @llchan!

JoeySchu:
Thanks for the work you guys have done on this. I'm currently working with a very large number of datasets, and I think the automatic conversion to bool would be a useful feature.

If there is no bitfield, I can access a dataset with:

`data = dset[:]`

and do not need to explicitly state any data types.

However, if there is even one bitfield, I need to write something like:

```
dt = [('Time', '<f8'),
        ('SubsetA', [('DataA1', '?'), ('DataA2', '<f8')]),
        ('SubsetB', [('DataB1', '<f8'), ('DataB2', '<f8')]),
        ('SubsetC', [('DataB1', '<f8'), ('DataB2', '<f8'), ('DataB3', '<f8'), ('DataB4', '<f8')])]
with dset.astype(dt):
    data = dset[:]
```

I can't read compound datasets in a h5 file which have bitfields without explicitly casting, and when working with a large number of distinct datasets (100+) most of which have bitfields, it would improve the usability if it would automatically cast to bool as I could simply use `data = dset[:]`. Any further work done on this would be greatly appreciated :)

llchan:
@JoeySchu would you mind opening a new issue for that? Based on the discussion in this thread it seemed like there was some hesitation from other maintainers/contributors to do that in any global way, whether on-by-default or via a config parameter, so I decided to keep this PR pretty minimal.  I think we'll need to reach consensus in a separate discussion before implementing any automatic conversions.

takluyver:
Does numpy itself have a more convenient way to convert one field in a compound datatype to a different type?

+1 to having a new issue to discuss defaults or other convenience features. I'd still push back on converting Bitfield8 to bool by default - it doesn't seem to match the intent of the HDF5 type, and it would be an odd one out if we added conversion of the other bitfield types to unsigned integers.

danieljfarrell:
I have also run into this problem. I'm trying to move away from pytables to h5py, but I'm reading old H5 files with `TypeBitfieldID` types.

What's the current status of this issue? Should I hold off for the next release when this is included?

takluyver:
I believe this is in a release already (2.10). You can use code like this to read the dataset:

```python
with dset.astype(np.bool):
    data = dset[:]

**Comment by danieljfarrell** [comment]
Thanks for replying so quickly. I have installed 2.10 and yes, I can see this is now working when reading dataset or arrays of groups!

However, I have booleans stored as attributes of groups in some H5 files and I can't see a way of reading those without raising the same exception. Any ideas?

Here is a test that shows what I mean. It does the following:
1. Uses `pytables` to write H5 file
2. Creates a single group that contains a boolean array
3. Adds various boolean attributes to the group.
4. Open the file with `h5py`
5. Uses `with dset.astype(np.bool)` casting to read the array
6. Fails read the bool attributes because it seems they cannot be cast.

```python
import h5py
import tables
import numpy as np


def write_test_file():
    with tables.open_file("test.h5", mode="w", title="Test file") as f:
        group = f.create_group("/", 'test', 'Test group')
        # Write bool attributes to group
        group._v_attrs['pybool'] = False
        group._v_attrs['npbool'] = np.bool(False)
        group._v_attrs['npboolarr'] = np.array([False, False])
        # Write bool array to group
        f.create_array(group, 'npboolarr', np.array([False, False]), "Test bool array")


def read_test_file():
    with h5py.File("test.h5", "r") as f:

        group = f['test']

        # Try without casting
        try:
            dset = group['npboolarr']
            data = dset[:]
            assert np.allclose(data, np.array([False, False]))
        except Exception as e:
            print("[EXPECTED TO FAIL] FAILED numpy bool array without casting ({})".format(str(e)))
        else:
            print("PASSED numpy bool array without casting.")

        # Try and read back the array
        try:
            dset = group['npboolarr']
            data = dset
            with dset.astype(np.bool):
                data = dset[:]
        except Exception as e:
            print("FAILED numpy bool array with casting ({})".format(str(e)))
        else:
            print("PASSED numpy bool array with casting")


        attrs = group.attrs
        # Try and read back in the attributes
        try:
            attrs['pybool']
        except Exception as e:
            print("FAILED Python bool attribute ({})".format(str(e)))
        else:
            print("PASSED Python bool attribute")

        try:
            attrs['npbool']
        except Exception as e:
            print("FAILED Numpy bool attribute ({})".format(str(e)))
        else:
            print("PASSED Numpy bool attribute")

        try:
            attrs['npboolarr']
        except Exception as e:
            print("FAILED Numpy bool array attribute ({})".format(str(e)))
        else:
            print("PASSED Numpy bool array attribute")


if __name__ == '__main__':
    write_test_file()
    read_test_file()

# Expected outout:
# [EXPECTED TO FAIL] FAILED numpy bool array without casting (No NumPy equivalent for TypeBitfieldID exists)
# PASSED numpy bool array with casting
# FAILED Python bool attribute (No NumPy equivalent for TypeBitfieldID exists)
# FAILED Numpy bool attribute (No NumPy equivalent for TypeBitfieldID exists)
# FAILED Numpy bool array attribute (No NumPy equivalent for TypeBitfieldID exists)

```

takluyver:
Hmmm, attributes might be trickier. Try something like this:

```python
attr = attrs.get_id('pybool')
arr = np.zeros(1, dtype=bool)
attr.read(arr)
```

[attr.get_id()](http://docs.h5py.org/en/stable/high/attr.html#AttributeManager.get_id) gets a low-level [AttrID](http://api.h5py.org/h5a.html#h5py.h5a.AttrID) for a specific attribute. Then `attr.read()` should copy the attribute's data into a numpy array.

danieljfarrell:
@takluyver that worked perfectly!

Updated script for future people,

```python
import h5py
import tables
import numpy as np
print("h5py: ", h5py.__version__)
print("tables: ", tables.__version__)
print("np: ", np.__version__)


def write_test_file():
    with tables.open_file("test.h5", mode="w", title="Test file") as f:
        group = f.create_group("/", 'test', 'Test group')
        # Write bool array to group
        f.create_array(group, 'npboolarr', np.array([False, False]), "Test bool array")
        # Write bool attributes to group
        group._v_attrs['pybool'] = False
        group._v_attrs['npbool'] = np.bool(False)
        group._v_attrs['npboolarr'] = np.array([False, False])



def read_test_file():
    with h5py.File("test.h5", "r") as f:

        group = f['test']

        # Try without casting
        try:
            dset = group['npboolarr']
            data = dset[:]
            assert np.allclose(data, np.array([False, False]))
        except TypeError as e:
            if e.args[0] != "No NumPy equivalent for TypeBitfieldID exists":
                print("FAILED Expected TypeBitfieldID error was generated without casting ({})".format(str(e)))
            else:
                print("PASSED Expected TypeBitfieldID error was generated without casting")
        else:
            print("FAILED Expected TypeBitfieldID error was generated without casting")

        # Try and read back the array
        try:
            dset = group['npboolarr']
            data = dset
            with dset.astype(np.bool):
                data = dset[:]
        except Exception as e:
            print("FAILED numpy bool array with casting ({})".format(str(e)))
        else:
            print("PASSED numpy bool array with casting")


        attrs = group.attrs
        # Try and read back in the attributes
        try:
            attr = attrs.get_id('pybool')
            arr = np.zeros(1, dtype=bool)
            attr.read(arr)
            assert np.isclose(arr, np.array([False]))
        except Exception as e:
            print("FAILED Python bool attribute ({})".format(str(e)))
        else:
            print("PASSED Python bool attribute")

        try:
            attrs.get_id('npbool')
            arr = np.zeros(1, dtype=np.bool)
            attr.read(arr)
            assert np.isclose(arr, np.array([False]))
        except Exception as e:
            print("FAILED Numpy bool attribute ({})".format(str(e)))
        else:
            print("PASSED Numpy bool attribute")

        try:
            attrs.get_id('npboolarr')
            arr = np.zeros(2, dtype=np.bool)
            attr.read(arr)
            assert np.allclose(arr, np.array([False]))
        except Exception as e:
            print("FAILED Numpy bool array attribute ({})".format(str(e)))
        else:
            print("PASSED Numpy bool array attribute")


if __name__ == '__main__':
    write_test_file()
    read_test_file()

# Output:
# h5py:  2.10.0
# tables:  3.5.1
# np:  1.17.4
# PASSED Expected TypeBitfieldID error was generated without casting
# PASSED numpy bool array with casting
# PASSED Python bool attribute
# PASSED Numpy bool attribute
# PASSED Numpy bool array attribute
```

takluyver:
Great! You should also be able to use `attr.shape` to create an array of the right shape automatically.

takluyver:
As in: `np.zeros(attr.shape, dtype=np.bool)`

PR #835, opened 2017-01-25
Fix version_tuple to make it PEP440 compliant

### LINKED ISSUE #834: ValueError: invalid version number '2.7.0rc3'
See logs in [travis](https://travis-ci.org/h5py/h5py/jobs/195097900) or [Debian](http://debomatic-amd64.debian.net/distribution#unstable/h5py/2.7.0~rc3-1/buildlog) builds.

aragilar:
Fixes #834. We could switch to https://github.com/warner/python-versioneer, which would simplify releases, and actually validate versions (unlike what I've done), but that's a decision for after 2.7 is released.

### DIFF
--- a/h5py/version.py
+++ b/h5py/version.py
@@ -13,22 +13,30 @@
 
 from __future__ import absolute_import
 
+from collections import namedtuple
 from . import h5 as _h5
-from distutils.version import StrictVersion as _sv
 import sys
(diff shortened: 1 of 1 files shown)


tacaswell:
Gah, sorry about this :cry: I did not build because changing the strings seemed safe.

Issue #818, opened 2017-01-10
Encoding error on Windows I/O with Python 3.6


JGoutin:
I use H5py 2.7rc2 with Python 3.6.0 64bits on Windows 7 64bits.

In this example, I try to write and read two files named `test_e.h5` and `test_é.h5` : 

Theses problems are new (Only since I updated to Python 3.6).

Maybe linked to [PEP529](https://docs.python.org/3/whatsnew/3.6.html#pep-529-change-windows-filesystem-encoding-to-utf-8) change.

#### Write test : 

```python
f = h5py.File('test_e.h5', 'w')
f.close()
f = h5py.File('test_é.h5', 'w')
f.close()
print('Directory content after test:', os.listdir())
```
Result : 
```python
Directory content after test:  ['test_e.h5', 'test_Ã©.h5']
```
Bad file name: `test_Ã©.h5` in place of `test_é.h5`

#### Read Test

```python
print('Directory content before test: ', os.listdir())
f = h5py.File('test_e.h5', 'r')
f.close()
f = h5py.File('test_é.h5', 'r')
f.close()
```

Result :
```python
Directory content before test:  ['test_e.h5', 'test_é.h5']

Traceback (most recent call last):

  File "D:/Dev/format_Hdf5.py", line 16, in <module>
    f = h5py.File('test_é.h5', 'r')

  File "d:\app\python36\lib\site-packages\h5py\_hl\files.py", line 271, in __init__
    fid = make_fid(name, mode, userblock_size, fapl, swmr=swmr)

  File "d:\app\python36\lib\site-packages\h5py\_hl\files.py", line 101, in make_fid
    fid = h5f.open(name, flags, fapl=fapl)

  File "h5py\_objects.pyx", line 54, in h5py._objects.with_phil.wrapper (D:\Build\h5py\h5py-2.7.0rc2\h5py\_objects.c:2853)

  File "h5py\_objects.pyx", line 55, in h5py._objects.with_phil.wrapper (D:\Build\h5py\h5py-2.7.0rc2\h5py\_objects.c:2811)

  File "h5py\h5f.pyx", line 78, in h5py.h5f.open (D:\Build\h5py\h5py-2.7.0rc2\h5py\h5f.c:2130)

OSError: Unable to open file (Unable to open file: name = 'test_é.h5', errno = 2, error message = 'no such file or directory', flags = 0, o_flags = 0)
```
Existing file `test_é.h5` not found.

andreabedini:
[...]

PR #836, opened 2017-01-28
Fix encoding error on Windows with Python 3.6

JGoutin:
Fix for #818 

And tried to add Python 3.6 to AppVeyor test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,8 @@ env:
   - TOXENV=py34-test-mindeps
 #  - TOXENV=py35-test-deps
 #  - TOXENV=py35-test-mindeps
+#  - TOXENV=py36-test-deps
+#  - TOXENV=py36-test-mindeps
 # commented out because of https://github.com/travis-ci/travis-ci/issues/4794
   - TOXENV=py26-test-deps
     TOX_TESTENV_PASSENV=LANG LC_ALL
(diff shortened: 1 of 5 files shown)


aragilar:
There are a few things missing from the 3.6 CI support, I've created a separate pull request #838 for that.

The tests are failing on python 3 as the new test needs to be skipped on systems without unicode support, and incorrectly passing on python 2 as they are being interpreted as bytes (need u'' prefix).

The changes to _fscodec are quite big, which will make it a pain to remove when python 2.6/2.7 is dropped. Making two new functions would reduce the patch (there are only four places which use fsencode, and two which use fsdecode), and make it easier to maintain the compat code (we can remove _fscodec on python 2.6/2.7 being dropped, and fspath on python 3.5 being dropped). Also, it'd make it clearer for new code which needs to deal with filenames (e.g. #809).

@JGoutin can you make the necessary changes, or do you want me to go ahead and do them?

JGoutin:
I make it for have the following easy:

To remove the `fscodec ` patch to drop Python 2.6/2.7, you only need to remove lines 43 to 93 and replace them by 
```
from os import fsencode as os_fsencode
from os import fsdecode as os_fsdecode
```

I didn't change anything to `_fscodec`, only change the try/except to define function only if needed (Previously, functions were defined everytime), and renamed them to `os_fsencode`, `os_fsdecode`

To remove `fspath `patch to drop Python 3 < 3.6, you only need to remove lines 8 to 40 and replace them by `from os import fspath`.

Code for `fscodec ` patch for Python >= 3.6 is only after line 96.


You can go ahead if you have a better idea in mind.  :wink:


Changes I did:
+ Updated with the 'u' prefix.
+ Fixed Python 3.6 CI with your PR.
+ Not skipped test for systems without Unicode support, Is there as function for detect this ?

JGoutin:
Updated in #841

PR #840, opened 2017-01-31
Use u prefix instead of six.u

aragilar:
As we dropped Python 3.2 support in 2.7, it makes sense to remove unneeded compatibility code. This removes calling six.u, which is only needed for Python 3.0-3.2.


tacaswell:
Using unicode literals simplifies the code a bit, but can lead to some nasty side-effects.  :+1: on this approach.

PR #841, opened 2017-01-31
Work around changes in PEP 529, and document how to pass filenames to h5py

### LINKED ISSUE #839: Support the entire set of filename on Windows
This is related to #818, but is for tracking a complete solution to how filenames are dealt with on Windows, which we currently don't have. Note that this going to be a breaking change, but fairly limited (users of external links and VDS on Windows). A summary of the issues can be found in https://www.python.org/dev/peps/pep-0528/ and https://www.python.org/dev/peps/pep-0529/, which caused this issue to come up again. A discussion of unicode and its encoding, especially on Windows, can be found at http://utf8everywhere.org/.

Background:
Nearly every Windows still used (including XP), unlike Unix-like systems, has two different interfaces to call filesystem functions, a ANSI (char) interface and a unicode (wchar) interface. Additionally, Windows has for some time now used (effectively, see https://simonsapin.github.io/wtf-8/ for caveats) the utf-16 encoding of the unicode path as the canonical path.

The ANSI interface, similar to Unix-like locale, uses the concept of a current encoding (which for python is represented by the mbcs encoding), which varies from system to system, but is almost never one of the utf-* encodings (instead being latin-1, or being an encoding native to the user's language). This means most filepaths cannot be represented on a single system. This creates interoperability problems which are sometimes hard to diagnose.

Issue:
Python since 2.3 (https://www.python.org/dev/peps/pep-0277/), has supported the unicode interfaces, and has since Python 3 (until Python 3.6, see PEP 529), marked the use of bytes (i.e. the ANSI interfaces) for filepaths as deprecated (https://docs.python.org/3/library/os.path.html).

HDF5, most likely due to its Unix-heritage, has only supported the ANSI interfaces (or rather, has not explicitly added support for. HDF5 has also previously deemed unicode filesystem support out of scope (https://support.hdfgroup.org/HDF5/doc1.8/Advanced/UsingUnicode/index.html, see Filenames under Caveats, Pitfalls, and Things to Watch For). While this does create problems, see e.g. https://tschoonj.github.io/blog/2014/11/06/hdf5-on-windows-utf-8-filenames-support/ and the other links in https://github.com/h5py/h5py/issues/818#issuecomment-271573752, the major problem comes with external links (and VDS, but that is newer and probably less widely used at the moment), which do not specify what encoding the filename is in.

Solution:
Add support for the unicode interfaces, which given the current API of the external links and VDS would be using UTF-8 for storage, and for HDF5 to call out to the unicode interfaces rather than the ANSI interfaces. Also, we need to deal with old files which may have different encodings. This means we need to complete the following tasks:

- [ ] Create a working patch for HDF5 which adds support for the unicode interfaces using UTF-8, raising errors where needed
- [ ] Have the external link code in h5py raise an error when there's a problem, and in such a way that working around different encodings is possible, and is documented
- [ ] Document how to deal with filepaths in user code (e.g. unicode over bytes for portability)

aragilar:
This is a smaller version of #836, which will allow easier cleanup in compat.py, and makes it easier to fix #839.

### DIFF
--- a/docs/high/file.rst
+++ b/docs/high/file.rst
@@ -118,6 +118,52 @@ of the HDF5 library.  However, once the file is closed you are free to read and
 write data at the start of the file, provided your modifications don't leave
 the user block region.
 
+
+.. _file_filenames:
+
+Filenames on different systems
+------------------------------
+
+Different operating systems (and different file systems) store filenames with
+different encodings. Additionally, in Python there are at least two different
+representations of filenames, as encoded bytes (via str on Python 2, bytes on
+Python 3) or as a unicode string (via unicode on Python 2 and str on Python 3).
+The safest bet when creating a new file is to use unicode strings on all
+systems.
+
+macOS (OSX)
+...........
+macOS is the simplest system to deal with, it only accepts UTF-8, so using
+unicode paths will just work (and should be preferred).
+
+Linux (and non-macOS Unix)
+..........................
+Unix-like systems use locale settings to determine the correct encoding to use.
+These are set via a number of different environment variables, of which ``LANG``
+and ``LC_ALL`` are the ones of most interest. Of special interest is the ``C``
+locale, which Python will interpret as only allowing ASCII, meaning unicode
+paths should be preencoded. This will likely change in Python 3.7 with
+https://www.python.org/dev/peps/pep-0538/, but this will likely be backported by
+distributions to earlier versions.
+
+To summarise, use unicode strings where possible, but be aware that sometimes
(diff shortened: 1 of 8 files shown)


tacaswell:
Should this be merged for 2.7.0 ?

The time pressure on that is off a bit b/c we have code in flight with debian and this is a windows bug.

aragilar:
This can wait until 2.7.1, I didn't think we were worrying about python 3.6 in general for 2.7, but I got the impression you wanted this in 2.7 (I think I've got a few more minor cleanups like #840 sitting on one of my systems, which I'll be holding off for 2.7.1)

PR #848, opened 2017-02-19
Adds `create_dataset_like` function.

lucasb-eyer:
I've been using this function as part of my utilities for a while and thought I'd see if you're interested in having it, or something similar?

Inspired by numpy's `{empty,zeros,full,...}_like` functions.

### DIFF
--- a/docs/high/group.rst
+++ b/docs/high/group.rst
@@ -363,6 +363,23 @@ Reference
 
         :keyword exact:     Require shape and type to match exactly (T/**F**)
 
+
+    .. method:: create_dataset_like(name, other, **kwds)
+
+        Create a dataset similar to `other`, much like numpy's `_like` functions.
+
+        :param name:
+            Name of the dataset (absolute or relative).  Provide None to make
+            an anonymous dataset.
+        :param other:
+            The dataset whom the new dataset should mimic. All properties, such
+            as shape, dtype, chunking, ... will be taken from it, but no data
+            or attributes are being copied.
+
+        Any dataset keywords (see create_dataset) may be provided, including
+        shape and dtype, in which case the provided values take precedence over
+        those from `other`.
+
     .. attribute:: attrs
 
         :ref:`attributes` for this group.
--- a/h5py/_hl/group.py
+++ b/h5py/_hl/group.py
@@ -158,12 +158,12 @@ class Group(HLObject, MutableMappingHDF5):
         shape and dtype, in which case the provided values take precedence over
         those from `other`.
         """
-        kw = {k: kwupdate.get(k, getattr(other, k)) for k in [
-            'shape', 'dtype', 'chunks', 'maxshape', 'compression',
-            'compression_opts', 'scaleoffset', 'shuffle', 'fletcher32', 'fillvalue'
-        ]}
+        for k in ('shape', 'dtype', 'chunks', 'maxshape', 'compression',
+                  'compression_opts', 'scaleoffset', 'shuffle', 'fletcher32',
+                  'fillvalue'):
+            kwupdate.setdefault(k, getattr(other, k))
 
-        return self.create_dataset(name, **kw)
+        return self.create_dataset(name, **kwupdate)
 
     def require_group(self, name):
         """ Return a group, creating it if it doesn't exist.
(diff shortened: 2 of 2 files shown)


aragilar:
If you can avoid using a dictionary comprehension to make this Python 2.6 compatible, and add the function to the documentation, then this could be merged.

lucasb-eyer:
Makes sense, did both in follow-up commits. I don't have a py2.6 to test though.

lucasb-eyer:
Wait, I just backtraced a problem I've been having to this. It seems to create chunks when it shouldn't. I've added a statement to print the `kw` dict and the following code demonstrates the problem. I need pointers.

```
In [2]: import h5py

In [3]: f = h5py.File('/tmp/test.h5', 'w')                                                                                                                                                                                                                     

In [4]: import numpy as np

In [5]: f['lol'] = np.arange(25).reshape(5,5)

In [6]: f.create_dataset_like('like_lol', f['lol'])
{'maxshape': (5, 5), 'dtype': dtype('int64'), 'fletcher32': False, 'fillvalue': 0, 'compression_opts': None, 'scaleoffset': None, 'compression': None, 'shuffle': False, 'chunks': None, 'shape': (5, 5)}
Out[6]: <HDF5 dataset "like_lol": shape (5, 5), type "<i8">

In [7]: f['like_lol'].chunks
Out[7]: (5, 5)

In [8]: f['lol'].chunks

```

Notice how `chunks is None` for the original `lol` (no IPython output line) whereas it's `(5,5)` with `like_lol`, even though the print clearly shows that `None` is passed for that. Ideas?

I think I'll have to add a couple unit-tests.

lucasb-eyer:
(Sorry for posting updates as I go, but I want to avoid anyone losing time duplicating my work just because I don't post.)

Going through `dataset.make_new_dset`, and specifically following into `filters.generate_dcpl`, it seems like [when `maxshape` is set, a `chunks` will be guessed](https://github.com/h5py/h5py/blob/master/h5py/_hl/filters.py#L166-L168). This is at odds with the behaviour of `'lol'` in my example code above, where `lol` does get a `maxshape`, but doesn't get `chunks`. Any suggestions?

aragilar:
If passing maxshape enforces chunks, then maybe the best thing would be to see if the dataset is chunked or not, and if it's not, don't pass maxshape.

If you can add some tests, and sort out what's happening with maxshape, then I'm happy to merge this.

takluyver:
Is it OK for me to take this branch and make a follow-up PR? Or @lucasb-eyer are you planning to come back to it?

Would it also make sense for the method to accept a numpy array? Obviously in that case it would only copy the shape and dtype. The symmetry with numpy functions is appealing.

aragilar:
@takluyver If you want to finish this PR up go ahead.

With the accepting of a numpy array, what's the advantage of using `create_dataset_like` over just using `create_dataset`, given `create_dataset` already looks at the shape/dtype unless otherwise specified? I see no reason to block the `data` keyword being passed via `create_dataset_like`, so creating a dataset with the same properties as another dataset but with different data is still one line.

takluyver:
I'm imagining that you might want to create a dataset matching the shape and the dtype of a numpy array, and then fill it by broadcasting a smaller array. I don't have an obvious use case for that, I just suspect that the symmetry with numpy's `_like` functions will lead to people trying to pass an array to it.

But it's always easier to add a feature later than to remove something that proved to be a mistake, so I'm happy to leave it for now.

takluyver:
I went to add a test for it, but I'm not sure where I should do so. `h5py.tests.hl` seems to have very few tests in; most of the tests are in `h5py.tests.old` - but that's 'old', so maybe new tests don't belong there. It's not immediately clear what is the problem with the 'old' tests.

aragilar:
My understanding (which could be wrong, as the tests have been like that since before I started contributing to h5py), is that `h5py.tests.hl` is for tests of the high-level interface (i.e. `h5py._hl`), and `h5py.tests.old` are those which haven't migrated to `h5py.tests.hl` (or if the test is not for the high-level interface, then somewhere else?). https://github.com/h5py/h5py/blob/master/h5py/tests/hl/test_file.py would be the most logical place to put this I think, but I wouldn't be overly concerned about whether it under `hl` or `old`.

PR #864, opened 2017-04-03
Update README and installation docs.

aragilar:
This expands and updates the installation docs, and simplifies the installation suggestions in the README. It also removes references to outdated information (Python XY seems dead, we no longer appear to distribute Windows installers).

Opinions?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -5,6 +5,7 @@
 
 HDF5 for Python
 ===============
+`h5py` is a thin, pythonic wrapper around the `HDF5 <https://support.hdfgroup.org/HDF5/>`_, which runs on Python 2 (2.6-2.7), and Python 3 (3.3-3.6).
 
 Websites
 --------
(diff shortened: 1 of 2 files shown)


aragilar:
Also, it appears we never actually test that the docs build, we should do that...

README.rst
```diff
@@ -13,43 +14,23 @@ Websites
...
+------------
 
-Prerequisites
--------------
+Pre-build `h5py` can either be installed via your Python Distribution (e.g.
+`Continuum Anaconda`_, `Enthought Canopy`_) or from `PyPI`_ via `pip`_.
 
-You need, at a minimum:
+More detailed installation instructions, including how to install `h5py` with
+MPI support, can be found at: http://docs.h5py.org/en/latest/build.html.
```
tacaswell: I would use internal links here 

```
:ref:`build`
```
here

and 

```
.. _build:
```

over the section you want to link ot.

docs/build.rst
```diff
@@ -3,45 +3,90 @@
...
+Wheels
+......
+If you have an existing Python installation (e.g. a python.org download,
+or one that comes with your OS), then on Windows, MacOS/OSX, and
+Linux on Intel computers, pre-built h5py wheels can be installed via pip from
+PyPI::
 
     $ pip install h5py
 
+Additionally, for Windows users, `Chris Gohlke provides third-party wheels
```
tacaswell: Cristoph builds the windows wheels we have on pypi for us.  I wonder if those are complied against mkl.

README.rst
aragilar: Can internal links be used when the documents are used in different places (README isn't used by sphinx, only github/PyPI)?

docs/build.rst
aragilar: As far as I know, HDF5 doesn't care about BLAS, so I suspect it's going to come down to how much numpy ABI is affected by using MKL (numpy on PyPI does not use MKL afaik, for the same reason there's no windows scipy wheel).

README.rst
tacaswell: oh, never mind then.

PR #868, opened 2017-04-04
Fix py2 deadlocks

tacaswell:
attn @pitrou 

`phil` is a re-entrant lock so I am still not sure that this is safe (that is these fixed methods are not called from other decorated methods.  Could you try running a version with this patch on the system where you found this?

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -16,6 +16,8 @@ from __future__ import absolute_import
 import posixpath as pp
 import sys
 
+from threading import local
+
 import six
 from six.moves import xrange    # pylint: disable=redefined-builtin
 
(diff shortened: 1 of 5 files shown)


pitrou:
This was found on a customer's system and they have applied a workaround already. Still, I will point them to this PR just in case.

h5py/_hl/dataset.py
```diff
@@ -310,21 +311,20 @@ def fillvalue(self):
         self._dcpl.get_fill_value(arr)
         return arr[0]
 
-    @with_phil
     def __init__(self, bind):
         """ Create a new Dataset object by binding to a low-level DatasetID.
         """
         from threading import local
```
pitrou: This one could probably be moved to the toplevel (no circular import here).

pitrou:
No particular feedback from our customer. This PR looks good to me and I would recommend merging.

Issue #379, opened 2013-12-27
python3: attributes are byte strings


jeremysanders:
There seems to be an attribute problem with python3 h5py (2.2.1). Reading text attributes produces bytestrings:

```
$ python2
Python 2.7.5+ (default, Sep 19 2013, 13:48:49)
[GCC 4.8.1] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import h5py
>>> f = h5py.File('test.hdf5')
>>> f['testds'] = [1,2,3]
>>> f['testds'].attrs['test'] = 'hello world'

$ python3
Python 3.3.2+ (default, Oct  9 2013, 14:50:09)
[GCC 4.8.1] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import h5py
>>> f = h5py.File('test.hdf5')
>>> f['testds'].attrs['test']
b'hello world'
```

titusjan:
Can I politely inform about the status of this bug? Is it a difficult issue to solve, or are there other priorities?

profxj:
Indeed this is causing me some havoc
+1

shoyer:
See https://github.com/h5py/h5py/pull/871 for a fix. There are a few options for the desired behavior, so if anyone has opinions about how to handle deserializing invalid ASCII, please take a look.

tacaswell:
Going back to the original post, in both python2 and python3 you are getting back bytes and you can unconditionally call `decode` on the returned object.

aimerson:
Has this incompatibility issue with strings been fixed yet? I'm using h5py 2.8.0 with python 3.7 and am finding that this bug is causing me so many nightmares. (Starting to wish I'd stuck with python 2.7... at least my code worked back then.)

shoyer:
I don't think anything has changed yet, but take a look at the linked
issues. This turned out to be a bit more complicated than I initially
anticipated and I didn't have time to finish my version of the did.

vasole:
Is it really a bug? If one wants to make sure what is written under python 2 is properly read under python 3 one does not have to write ``f['testds'].attrs['test'] = 'hello world'`` but ``f['testds'].attrs['test'] = u'hello world'`` Strings under python2 where indeed bytes unless using unicode strings. To read them back as bytes seems to me the most correct thing that can be done. One has to be aware about what is being written. For single string attributes is very straightforward to write utf-8 unicode under python2. 

[...]

Issue #585, opened 2015-06-12
Read unicode strings in attributes


tritemio:
Hi,

seems like h5py cannot read unicode strings from attributes. I have files created with pytables with unicode strings in attributes that load correctly on HDFView but fail to load on h5py:

```
---------------------------------------------------------------------------
IOError                                   Traceback (most recent call last)
<ipython-input-7-4388c265af37> in <module>()
----> 1 h5file.attrs['format_name']

h5py\_objects.pyx in h5py._objects.with_phil.wrapper (D:\Build\h5py\h5py-2.4.x\h5py\_objects.c:2413)()

h5py\_objects.pyx in h5py._objects.with_phil.wrapper (D:\Build\h5py\h5py-2.4.x\h5py\_objects.c:2370)()

C:\Users\laser2002j\Anaconda\lib\site-packages\h5py\_hl\attrs.pyc in __getitem__(self, name)
     58 
     59         arr = numpy.ndarray(attr.shape, dtype=rtdt, order='C')
---> 60         attr.read(arr)
     61 
     62         if len(arr.shape) == 0:

h5py\_objects.pyx in h5py._objects.with_phil.wrapper (D:\Build\h5py\h5py-2.4.x\h5py\_objects.c:2413)()

h5py\_objects.pyx in h5py._objects.with_phil.wrapper (D:\Build\h5py\h5py-2.4.x\h5py\_objects.c:2370)()

h5py\h5a.pyx in h5py.h5a.AttrID.read (D:\Build\h5py\h5py-2.4.x\h5py\h5a.c:4896)()

h5py\_proxy.pyx in h5py._proxy.attr_rw (D:\Build\h5py\h5py-2.4.x\h5py\_proxy.c:853)()

IOError: Unable to read attribute (No appropriate function for conversion path)
```

The problem is both with python 2 and 3. I'm pushed to use unicode strings because under python 3 pytables can save only unicode strings in the TITLE attribute.

Any workaround?

subhacom:
I encountered this issue as well when reading PyTables generated file. Is it possible to get the attribute data as binary blob and do the conversion outside h5py?

andrewcollette:
Hard to say without knowing exactly what PyTables stores in the file.  What does a minimal example look like in h5dump?

(i.e. ``h5dump file.hdf5``)

subhacom:
Here is a minimal example:

```python
import h5py as h5
import tables

fd = tables.File('test.h5', 'w')
[...]

PR #871, opened 2017-04-08
Ensure scalar ASCII decodes to strings on Python 3

shoyer:
Fixes https://github.com/h5py/h5py/issues/379

This is consistent with the documented behavior for strings under [Exceptions for Python 3]( http://docs.h5py.org/en/latest/strings.html#exceptions-for-python-3) and the discussion over in #379, so in principle this could be considered a bug fix -- but it's a bug that's been in the wild for quite some time.

I do have one unresolved concern. For "Fixed-length ASCII", the documentation says:

> Technically, these strings are supposed to store only ASCII-encoded text, although in practice anything you can store in NumPy will round-trip. But for compatibility with other progams using HDF5 (IDL, MATLAB, etc.), you should use ASCII only.

The current version of my fix will break this guarantee on Python 3, because the scalar version of fixed-length ASCII is `bytes`, which this fix will try (and fail) to decode. I see a few options:

1. Keep my current fix and update the documentation to note that you can no longer read arbitrary bytes on Python 3.
2. Update my fix to explicitly decode only variable length ASCII, by using the low level API. This will also require a documentation update, because it won't be true that "when reading or writing scalar string attributes, on Python 3 they will always be returned as type str".
3. Add some sort of terrible fallback logic that *attempts* to decode ASCII, returning the original scalar (`np.string_`) if decoding fails. This is maximally backwards compatible, but also results in difficult to predict behavior (though we can document it).

Reluctantly, I think option (3) is probably the most pragmatic choice. Any other opinions?

Finally, it might be time to update the strings documentation to special case "Exceptions for Python 2" rather than Python 3. I'll update that if I have time.

### DIFF
--- a/docs/strings.rst
+++ b/docs/strings.rst
@@ -72,10 +72,19 @@ In the file, these map to fixed-width ASCII strings.  One byte per character
 is used.  The representation is "null-padded", which is the internal
 representation used by NumPy (and the only one which round-trips through HDF5).
 
-Technically, these strings are supposed to store `only` ASCII-encoded text,
-although in practice anything you can store in NumPy will round-trip.  But
-for compatibility with other progams using HDF5 (IDL, MATLAB, etc.), you
-should use ASCII only.
+How these strings are read depends upon whether they are scalar attributes or
+not:
+
+- Scalar attributes are read as native Python strings. On Python 3, that means
+  decoding from ASCII to unicode. If decoding fails, the
+  current version of h5py will issue a ``DeprecationWarning`` and return a
+  ``np.string_``. In the future, this will be an error and such strings will
+  not be readable with h5py's high level API.
+- All other values (non-scalar attributes and datasets) remain arrays with the
+  ``S`` dtype. This means that even though technically these strings are
+  supposed to store `only` ASCII-encoded text, in practice anything you can
+  store in a NumPy array will round-trip. But for compatibility with other
+  programs using HDF5 (IDL, MATLAB, etc.), you should use ASCII only.
 
 .. note::
 
@@ -103,6 +112,10 @@ only store ASCII-encoded text, without NULL bytes::
 In the file, these are created as variable-length strings with character set
 H5T_CSET_ASCII.
 
+Similarly to how fixed-length ASCII is handled, scalar attributes are read
+as native Python strings (and if that fails, a warning or error will be issued).
+Non-scalar attributes and datasets are left as arrays of bytes.
+
 
 Variable-length UTF-8
 ^^^^^^^^^^^^^^^^^^^^^
@@ -119,21 +132,6 @@ or if you create a dataset with an explicit ``unicode`` vlen type:
 They can store any character a Python unicode string can store, with the
 exception of NULLs.  In the file these are created as variable-length strings
 with character set H5T_CSET_UTF8.
-
-
-Exceptions for Python 3
-^^^^^^^^^^^^^^^^^^^^^^^
-
-Most strings in the HDF5 world are stored in ASCII, which means they map to
-byte strings.  But in Python 3, there's a strict separation between `data` and
-`text`, which intentionally makes it painful to handle encoded strings
-directly.
-
-So, when reading or writing scalar string attributes, on Python 3 they will
-`always` be returned as type ``str``, regardless of the underlying storage
-mechanism.  The regular rules for writing apply; to get a fixed-width ASCII
-string, use ``numpy.string_``, and to get a variable-length ASCII string, use
-``bytes``.
 
 
 What about NumPy's ``U`` type?
--- a/h5py/_hl/attrs.py
+++ b/h5py/_hl/attrs.py
@@ -16,6 +16,8 @@
 
 from __future__ import absolute_import
 
+import warnings
+
 import six
 import numpy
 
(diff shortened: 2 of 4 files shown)


aragilar:
I'd suggest decoding as UTF-8 (as UTF-8 is a superset of ASCII), and we deal with #585 by treating UTF-8 encoded strings and ASCII encoded strings the same (in a different PR).

I'd also suggest going with option 3, and we explicitly document that reading arbitrary data from non-opaque type, while it may work, will be dropped in a future release, and here is a workaround for old files. We could go with option 1, and give the workaround, and would force people to deal with the incorrectly encoded data sooner...

shoyer:
> I'd suggest decoding as UTF-8 (as UTF-8 is a superset of ASCII), and we deal with #585 by treating UTF-8 encoded strings and ASCII encoded strings the same (in a different PR).

Yeah, this is probably the maximally cross-compatible thing to do. No need to support writing these arrays.

> I'd also suggest going with option 3, and we explicitly document that reading arbitrary data from non-opaque type, while it may work, will be dropped in a future release, and here is a workaround for old files.

Sounds good -- but in that case, we'll need to add another method for reading raw binary attributes on `.attrs`. Maybe `attrs.read_bytes(name)` or `attrs.read_raw(name)`?

Also, if we're going to remove this behavior in the future, we should start issuing a warning when decoding fails.

shoyer:
Another option is adding a keyword argument to `get`, e.g., `attr.get(name, decode=False)`.

aragilar:
By workaround I mean using the low-level interface (e.g. using h5a.*), which makes it obvious that the code producing the file is invalid (and if someone is storing random bytes in strings, then they know how to decode it anyway). I'd also add a warning which appears when the fallback logic is used, so that users are aware that they are relying on a hack.

andreabedini:
@shoyer @aragilar Sounds like a plan! great work.

andreabedini:
> Finally, it might be time to update the strings documentation to special case "Exceptions for Python 2" rather than Python 3. I'll update that if I have time.

💯 this!

shoyer:
> By workaround I mean using the low-level interface (e.g. using h5a.*), which makes it obvious that the code producing the file is invalid (and if someone is storing random bytes in strings, then they know how to decode it anyway).

We *could* do this, but I think we should consider a high-level interface that allows for reading bytes without decoding to strings. I don't think it's that bit of a deal to add a `decode` option to the `attrs.get` (in contrast, switching to the low-level h5py API is quite a switch).

The issue is that some people *do* want to potentially use the full 8-bits for ASCII character encoding, e.g., for netCDF: https://github.com/Unidata/netcdf-c/issues/298#issuecomment-238135821

The HDF5 docs note that this is not supported, but also not enforced by the HDF5 library:
"Extended ASCII encodings (for example, ISO 8859) are not supported. This encoding policy is not enforced by the HDF5 Library. Using encodings other than ASCII and UTF-8 can lead to compatibility and usability problems."
https://support.hdfgroup.org/HDF5/doc/RM/RM_H5T.html#Datatype-SetCset

shoyer:
Maybe another argument to `.get()` is not entirely ideal, because it adds a default value, too. I'm not entirely sure this is worth adding another high level method. I guess we can wait and see if anyone complains about the deprecation warning.

shoyer:
Do we want to switch this behavior to starting decoding and warning on Python 2, as well? That would let us a slightly more consistent story for how h5py handles strings, though it could needlessly annoy Python 2 users.

shoyer:
I pushed a partial update updating the strings docs and adding a warning. (Tests still need to be updated.)

I'm looking at the handling of variable length ASCII. If we're treating the scalars as strings, I really don't see a good reason to handle these arrays as anything other than unicode strings -- we are already converting to native Python strings in `conv_vlen2str`. It's different for fixed width ASCII, which can be converted without a copy to `dtype='S'`.

h5py/_hl/attrs.py
```diff
@@ -74,12 +77,28 @@ def __getitem__(self, name):
...
+                # The low-level interface gives HDF5 strings with ASCII
+                # character sets NumPy's 'S' dtype, for which the corresponding
+                # builtin scalar type is bytes. Decode them as ASCII to convert
+                # these values to strings on Python 3 (on Python 2, bytes is
+                # already a string type).
+                try:
+                    value = value.decode('utf-8')
+                except UnicodeDecodeError:
+                    warnings.warn(
+                        'Decoding ASCII attribute failed, falling back to '
```
tacaswell: Warning needs to be updated to utf8

docs/strings.rst
```diff
@@ -72,10 +72,19 @@ In the file, these map to fixed-width ASCII strings.  One byte per character
...
 
-Technically, these strings are supposed to store `only` ASCII-encoded text,
-although in practice anything you can store in NumPy will round-trip.  But
-for compatibility with other progams using HDF5 (IDL, MATLAB, etc.), you
-should use ASCII only.
+How these strings are read depends upon whether they are scalar attributes or
+not:
+
+- Scalar attributes are read as native Python strings. On Python 3, that means
+  decoding from ASCII to unicode. If decoding fails, the
```
tacaswell: acscii -> utf8

tacaswell:
Left notes on docs just to make sure that did not get lost.

docs/strings.rst
shoyer: Thanks for noting this. This was actually intentional: I didn't want to give users ideas about storing data UTF-8 in HDF5 ASCII strings. But it's probably better to keep the keep the docs in sync with the actual implementation.

tacaswell:
I am a bit confused though, https://github.com/h5py/h5py/pull/871/files#diff-f52ed52d97865d1d1738ad9926834f1bR178 says that unicode already round-tripped correctly.  If the user gives us bytes directly, up-casting that to unicode on retrieval seems wrong (in part because we do not provide any way to pick the encoding).  This would be problematic if the user (for what ever reason) is using some other encoding.

It might be worth _forbidding_ bytes into `__setitem__` so that we can provide the enforcement (that hdf5 does not provide) that everything is encoded as utf8.

shoyer:
> It might be worth forbidding bytes into __setitem__ so that we can provide the enforcement (that hdf5 does not provide) that everything is encoded as utf8.

This does seem more consistent. I guess we can still use `np.string_` for ASCII, but now how do we support variable length ASCII? Even writing `h5py.special_dtype(vlen=bytes)` seems wrong, because these aren't really `bytes` -- they're ASCII encoded strings.

Maybe `h5py.special_dtype(vlen=h5py.ascii)`? Hypothetically, `h5py.ascii` could be a dummy subclass of `str` (without additional validation).

If only we had that `dtype='s'` ASCII string dtype we've been waiting for in NumPy...

takluyver:
Is anyone interested in picking this up again before 2.9?

HariSeldon11988:
So six-year-old issue is still not solved...
Should I remind you that Python 2 is no longer supported, so more and more people will encounter the issue when switching to Python 3?

scopatz:
It seems like there are conflicts with this now. Please feel free to update and put in a new PR!

aragilar:
I looked at rebasing this: most of the conflicts are due to whitespace cleanups.

This currently only changes scalars, which would make them a special case (and would prevent round-tripping). As I think from #1338 we've decided against this, I'm going to close this (and note that the "bug" was in the documentation, not the code).

PR #884, opened 2017-05-13
Remove pre-1.10 warnings

cpascual:
The intro paragraph is outdated since the official release of 
HDF5 1.10.0 more than one year ago. Remove the outdated
paragraphs and warnings.


aragilar:
This is good, but could you add a sentence or so to the first paragraph stating that HDF5 1.10 is required (and that this can be checked via `h5py.version`)? Otherwise this should be good to go.

cpascual:
Done

aragilar:
Thanks, merging

PR #887, opened 2017-05-24
Deduplicate and cleanup old.common and common in tests

aragilar:
This should reduce confusion about where common code should go, and reduce the number of places where unittest vs. unittest2 makes a difference.


andreabedini:
LGTM

h5py/tests/common.py
```diff
@@ -66,35 +63,33 @@ def mktemp(self, suffix='.hdf5', prefix='', dir=None):
...
-                    raise AssertionError("Item '%s' appears in a but not b" % x)
-
-            for x in b:
-                match = False
-                for y in a:
-                    if x == y:
-                        match = True
-                if not match:
-                    raise AssertionError("Item '%s' appears in b but not a" % x)
+    def assertSameElements(self, a, b):
```
tacaswell: From the comment it looks like this code was meant to be temporary and now we have just removed the conditional.  Is there no chance of backport or are we just ove-riding no matter what to be on the safe side?

aragilar: So this appeared in python2.7 (or at least I think it did, 2.7 made a whole bunch of changes to unittest, resulting in unittest2 to backport changes, which has kept up with unittest changes, so current unittest2 matches python 3.6), was deprecated in python 3.2, and dropped in python 3.3, so it's simpler if we just use our custom version across all versions (I considered replacing this function entirely, but that would complicate this patch, which was mainly about removing the duplication).

PR #902, opened 2017-06-30
Specify dtype to avoid Dataset size overflow on Windows

### LINKED ISSUE #901: Dataset size may overflow for large arrays on Windows
On Windows, `h5py.Dataset.size` may return negative values for large datasets:

```
>>> type(dataset)
h5py._hl.dataset.Dataset

>>> dataset.size
-1654967296

>>> dataset.shape
(7500000, 352)
```

This is caused by the default int being `int32` in numpy.
```
>>> import numpy
>>> numpy.prod((1000, 100, 1000, 100, 100))
-727379968
```
An easy fix would be to specify the `dtype` in `numpy.prod` when computing `size`.

kitchoi:
Fix #901

### DIFF
--- a/h5py/tests/hl/test_dataset_getitem.py
+++ b/h5py/tests/hl/test_dataset_getitem.py
@@ -472,7 +472,7 @@ class TestVeryLargeArray(TestCase):
 
     def setUp(self):
         TestCase.setUp(self)
-        self.data = np.ones((2**15, 2**16), dtype='i4')
+        self.data = np.empty((2**15, 2**16), dtype=np.bool)
         self.dset = self.f.create_dataset('x', data=self.data)
 
     @ut.skipIf(sys.maxsize < 2**31, 'Maximum integer size >= 2**31 required')
--- a/h5py/tests/hl/test_dataset_getitem.py
+++ b/h5py/tests/hl/test_dataset_getitem.py
@@ -472,8 +472,7 @@ class TestVeryLargeArray(TestCase):
 
     def setUp(self):
         TestCase.setUp(self)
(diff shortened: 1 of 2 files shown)


h5py/tests/hl/test_dataset_getitem.py
```diff
@@ -452,18 +453,28 @@ def setUp(self):
...
         """ see issue #473 """
         self.assertNumpyBehavior(self.dset, self.data, np.s_[:,[0,1,2]])
 
-        
+
+class TestVeryLargeArray(TestCase):
+
+    def setUp(self):
+        TestCase.setUp(self)
+        self.data = np.empty((2**15, 2**16), dtype=np.bool)
```
tacaswell: You can create a data set without data, that is probably a better option 

```python

In [23]: d = F.create_dataset('huge3', (2**30, 2**30))

In [24]: d.size
Out[24]: 1152921504606846976

In [25]: d.shape
Out[25]: (1073741824, 1073741824)
```

kitchoi:
Thanks @tacaswell 
I  just discovered there is a `test_dataset` under `h5py.tests.old`.  The test in this PR looks like it belongs there.  However the module is called `old`, so I am less sure because of that.  I will leave the test as it is now but I am happy to move it across if that's more appropriate.

The test that fails on CI (`h5py.tests.old.test_file.TestFileOpen`) seems to have nothing to do with `Dataset`, so I am not sure how it is related.

aragilar:
That looks like we hit #888 (or at least something similar to it), I've restarted the appveyor test, let's see if it passes this time...

tacaswell:
'power cycled' to re-run CI against current master.

PR #904, opened 2017-07-10
Avoid unaligned memory accesses in conversion functions

jrtc27:
No guarantees are provided for the alignment of the buffers used for the various conversion functions, and in the case of packed structs (easily creatable with np.dtype, and in fact the test suite does so on the odd occasion) they definitely can be unaligned. Therefore exclusively use memcpy for multi-byte accesses to these buffers. The test suite now passes on Linux sparc64, which was previously crashing due to unaligned accesses.

### DIFF
--- a/h5py/_conv.pyx
+++ b/h5py/_conv.pyx
@@ -162,30 +162,35 @@ cdef int conv_vlen2str(void* ipt, void* opt, void* bkg, void* priv) except -1:
     cdef char** buf_cstring = <char**>ipt
     cdef PyObject* temp_obj = NULL
     cdef conv_size_t *sizes = <conv_size_t*>priv
+    cdef PyObject* bkg_obj0
+    cdef char* buf_cstring0
+
+    memcpy(&bkg_obj0, bkg_obj, sizeof(bkg_obj0))
+    memcpy(&buf_cstring0, buf_cstring, sizeof(buf_cstring0))
 
     # When reading we identify H5T_CSET_ASCII as a byte string and
     # H5T_CSET_UTF8 as a utf8-encoded unicode string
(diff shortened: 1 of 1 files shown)


jrtc27:
AppVeyor failure is due to the Python35 (32-bit) build:

> Failed to download hdf5 version 1.8.17, exiting

aragilar:
Reran appveyor, LGTM

tacaswell:
Thanks!

I'll let this set a bit longer to see if it attracts any other attention, but :+1: to merging.

Between this and #903 I think we should to a 2.7.1 next week.

aragilar:
@tacaswell That sounds good

ghisvail:
@tacaswell Feel free to ping me when `2.7.1` is out so I can update the Debian packaging. Thanks.

tacaswell:
@ghisvail If I recall correctly we were seeing failures on sparc in the debian build system.  Do I recall correctly and would it be possible to get this run through that before merging this?  Is the cost of getting that tested an rc ;) ?

jrtc27:
@tacaswell I actually tested this patch by adding it to the (admittedly, 2.7.0) Debian package and building on a sparc64 machine

ghisvail:
I fully trust James on this. I don't think an RC is necessary, better just to roll a point release out and iterate if need be.

ghisvail:
@tacaswell Are you planning a new release, or shall I apply the patch to the packaging?

tacaswell:
Hopefully this week.....I am thoroughly over-subscribed

jrtc27:
Friendly ping :)

tacaswell:
Making progress, see #913

PR #912, opened 2017-08-10
Don't capitalize error messages before printing

### LINKED ISSUE #854: Missing file error message misleadingly turned to lowercase
When trying to open a non-existing file, such as
`h5py.File('FOO.h5', 'r')  # note the uppercase`
The returned error message is:
`OSError: Unable to open file (Unable to open file: name = 'foo.h5', errno = 2, error message = 'no such file or directory', flags = 0, o_flags = 0)`
The fact that the error message mentions `foo.h5` in lowercase is extremely misleading. It is even more the case if the file foo.h5 actually exists.
The same happens when providing an absolute path: it is changed to lowercase.

evanpw:
This prevents mangling of filenames containing capital letters. Resolves #854.

### DIFF
--- a/h5py/_errors.pyx
+++ b/h5py/_errors.pyx
@@ -120,7 +120,7 @@ cdef int set_exception() except -1:
     if desc_bottom is NULL:
         raise RuntimeError("Failed to extract bottom-level error description")
 
-    msg = ("%s (%s)" % (desc.decode('utf-8').capitalize(), desc_bottom.decode('utf-8').capitalize())).encode('utf-8')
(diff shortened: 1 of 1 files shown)


tacaswell:
There is still a capitalization in the first half.

The capitalization was added in b95551234bd22ab67884cfd73a8cc73eceb266d8 by @andrewcollette I am worried _why_.

evanpw:
I haven't seen the first half contain a filename, so it capitalizing it doesn't break anything as far as I know.

PR #915, opened 2017-08-28
Updated the installation directions for parallel h5py with custom builds of MPI and HDF5  

leonfoks:


docs/build.rst
```diff
@@ -192,19 +192,42 @@ HDF5 features in h5py itself::
     $ pip install --no-binary=h5py h5py
 
 If you want access to the full Parallel HDF5 feature set in h5py
-(:ref:`parallel`), you will further have to build in MPI mode.  This can either
-be done with command-line options from the h5py tarball or by::
+(:ref:`parallel`), you will further have to build in MPI mode.
 
-    $ export HDF5_MPI="ON"
```
tacaswell: why is this export no longer needed?

leonfoks: I thought that by using --mpi in the configure step, the environment variable export was superfluous?

aragilar:
There's a number of problems with the changes:
 1. Using setup.py over pip. This is going to conflict with new packaging standards, and is going to cause people to run into more packaging bugs.
 2. The suggestion is to use a git checkout, which isn't necessary, and is again going to cause headaches for users.
 3. The no-cache option is not the option you want (it probably fixed the problem you ran into, which was you had an incorrectly built mpi4py/h5py in the cache), the option is no-binary=<pkg-name>, which forces building from source (see https://pip.pypa.io/en/stable/reference/pip_install/#cmdoption-no-binary), as a wheel will still be used if it's found with no-cache (and the better thing to do if you have a cache problem is delete the files in the cache).

An short intro on getting mpi4py installed, explicitly referring to the mpi4py installation instructions, is something worth having, as is the use of ldd to debug linking problems, though we probably want to mention the equivalent MacOS command (the most compatible way (due to virtualenv being built off py26 site.py) of finding the site-package path is
``` python
>>> from distutils.sysconfig import get_python_lib
>>> get_python_lib()
```
)

We probably also want to move the installation instructions from the mpi page to installation page, to reduce duplication and confusion.

takluyver:
I'm closing this, because there's been no update since @aragilar's comments well over a year ago.

@leonfoks if you want to come back to it, let us know, and we can reopen the PR. But having PRs lingering which aren't going to be merged or fixed makes it harder to see what's going on.

PR #916, opened 2017-08-28
Fix compounds structs with variable-size members

### LINKED ISSUE #862: Problem with variable length datatypes
Dear h5py team,

I just ran into an issue concerning variable length datatypes. The following minimal example demonstrates the issue:

```
import numpy as np
import h5py
h5f = h5py.File('test.h5') 
varbindt = h5py.special_dtype(vlen=np.uint8)
dtype = np.dtype([('c1', varbindt), ('c2', varbindt), ('c3', varbindt)])
h5f.create_dataset('test3', dtype=dtype)
```
ValueError: Unable to insert member (Member overlaps with another member)

Curiously, it works with two columns, but not three or more. Am I missing out here something? For variable length strings, it works. I found #711 and #701, but didn't see the relation to this issue.

sethrj:
Fixes #862 .

### DIFF
--- a/h5py/h5t.pyx
+++ b/h5py/h5t.pyx
@@ -14,7 +14,6 @@
     subclasses which represent things like integer/float/compound identifiers.
     The majority of the H5T API is presented as methods on these identifiers.
 """
-
 # Pyrex compile-time imports
 include "config.pxi"
 from _objects cimport pdefault
@@ -1464,21 +1463,21 @@ cdef TypeCompoundID _c_compound(dtype dt, int logical, int aligned):
     # Compound datatypes
 
     cdef hid_t tid
-    cdef TypeID type_tmp
-    cdef dtype dt_tmp
-    cdef size_t offset
-    cdef size_t accum_offset = 0
+    cdef TypeID member_type
+    cdef dtype member_dt
+    cdef size_t member_offset = 0
 
     cdef dict offsets = {}
     cdef list fields = []
 
     # The challenge with correctly converting a numpy/h5py dtype to a HDF5 type
     # which is composed of subtypes has three aspects we must consider
-    # 1. numpy/h5py dtypes do not always have the same size and HDF5, even when
+    # 1. numpy/h5py dtypes do not always have the same size as HDF5, even when
     #   equivalent (can result in overlapping elements if not careful)
     # 2. For correct round-tripping of aligned dtypes, we need to consider how
-    #   much padding we need
+    #   much padding we need by looking at the field offsets
     # 3. There is no requirement that the offsets be monotonically increasing
+    #  (so we start by sorting the names as a function of increasing offset)
     #
     # The code below tries to cover these aspects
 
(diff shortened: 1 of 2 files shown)


sethrj:
@aragilar Since this relates to work you did (which was approved by @andreabedini ) could you take a look at the new compound implementation?

h5py/h5t.pyx
```diff
@@ -1464,57 +1463,58 @@ cdef TypeCompoundID _c_compound(dtype dt, int logical, int aligned):
...
+    for name in sorted(dt.names, key=offsets.__getitem__):
+        field = dt.fields[name]
+        name = name.encode('utf8') if isinstance(name, unicode) else name
+
+        # Get HDF5 data types and set the offset for each member
+        member_dt = field[0]
+        member_offset = max(member_offset, field[1])
+        member_type = py_create(member_dt, logical=logical, aligned=aligned)
+        if aligned:
+            if member_offset > field[1] or member_dt.itemsize != member_type.get_size():
```
aragilar: Don't we only care if the member is too big, not too small? (in reference to `member_dt.itemsize != member_type.get_size()`)

aragilar:
This looks good, tests look good. So long we sort out whether we need to worry about the member being too small, I think this can be merged (it's definitely cleaner than my version).

h5py/h5t.pyx
sethrj: Thanks for checking this out. I think that if the compound is aligned, the type sizes *should* be equivalent.  I could envision something weird happening if the numpy and h5py sizes aren't the same but are padded to the same alignment. In the case of the vlen members that this tries to fix, the vlen `void*` in numpy is the size of two `void*` in HDF5. If (for whatever reason) your numpy array is padded so that the next member after a variable-length member is an extra `sizeof(void*)`, then I think we need to check that both the offset isn't too big and that the member is the same size.

But you did make me realize that there's an inconsistency in the final member; if I'm doing
```py
        member_offset = max(member_offset, field[1])
```
then I should also be doing after the loop
```py
    member_offset = max(member_offset, dt.itemsize)
```

I'll push that change.

sethrj:
Also @aragilar, what are the chances of getting this merged before the 2.7.1 patch? 😬

h5py/h5t.pyx
```diff
@@ -14,6 +14,7 @@
     subclasses which represent things like integer/float/compound identifiers.
     The majority of the H5T API is presented as methods on these identifiers.
 """
+from __future__ import print_function
```
aragilar: I'm presuming this is left over from debugging?

h5py/h5t.pyx
aragilar: I'd be concerned if the numpy and h5py sizes were different when aligned, as that would imply some underlying bug in either numpy or hdf5 (as they should both be using cpu specific types, or well defined generic types, both of which should have a single representation), I was more concerned about unusual padding, but that would also imply a bug in either numpy or hdf5, so ignore my previous comment.

aragilar:
@sethrj We definitely should make sure this gets into 2.7.1.

h5py/h5t.pyx
sethrj: Yeah, sorry. Pushing its removal.

sethrj:
@aragilar Excellent, this should be good to go then.

sethrj:
@tacaswell Can you add to the 2.7.1 milestone and merge when appropriate? Thanks!

PR #857, opened 2017-03-22
Simplifies tox by using factors

aragilar:
This makes the tox file much cleaner (plus it allows for testing against multiple python version against mpi without any major changes).

PR #922, opened 2017-09-11
Update Travis CI to correctly use Ubuntu Trusty, plus additional CI/Testing improvements

aragilar:
This is a mix of different things:
 * HDF5 1.10 support (from #894)
 * Updating Travis to use Ubuntu Trusty (from #894) - Travis has now migrated to only Trusty, which is currently causing travis to fail.
 * A simplification of tox.ini (from #857)


aragilar:
Closed and reopened to cycle CIs, there should be a better way...

aragilar:
Appveyor appears to be playing up, the ci keeps getting stuck at removing a non-empty directory with rmtree...

aragilar:
So the cause appears to be that shutil.rmtree doesn't always work on windows, which causes problems with removing the tmpdirs (see https://bugs.python.org/issue29982), I've dropped the non-WIndows support from get_hdf5.py (and so have moved it under the appveyor dir), and have wrapped the whole thing in an try..except with OSError, but if we still have problems, then I might just not use tmpdirs, and let the fact that this is only running on appveyor mean that I won't bother about cleanup code.

tacaswell:
:tada: Congrats on defeating both CI services!

PR #926, opened 2017-09-18
ENH: Deal with multiple float types of the same size

aragilar:
We currently assume that there is a numpy float type per size. If/when numpy gets true IEEE-128bit floats, there will be a conflict between the long double and IEEE-128bit floats on systems where longdouble is padded to 128 bits (e.g. x86-64), or where longdouble isn't a IEEE-128bit float.

### DIFF
--- a/h5py/h5t.pyx
+++ b/h5py/h5t.pyx
@@ -32,12 +32,14 @@ from warnings import warn
 from h5 import get_config
 import numpy as np
 from ._objects import phil, with_phil
+from .h5py_warnings import H5pyDeprecationWarning
 import platform
 
 try:
     from collections.abc import Mapping
 except ImportError:
     from collections import Mapping
+
 
 cfg = get_config()
 
@@ -269,15 +271,15 @@ class _DeprecatedMapping(Mapping):
         self._message = message
 
     def __len__(self):
-        warn(self._message, DeprecationWarning)
+        warn(self._message, H5pyDeprecationWarning)
         return len(self._mapping)
 
     def __iter__(self):
-        warn(self._message, DeprecationWarning)
+        warn(self._message, H5pyDeprecationWarning)
         return iter(self._mapping)
 
     def __getitem__(self, key):
-        warn(self._message, DeprecationWarning)
+        warn(self._message, H5pyDeprecationWarning)
         return self._mapping[key]
 
 available_ftypes = dict()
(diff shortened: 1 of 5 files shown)


h5py/h5t.pyx
```diff
@@ -222,10 +235,23 @@ cdef dict _order_map = { H5T_ORDER_NONE: '|', H5T_ORDER_LE: '<', H5T_ORDER_BE: '
...
+            available_ftypes[np.dtype(ftype).itemsize].append(ftype)
+
+    sorted_ftypes = []
+    for size, ftypes in sorted(available_ftypes.items()):
+        for ftype in sorted(ftypes, key=cmp_ftype):
+            if ftype not in sorted_ftypes:
+                sorted_ftypes.append((ftype, np.finfo(ftype), size))
+    return tuple(sorted_ftypes)
+
+available_ftypes = _get_available_ftypes()
```
tacaswell: This is an API change, `h5py.h5t.available_ftypes` is accessible to users which means someone is depending on it.  I think the right thing to do here is to leave this dict as-is and create a new private attribute for this data structure.

aragilar: I've replaced `available_ftypes` with `_available_ftypes`, so it's explicit that the variable is internal, and to avoid shadowing the old dict.

The problem with keeping the old dict is that it's not just read only, so users may expect to be able to add additional types to it and have them be supported by h5py (which won't be true). We could make `available_ftypes` a read-only dict which warns when accessed, so that users know not to use `available_ftypes`, and that they can't modify it, but as users tend to ignore warnings, and can still break user code, I'm not sure whether that's an improvement.

h5py/h5t.pyx
```diff
@@ -222,10 +235,23 @@ cdef dict _order_map = { H5T_ORDER_NONE: '|', H5T_ORDER_LE: '<', H5T_ORDER_BE: '
 cdef dict _sign_map  = { H5T_SGN_NONE: 'u', H5T_SGN_2: 'i' }
 
 # Available floating point types
-available_ftypes = dict()
```
tacaswell: This is still missing at the module level.

aragilar:
Got the deprecation stuff working, so that we break as little code as possible.

trDjango:
how do i solve my problem? python:3.6 numpy:1.14 thans for helps.
```
/usr/local/lib/python3.6/dist-packages/h5py/__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.
  from ._conv import register_converters as _register_converters
```

aragilar:
@tacaswell Was there anything else that needed to be added before merging this?

2017develper:
i have the same problem , anyone help me please

thomasaarholt:
Over at hyperspy we would like this merged due to the float/np.floating warnings.
The PR looks good with respect to these, but I can't comment on the rest of the code.

charlesreid1:
For reference, get 2.8.0rc1 on the [releases page](https://github.com/h5py/h5py/releases), or with pip:

```
pip install "h5py==2.8.0rc1"
```

h5py/tests/common.py
```diff
@@ -20,11 +20,11 @@
 import numpy as np
 import h5py
 
-if sys.version_info[:2] == (2, 6):
+if sys.version_info[0] == 2:
```
jakirkham: Why is this change needed?

jakirkham: Ah ok. Was not aware that new Python 3 behavior was also being added to `unittest`. Was under the impression that it was only Python 2.7 backports to older Python versions. Thanks for the pointer.

PR #936, opened 2017-10-10
FIX: Make libver default to 'earliest' as documented

### LINKED ISSUE #933: libver=earliest is not the default appearently
When opening a hdf5 file created with an older version of hdf5 C library it gets corrupt.

Please find details at [this stackoverflow thread](https://stackoverflow.com/questions/46527176/hdf5-made-with-h5py-py2-corrupted-after-opening-with-h5py-in-py3/46527951#46527951)

When I explicitly open with `libver=earliest` it works.

This seems to contradict the [documentation](http://docs.h5py.org/en/latest/high/file.html#version-bounding).

## System

Fedora 25 64-bit python 3.6 h5py 2.7.0 from conda

aragilar:
Fixes #933. I remain confused what was actually happening, as as far as I can see, we were using the default settings of HDF5 (see https://support.hdfgroup.org/HDF5/doc/RM/RM_H5P.html#Property-SetLibverBounds), which were what we said they were. In any case, we're consistent with our docs now, if HDF5 ever changes its default.

### DIFF
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -49,7 +49,10 @@ def make_fapl(driver, libver, **kwds):
             high = h5f.LIBVER_LATEST
         else:
             low, high = (libver_dict[x] for x in libver)
-        plist.set_libver_bounds(low, high)
+    else:
+        # we default to earliest
+        low, high = h5f.LIBVER_EARLIEST, h5f.LIBVER_LATEST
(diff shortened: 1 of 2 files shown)


aragilar:
Cycling CIs, appveyor apparently failed to download the zip file correctly: https://ci.appveyor.com/project/h5py/h5py/build/1.0.214/job/ylrfbyni59cofb95

PR #938, opened 2017-10-12
MAINT: Add PEP-518 (a.k.a pyproject.toml) support

aragilar:
This adds PEP-518 support, which is a much improved replacement for setup_requires. Whereas setup_requires is setuptools specific, and always calls easy_install (meaning no wheels, no recent improvements to packaging etc.), PEP-518 is build-system independent (so pip and other tools can automatically get the needed dependencies). One advantage of PEP-518 for use is that by having pip (or at least newer versions of pip) deal with all of the installation, `python_depends` will be checked on all our dependencies, which means we should have less problems with our dependencies dropping older versions of python. Numpy already has `python_depends` set, and I'm sending a patch to both cython and pkgconfig.

There's a good explanation of PEP-518 at https://snarky.ca/clarifying-pep-518/, but I don't think we (h5py) should change our current setup.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -8,6 +8,7 @@ include setup_build.py
 include setup_configure.py
 include tox.ini
 include pytest.ini
+include pyproject.toml
 
 recursive-include docs *
 prune docs/_build
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,4 +2,11 @@
 # this should match SETUP_REQUIRES in setup.py, with the addition of setuptools
 # and wheel (in case some point in the future pip does not assume setuptools and
 # wheel)
-requires = ["setuptools", "wheel", "numpy>=1.7", "Cython>=0.23", "pkgconfig"]
+requires = [
+  "setuptools",
+  "wheel",
+  "numpy==1.12.0; python_version=='3.6'",
+  "numpy==1.14.3; python_version>='3.7'",
+  "Cython>=0.23",
+  "pkgconfig"
+]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,5 @@
+[build-system]
+# this should match SETUP_REQUIRES in setup.py, with the addition of setuptools
+# and wheel (in case some point in the future pip does not assume setuptools and
+# wheel)
+requires = ["setuptools", "wheel", "numpy>=1.7", "Cython>=0.23", "pkgconfig"]
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,12 @@ RUN_REQUIRES = [NUMPY_DEP,]
 # these are required to build h5py
 # RUN_REQUIRES is included as setup.py test needs RUN_REQUIRES for testing
 # RUN_REQUIRES can be removed when setup.py test is removed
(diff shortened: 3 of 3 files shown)


takluyver:
Reopening to rerun tests

takluyver:
Fewer test failures now, but there are problems on the `mindeps` test runs on Travis. I suspect it's compiling with one version of numpy and then importing a different one.

Python 3.5 and 3.6 give this error:

```
RuntimeError: module compiled against API version 0xc but this version of numpy is 0xa
  File "/home/travis/build/h5py/h5py/.tox/py36-test-mindeps/lib/python3.6/site-packages/h5py/__init__.py", line 36, in <module>
    from ._conv import register_converters as _register_converters
  File "h5py/h5t.pxd", line 14, in init h5py._conv
  File "h5py/h5t.pyx", line 1, in init h5py.h5t
SystemError: execution of module h5py.utils raised unreported exception
```

Python 2.7 and 3.4 give this error:

```
  File "/home/travis/build/h5py/h5py/.tox/py34-test-mindeps/lib/python3.4/site-packages/h5py/__init__.py", line 36, in <module>
    from ._conv import register_converters as _register_converters
  File "h5py/h5t.pxd", line 14, in init h5py._conv
  File "h5py/numpy.pxd", line 66, in init h5py.h5t
ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject
```

takluyver:
Dropping the 2.10 milestone as this has problems as stands and it's not clear what needs doing.

takluyver:
This appears to be failing on the `mindeps` tests in various environments. I think this is because it builds with a newer numpy version, and then finds an older one in the test environment.

This is highlighting a problem that's going to bite us sooner or later as it becomes more common to separate the build from the target install environment: the built package should have more tightly specified dependencies than the sdist. Specifically, it needs to depend on a version of numpy that's binary-compatible with the one it was built with.

takluyver:
I realised there's a bigger problem to solve with build isolation, even if we make the build process correctly indicate a dependency on numpy. I've started a thread on the Python forum about it: https://discuss.python.org/t/pep-517-how-to-pick-compatible-build-dependencies/2460

takluyver:
I hope you don't mind me butting in and changing your PR. Over on Discourse, Joris van den Bossche pointed out that scipy solves this by [specifying the oldest supported numpy as a build requirement](https://github.com/scipy/scipy/blob/master/pyproject.toml), so I've had a go at that.

takluyver:
It appears to be just the MPI test failing now, but I don't really understand what's happening with that.

aragilar:
I think there's some kind of conflict between versions of numpy (mpi4py I think wants 1.14 based on the error). I suspect we're going to have the same issue with mpi4py as we do for numpy (in terms of locking down the ABI), which is probably going to be all kinds of pain when dealing with MPI, so maybe we should hold off on this until something is decided at the PyPA level or people actually run into problems (I'm pretty sure wheels cover most people, and those using MPI will need to configure their environment correctly anyway)?

aragilar:
Closing this in favour of #1619.

PR #942, opened 2017-10-25
ENH: Call np.issubdtype with np.floating

effigies:
Numpy is deprecating this usage (see numpy/numpy#9505).


effigies:
Ah, just noticed this is already covered in #926.

Issue #946, opened 2017-11-06
pkgconfig and hardcoded paths unconditionally used even with explicit `--hdf5`


zao:
* Ubuntu 16.04 LTS
 * Python  2.7.12 (default, Oct 17 2017, 19:04:42) [GCC Intel(R) C++ gcc 5.4 mode]
 * Self-built Python via EasyBuild, intelcuda/2016.11 toolchain.
 * h5py 2.7.0, 2.7.1
 * Self-built HDF5 1.10.0-patch1 via EasyBuild, same toolchain.

I have several concerns about how paths are gathered in `h5py`'s configure/build scripts.
There's two sources of paths that cannot be disabled, both the hardcoded `/usr/local/include` + `/opt/local/include` and the unconditional attempt to run `pkgconfig`.

This is fine if you're trying to catch all the places where HDF5 may be installed, but on my cluster, there's a system-installed HDF5 1.8.x that should never be used by non-distro software. It's only there because it's a mandatory dependency of some other package.

I have my own HDF5 build and specify it with `--hdf5=/foo` when invoking the setup. That installation does not have any `hdf5.pc` file, and as such, doesn't shadow the file from the system installation and the pkgconfig invocation "erroneously" picks up the one from the system.

I would like to argue that if a path for an HDF5 installation is explicitly specified, the scripts should not consider the hardcoded paths nor run pkgconfig.

tacaswell:
Sounds reasonable,  would you mind taking a pass at fixing that?

zao:
Sure, I'll try to make a PR once I've sorted out the debris from building software dependent on h5py.

aragilar:
Closed via #947

PR #947, opened 2017-11-22
Prefer custom HDF5 over pkg-config/fallback paths

zao:
When specifying a custom HDF5 location, alternative locations for the
library from system pkg-config files and the hardcoded fallback paths
can interfere with proper operation of the specified library.

If a location is specified, avoid invoking pkg-config and the fallback
paths to honor the user intent.

This addresses GitHub issue #946.


aragilar:
Looks good, merging.

PR #956, opened 2017-12-05
ENH: allow registration of new drivers

llllllllll:
I am currently developing a driver for hdf5 to transparently store and read files from Amazon S3 and would like to expose bindings in Python with h5py. I searched briefly, but did not see any way to add support for a new driver in h5py so I have tried to implement a simple registration API.

This would probably need some tests and may not be the desired API, but I wanted to open this PR early to gauge interest in this feature. I have manually tested this feature against my driver and this does work, though it might not be comprehensive.

For those interested, the driver implementation is available [here](https://github.com/h5s3/h5s3). With this patch applied we have been able to read and write files, though we still have a lot of work to do before I would advise using this.

### DIFF
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -39,6 +39,53 @@ libver_dict = {'earliest': h5f.LIBVER_EARLIEST, 'latest': h5f.LIBVER_LATEST}
 libver_dict_r = dict((y, x) for x, y in six.iteritems(libver_dict))
 
 
+def _set_fapl_mpio(plist, **kwargs):
+    kwargs.setdefault('info', mpi4py.MPI.Info())
+    plist.set_fapl_mpio(**kwargs)
+
+
+_drivers = {
+    'sec2': lambda plist, **kwargs: plist.set_fapl_sec2(**kwargs),
+    'stdio': lambda plist, **kwargs: plist.set_fapl_stdio(**kwargs),
+    'core': lambda plist, **kwargs: plist.set_fapl_core(**kwargs),
+    'family': lambda plist, **kwargs: plist.set_fapl_family(
+        memb_fapl=plist.copy(),
+        **kwargs
+    ),
+    'mpio': _set_fapl_mpio,
+}
+
+
+def register_driver(name, set_fapl):
+    """Register a custom driver.
+
+    Parameters
+    ----------
+    name : str
+        The name of the driver.
+    set_fapl : callable[PropFAID, **kwargs] -> NoneType
+        The function to set the fapl to use your custom driver.
+    """
+    _drivers[name] = set_fapl
+
+
+def unregister_driver(name):
+    """Unregister a custom driver.
+
+    Parameters
+    ----------
+    name : str
+        The name of the driver.
+    """
+    del _drivers[name]
+
+
+def registered_drivers():
+    """Return a frozenset of the names of all of the registered drivers.
+    """
+    return frozenset(_drivers)
+
+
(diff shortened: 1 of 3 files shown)


h5py/_hl/files.py
```diff
@@ -39,6 +39,47 @@
...
+    plist.set_fapl_mpio(**kwargs)
+
+
+drivers = {
+    'sec2': lambda plist, **kwargs: plist.set_fapl_sec2(**kwargs),
+    'stdio': lambda plist, **kwargs: plist.set_fapl_stdio(**kwargs),
+    'core': lambda plist, **kwargs: plist.set_fapl_core(**kwargs),
+    'family': lambda plist, **kwargs: plist.set_fapl_family(
+        memb_fapl=plist.copy(),
+        **kwargs,
```
aragilar: Extra comma (which is causing the tests to fail).

h5py/_hl/files.py
```diff
@@ -39,6 +39,47 @@
 libver_dict_r = dict((y, x) for x, y in six.iteritems(libver_dict))
 
 
+def _set_fapl_mpio(plist, **kwargs):
+    kwds.setdefault('info', mpi4py.MPI.Info())
+    plist.set_fapl_mpio(**kwargs)
+
+
+drivers = {
```
aragilar: Should be private (a.k.a. `_drivers`)

h5py/_hl/files.py
```diff
@@ -39,6 +39,47 @@
...
+
+
+def register_driver(name, set_fapl):
+    """Register a custom driver.
+
+    Parameters
+    ----------
+    name : str
+        The name of the driver.
+    set_fapl : callable[PropFAID, **kwargs, NoneType]
```
aragilar: I'm confused by the NoneType, why is it there?

aragilar:
This looks good. Can you add some tests, and document how to use this at the end of http://docs.h5py.org/en/latest/high/file.html#file-drivers?

h5py/_hl/files.py
llllllllll: ah, only tested on 3.6

h5py/_hl/files.py
llllllllll: This was saying that the signature is PropFAID, then keyword arguments, with a return type of NoneType. I don't see a lot of examples for documenting callbacks so this is what I have used. I would be happy to change this to whatever you prefer. One option would just be "callable" and then describe the signature below.

llllllllll:
I am not totally sure how to test this without implementing a small driver. I could whitebox test it by mocking out set_fapl and checking that it gets called on open though.

aragilar:
You could unregister and register a builtin driver. Mocking would also work.

h5py/_hl/files.py
aragilar: If you have multiple versions of python on your path, using tox helps catch that kinda thing (I know it's saved me from creating python 3 only fixes). Enabling travis and appveyor on your fork also helps (esp. appveyor if you lack easy access to windows).

h5py/_hl/files.py
aragilar: Is `callable[PropFAID, **kwargs] -> NoneType` a thing, or is my knowledge of the new typing syntax just wrong?

llllllllll: I don't think typing uses this at all, but that is more clear than what I wrote.

llllllllll:
added some tests for registration and unregistration

llllllllll:
Thank you!

Issue #289, opened 2013-04-25
TypeError raised when setting list of unicodes as attribute


ppinard:
I get the following TypeError when trying to set an attribute as a list of unicodes. 

f = h5py.File("foo.h5", "w")
f.attrs['bar'] = [u'\u03b2', u'\u03b2', u'\u03b3']

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/h5py-2.1.2-py2.7-linux-x86_64.egg/h5py/_hl/attrs.py", line 56, in __setitem__
    self.create(name, data=value, dtype=base.guess_dtype(value))
  File "/usr/local/lib/python2.7/dist-packages/h5py-2.1.2-py2.7-linux-x86_64.egg/h5py/_hl/attrs.py", line 102, in create
    htype = h5t.py_create(dtype, logical=True)
  File "h5t.pyx", line 1361, in h5py.h5t.py_create (h5py/h5t.c:11451)
  File "h5t.pyx", line 1439, in h5py.h5t.py_create (h5py/h5t.c:11361)
TypeError: No conversion path for dtype: dtype('<U1')

h5py: 2.1.2
python: 2.7.3

alexbw:
+1

jakirkham:
+1

matteobachetti:
+1

graingert:
>  We are discussing ways to address this.

@andrewcollette Where are those discussions?

andrewcollette:
@graingert there were periodic discussions on the mailing list, although that comment was 3 years ago.  Unicode in HDF5 is something of a mess, unfortunately.  There simply isn't a destination type for "U", so logic would have to be added to store as vlen strings instead.

guddulrk:
if use_htype is None:
                htype = h5t.py_create(original_dtype, logical=True)
                htype2 = h5t.py_create(original_dtype)  # Must be bit-for-bit representation rather than logical
            else:
                htype = use_htype
                htype2 = None


and I am getting following error 
TypeError: No conversion path for dtype: dtype('<U12')

Any help please..

aragilar:
@guddulrk You need to pre-encode your strings as UTF-8, as per https://github.com/h5py/h5py/issues/289#issuecomment-17015083. See also http://docs.h5py.org/en/latest/strings.html.

jhollowed:
[...]

PR #957, opened 2017-12-06
Document error storing utf-8 strings with python 3

achalddave:
This documents issue #289 in the "Exceptions for Python 3" section of the documentation. I'm not sure if I described it well, or in the right place, but I'm happy to modify the pull request.

### DIFF
--- a/docs/strings.rst
+++ b/docs/strings.rst
@@ -135,6 +135,10 @@ mechanism.  The regular rules for writing apply; to get a fixed-width ASCII
 string, use ``numpy.string_``, and to get a variable-length ASCII string, use
 ``bytes``.
 
+Since Python 3 strings are encoded as UTF-8 by default. This causes some issues
+when storing a list of strings. For more information, and work arounds, see
+[issue #289](https://github.com/h5py/h5py/issues/289).
+
 
 What about NumPy's ``U`` type?
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^


aragilar:
So there's two parts to fixing #289 (and a bunch of similar issues), one is providing a solution to the problem (which currently is to encode the strings before using them with h5py, as mentioned in the bug), and the second is understanding why this doesn't "just work".

The reason this doesn't "just work" is that the high-level api for h5py has two principles about how it operates:
1. Types should always be written and read in the same format, so roundtripping from python->hdf5->python should always work (there's an exception that hdf5->python->hdf5 may not work if the details of how type setting differ from h5py's defaults e.g. null-pad vs. null-term vs space-pad)
2. data is (mostly) encoded in numpy arrays, that means lists/tuples etc. are converted to numpy arrays.

In this case, lists of strings (on python 2 lists of unicode) are converted to arrays with the "unicode" (a.k.a "U") dtype, which is UTF-32. h5py explicitly does not support the "unicode" dtype, as there is no way to know whether we should leave the string as ASCII/UTF-8, or convert to UTF-32 (which increases memory usage significantly).

What makes sense to do is to have an additional line at the beginning explaining how lists/arrays of strings are written by h5py, and reminding users about this in the python 3 section. Both sections should link to the the "unicode" non-support section.

docs/strings.rst
```diff
@@ -135,6 +135,10 @@ mechanism.  The regular rules for writing apply; to get a fixed-width ASCII
 string, use ``numpy.string_``, and to get a variable-length ASCII string, use
 ``bytes``.
 
+Since Python 3 strings are encoded as UTF-8 by default. This causes some issues
```
aragilar: FYI, the python 3 str/python 2 unicode types have variable encodings (see https://www.python.org/dev/peps/pep-0393/), I would avoid mentioning how they are encoded.

tacaswell:
Thank you for your work @achalddave however I do not think this should be merged so I am closing.

PR #963, opened 2017-12-18
FIX: Don't pass non-boolean ndarrays to PointSelection

### LINKED ISSUE #586: bug when indexing with a tuple of slices
The following works for N-dimensional datasets, as long as N > 1:

```
index_tuple = (slice(None), ) * (N-1) + (numpy.arange(3), )
n_dimensional_dataset[index_tuple])  # fine
```

If N == 1, however, it fails:

```
index_tuple = (numpy.arange(3), )
one_dimensional_dataset[index_tuple]

*** TypeError: PointSelection __getitem__ only works with bool arrays
```

It works fine if I stick an `Ellipsis` object at the end of `index_tuple`:

```
index_tuple = (numpy.arange(3), Ellipsis)
one_dimensional_dataset[index_tuple]  # works
```

I think the problem might be that when you give a 1-element tuple to `Dataset.__getitem__`, it for some reason takes the element out and uses that as the index, rather than the tuple. I think this because I get the same error if I do that myself:

```
index_tuple = (numpy.arange(3))
one_dimensional_dataset[index_tuple[0]]

*** TypeError: PointSelection __getitem__ only works with bool arrays
```

This guess is somewhat similar to issue #584, where `Dataset.__getitem__` erroneously treats a length-1 integer array the same way as the integer it contains. Maybe this is another manifestation of the same bug?

I am using h5py 2.2.1, the version that ships with Ubuntu Linux 14.04.

aragilar:
Fixes #586. I need to add tests, both for checking boolean and non-boolean arrays, so WIP.

### DIFF
--- a/h5py/_hl/selections.py
+++ b/h5py/_hl/selections.py
@@ -67,7 +67,7 @@ def select(shape, args, dsid):
                 raise TypeError("Mismatched selection shape")
             return arg
 
-        elif isinstance(arg, np.ndarray):
+        elif isinstance(arg, np.ndarray) and arg.dtype.kind == 'b':
             sel = PointSelection(shape)
             sel[arg]
             return sel
--- a/h5py/tests/hl/test_dataset_getitem.py
+++ b/h5py/tests/hl/test_dataset_getitem.py
@@ -407,6 +407,9 @@ class Test1DFloat(TestCase):
     def test_indexlist_simple(self):
         self.assertNumpyBehavior(self.dset, self.data, np.s_[[1,2,5]])
 
+    def test_indexlist_numpyarray(self):
+        self.assertNumpyBehavior(self.dset, self.data, np.s_[np.array([1, 2, 5])])
+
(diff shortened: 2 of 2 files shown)


takluyver:
@aragilar do you think you'll get round to this soon, or shall we bump it to a later milestone?

aragilar:
I won't have time to finish this for 2.9, pushing this off to 2.10.

takluyver:
@aragilar do you think you'll get to this for 2.10? I might have a look at adding some tests if you're not going to get to it.

takluyver:
I've added a test for selecting data on a single axis with an array of integer positions, which is the case described in #586. This test fails when run against master.

There's definitely an argument for testing the selection functionality (`h5py._hl.selections`) separately from the actual dataset interface. I can't see any direct tests of that module at the moment. But I don't think this relatively straightforward PR needs to wait for someone to write a load of tests. The functionality is already being tested through the dataset interface..

PR #964, opened 2017-12-18
ENH: Merge in doc improvements that were only in 2.5 branch.

aragilar:
This merges in the extra work which was done on the 2.5 branch, which is not included in mainline master. If there's a better way of doing this, feel free to close this PR.


tacaswell:
Power-cycled to update CI, but realize that none of this is tested!

PR #968, opened 2017-12-26
Track object creation order under group

cyfdecyf:
Changes made in this pull request:

1. Add an option `track_order` in `Group.create()` to enable object creation order tracking. Because of HDF5 limitation, this only works for group, not for root level objects.
2. If group has creation order tracking property, iterate over objects with creation order.

### DIFF
--- a/h5py/tests/old/test_group.py
+++ b/h5py/tests/old/test_group.py
@@ -395,6 +395,20 @@ class TestIter(BaseMapping):
         finally:
             hfile.close()
 
+class TestTrackOrder(BaseGroup):
+    def test_track_order(self):
+        g = self.f.create_group('order', track_order=True)
+        for i in xrange(100):
+            # Mix group and dataset creation.
+            if i % 10 == 0:
+                g.create_group(str(i))
+            else:
+                g[str(i)] = [i]
+
+        objs = [str(i) for i in xrange(100)]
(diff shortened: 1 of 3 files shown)


h5py/tests/old/test_group.py
```diff
@@ -395,6 +395,20 @@ def test_iter_zero(self):
         finally:
             hfile.close()
 
+class TestTrackOrder(BaseGroup):
+    def test_track_order(self):
+        g = self.f.create_group('order', track_order=True)
+        for i in xrange(100):
```
aragilar: Use range not xrange for python 3.

h5py/tests/old/test_group.py
```diff
@@ -395,6 +395,20 @@ def test_iter_zero(self):
...
+            # Mix group and dataset creation.
+            if i % 10 == 0:
+                g.create_group(str(i))
+            else:
+                g[str(i)] = [i]
+
+        objs = [str(i) for i in xrange(100)]
```
aragilar: Use range not xrange for python 3.

aragilar:
This looks fine, apart from failing on python 3. Once the tests pass, I'll merge this.

cyfdecyf:
Thanks for pointing out the problem. Fix committed.

PR #962, opened 2017-12-18
ENH: Add missing py36 setting to tox

aragilar:
Minor update, missed 36 in numpy deps.

PR #970, opened 2017-12-28
FIX: Don't reorder compound types, breaks on numpy 1.14 

aragilar:
This removes the reordering of compound dtypes when written or read, needed on order for h5py to work on numpy 1.14 and later.

This includes #962, so that all the tests will pass on Travis. Appveyor may fail due to #954.

### DIFF
--- a/h5py/h5t.pyx
+++ b/h5py/h5t.pyx
@@ -1136,12 +1136,6 @@ cdef class TypeCompoundID(TypeCompositeID):
         else:
             if sys.version[0] == '3':
                 field_names = [x.decode('utf8') for x in field_names]
-            if len(field_names) > 0:
-                collated_fields = zip(field_names, field_types, field_offsets)
-                ordered_fields = sorted(
-                    collated_fields, key=operator.itemgetter(2))
-                field_names, field_types, field_offsets = \
-                    map(list, zip(*ordered_fields))
             typeobj = dtype({
                 'names': field_names,
                 'formats': field_types,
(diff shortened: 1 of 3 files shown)


tacaswell:
Do we need to document this as an API change?

aragilar:
It should be strongly highlighted part of a release announcement (where we should recommend people upgrade to numpy 1.14 if they're using compound types). It's definitely an API change (the main breaking part will be when writing out out-of-order compound types), but it's fairly minimal in terms of actual breakage (reading should be unaffected, and in-order compound types are completely unaffected).

tacaswell:
power-cycled to re-run CI.

Given the amount of other stuff this is broken we should probably merge this and cut a release ASAP.

Do we want to call the next release 2.7.2 or 2.8.0?

aragilar:
Unless we only cherry-pick only the needed changes, calling a 2.8 release makes more sense.

@tacaswell Have you seen https://bors.tech/, I came across it recently, it allows retrying of tests and tests the merge, not just what's in the PR (rather than open/closing to retry, and having to rebase to check that the tests actually pass). It may be worth setting up after the 2.8 release.

tacaswell:
Bumping to 2.8 does make some sense as it is easier to say (if numpy > =1.14 then h5py >=2.8) than on a micro.

tacaswell:
Used the github web interface to fix the confilct.  Not _super_ happy about having master merged into the feature branch, but :man_shrugging: .

Will merge once this passes.

PR #985, opened 2018-01-22
FIX: Fix check

larsoner:
On latest `NumPy` dev I get a lot of these warnings when using `h5py`:
```
h5py/__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.
```
This gets rid of them, but I'm not sure if the check should be more or less specific than this, but I think this is the change suggested by the warning message.

### DIFF
--- a/h5py/h5t.pyx
+++ b/h5py/h5t.pyx
@@ -224,7 +224,7 @@ cdef dict _sign_map  = { H5T_SGN_NONE: 'u', H5T_SGN_2: 'i' }
 # Available floating point types
 available_ftypes = dict()
 for ftype in np.typeDict.values():
-    if np.issubdtype(ftype, float):
+    if np.issubdtype(ftype, np.floating):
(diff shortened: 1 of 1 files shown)


larsoner:
... looks like this fix is already part of #926. Feel free to close if it's better to leave it to that PR.

aragilar:
I'm merging this, and it should be backported to a 2.7.2 release (my PR is more complicated, so that should only go in a potential 2.8 release).

larsoner:
Let me know if you need me to do anything for the 2.7.2 backport like open a separate PR

Erotemic:
Is there any change of getting this into a release anytime soon? This warning has been appearing for months now.

aragilar:
@Erotemic See https://groups.google.com/forum/#!topic/h5py/liapKtBCmlE, the final release should be coming out soon.

Issue #719, opened 2016-06-13
OSError: Unable to read attribute (No appropriate function for conversion path) 


mangecoeur:
I'm trying to open a netCDF4 file created by Xarray (made by opening a netCDF3 file and re-saving as netCDF4 using `.to_netcdf`) and I get the error below.

``` python
import h5py as h5
with h5.File('test.nc', 'r') as fd:
    print(list(fd['TMP_L103'].attrs.items()))
```

```
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
<ipython-input-5-bd3188829b86> in <module>()
      3 fd = h5.File('test.nc', 'r')
      4 
----> 5 print(list(fd['TMP_L103'].attrs.items()))

/Users/<me>/anaconda/lib/python3.5/site-packages/h5py/_hl/base.py in __iter__(self)
    336         with phil:
    337             for key in self._mapping:
--> 338                 yield (key, self._mapping.get(key))
    339 
    340 

/Users/<me>/anaconda/lib/python3.5/_collections_abc.py in get(self, key, default)
    593         'D.get(k[,d]) -> D[k] if k in D, else d.  d defaults to None.'
    594         try:
--> 595             return self[key]
    596         except KeyError:
    597             return default

h5py/_objects.pyx in h5py._objects.with_phil.wrapper (-------src-dir--------/h5py/_objects.c:2582)()

h5py/_objects.pyx in h5py._objects.with_phil.wrapper (-------src-dir--------/h5py/_objects.c:2541)()

/Users/<me>/anaconda/lib/python3.5/site-packages/h5py/_hl/attrs.py in __getitem__(self, name)
     77 
     78         arr = numpy.ndarray(shape, dtype=dtype, order='C')
---> 79         attr.read(arr, mtype=htype)
     80 
     81         if len(arr.shape) == 0:

h5py/_objects.pyx in h5py._objects.with_phil.wrapper (-------src-dir--------/h5py/_objects.c:2582)()

h5py/_objects.pyx in h5py._objects.with_phil.wrapper (-------src-dir--------/h5py/_objects.c:2541)()

h5py/h5a.pyx in h5py.h5a.AttrID.read (-------src-dir--------/h5py/h5a.c:5123)()

h5py/_proxy.pyx in h5py._proxy.attr_rw (-------src-dir--------/h5py/_proxy.c:915)()

OSError: Unable to read attribute (No appropriate function for conversion path)
```

[...]

Issue #973, opened 2018-01-07
RFC: implement proper support for strings


wavexx:
I know I can read utf8 and somewhat work-around writing by manually encoding(), but I *need* to be able to write compound datatypes with strings with the *proper* encoding set (usually utf8) in order to write compliant hd5 files. This is currently impossible.

As outlined in other PRs and issues I've seen, python2 handling of strings should be the exception, not the norm.

After reading the source a bit, I'd propose the following scheme:

- Actually use numpy's "V" type for bytes (vlen or not) and map it properly to H5T_OPAQUE.
- Map str in python3 (or unicode in python2) to type "S", but include an extra attribute in the constructor of the dtype: ``dtype("S", metadata={"encoding": encoding})``.
- The encoding should default to CSET_UTF8 in python3, and CSET_ASCII for plain 'str' objects in python2.
- Assigning an encoding manually (for example, to write ascii string columns in python3) can be done by explicitly constructing the dtype, as currently done for vlen strings. Combining encoding and vlen information becomes the norm, not the exception. No special handling required.
- Reading is unambiguous: H5T_STRING always yields an "S" dtype, an unicode object in python2 and an str object in python3. We can now use the encoding attribute to decode the string to the current python str's object correctly.
- Type 'U' still has no equivalent mapping in h5t and should still be rejected.

This scheme allows stuffing padded, variable-length UTF8-encoded strings in numpy matrices/records without losing any encoding information. We always preserve the encoding via the metadata dict.

Is is debatable whether reading a ``CSET_ASCII`` string in python2 should yield a str or not. I'd go for unicode, always. The reasoning is that reading and later writing the very same objects should yield the same on-disk representation.

Manually reading the binary content via read_bytes is already uniform in both python2 and python3.

Any issues or insights for the above proposal? I did a few tests to make sure it's feasible, and I could work on implementing it in the next weeks.

aragilar:
@wavexx I'll need to have a look at this in detail (and so my response may not be relevant), but some things which need to keep in mind are backwards compatibility, and unix vs. windows string handling. I'm also not inclined to completely change how strings work on Python 2, given that Python 2 is going away soon (and so in the unlikely event we change a whole bunch of things about strings, we'd want to do it post dropping Python 2).

aragilar:
Going through each of the points:

> Actually use numpy's "V" type for bytes (vlen or not) and map it properly to H5T_OPAQUE.

[...]

PR #988, opened 2018-01-24
FIX: Enable reading fixed-length UTF-8 strings

ajelenak-thg:
This adds a datatype conversion function that enables reading fixed-length UTF-8 strings into the h5py's chosen fixed-length null-padded ASCII string datatype as reported in issue #719.

The PR is meant to check whether this is the correct approach and, if not, what would be a more suitable one. The new conversion function avoids changing how h5py handles strings in any way and targets only fixed-length UTF-8 strings stored in HDF5 files.

### DIFF
--- a/h5py/_conv.pyx
+++ b/h5py/_conv.pyx
@@ -23,7 +23,7 @@ np.import_array()
 
 # Minimal interface for Python objects immune to Cython refcounting
 cdef extern from "Python.h":
-    
+
     # From Cython declarations
     ctypedef int PyTypeObject
     ctypedef struct PyObject:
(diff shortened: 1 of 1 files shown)


wavexx:
This would be relevant for #973 as well. Any reason it has not been accepted yet?

aragilar:
@ajelenak-thg This fell off my todo list to review, sorry.

This feels like a hack which is going to bite us later. A better plan would be to do what's suggested in #973, which would be to create a proper "type" to handle utf-8 (via dtype's metadata), and if necessary, write a conversion function for that (once we can sensibly read and write fixed length strings of arbitrary encodings, then we can fix many of the issues listed at https://github.com/h5py/h5py/labels/string%2Funicode%20usage).

t-b:
@ajelenak-thg Does #1132 make this PR obsolete?

scopatz:
Looks like there are conflicts with master

ajelenak-thg:
Withdrawing this PR because of #1132.

Issue #945, opened 2017-11-06
Add H5Pset_external


GregoryEAllen:
`H5Pset_external` adds an external file to the list of external files. It is not exposed in current versions of h5py.

tacaswell:
@GregoryEAllen Can you take a crack at doing this?  It should probably go someplace around https://github.com/h5py/h5py/blob/309c4740da04fd89e0879b1dfea22ab3ab988e08/h5py/api_functions.txt#L253

GregoryEAllen:
I'd like to. My pressing urge was also worked around with `ctypes`. Here's a [gist](https://gist.github.com/GregoryEAllen/75f8dd69216a0cc2797525578199d8e1) that may help others.

Integration with higher layers would also be nice.

tacaswell:
The high-level interface on this seems like it has the same issues that virtual dataset API in that you need to provide pretty low-level information that does not have an obvious way to simplify / make terser in a general way.

GregoryEAllen:
Adding `H5Pset_external` was straightforward. Where I got stuck was on `H5Pset_efile_prefix`. I think a complete implementation would include both. Even so, I've decided to proceed for now without the latter.

My plan is to add `dataset.external` that returns a list of tuples.

    [(file0, offset0, size0), (file1, offset1, size1), (file2, offset2, size2), ... ]

I typically there will only be one. It returns `None` if there's None.

I also plan to add `file.create_dataset(external=None)`

Setting external to a list like above will set these properties. I'd also support a single list/tuple or a single file name. `offset` defaults to 0 and `size` defaults to `H5F_UNLIMITED` if not specified.

PR #1000, opened 2018-02-18
Add external dataset support

GregoryEAllen:
Enhancement for #945


aragilar:
This looks good. One question though, which HDF5 release was this introduced? If it's after 1.8.4, we should add a guard to enable this only on releases of HDF5 which support it.

GregoryEAllen:
It's in [hdf5-1.2.2](https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-1.2.2.tar.gz), the oldest I could find.

aragilar:
Merging, thanks for this.

PR #1002, opened 2018-02-19
Whitespace cleanup

### LINKED ISSUE #1001: Whitespace cleanup
In trying to address #973 I noticed that the sources are somewhat riddled with trailing whitespace.
It's common with both vim and (in my case) emacs for those to get stripped automatically, so they end up showing as diff. Several open PRs show useless whitespace changes for the same reason.

Would you accept a zero-change whitespace cleanup PR?

wavexx:
Touching only python sources, I've stripped trailing whitespace and trailing empty newlines at the end of each file. Expanded tabs to follow usual python conventions. Double-checked the expansions, and it looks like there was only one example file with weird indentation (!=4), which I fixed nonetheless.

Closes #1001


tacaswell:
Can you also add a check to the CI the makes sure the whitespace does not come back?

wavexx:
Is it actually being done in Pandas/NumPy? I honestly never seen this happen frequently in python. Just keeping an eye should prevent this, instead of doing some enforcement.

wavexx:
Can we move on this and add another issue to add style checks in the CI?

tacaswell:
Matplotlib has style checking turned on.  

If we are going to apply these sort of big changes (which cause a lot of code churn for no _functional_ change) we should put the tests in prevent the issues from coming back.

aragilar:
Replaced by #1048 (now merged), closing.

PR #1003, opened 2018-02-20
ENH: Use pytest to find and run tests

aragilar:
Part of a plan to move from unittest to pytest (giving nicer reporting, use of pdb for debugging, and hopefully easier testing). ~~WIP until I check there are no regressions, and that the setup is robust against pytest changes.~~ Should now be fine to use.

### DIFF
--- a/h5py/tests/__init__.py
+++ b/h5py/tests/__init__.py
@@ -7,35 +7,19 @@
 # License:  Standard 3-clause BSD; see "license.txt" for full license terms
 #           and contributor agreement.
 
-from __future__ import print_function
-
-import sys
-from .common import ut
-
-from . import old, hl
-
-MODULES = old.MODULES + hl.MODULES
+from __future__ import absolute_import
 
 
-def mname(obj):
-    """ Get the full dotted name of the test method """
-    mod_name = obj.__class__.__module__.replace('h5py.tests.','')
-    return "%s.%s.%s" % (mod_name, obj.__class__.__name__, obj._testMethodName)
-
-
-def run_tests(verbose=False):
-    """ Run tests with TextTestRunner.  Returns a TestResult instance.
-
-    """
-    suite = ut.TestSuite()
-    for m in MODULES:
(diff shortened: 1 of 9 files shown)


aragilar:
Comments on the setup used:
 * Running pytest in the existing python process (as earlier versions of the code did, and how the tests are currently run with unittest) causes incorrect values for coverage, and a lot of pain when trying to work on windows with additional pytest arguments (and with additional arguments generally).
 * Coverage is collected, but currently there's no online coverage service connected to CI, opinions on what's the best service to use, if any (coveralls?, codecov?)?
 * I haven't added any tests yet to ensure that `h5py.run_tests()` keeps working, I could add an additional tox which runs setup.py test (as that depends on `h5py.run_tests()` and we're not checking that either, do people thing that's sufficient?

tacaswell:
matplotlib uses codecov and have been pretty happy with it.

h5py/tests/__init__.py
```diff
@@ -7,35 +7,18 @@
...
-
-    return result
+from __future__ import absolute_import
+
+
+def run_tests(args=''):
+    try:
+        from pytest import main
+    except ImportError:
+        print("Tests require pytest, pytest not installed")
```
takluyver: This should probably have a `return 1` so that the exit code indicates something went wrong.

aragilar: Good catch, I've added that now.

.gitignore
```diff
@@ -24,4 +24,3 @@ MANIFEST
 .eggs/
 *.so
 *~
-.pytest_cache
```
takluyver: This is being removed from `.gitignore`? Shouldn't it be there? I have something similar for other projects using pytest.

takluyver:
:+1:

.gitignore
aragilar: That was me screwing up the rebase, I've dropped that change (so `.pytest_cache` is still being ignored).

PR #1014, opened 2018-03-12
Lazy-loading of run_tests to avoid strong dependency on unittest2

### LINKED ISSUE #1013: unittest2 is required with python 2.x but it is not in install_requires
With h5py 2.8.0rc1 (installed from pypi wheels) on Python2.7, `unittest2` is required to `import h5py` but it is not listed in setup.py's `install_requires`.

I tested on Linux and Windows but it should affect all platform with python 2.x:
```
>>> import h5py
  File "[...]/lib/python2.7/site-packages/h5py/__init__.py", line 67, in <module>
    from .tests import run_tests
  File "[...]/lib/python2.7/site-packages/h5py/tests/__init__.py", line 13, in <module>
    from .common import ut
  File "[...]/python2.7/site-packages/h5py/tests/common.py", line 27, in <module>
    raise ImportError( "unittest2 is required to run tests with Python 2")
ImportError: unittest2 is required to run tests with Python 2
```

t20100:
This PR adds a `run_tests` function in top level module to defer import of `h5py.tests.run_tests` in order to avoid a strong dependency on `unittest2` with python2.x.

closes #1013

### DIFF
--- a/h5py/__init__.py
+++ b/h5py/__init__.py
@@ -64,7 +64,6 @@ from .h5t import special_dtype, check_dtype
 from . import version
 from .version import version as __version__
 
-from .tests import run_tests
 
 if version.hdf5_version_tuple != version.hdf5_built_version_tuple:
     _warn(("h5py is running against HDF5 {0} when it was built against {1}, "
@@ -72,6 +71,14 @@ if version.hdf5_version_tuple != version.hdf5_built_version_tuple:
             '{0}.{1}.{2}'.format(*version.hdf5_version_tuple),
             '{0}.{1}.{2}'.format(*version.hdf5_built_version_tuple)
     ))
+
+
+def run_tests(verbose=False):
+    """Run tests with TextTestRunner and returns a TestResult instance.
+    """
+    # Lazy-loading of tests package to avoid strong dependency on unittest2
+    from .tests import run_tests
+    return run_tests(verbose=verbose)
 
 
 def enable_ipython_completer():


tacaswell:
Much better than my suggestions!

aragilar:
@t20100 Thanks for this.

cdeil:
Just to note: this (I think accidentally) introduced a change / regression concerning the deprecated (but used e.g. by Astropy) `h5py.highlevel`: see #1046

t20100:
This was unintended, and indeed breaks the indirect loading (through some tests) of `h5py.highlevel`.

vasole:
Could one use the same approach to lazy-loading highlevel while at the same time issuing a deprecation warning about the use of highlevel?

cdeil:
> Could one use the same approach to lazy-load highlevel while at the same time issuing a deprecation warning about the use of highlevel?

I'm sure it's possible to lazy import `h5py.highlevel` on attribute access.
But it might not be simple / code that you want to put in `h5py/__init__.py`.
I'm not an expert on this / can't give a recommendation, but here's an example I found with a quick Google search: 
https://pypi.org/project/lazy-import/

The simpler solution would be to just put the `import highlevel`  back in `h5py/__init__.py` as mentioned already in https://github.com/h5py/h5py/issues/1046#issuecomment-393905108 and then in ~ 2 years or so do the break that was accidentally done here in v2.8 .
You could still do a soft deprecation of announcing the change in `h5py` release notes, encouraging projects to check if they rely on this.
IMO this would be the most user-friendly way forward, deprecation warnings can be very annoying and if libraries like even the recent Astropy 3.0 import `h5py.highlevel` many users see them yet have no way to fix that.

tacaswell:
@effigies Sorry, I missed this comment in the run up to 2.8.

https://github.com/h5py/h5py/pull/1047/files implements a lazy import.

PR #1023, opened 2018-04-03
Remove unnecessary import

glikson:
Import of 'run_tests' seems unnecessary in __init__.py


aragilar:
The purpose of `run_tests` is to make it easy to run the tests on an installed h5py (see http://docs.h5py.org/en/latest/build.html#installation). In order to make sure that the tests are run the same way as suggested for users, the CI uses `run_tests`. I'm going to close this as `run_tests` is explicit part of the API.

Issue #441, opened 2014-05-06
Python3 error storing a tuple of strings as attribute


monkeybutter:
I'm in the process of making my code python3 compatible and I'm getting an error in the following line:

```
dset.attrs['Column Names'] = ('X', 'Y', 'Z', 'Intensity', 'Distance', 'Angle')
```

The error shows:

```
Traceback (most recent call last):
  File "test.py", line 10, in write_to_hdf5
    dset.attrs['Column Names'] = ('X', 'Y', 'Z', 'Intensity', 'Distance', 'Angle')
  File "/home/roz016/anaconda/envs/py3k/lib/python3.3/site-packages/h5py/_hl/attrs.py", line 71, in __setitem__
    self.create(name, data=value, dtype=base.guess_dtype(value))
  File "/home/roz016/anaconda/envs/py3k/lib/python3.3/site-packages/h5py/_hl/attrs.py", line 108, in create
    htype = h5t.py_create(dtype, logical=True)
  File "h5t.pyx", line 1379, in h5py.h5t.py_create (h5py/h5t.c:12683)
  File "h5t.pyx", line 1457, in h5py.h5t.py_create (h5py/h5t.c:12578)
TypeError: No conversion path for dtype: dtype('<U9')
```

This code works perfectly fine using python2.7. Isn't it possible to store string tuples as attributes using python3?

guddulrk:
I am making python code compatible to python3.5 and getting following error in this line of code:
    db['data'][dname].attrs['txt'] = res[i]['txt']

Traceback (most recent call last):
  File "C:/python_projects/SynthText-master/SynthText-master/gen.py", line 122, in main
    add_res_to_db(imname,res,out_db)
  File "C:/python_projects/SynthText-master/SynthText-master/gen.py", line 73, in add_res_to_db
    db['data'][dname].attrs['txt'] = res[i]['txt']
  File "h5py\_objects.pyx", line 54, in h5py._objects.with_phil.wrapper (C:\Minonda\conda-bld\h5py_1474482825505\work\h5py\_objects.c:2705)
  File "h5py\_objects.pyx", line 55, in h5py._objects.with_phil.wrapper (C:\Minonda\conda-bld\h5py_1474482825505\work\h5py\_objects.c:2663)
  File "C:\Program Files\Anaconda3\lib\site-packages\h5py\_hl\attrs.py", line 93, in __setitem__
    self.create(name, data=value, dtype=base.guess_dtype(value))
[...]

PR #1031, opened 2018-05-10
Support lists of unicode strings as attributes

chrisjbillington:
Store attributes that are lists of unicode strings as numpy bytestring arrays with character set set to UTF8 to indicate the data is encoded text.

This is compatible with other programs reading the file, which will simply see an array of bytestrings and the character set flag (which they can ignore if they like)

This addresses #441


chrisjbillington:
Actually I've changed my mind about this. Probably prefer the approach in pull request #1032 as much simpler.

PR #1032, opened 2018-05-11
Allow list of unicode strings to be saved as attrs

chrisjbillington:
This addresses issue #441

Store attributes that are lists of unicode strings as arrays of vlen unicode type

This is only done if the array was auto converted from a tuple or similar.
If the user passes in a numpy array, then its type will be respected and not auto converted.

With this change the following works without error:

```python
with h5py.File('test', driver='core', backing_store=False) as f:
    f.attrs['x'] = ['ü', 'y']
    print(repr(f.attrs[u'x']))
```
printing:

```python
array(['ü', 'y'], dtype=object)
```

### DIFF
--- a/docs/strings.rst
+++ b/docs/strings.rst
@@ -148,6 +148,28 @@ this and "pretend" to support it, h5py will raise an error when attempting
 to create datasets or attributes of this type.
 
 
+Handling of iterables of strings as attributes
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If you set an attribute equal to a Python list or tuple of unicode strings,
+such as the following:
+
+    >>> f.attrs['x'] = (u'a', u'b') # PY2
+    >>> f.attrs['x'] = ('a', 'b') # PY3
+
+h5py will save these as arrays of variable-length strings with character set
+H5T_CSET_UTF8. When read back, the result will be numpy arrays of dtype
+``'object'``, as if the original data were written as:
+
+    # PY2
+    >>> f['x'] = np.array((u'a', u'b'), dtype=h5py.special_dtype(vlen=unicode))
+
+    # PY3
+    >>> f['x'] = np.array(('a', 'b'), dtype=h5py.special_dtype(vlen=str))
+    
+h5py will only perform this conversion if you set an attribute that is not a numpy array.
+
+
 Object names
 ------------
 
--- a/docs/strings.rst
+++ b/docs/strings.rst
@@ -154,15 +154,14 @@ Handling of iterables of strings as attributes
 If you set an attribute equal to a Python list or tuple of unicode strings,
 such as the following:
 
-    >>> f.attrs['x'] = (u'a', u'b') # PY2
-    >>> f.attrs['x'] = ('a', 'b') # PY3
+    >>> f.attrs['x'] = (u'a', u'b')
 
 h5py will save these as arrays of variable-length strings with character set
 H5T_CSET_UTF8. When read back, the result will be numpy arrays of dtype
 ``'object'``, as if the original data were written as:
 
-    >>> f['x'] = np.array((u'a', u'b'), dtype=h5py.special_dtype(vlen=unicode)) # PY2
     >>> f['x'] = np.array(('a', 'b'), dtype=h5py.special_dtype(vlen=str)) # PY3
+    >>> f['x'] = np.array((u'a', u'b'), dtype=h5py.special_dtype(vlen=unicode)) # PY2
     
 h5py will only perform this conversion if you set an attribute that is not a numpy array.
 
--- a/docs/strings.rst
+++ b/docs/strings.rst
@@ -148,8 +148,8 @@ this and "pretend" to support it, h5py will raise an error when attempting
 to create datasets or attributes of this type.
 
 
-Handling of iterables of strings as attributes
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Handling of lists/tuples of strings as attributes
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 If you set an attribute equal to a Python list or tuple of unicode strings,
 such as the following:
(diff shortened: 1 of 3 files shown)


aragilar:
I think this is best approach for handling unicode on attributes (as it leaves open improvements to the internals of h5py). I'll leave some more comments on the issue thread (#441), but only additional changes I would want are tests (to avoid regression), and updating the string usage documentation.

chrisjbillington:
Excellent, I'm glad this approach is agreeable. I can make changes to add tests and update docs about three weeks from now (my thesis is due soon), so I'll just leave this for the moment. Anyone reading this is free to do it for me if they are impatient.

tacaswell:
@chrisjbillington Any interesting in adding the tests + docs?

chrisjbillington:
Yes! Apologies for the delay, I'll get to this in the next couple of days.

takluyver:
Ping @chrisjbillington - we're aiming for a 2.9 release fairly soon, so if you can find time to write some tests and docs, it sounds like people are happy for this to go in.

chrisjbillington:
Great, working on this today.

docs/strings.rst
```diff
@@ -148,6 +148,28 @@ this and "pretend" to support it, h5py will raise an error when attempting
...
 
 
+Handling of iterables of strings as attributes
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If you set an attribute equal to a Python list or tuple of unicode strings,
+such as the following:
+
+    >>> f.attrs['x'] = (u'a', u'b') # PY2
+    >>> f.attrs['x'] = ('a', 'b') # PY3
```
takluyver: The `u` prefix is valid in Python 3 as well, if you want to simplify the example.

docs/strings.rst
```diff
@@ -148,6 +148,25 @@ this and "pretend" to support it, h5py will raise an error when attempting
...
+
+h5py will save these as arrays of variable-length strings with character set
+H5T_CSET_UTF8. When read back, the result will be numpy arrays of dtype
+``'object'``, as if the original data were written as:
+
+    >>> f['x'] = np.array((u'a', u'b'), dtype=h5py.special_dtype(vlen=unicode)) # PY2
+    >>> f['x'] = np.array(('a', 'b'), dtype=h5py.special_dtype(vlen=str)) # PY3
```
takluyver: I'd put the Python 3 example first.

takluyver:
The automated checks are failing on some whitespace issues:

```
Fix End of Files.........................................................Failed
hookid: end-of-file-fixer
Files were modified by this hook. Additional output:
Fixing h5py/tests/hl/test_attribute_create.py
Trim Trailing Whitespace.................................................Failed
hookid: trailing-whitespace
Files were modified by this hook. Additional output:
Fixing docs/strings.rst
Fixing h5py/_hl/attrs.py
```

takluyver:
No problem & no rush :-)

takluyver:
Thanks, looks like CI is happy. @aragilar @tacaswell are you happy for me to merge this before 2.9?

h5py/_hl/attrs.py
```diff
@@ -116,10 +116,18 @@ def create(self, name, data, shape=None, dtype=None):
...
         with phil:
 
-            # First, make sure we have a NumPy array.  We leave the data
-            # type conversion for HDF5 to perform.
+            # First, make sure we have a NumPy array.  We leave the data type
+            # conversion for HDF5 to perform (other than the below exception).
             if not isinstance(data, Empty):
+                # Whether we were already given a numpy array or whether we are
+                # converting to a numpy array ourselves:
+                auto_convert_to_array = not isinstance(data, numpy.ndarray)
```
takluyver: Just a brief thought: is this check sufficient? There may be other array-like things (such as h5py datasets, or pandas series) which aren't an ndarray, but have similar constraints. Maybe we should more explicitly check for a list/tuple of strings?

chrisjbillington: I was aiming to support setting as wide a range of things as possible: to auto-convert anything that is not a numpy array, and that, when converted to a numpy array, would have the numpy `U` datatype. I wouldn't want to limit it to just strings and tuples because there are other iterables, including custom ones that are not subclasses of the builtins. I figure passing in anything that is not a numpy array or a subclass of a numpy array is safe to treat as consent to data types being changed. But I might be wrong: I suppose things that it is arguable shouldn't be auto converted are types that are sufficiently numpy-array-like that their datatype should be considered a deliberate choice and respected (where 'respected' just means 'error if it's `U` type instead of silent conversion').

Looks like you can make a pandas Series with vlen str dtype:

```pandas.Series(np.array(['a', 'b'], dtype=h5py.special_dtype(vlen=str)))```

and this can be saved as a dataset or attribute just fine, with or without this patch.

But pandas series don't seem to support the numpy `U` type though ([bug report](https://github.com/pandas-dev/pandas/issues/5261)), so at the moment a user cannot try to use a pandas Series to intentionally write an array with `U` dtype. So there is no risk (yet) that they will get silent conversion instead of an error.

I think there's no need to exclude h5py datasets though because we know for a fact that these, when converted to numpy arrays, will not have `U` datatype (since HDF5 has no such datatype).

If we want to be paranoid about the future of pandas, if they decide to support the `U` type, we could put in something like `if isinstance(getattr(data, 'dtype', None), np.dtype) and ...` instead of the check for `isinstance(data, numpy.ndarray)`. But I suspect this is getting ahead of ourselves, and would basically just be for pandas since other classes out there that have a numpy dtype but not as an attribute called `.dtype`.

What other things are there out there that have numpy datatypes (and therefore might have the `U` datatype), but aren't actual numpy arrays or subclasses? I'm guessing few enough to not be not worth worrying about.

All in all I think it's not too bad for h5py to demand that if you want your datatype respected (again, where that just means error instead of silent conversion), you need to pass in a numpy array. 

I could change the documentation though to make it sound more like it's about anything that isn't a numpy array. It says that in the text, but the title definitely makes it sound like it's about only lists and tuples.

I'm just thinking out loud, but I'll make any change you want really!

takluyver: My rationale for wanting to restrict it more is twofold:

1. Lots of things are iterable, but it doesn't necessarily make sense to store them as attributes, especially if they won't roundtrip. If I try to store some container object with special properties, I'd rather it just refused than quietly converted it to something else. In this particular context, I think duck-typing is more of a hindrance than a help.
2. If we we allow only selected types (lists, tuples) for now, it's easy to add more later when a need becomes apparent. It's much trickier to restrict an initially broad category later if problems come up, because that might break code already in use. So my instinct is to start with a narrow list of types.

It's not my project, though - I've only just joined as a maintainer. I'd like @tacaswell or @aragilar to weigh in on this.

aragilar: I agree we should only do this on tuples/lists only, this is more of a convenience than anything else, as with strings and h5py the best thing to do is to do the conversion beforehand (so that you know what is written and going to be read out).

takluyver:
Thanks all :-)

PR #1042, opened 2018-05-19
Drop support for python 2.6 and 3.3

tacaswell:
Both are no longer being supported by upstream python.


aragilar:
Appveyor was playing up so I ignored it and merged this to avoid further failures on travis.

PR #1047, opened 2018-05-30
ENH: Add deprecation warnings where needed.

aragilar:
There's a few places where we've deprecated something, but not warned about it (based on grepping the h5py folder, so there may be other places where we should be warning). This adds those warnings.


rainwoodman:
This is unrelated -- but when I tried to test this against astropy, installing this branch with
```
https://github.com/h5py/h5py/archive/f8156cf139180d84d1ce6a63151c4a82b7b5b3b6.zip
```
on anaconda with hdf library, but it faild not finding `hdf5.h`. Is there anything special I need to be aware pip installing h5py on anaconda?

aragilar:
@rainwoodman You need to define `HDF5_DIR` as the path to the anaconda HDF5 installation.

PR #1060, opened 2018-06-23
Virtual datasets interface

takluyver:
This builds on #894, but I've redesigned the interface. Here's the TLDR (updated following comments below):

```python
# The 'layout' is the structure of the virtual dataset we want to build
layout = h5py.VirtualLayout(shape=(4, 100), dtype='i4')

for n in range(1, 5):
    filename = "{}.h5".format(n)
    # Sources represent a source dataset to include
    vsource = h5py.VirtualSource(filename, 'data', shape=(100,))
    # Map the source dataset (or a slice thereof) into a slice of the layout
    layout[n - 1] = vsource

# Add virtual dataset to output file
with h5py.File("VDS.h5", 'w', libver='latest') as f:
    f.create_virtual_dataset('data', layout, fillvalue=-5)
```

`VirtualSource` can be instantiated in two ways:

```python
# By file path and dataset name (doesn't read the file)
vsource = h5py.VirtualSource(filename, 'data', shape=(100,))

# From a Dataset object (gets the shape automatically)
with h5py.File(filename, 'r') as f:
    vsource = h5py.VirtualSource(f['data'])
```


takluyver:
Looks like there are merge conflicts - do you prefer me to rebase, or to merge from master?

h5py/tests/hl/test_vds/test_virtual_source.py
```diff
@@ -0,0 +1,140 @@
...
+
+    # def test_full_slice_inverted(self):
+    #     dataset = h5.VirtualSource('test','test',(20,30,30))
+    #     sliced = dataset[:,:,::-1]
+    #     self.assertEqual(dataset.shape,sliced.shape)
+    #
+    # def test_subsampled_slice_inverted(self):
+    #     dataset = h5.VirtualSource('test','test',(20,30,30))
+    #     sliced = dataset[:,:,::-2]
+    #     self.assertEqual((20,30,15),sliced.shape)
```
takluyver: I commented out several tests in this file that were using either negative steps or `start > stop` reversed slices. These are rejected by the `SimpleSelection` machinery I switched to apply slices. I'm not sure if reversed slices are valid for VDS, but I wanted to prioritise getting something simple working, even if that means neglecting some corner cases.

takluyver:
I saw on #809 that you asked for a rebase, so I've rebased the changes again onto master.

Tests have passed on Travis, and it looks like we're on the way to passing tests on Appveyor too.

h5py/tests/hl/test_vds/test_virtual_source.py
tacaswell: From the user point of view a VDS vs normal dataset should be identical.  I suspect we will want to handle the full set of 'fancy' slicing.  

That said, on access users should just see a normal dataset, this is only used to construct the VDS?  If that is the case, I agree that super fancy strides are important on a first pass at this.

tacaswell:
I gave this a quick skim and very much like the proposed API :+1: .

h5py/tests/hl/test_vds/test_virtual_source.py
takluyver: Yup, the `VirtualSource` class this was testing is only used when preparing a virtual dataset; access is just like a normal dataset. Though you can't access a dataset with a negative slice either (`ValueError: Step must be >= 1 (got -1)`).

I was debugging code added in the PR to translate Python slicing into an HDF5 selection, when I realised that h5py must already have that functionality to access slices of datasets. Sure enough, I found it in `h5py._hl.selections`. So I discarded the new code and used the selections module in its place. I suspect that if that module doesn't allow negative steps, it's because HDF5 selections don't support that.

takluyver:
There were a couple of design points I wasn't sure about; I hope this doesn't kick off a long naming debate, but maybe you have clearer ideas having worked with HDF5 for longer:

1. Naming for the class currently called `VirtualTarget`. This represents the virtual dataset 'under construction'. I don't want to call it `VirtualDataset`, because it's not usable as a dataset. It could be `VirtualSpace`, `VirtualMap`, `VirtualLayout`. "Target" has a neat symmetry with source, but I always find those terms confusing with symlinks (is the link pointing to the target, or made from the source?).
2. In the two step construction process, which parameters are passed to `VirtualTarget`, and which to `create_virtual_dataset`? The shape certainly needs to be in the first step, and having the dtype there may let us check that the source dtype is compatible. But e.g. `fillvalue` could move to the second step. I don't have a strong feeling either way.

takluyver:
I see from the docs that you can create/link a dataset with an assignment like `grp["name"] = array`. It would seem natural for assigning a VirtualTarget object like that to create a virtual dataset. That points to keeping most parameters in the `VirtualTarget` constructor rather than the `create_virtual_dataset` method.

tacaswell:
The tension in 2 is I think what drove me to making `create_virtual_dataset` have cranky verbose API + helper functions.   I am biased towards pushing anything that the `VirtualTarget` object does not need to know to be a parameter to `create_virtual_dataset`.   I think of `VirtualTarget` being a fancy object as a consequence of needing to encode the data structure of how to map values in the VDS to values in (many) source data sets rather than "bucket for all the config".

I agree calling the contstructor_helper_object `VirtualDataset` is not good (because that makes it sound like you can get data _out_ of it). `VirtualMap` sounds like it should be a dict proxy and `VirtualSpace` collides in my brain to much with all of the other `*Space` names in hdf5.  

Given the above discussion I like `VirtualLayout` better than `VirtualTarget` as it captures that the object is just the layout related information and fits well with pushing the non-layout things (fillvalue) don't belong it in.  It also helps break the source vs target distinction. "You assemble a bunch of sources into a layout that you then use to create the dataset" is a reasonably understandable English sentence.

takluyver:
Is there anything else I should be doing on this, by the way?

appveyor.yml
```diff
@@ -98,6 +98,49 @@ environment:
       HDF5_DIR: "C:\\hdf5\\%HDF5_VERSION%\\%HDF5_VSVERSION%"
       HDF5_VERSION: "1.10.1"
```
aragilar: This appears to be duplicated?

aragilar:
This looks fine to me, apart from the duplication in the appveyor config. There also doesn't appear to be any tests for unicode paths, but that can be put off for a later PR.

appveyor.yml
takluyver: Good catch, probably I made a mistake with the rebase.

jangarrevoet:
@takluyver I found another thing. When you do not create the virtual data set on the root level but under a deeper level you get the following.
In vds_simple.py I changed:
`f.create_virtual_dataset('data', layout, fillvalue=-5)`
to
`f.create_virtual_dataset('/data/data', layout, fillvalue=-5)`

ValueError                                Traceback (most recent call last)
/tmp/vds_exp.py in <module>()
     18 # Add virtual dataset to output file
     19 with h5py.File("VDS.h5", 'w', libver='latest') as f:
---> 20     f.create_virtual_dataset('/data/data', layout, fillvalue=-5)
     21     print("Virtual dataset:")
     22     print(f['data'][:, :10])

/home/p06user/PythonPath/lib/python/h5py-2.8.0.post0-py2.7-linux-x86_64.egg/h5py/_hl/group.pyc in create_virtual_dataset(self, name, layout, fillvalue)
    158                                   tid=h5t.py_create(layout.dtype, logical=1),
    159                                   space=virt_dspace,
--> 160                                   dcpl=dcpl)
    161             return dset
    162 

h5py/_objects.pyx in h5py._objects.with_phil.wrapper (/tmp/h5py/h5py/_objects.c:2851)()

h5py/_objects.pyx in h5py._objects.with_phil.wrapper (/tmp/h5py/h5py/_objects.c:2809)()

h5py/h5d.pyx in h5py.h5d.create (/tmp/h5py/h5py/h5d.c:2129)()

ValueError: Unable to create dataset (component not found)

takluyver:
@jangarrevoet Good catch, I fixed that.

jangarrevoet:
@takluyver There also might still be something with the maxshape kwarg. When adapting vds_simple.py
'layout = h5py.VirtualLayout(shape=(4, 100), dtype='i4',maxshape=(None,100))'
 I get the following.
/tmp/vds_simple.py in <module>()
     24 # Add virtual dataset to output file
     25 with h5py.File("VDS.h5", 'w', libver='latest') as f:
---> 26     f.create_virtual_dataset('data', layout, fillvalue=-5)
     27     print("Virtual dataset:")
     28     print(f['data'][:, :10])

/home/p06user/PythonPath/lib/python/h5py-2.8.0.post0-py2.7-linux-x86_64.egg/h5py/_hl/group.pyc in create_virtual_dataset(self, name, layout, fillvalue)
    147                 dcpl.set_fill_value(numpy.array([fillvalue]))
    148 
--> 149             virt_dspace = h5s.create_simple(layout.shape, layout.maxshape)
    150 
    151             for vspace, fpath, dset, src_dspace in layout.sources:

h5py/_objects.pyx in h5py._objects.with_phil.wrapper (/tmp/h5py/h5py/_objects.c:2851)()

h5py/_objects.pyx in h5py._objects.with_phil.wrapper (/tmp/h5py/h5py/_objects.c:2809)()

h5py/h5s.pyx in h5py.h5s.create_simple (/tmp/h5py/h5py/h5s.c:1716)()

h5py/utils.pyx in h5py.utils.convert_tuple (/tmp/h5py/h5py/utils.c:1996)()

TypeError: Can't convert element 0 (None) to hsize_t

takluyver:
I've made the code creating a virtual dataset more closely parallel the code creating a regular dataset: the main work is done by a function which returns a low-level dataset ID, and then the `Group.create_virtual_dataset()` method links that into the group and wraps it in an h5py Dataset object.

Having a `None` dimension in maxshape is now possible, though I haven't done any testing of how it behaves when growing datasets. This doesn't seem very high priority to me: because virtual datasets are a lightweight list of references, it's cheap to recreate them if the underlying datasets change.

jangarrevoet:
@takluyver I will do some testing tonight if I find the time to see if I can append.

takluyver:
Any more feedback on this? :)

jangarrevoet:
I did a test to append but this gave an error. Wanted to investigate a bit more before reporting back but did not find the time yet. It was complaining about a chunk size setting.

takluyver:
Thanks. Is resizing a virtual dataset something that's meant to be supported? The [docs for H5Dset_extent](https://support.hdfgroup.org/HDF5/doc/RM/RM_H5D.html#Dataset-SetExtent) don't list virtual datasets as a supported option. From my understanding, it doesn't seem like an important feature, because it's cheap to recreate a virtual dataset with a new size if necessary.

If we don't need to support resizing virtual datasets, I think that error message is probably OK: a virtual dataset is not a chunked dataset.

jangarrevoet:
Looking at the examples in [this document](https://support.hdfgroup.org/HDF5/docNewFeatures/VDS/HDF5-VDS-requirements-use-cases-2014-12-10.pdf) it seems like you might be able to do resizing although not fully clear to me if that is what they intend to achieve but why otherwise make an unlimited dimension of the vds.
I think users will expect the feature to be there.

jangarrevoet:
@takluyver I think one can merge.

takluyver:
Thanks :-)

aragilar:
@tacaswell Anything else that needs changing before we merge this?

examples/eiger_use_case.py
```diff
@@ -0,0 +1,23 @@
...
+
+import h5py
+import numpy as np
+
+
+files = ['1.h5', '2.h5', '3.h5', '4.h5', '5.h5']
+entry_key = 'data' # where the data is inside of the source files.
+sh = h5py.File(files[0], 'r')[entry_key].shape # get the first ones shape.
+
+layout = h5py.VirtualLayout(shape=(len(files),) + sh, dtype=np.float)
```
tacaswell: Isn't this going to produce a 4d layout (as `sh` is already length 3)?  I think this should be `(len(files) * sh[0], ) + sh[1:]`.

takluyver: You're right, done. Thanks!

takluyver:
CI failure was some issue downloading Python; reopening to rerun.

takluyver:
Was there anything else I should be doing on this?

tacaswell:
Sorry, I keep trying to sit down and read this top-to-bottom one last time before merging and keep getting interrupted / pulled away!  On my to-do list for today (again).

takluyver:
Thanks!

tacaswell:
I could not figure out how to make the fprint style formatting (described at https://support.hdfgroup.org/HDF5/doc/RM/RM_H5P.html#Property-SetVirtual) to work, but do not want to hold this PR over it.

The two things I am most concerned about is raising if inputs are ignored an the API change on the low-level API.

tacaswell:
Actually, I'll just push commits with those changes.

h5py/_hl/vds.py
```diff
@@ -0,0 +1,137 @@
...
+        self.sources = []
+
+    def __setitem__(self, key, source):
+        sel = SimpleSelection(self.shape)[key]
+        self.sources.append(VDSmap(sel.id,
+                               source.path,
+                               source.name,
+                               source.sel.id))
+
+    def _make_block_shape(self, source):
```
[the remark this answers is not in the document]
takluyver: No, I think you're right. I think it's a leftover from a previous implementation, before I realised it was possible to use `SimpleSelection`.

h5py/h5p.pyx
```diff
@@ -834,7 +834,7 @@ cdef class PropDCID(PropOCID):
             return count
 
         @with_phil
-        def get_virtual_dsetname(self, size_t index=0):
+        def get_virtual_dsetname(self, size_t index):
```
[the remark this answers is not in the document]
takluyver: I'm fine with leaving the default. It was actually your commit that removed it - I guess it's a balance between compatibility and explicitness.

51eb0b30cb41e3845b76c8f7702b482e362fa54b

takluyver:
Thanks! I'll leave the changes to you for now, but I'm happy to finish them up if you run out of time - just let me know.

takluyver:
For the printf-style names, I think it should work if you create a `VirtualSource` with the special `%0b` parts in either the filename or dataset name. It passes the names through to the low-level implementation. I haven't tried it, though.

The printf thing feels like it was bolted on late without a lot of thought. E.g. we have numbered sequence files a lot like the example use cases, but our numbers are zero-padded, like `00001`, and I can't see any documentation on how you'd format numbers like that. The 'virtual printf gap' property also feels inelegant.

tacaswell:
I naively tried passing `'%0b.h5'` in for a file name and it did not work (but I may have been doing something else wrong....)

tacaswell:
I tried to split my commits up, feel free to drop and force-push any you do not like.

h5py/_hl/vds.py
```diff
@@ -51,9 +51,21 @@ class VirtualSource(object):
...
+    def __init__(self, path_or_dataset, name=None,
+                 shape=None, dtype=None, maxshape=None):
         from .dataset import Dataset
         if isinstance(path_or_dataset, Dataset):
+            if any(inp is not None
+                   for inp in [name, shape, dtype, maxshape]):
+                failed = {k: v
+                          for k, v in
+                          {'name': name, 'shape': shape,
+                           'dtype': dtype, 'maxshape': maxshape}.items()
```
takluyver: Slightly more concise: construct this dict before the `if`, then just check `if failed:`.

h5py/_hl/vds.py
```diff
@@ -51,9 +51,21 @@ class VirtualSource(object):
...
+                          if v is not None}
+
+                raise TypeError("If a Dataset is passed as the first argument "
+                                "then no other arguments maybe passed.  You "
```
takluyver: maybe -> may be

takluyver:
:+1:

tacaswell:
@aaron-parsons Are you happy with where this landed?

I think this will come down to @aragilar to merge (as I have commits in here).

aragilar:
I'm happy to merge this, assuming the tests pass (appveyor hasn't started yet, it's still in the queue...).

tacaswell:
We should turn on the auto-cancel feature on appveyor...

takluyver:
I looked into the printf-style name thing. I think it depends on the ability to make a selection unlimited in one dimension, so that you can assign to an unlimited selection of the virtual layout.

Unlimited selections were the part that I asked be left out here for a separate pull request. I imagine they could work something like this:

```python
layout[:UNLIMITED] = source
# Or possibly:
layout[:...] = source
```

takluyver:
Appveyor has now caught up and the tests have passed :-)

aragilar:
Merged and rolling builds enabled (appveyor's name for auto-cancel).

takluyver:
Thanks! :tada:

PR #1075, opened 2018-08-13
Add open_file_image for opening a file from an image

mdietz94:
Adds access to H5LTopen_file_image from the low level interface. This will allow opening files from byte arrays without creating an empty file, as opening a new file and then calling set_file_image would.

### DIFF
--- a/h5py/api_functions.txt
+++ b/h5py/api_functions.txt
@@ -150,6 +150,7 @@ hdf5:
 
   # File Image Operations
   1.8.9     ssize_t H5Fget_file_image(hid_t file_id, void *buf_ptr, size_t buf_len)
+  1.8.9     hid_t H5LTopen_file_image(void *buf_ptr, size_t buf_len, unsigned int flags)
 
   # MPI functions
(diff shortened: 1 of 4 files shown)


tacaswell:
How does this relate to #1061 ?

mdietz94:
My understanding of #1061 is that it allows users to construct files with file-like objects instead of paths, whereas this change would allow users to construct files from byte arrays directly using the image api.

It would be possible to rewrite this change in terms of the other, by creating a fake file over the byte array, or vice versa by reading out the file and then using this api, but I think using the low level interface more directly should be more user friendly and predictable, especially as this change will expose the remainder of the image api that was partially exposed by the existing get_file_image and set_file_image methods.

mdietz94:
Friendly ping.

aragilar:
I'm happy to merge this, given that it's exposing existing HDF5 functions. If #1061 can benefit from this in some way, all the better.

mdietz94:
Great! Thanks!

mdietz94:
@aragilar ?

aragilar:
Merging, @mdietz94 thanks for this!

mdietz94:
Great, thanks so much!

PR #1085, opened 2018-08-24
New method create_dataset_like

takluyver:
Continuation of #848.

I added the test under the `old` subdirectory because that's where all the other tests of creating datasets seemed to be. I can move it if preferred, but it wasn't clear to me what qualifies a test as not 'old'.


aragilar:
This seems fine to me, merging.

takluyver:
Thanks @aragilar :-)

PR #1109, opened 2018-10-06
Fixed incorrect file mode as pointed in github issue #1071

### LINKED ISSUE #1071: Wrong filename in quickstart docs?
It appears that the wrong file name is given in the [Groups and Hierarchical Organization](http://docs.h5py.org/en/stable/quick.html#groups-and-hierarchical-organization) section of the docs. The docs read:

capture

But this produces an OSError. Replacing 'mydataset.hdf5' with 'mytestfile.hdf5' works.

pharshalp:
closes #1071

### DIFF
--- a/docs/quick.rst
+++ b/docs/quick.rst
@@ -107,9 +107,9 @@ created is itself a group, in this case the `root group`, named ``/``:
     >>> f.name
     u'/'
 
-Creating a subgroup is accomplished via the aptly-named ``create_group``. But we need to open the file in read/write mode first ::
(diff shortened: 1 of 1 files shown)


tacaswell:
Do we want to use `'w'` which will over-write any file there and start from an empty file or `'a'` which will append to any file that is there or create one if it has to?

pharshalp:
you are right... changing the file object mode to `'a'` would be safer. pushed the change.

tacaswell:
Thanks @pharshalp.  I see this is your first merged PR to h5py :tada: Hopefully we will hear from you again!

Issue #1110, opened 2018-10-10
The numpy.copy function fails on large (16GB) h5py datasets


cmeyer:
* Operating System (Windows 10)
 * Python version (tested on both 3.6, 3.7)
 * Where Python was acquired (Miniconda)
 * h5py version (2.8)
 * HDF5 version (1.10.2)

Here are Python commands to reproduce. Also included are alternate (working) ways of copying the data. Also attached raw script. Obviously you need a bunch of RAM (64GB) to run this test.

```
Python 3.6.6 |Anaconda, Inc.| (default, Jun 28 2018, 11:27:44) [MSC v.1900 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy
>>> import h5py
>>>
>>> print('H5PY info:', h5py.version.info)
H5PY info: Summary of the h5py configuration
---------------------------------

h5py    2.8.0
HDF5    1.10.2
Python  3.6.6 |Anaconda, Inc.| (default, Jun 28 2018, 11:27:44) [MSC v.1900 64 bit (AMD64)]
sys.platform    win32
sys.maxsize     9223372036854775807
numpy   1.15.2

>>>
>>> data = numpy.zeros((32, 32, 2048, 2048), dtype=numpy.float32)
>>> data[20, 20, 20, 20] = 1
>>>
>>> h5_file = h5py.File('test.h5', mode='w')
>>>
>>> h5_source = h5_file.create_dataset('Data', data=data)
>>>
>>> data_copy = numpy.copy(h5_source)
>>> print(data_copy[20, 20, 20, 20])
0.0
>>>
>>> data_copy2 = h5_source[:]
>>> print(data_copy2[20, 20, 20, 20])
1.0
>>>
>>> data_copy3 = numpy.empty(data.shape, data.dtype)
>>> h5_source.read_direct(data_copy3)
>>> print(data_copy3[20, 20, 20, 20])
1.0
```

```
import numpy
import h5py

print('H5PY info:', h5py.version.info)

data = numpy.zeros((32, 32, 2048, 2048), dtype=numpy.float32)
data[20, 20, 20, 20] = 1

h5_file = h5py.File('test.h5', mode='w')

h5_source = h5_file.create_dataset('Data', data=data)

data_copy = numpy.copy(h5_source)
print(data_copy[20, 20, 20, 20])

data_copy2 = h5_source[:]
print(data_copy2[20, 20, 20, 20])

[...]

PR #1111, opened 2018-10-10
Fixed integer overflows flattening shapes (e.g., 32,32,2048,2048).

cmeyer:
Fixes issue #1110.


cmeyer:
Sure, thanks for reviewing - and just to make sure I understand: use numpy.ulonglong instead of numpy.uint64, right?

aparamon:
Correct!

PR #1112, opened 2018-10-10
MAINT Numpy 1.16 deprecates asscalar

ksunden:
Just a quick refactor to avoid using deprecated function that is deprecated in numpy/numpy#12123


tacaswell:
Thanks @ksunden !

PR #1115, opened 2018-10-14
FIX: incorect casting to lower precision floats

### LINKED ISSUE #819: erroneous float64-to-float16 conversions
I have a float64 dataset in which, after creating a version of it with a dtype of float16, results in a few float16 values that are erroneous, i.e., not correctly cast. In code:

```
>>> test.create_dataset('data', data=x, chunks=True, shuffle=True, compression='gzip')
<HDF5 dataset "data": shape (289296,), type "<f8">

>>> test.create_dataset('data16', data=test['data'].value, dtype='float16', chunks=True, shuffle=True, compression='gzip')
<HDF5 dataset "data16": shape (289296,), type "<f2">

>>> np.nanmax(np.abs(test['data'].value - test['data16'].value))
127.9551361

>>> np.nanmax(np.abs(test['data'].value - np.float16(test['data'].value)))
0.12475106300001926
```

This is with h5py-2.3.1 and hdf5-1.8.18. It may well be a bug in the hdf5 library itself, but I first wanted to check whether it may be due to an issue in h5py.

Let me know how to share the test file, because uploading here doesn't seem to work.

tacaswell:
Not clear yet if this is a bug in `dset_id.write`, a bug in hdf5, or
incorrect use of `dset_id.write`.  However, this fixes it.

Closes #819

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -55,8 +55,8 @@ def readtime_dtype(basetype, names):
 
 
 def make_new_dset(parent, shape=None, dtype=None, data=None,
-                 chunks=None, compression=None, shuffle=None,
-                    fletcher32=None, maxshape=None, compression_opts=None,
+                  chunks=None, compression=None, shuffle=None,
+                  fletcher32=None, maxshape=None, compression_opts=None,
                   fillvalue=None, scaleoffset=None, track_times=None,
                   external=None):
     """ Return a new low-level dataset identifier
@@ -67,7 +67,20 @@ def make_new_dset(parent, shape=None, dtype=None, data=None,
     # Convert data to a C-contiguous ndarray
     if data is not None and not isinstance(data, Empty):
         from . import base
-        data = numpy.asarray(data, order="C", dtype=base.guess_dtype(data))
+        # normalize strings -> np.dtype objects
+        if dtype is not None:
+            _dtype = numpy.dtype(dtype)
+        else:
+            _dtype = None
+
+        # if we are going to a f2 datatype, pre-convert in python
+        # to workaround a possible h5py bug in the conversion.
+        is_small_float = (_dtype is not None and
+                          _dtype.kind == 'f' and
+                          _dtype.itemsize == 2)
+        data = numpy.asarray(data, order="C",
+                             dtype=(_dtype if is_small_float
+                                    else base.guess_dtype(data)))
 
     # Validate shape
     if shape is None:
--- a/h5py/tests/hl/test_datatype.py
+++ b/h5py/tests/hl/test_datatype.py
@@ -121,6 +121,24 @@ class TestVlen(TestCase):
                          h5py.check_dtype(enum=h5py.check_dtype(vlen=dt2)))
 
 
+class TestExplicitCast(TestCase):
+    def test_f2_casting(self):
+        fname = self.mktemp()
+
+        np.random.seed(1)
+        A = np.random.rand(1500, 20)
+
+        # Save to HDF5 file
+        with h5py.File(fname, "w") as Fid:
+            Fid.create_dataset("Data", data=A, dtype='f2')
+
+        with h5py.File(fname, "r") as Fid:
+            B = Fid["Data"][:]
+
+        # Compare
+        self.assertTrue(np.all(A.astype('f2') == B))
+
+
 class TestOffsets(TestCase):
     """
         Check that compound members with aligned or manual offsets are handled


h5py/_hl/dataset.py
```diff
@@ -67,7 +67,11 @@ def make_new_dset(parent, shape=None, dtype=None, data=None,
     # Convert data to a C-contiguous ndarray
     if data is not None and not isinstance(data, Empty):
         from . import base
-        data = numpy.asarray(data, order="C", dtype=base.guess_dtype(data))
+        if dtype is None:
+            _dtype = base.guess_dtype(data)
+        else:
+            _dtype = dtype
+        data = numpy.asarray(data, order="C", dtype=_dtype)
```
kalvdans: This change looks valid. There is no point in converting to an intermediate type if the user has given a specific type as keyword.

The code could be shortened as `data = numpy.asarray(data, order="C", dtype=dtype or base.guess_dtype(data))`

tacaswell:
Thanks for the review @kalvdans !  Simplified as you suggested and force-pushed.

tacaswell:
I am less sure this is the 'right' solution in the harsh light of day.

In `make_new_dataset` we have

https://github.com/h5py/h5py/blob/2e284d395a9b5160e04cf6a7fc2ee5165a3c690d/h5py/_hl/dataset.py#L147-L148

which calls

https://github.com/h5py/h5py/blob/2e284d395a9b5160e04cf6a7fc2ee5165a3c690d/h5py/h5d.pyx#L185-L221
which should be creating a memory-type based on the dtype of the array that comes in.  If I am understanding this correctly, this should then have hdf5 do the conversion on the way to disk (rather than us doing it in memory ahead of time) which will save memory / performance.

aparamon:
@tacaswell Yes, HDF5 is capable of efficiently converting Datatypes on serialization, and it's a good idea to leverage this. However HDF5 doesn't currently support 16-bit floats, to my best knowledge.

tacaswell:
@aparamon can you take a look at the bits of code I flagged in #819 ?  I am not sure we are doing it in the most efficient way at the bottom anyway...

aparamon:
@tacaswell I have looked into but failed to grok the code :-/

My expectation was that there are some `H5Tregister()` calls specifying conversion mechanisms between "orthodox" (32-bit/64-bit) and half-precision floats; but apparently there are none.

In your last test example in https://github.com/h5py/h5py/issues/819, what is the actual HDF5 storage datatype, and what is `B.dtype`? What is the expected HDF5 storage datatype (float16, I presume)? Does HDFView display the data, and how?

tacaswell:
A cut-down version:

```
13:14 $ h5dump /tmp/test.h5 
HDF5 "/tmp/test.h5" {
GROUP "/" {
   DATASET "Data" {
      DATATYPE  16-bit little-endian floating-point
      DATASPACE  SIMPLE { ( 10, 1 ) / ( 10, 1 ) }
      DATA {
      (0,0): 0.416992,
      (1,0): 0.720215,
      (2,0): 0.000114381,
      (3,0): 0.302246,
      (4,0): 0.146729,
      (5,0): 0.0923462,
      (6,0): 0.186279,
      (7,0): 0.345459,
      (8,0): 0.396729,
      (9,0): 0.538574
      }
   }
}
}

```

I think these types are defined here https://github.com/h5py/h5py/blob/11da4d6813b7830cb6cb04478e7d017b341d064a/h5py/h5t.pyx#L211-L216

Which maybe the real source of the error (if the float is defined wrong)!

aparamon:
@tacaswell Hmm, the data in the file doesn't look particularly similar to `np.linspace`... :-)
I think it dichotomically localizes the issue into writing.
Indeed, does
https://github.com/h5py/h5py/blob/11da4d6813b7830cb6cb04478e7d017b341d064a/h5py/h5t.pyx#L213
or
https://github.com/h5py/h5py/blob/11da4d6813b7830cb6cb04478e7d017b341d064a/h5py/h5t.pyx#L218
correspond to [numpy float16](
https://docs.scipy.org/doc/numpy-1.15.1/user/basics.types.html#array-types-and-conversions-between-types)?
(hopefully the arguments are correctly passed to [H5Tset_fields](https://support.hdfgroup.org/HDF5/doc/RM/RM_H5T.html#Datatype-SetFields))
Is my understanding correct that on your PCs numpy.float16 is little-endian in memory?

tacaswell:
> @tacaswell Hmm, the data in the file doesn't look particularly similar to `np.linspace`... :-)
> I think it dichotomically localizes the issue into writing.

Sorry, that was from the tests which is using a random array!  

> correspond to [numpy float16](https://docs.scipy.org/doc/numpy-1.15.1/user/basics.types.html#array-types-and-conversions-between-types)?

Yes, I think so.  @andrewcollette added that code in b76a7db12dc3e51f5130d7ad6eca02e140b98538 for 2.2 in 2013.

from numpy:

> float16	Half precision float: sign bit, 5 bits exponent, 10 bits mantissa

from https://support.hdfgroup.org/HDF5/doc/RM/RM_H5T.html#Datatype-SetFields 

>Parameters:
>hid_t dtype_id    	IN: Identifier of datatype to set.
>size_t spos	IN: Sign position, i.e., the bit offset of the floating-point sign bit.
>size_t epos	IN: Exponent bit position.
>size_t esize	IN: Size of exponent in bits.
>size_t mpos	IN: Mantissa bit position.
>size_t msize	IN: Size of mantissa in bits.

Which means the call maps to
```python
{'spos': 15,
 'epos': 10,
 'esize': 5,
 'mpos': 0,
 'msize': 10}
```

which looks like it matches the numpy definition and 
https://en.wikipedia.org/wiki/Half-precision_floating-point_format.

This lead me down the rabbit hole of looking at the bit representation of the wrong returned floats and the correct floats from numpy which gave me

```python

import numpy as np
import h5py

A = np.linspace(0, 5, 100_001)

# Save to HDF5 file
with h5py.File("/tmp/Test.hdf5", "w") as Fid:

    Group01 = Fid.create_group("U")
    Group01.create_dataset("Data", data=A, dtype='f2')


with h5py.File("/tmp/Test.hdf5", "r") as Fid:
    B = Fid["U"+"/Data"][:]

# Compare
C = np.abs(A.astype('f2')-B)

failed = C > .001

def to_bits(v):
    bits = '{:016b}'.format(np.asarray(v, 'f2').view('uint16').item())
    return ' '.join((bits[:1], bits[1:6], bits[6:]))

print(f'{"32 bit value":^18s} | {"binary from hdf5":^18s} | {"binary from np":^18s}')
for inp, ret, correct in zip(A[failed], B[failed], A[failed].astype('f2')):
    print(f'{inp:.16f}', '|', to_bits(ret), '|', to_bits(correct))

```

which depending on the range and limit of linspace give you things like

```
   32 bit value    |  binary from hdf5  |   binary from np  
0.2499500000000000 | 0 01100 0000000000 | 0 01101 0000000000
0.4999000000000000 | 0 01101 0000000000 | 0 01110 0000000000
0.4999500000000000 | 0 01101 0000000000 | 0 01110 0000000000
0.9998000000000000 | 0 01110 0000000000 | 0 01111 0000000000
0.9998500000000000 | 0 01110 0000000000 | 0 01111 0000000000
0.9999000000000000 | 0 01110 0000000000 | 0 01111 0000000000
0.9999500000000000 | 0 01110 0000000000 | 0 01111 0000000000
1.9995500000000002 | 0 01111 0000000000 | 0 10000 0000000000
1.9996000000000000 | 0 01111 0000000000 | 0 10000 0000000000
1.9996500000000001 | 0 01111 0000000000 | 0 10000 0000000000
1.9997000000000000 | 0 01111 0000000000 | 0 10000 0000000000
1.9997500000000001 | 0 01111 0000000000 | 0 10000 0000000000
1.9998000000000000 | 0 01111 0000000000 | 0 10000 0000000000
1.9998500000000001 | 0 01111 0000000000 | 0 10000 0000000000
1.9999000000000000 | 0 01111 0000000000 | 0 10000 0000000000
1.9999500000000001 | 0 01111 0000000000 | 0 10000 0000000000
3.9990500000000000 | 0 10000 0000000000 | 0 10001 0000000000
3.9991000000000003 | 0 10000 0000000000 | 0 10001 0000000000
3.9991500000000002 | 0 10000 0000000000 | 0 10001 0000000000
3.9992000000000001 | 0 10000 0000000000 | 0 10001 0000000000
3.9992500000000000 | 0 10000 0000000000 | 0 10001 0000000000
3.9993000000000003 | 0 10000 0000000000 | 0 10001 0000000000
3.9993500000000002 | 0 10000 0000000000 | 0 10001 0000000000
3.9994000000000001 | 0 10000 0000000000 | 0 10001 0000000000
3.9994500000000004 | 0 10000 0000000000 | 0 10001 0000000000
3.9995000000000003 | 0 10000 0000000000 | 0 10001 0000000000
3.9995500000000002 | 0 10000 0000000000 | 0 10001 0000000000
3.9996000000000000 | 0 10000 0000000000 | 0 10001 0000000000
3.9996500000000004 | 0 10000 0000000000 | 0 10001 0000000000
3.9997000000000003 | 0 10000 0000000000 | 0 10001 0000000000
3.9997500000000001 | 0 10000 0000000000 | 0 10001 0000000000
3.9998000000000000 | 0 10000 0000000000 | 0 10001 0000000000
3.9998500000000003 | 0 10000 0000000000 | 0 10001 0000000000
3.9999000000000002 | 0 10000 0000000000 | 0 10001 0000000000
3.9999500000000001 | 0 10000 0000000000 | 0 10001 0000000000

```

The binary values are (sign) (exponent) (mantissa). for the 16 bit floats.  As you can see the exponent read out of hdf5 is off-by-one (it is smaller, but because the exponent is stored with a 15 offset, this gives us bigger numbers by a factor of 2 which is what we see!).

Oddly, don't see this issue with negative numbers.

Given the pattern we see in this, I am pretty sure that this is a bug in hdf5 (but I suppose I will have to write some c-code to be sure).

takluyver:
@tacaswell where did this get to? In some of your comments you seemed unsure if this was the right way to go about things. Do you think it should go in as is, or wait for further consideration?

tacaswell:
I think this one should go in as-is and we need to follow up with hdf5 group about what I think is a bug in their generic custom float -> custom float conversion code.

takluyver:
OK; is there any particular review you'd like before my itchy finger clicks the merge button? ;-)

PR #1116, opened 2018-10-15
Document wrapping a file-like object

takluyver:
Follow-up to #1061.


tacaswell:
Is it worth also showing write / close / read from a BytesIO object?

takluyver:
Added an example script, and included it into the docs.

Issue #1093, opened 2018-09-11
NameError: name 'mpi4py' is not defined when trying to use the "mpio" driver on a Linux cluster


tmaric:
I get  NameError: name 'mpi4py' is not defined when trying to use the "mpio" driver with "h5py" on a Linux cluster. I have installed "mpi4py" and "h5py" locally using 

```
pip install mpi4py --user
pip install h5py --user
```
I have tried to build both "mpi4py" as well as "h5py" (with MPI support) from source and succeeded, but then I get other MPI run-time errors. 

Is the "h5py" that is available via "pip" built with MPI support? If not, should I try to build "h5py" from source with "MPI" support and a path to a parallel HDF5 library?

 * Operating System CentOS Linux release 7.4.1708 (Core)
 * Python version Python 3.6.2
 * system Python Linux
 * h5py version 2.8.0
 * HDF5 version hdf5/parallel/1.10.2
 * The full traceback/stack trace shown

  1 --------------------------------------------------------------------------                                                                                                                
  2 A process has executed an operation involving a call to the                                                                                                                               
  3 "fork()" system call to create a child process.  Open MPI is currently                                                                                                                    
  4 operating in a condition that could result in memory corruption or                                                                                                                        
  5 other system errors; your job may hang, crash, or produce silent                                                                                                                          
  6 data corruption.  The use of fork() (or system() or other calls that                                                                                                                      
  7 create child processes) is strongly discouraged.                                                                                                                                          
  8                                                                                                                                                                                           
  9 The process that invoked fork was:                                                                                                                                                        
 10                                                                                                                                                                                           
 11   Local host:          [[24011,0],0] (PID 30576)                                                                                                                                          
 12                                                                                                                                                                                           
 13 If you are *absolutely sure* that your application will successfully                                                                                                                      
 14 and correctly survive a call to fork(), you may disable this warning                                                                                                                      
 15 by setting the mpi_warn_on_fork MCA parameter to 0.                                                                                                                                       
 16 --------------------------------------------------------------------------                                                                                                                
 17 --------------------------------------------------------------------------                                                                                                                
 18 A process has executed an operation involving a call to the                                                                                                                               
 19 "fork()" system call to create a child process.  Open MPI is currently                                                                                                                    
 20 operating in a condition that could result in memory corruption or                                                                                                                        
 21 other system errors; your job may hang, crash, or produce silent                                                                                                                          
 22 data corruption.  The use of fork() (or system() or other calls that                                                                                                                      
[...]

Issue #1079, opened 2018-08-15
Importing h5py before mpi4py causes OpenMPI warnings


1313e:
**Version information:**
- h5py    2.8.0
- HDF5    1.10.2
- Python  3.6.5 | packaged by conda-forge | (default, Apr  6 2018, 13:39:56)
- [GCC 4.8.2 20140120 (Red Hat 4.8.2-15)]
- sys.platform    linux
- sys.maxsize     9223372036854775807
- numpy   1.15.0

My 'problem' is really simple: Whenever I execute a script that imports h5py before mpi4py in MPI (on a cluster running UNIX), I get the following warning:
```
A process has executed an operation involving a call to the
"fork()" system call to create a child process.  Open MPI is currently
operating in a condition that could result in memory corruption or
other system errors; your job may hang, crash, or produce silent
data corruption.  The use of fork() (or system() or other calls that
create child processes) is strongly discouraged.

The process that invoked fork was:

  Local host:          [[22007,1],2] (PID 94545)

If you are *absolutely sure* that your application will successfully
and correctly survive a call to fork(), you may disable this warning
by setting the mpi_warn_on_fork MCA parameter to 0.
```

And, yes, this script can be as simple as:
```
import h5py
from mpi4py import MPI
```
and running such a script with `mpiexec -n 4 python -m mpi4py test.py`.

I do not get this warning whenever I import mpi4py before h5py.
I have seen this warning appear on multiple different clusters that use OpenMPI (it does not appear when using MPICH2, for example).

Now, I know that this warning can be prevented by simply importing mpi4py before h5py (or not using OpenMPI), and I have actually never seen any damage being caused or anything, but I simply wanted to raise this issue in case it is a sign of a problem.

Anybody else has ever seen a similar warning before?

takluyver:
I'm going to drop the milestone for now, because it's not clear what's going on, and until someone understands the problem properly, we can't really aim to fix it.

There's more discussion continuing on the already merged PR #1117.

msyriac:
Weird, I get the exact same error as @1313e except for the opposite order of imports! (I have a very similar configuration too)

```
from mpi4py import MPI                                                                                                                               
import h5py
print("done")


mpirun -np 1 script.py 

--------------------------------------------------------------------------
A process has executed an operation involving a call to the
"fork()" system call to create a child process.  Open MPI is currently
operating in a condition that could result in memory corruption or
other system errors; your job may hang, crash, or produce silent
[...]

PR #1117, opened 2018-10-15
FIX: hide mpi imports until as late as possible

### LINKED ISSUE #1081: import h5py initializes mpi prematurely
importing h5py inits mpi which then starts orted:

```sh
$ python -c 'import h5py, os; os.system("ps -x|grep orted")'
12485 pts/6    Sl+    0:00 python -c import h5py, os; os.system("ps -x|grep orted")
12530 ?        Ssl    0:00 orted --hnp --set-sid --report-uri 11 --singleton-died-pipe 12 -mca state_novm_select 1
12543 pts/6    S+     0:00 sh -c ps -x|grep orted
12545 pts/6    S+     0:00 grep orted
```

orted inherits the parent's open files so the open sockets are a mess:

```sh
$ lsof -i -P
COMMAND     PID    USER   FD   TYPE DEVICE SIZE/OFF NODE NAME
sirepo    12630 vagrant    3u  IPv4 379187      0t0  TCP *:8000 (LISTEN)
sirepo    12630 vagrant   24u  IPv4 379177      0t0  TCP *:1024 (LISTEN)
orted     12678 vagrant   16u  IPv4 378655      0t0  TCP *:51099 (LISTEN)
python2.7 12688 vagrant    3u  IPv4 379187      0t0  TCP *:8000 (LISTEN)
python2.7 12688 vagrant    4u  IPv4 379187      0t0  TCP *:8000 (LISTEN)
python2.7 12688 vagrant   24u  IPv4 379177      0t0  TCP *:1024 (LISTEN)
python2.7 12688 vagrant   30u  IPv4 379191      0t0  TCP *:1025 (LISTEN)
orted     12692 vagrant    3u  IPv4 379187      0t0  TCP *:8000 (LISTEN)
orted     12692 vagrant   19u  IPv4 378672      0t0  TCP *:39617 (LISTEN)
orted     12692 vagrant   24u  IPv4 379177      0t0  TCP *:1024 (LISTEN)
```

Fedora 27
2.7.14
   Windows)
pyenv
 * h5py version (e.g. 2.6)
2.8.0
 * HDF5 version (e.g. 1.8.17)
hdf5-openmpi-static-1.8.18-11.fc27.x86_64
hdf5-1.8.18-11.fc27.x86_64
hdf5-openmpi-devel-1.8.18-11.fc27.x86_64
hdf5-devel-1.8.18-11.fc27.x86_64
hdf5-openmpi-1.8.18-11.fc27.x86_64
n/a

python -c 'import h5py; print(h5py.version.info)'
```
python -c 'import h5py; print(h5py.version.info)'

h5py    2.8.0
HDF5    1.8.18
Python  2.7.14 (default, Jul  7 2018, 02:04:27)
[GCC 7.3.1 20180303 (Red Hat 7.3.1-5)]
sys.platform    linux2
numpy   1.14.5
```
where `python` should be substituted for the path to python used to install
`h5py` with.

tacaswell:
This may help all of the mpi order-related issues.

closes #1081

### DIFF
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -34,14 +34,13 @@ swmr_support = False
 if hdf5_version >= h5.get_config().swmr_min_hdf5_version:
     swmr_support = True
 
-if mpi:
-    import mpi4py
 
 libver_dict = {'earliest': h5f.LIBVER_EARLIEST, 'latest': h5f.LIBVER_LATEST}
 libver_dict_r = dict((y, x) for x, y in six.iteritems(libver_dict))
 
 
 def _set_fapl_mpio(plist, **kwargs):
+    import mpi4py
     kwargs.setdefault('info', mpi4py.MPI.Info())
     plist.set_fapl_mpio(**kwargs)
 
--- a/h5py/h5p.pyx
+++ b/h5py/h5p.pyx
@@ -29,14 +29,6 @@ from h5py import _objects
 
 from ._objects import phil, with_phil
 
-if MPI:
-    if MPI4PY_V2:
-        from mpi4py.libmpi cimport MPI_Comm, MPI_Info, MPI_Comm_dup, MPI_Info_dup, \
-                               MPI_Comm_free, MPI_Info_free
-    else:
-        from mpi4py.mpi_c cimport MPI_Comm, MPI_Info, MPI_Comm_dup, MPI_Info_dup, \
-                               MPI_Comm_free, MPI_Info_free
-
 # Initialization
 import_array()
 
@@ -1174,6 +1166,15 @@ cdef class PropFAID(PropInstanceID):
             0. The mpi4py.MPI.Comm Communicator
             1. The mpi4py.MPI.Comm Info
             """
+            if MPI4PY_V2:
+                from mpi4py.libmpi cimport (
+                    MPI_Comm, MPI_Info, MPI_Comm_dup, MPI_Info_dup,
+                    MPI_Comm_free, MPI_Info_free)
+            else:
+                from mpi4py.mpi_c cimport (
+                    MPI_Comm, MPI_Info, MPI_Comm_dup, MPI_Info_dup,
+                    MPI_Comm_free, MPI_Info_free)
+
             cdef MPI_Comm comm
             cdef MPI_Info info
 
--- a/h5py/h5p.pyx
+++ b/h5py/h5p.pyx
@@ -29,6 +29,17 @@ from h5py import _objects
 
 from ._objects import phil, with_phil
 
+if MPI:
+    if MPI4PY_V2:
+        from mpi4py.libmpi cimport (
+            MPI_Comm, MPI_Info, MPI_Comm_dup, MPI_Info_dup,
+            MPI_Comm_free, MPI_Info_free)
+    else:
+        from mpi4py.mpi_c cimport (
+            MPI_Comm, MPI_Info, MPI_Comm_dup, MPI_Info_dup,
+            MPI_Comm_free, MPI_Info_free)
+
+
 # Initialization
 import_array()
 
@@ -1166,15 +1177,6 @@ cdef class PropFAID(PropInstanceID):
             0. The mpi4py.MPI.Comm Communicator
             1. The mpi4py.MPI.Comm Info
             """
-            if MPI4PY_V2:
-                from mpi4py.libmpi cimport (
-                    MPI_Comm, MPI_Info, MPI_Comm_dup, MPI_Info_dup,
-                    MPI_Comm_free, MPI_Info_free)
-            else:
-                from mpi4py.mpi_c cimport (
-                    MPI_Comm, MPI_Info, MPI_Comm_dup, MPI_Info_dup,
-                    MPI_Comm_free, MPI_Info_free)
-
             cdef MPI_Comm comm
             cdef MPI_Info info
 


tacaswell:
xref https://github.com/h5py/h5py/issues/1095 https://github.com/h5py/h5py/issues/1093 https://github.com/h5py/h5py/issues/1081 https://github.com/h5py/h5py/issues/1079

attn @tmaric @robnagler @1313e does this fix any of your problems?

robnagler:
Yes, fixes the problem. Thanks!

tacaswell:
The thing I find confusing is that the offending import was added in 2013 (6a626ee04c975380f0d7220df1ee4569469586c5), but these problems have only begun to be reported recently.  Did mpi4py change something about how it initializes it's self?

robnagler:
I think we've had these problems "all along", or at least since 2.7.1: https://github.com/radiasoft/sirepo/issues/1151

tacaswell:
I would have expected to see these issues going back 4 years, that issue is only 6mo's ago ;)

tacaswell:
@1313e can you try going back to mpi4py v2.0.0?  I wonder if they change something about what they do on import?

1313e:
@tacaswell I just performed the test using mpi4py v2.0.0 and I still get the same results as mentioned above: OpenMPI warnings when importing mpi4py before h5py, but not when I import h5py before mpi4py (happens for both h5py v2.8.0 and your branch).
I am really confused now as to why the situation for that warning is swapped.

1313e:
@tacaswell Nothing happens, no warnings.
Although (please correct me if I am wrong), importing the same module twice has no effect due to the way that the Python importer works, right?

tacaswell:
I also expect there to be no effect from importing a module twice, as it should go just get the already imported module from `sys.modules`, but it seems that the second import of `mpi4py` when one of the imports is inside of h5py causes the warnings so something fishy is happening!

I don't really know enough MPI to understand https://mpi4py.readthedocs.io/en/stable/mpi4py.run.html
but the "At import time, mpi4py initializes the MPI execution environment calling MPI_Init_thread()" seems suspicious to me.  I have the sense that this is something you should not do twice.  Could both mpi4py and the underlying hdf5 library are both trying to be helpful and messing with each other?

Again, I know so little about MPI, do you have to re-compile to switch between OppenMPI and MPICH2?   Could there be a conflict between the version mpi4py / h5py / hdf5 are set up for?

Do we have any idea what is doing the forking (I assume mpiexec)?

1313e:
@tacaswell It might be mpi4py, but I only have this problem when using OpenMPI. Using MPICH2 does not cause this problem (and yes, mpi4py needs to be recompiled in order to switch between MPI distributions).
To be honest, that mpi4py automatically initializes the MPI environment has caused me enough headaches already. However, I do know that attempting to initialize the MPI environment when it is already initialized will not work. If you try to initialize it twice, the program itself will simply crash (one can easily check this by executing `from mpi4py import MPI` and `MPI.Init()`).
I also just checked what happens when I import mpi4py before h5py in an iPython console (so, not executed in an MPI environment) and I get the same warnings.

takluyver:
There appear to be several different issues:

- #1081 says importing h5py is doing too much MPI initialisation (possibly by importing mpi4py?)
- #1079 is a fork warning with importing mpi4py after h5py
- #1093, #1095, [this mailing list post](https://groups.google.com/forum/#!topic/h5py/AZRc35i7mGg) and @1313e's comments in this post all seem to refer to seeing the fork warning when importing mpi4py before h5py.

I'm still trying to make head or tail of this.

takluyver:
People who can get the fork warning reliably: can you try using [strace](https://linux.die.net/man/1/strace) to look for fork and exec* system calls and try to work out what is attempting to start a subprocess?

1313e:
@takluyver I can get that fork warning reliably, but using strace does not really give anything.
`fork` never occurs in the entire output and `exec` only once in the form of `mpiexec` (which is my own command).

takluyver:
That's curious. I had a look at the OpenMPI source code, and the warning seems to be hooked up by a handler registered with [pthread_atfork](http://man7.org/linux/man-pages/man3/pthread_atfork.3p.html), which is meant to register handlers to be called on `fork()`. I see only one reference to the message, and no indication that the fork handlers fire in any situation other than a fork. :confused:

takluyver:
In any case, I think delaying the import of mpi4py until the mpio driver is used makes sense. It appears to solve at least one of the issues related to MPI (#1081), and it doesn't appear to make other issues worse. So I suggest we merge this change and continue to investigate the causes of the fork warning separately.

1313e:
@takluyver Sounds like a good idea.

leofang:
I wandered to this page and wonder if the conflicts between `h5py` and `mpi4py` has been fully resolved. Our code also encountered the similar situation if `h5py` is built against an MPI-enabled hdf5. Our solution is to not initialize MPI and let `h5py` do the job:

```Python
import mpi4py
mpi4py.rc.initialize = False # undocumented switch
from mpi4py import MPI
assert not MPI.Is_initialized()
import h5py # must be loaded after mpi4py, or it'd initialize MPI if parallel HDF5 is built!
# ...and do the work
```

This, together with mpio, seems to work for us. In this case we used `mpi4py` v3.0.0 and `h5py` v2.7.1.

tacaswell:
@leofang can you try this with master branch?

leofang:
@tacaswell sorry I missed the notification. I just updated my earlier comment slightly. The code snippet above works for both h5py v2.8.0 and the current master branch.

I want to add that we never saw error messages as reported in #1079, and I'm wondering if the linked issues are related to how MPI, HDF5, etc are built and not to h5py itself. 

The problem we encountered is that there's a certain initialization order when using MPI together with (CUDA) GPU, and importing a parallel hdf5/h5py initialized MPI prematurely and thus messed up this order, as observed in #1081 and also revealed in the script below (tested with h5py v2.8.0 and master):
```python
import h5py # MPI is already initialized here...
import mpi4py
mpi4py.rc.initialize = False # undocumented switch
from mpi4py import MPI

# MPI.Is_initialized() is supposed to be False given the switch is turned off,
# but surprisingly it's True because h5py is imported first!
assert MPI.Is_initialized()
``` 
I would therefore suggest to load h5py always after mpi4py.

1313e:
I personally still have the feeling that this is something that occurs only when using OpenMPI (since it does not happen with MPICH2 or Microsoft MPI), which would not surprise me given that I have found several other 'problems' with it as well.
But, since `h5py` is the only package that does this with OpenMPI (using other packages that use `mpi4py` do not raise that warning), I figured it had something to do with `h5py`.
I do not know how however, so do not ask me that. :)

1313e:
> Just a bit of digression: I have no idea M$ has its own MPI!

It is the only MPI implementation that one can actually use on Windows.

DTHaydon:
I have a MPI HDF5 library compiled.

When I import h5py built from this library:

    import h5py
    # ---------- or ----------
    import h5py
    from mpi4py import MPI
    # ---------- or ----------
    from mpi4py import MPI
    import h5py
    # ---------- or ----------
    import mpi4py

    # combinations of the following
    mpi4py.rc.initialize = False
    mpi4py.rc.threads = False
    mpi4py.rc.thread_level = "single"

    from mpi4py import MPI
    import h5py
 
I always get the warning:

> A process has executed an operation involving a call to the
> "fork()" system call to create a child process.  Open MPI is currently
> operating in a condition that could result in memory corruption or
> other system errors; your job may hang, crash, or produce silent
> data corruption.  The use of fork() (or system() or other calls that
> create child processes) is strongly discouraged.

However I compiled and ran the C code example from:
https://support.hdfgroup.org/ftp/HDF5/current/src/unpacked/examples/ph5example.c
(which creates and writes HDF5 files in parallel)
using the same MPI HDF5 library that I built h5py and there is no warning.

Do we know if it's something `mpi4py` is doing of `h5py`?

------------
h5py 2.9.0
mpi4py 3.0.0
HDF5 1.10.4
Open MPI 3.0.0
Linux 3.10.0-693.5.2.el7.x86_64 x86_64

leofang:
Hi @DTHaydon, not sure if this would help, but could you please try newer version of Open MPI, say 3.1.3? According to its [changelog](https://raw.githubusercontent.com/open-mpi/ompi/v4.0.x/NEWS), in 3.0.1 they
> Fix a number of issues in MPI I/O found by the HDF5 test suite.
Unfortunately you'd also need to rebuild everything (HDF5, mpi4py, h5py) to ensure they all work on top of the new version I think.

Also, for all affected users, could you please report your Open MPI version? I don't find *any* in the linked issues. Thanks.

DTHaydon:
Unfortunately this is on a university cluster, so I don't have so much control of what MPI version I'm using.
However, if the Open MPI version is the issue, I would expect the C code to also give the same fork() warning.

Issue #910, opened 2017-08-08
Broadcasting for assignment with singleton dimensions


swehrwein:
Simple indexing with singleton dimensions seems to fail in a couple different ways; the issue seems to be in selections.py in the broadcasting code where it "skips over singleton dimensions" but doesn't pop them from the target shape if they're in there as ones. The code below demos two examples of this. As shown in the demo, a workaround is to np.squeeze() the right hand side first.

```
import numpy as np
import h5py

with h5py.File('foo.h5', 'w') as f:
    ds = f.create_dataset("foo", (2, 10, 10, 4))

    x = np.zeros((2, 10, 10, 1))
    ds[:, :, :, 2] = np.squeeze(x) # works
    ds[:, :, :, 2] = x # barfs: can't broadcast

with h5py.File('bar.h5', 'w') as f:
    ds = f.create_dataset("bar", (10, 4))

    x = np.zeros((10, 1))
    ds[:, 2] = np.squeeze(x) # works
    ds[:, 2] = x # barfs: can't prepare for writing data
```

Summary of the h5py configuration
---------------------------------
h5py    2.7.0
HDF5    1.8.18
Python  3.4.3 (default, Nov 17 2016, 01:08:31)
[GCC 4.8.4]
sys.platform    linux
sys.maxsize     9223372036854775807
numpy   1.13.1

takluyver:
Still reproducible on master:

```
$ python3 singleton_dims.py 
Traceback (most recent call last):
  File "singleton_dims.py", line 9, in <module>
    ds[:, :, :, 2] = x # barfs: can't broadcast
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "/home/takluyver/miniconda3/envs/h5py-vds/lib/python3.6/site-packages/h5py/_hl/dataset.py", line 685, in __setitem__
    for fspace in selection.broadcast(mshape):
  File "/home/takluyver/miniconda3/envs/h5py-vds/lib/python3.6/site-packages/h5py/_hl/selections.py", line 299, in broadcast
    raise TypeError("Can't broadcast %s -> %s" % (target_shape, count))
TypeError: Can't broadcast (2, 10, 10, 1) -> (2, 10, 10, 1)
```

takluyver:
[...]

PR #1121, opened 2018-10-24
Clearer broadcasting errors with singleton dimensions

takluyver:
From discussion on #910.

```python
import numpy as np
import h5py

with h5py.File('foo.h5', 'w') as f:
    ds = f.create_dataset("foo", (2, 10, 10, 4))

    ds[:, :, :, 2] = np.zeros((2, 10, 10, 1))  # Error A

with h5py.File('bar.h5', 'w') as f:
    ds = f.create_dataset("bar", (10, 4))

    ds[:, 2] = np.zeros((10, 1))  # Error B
```

Error A:

```
Before: TypeError: Can't broadcast (2, 10, 10, 1) -> (2, 10, 10, 1)
After:  TypeError: Can't broadcast (2, 10, 10, 1) -> (2, 10, 10)
```

Error B:

```
Before: OSError: Can't prepare for writing data (src and dest data spaces have different sizes)
After:  TypeError: Can't broadcast (10, 1) -> (10,)
```

### DIFF
--- a/h5py/_hl/selections.py
+++ b/h5py/_hl/selections.py
@@ -296,7 +296,13 @@ class SimpleSelection(Selection):
                 if t == 1 or count[-idx] == t:
                     tshape.append(t)
                 else:
-                    raise TypeError("Can't broadcast %s -> %s" % (target_shape, count))
+                    raise TypeError("Can't broadcast %s -> %s" % (target_shape, self.mshape))
+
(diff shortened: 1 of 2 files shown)


takluyver:
Somewhat cheekily milestoneing this for 2.9; feel free to bump it to a later milestone if you disagree.

tacaswell:
Thanks!

PR #1122, opened 2018-10-24
Expose the full H5Sselect_hyperslab API to _hl.selections

GDYendell:
This is an idea of how to implement 4-tuple (start, count, stride, block) selections in the high-level VDS API, as defined in the [HDF5 docs](https://support.hdfgroup.org/HDF5/Tutor/selectsimple.html).

I have a use case of interleaving frames written to multiple files into a single VDS. I initially did this with a per frame mapping, but by making use of the block selection I have now implemented it with a per file mapping, which is much quicker (4 maps in <1ms vs 1,000,000 maps in ~35 seconds, with 250,000 frames in each file) and produces much smaller files. This is implemented [here](https://github.com/dls-controls/vds-gen/commit/804352b429456850c690b6233c511d21268d3630#diff-b16ac9c817c68af1282779934f0ef4e3) by subclassing the VirtualLayout class and overriding `__set_item__`.

Here is an excerpt of its usage in my code with 3D (axis, height, width) datasets:

```
...
v_layout = AdvancedVirtualLayout(target_shape, source_meta.dtype)
...
for file_idx, file_path in enumerate(self.files):
    ...
    v_source = h5.VirtualSource(
        file_path,
        name="data", shape=source_shape, dtype=source_meta.dtype
    )
    v_layout[(start, count, stride, block), :, :] = v_source
...
```

I thought this might be more generally useful, so I have created this addition in h5py, which can achieve the same results using the existing VirtualLayout class.

Some things for discussion:

- I didn't know exactly where to implement this, I went with this as a balance of duplicating as little code as possible and not making invasive changes. There are other options:

   1. It could be made more vds specific moving it up into VirtualLayout, as I have in [my version](https://github.com/dls-controls/vds-gen/commit/804352b429456850c690b6233c511d21268d3630#diff-b16ac9c817c68af1282779934f0ef4e3) (i.e. by avoiding using SimpleSelection at all if a tuple element is given and directly creating a dataspace and calling select_hyperslab instead).

   2. It could be made more generic by moving it into the the SimpleSelection class directly (depending on the definition of _simple_). This would mean it could be used outside of the vds context; I don't know if this would be useful or not. This itself could be done in two ways:

      - Checking for tuple elements in `__set_item__` and calling `_handle_simple` or `_handle_advanced`
      - Replacing `_handle_simple`, as `_handle_advanced` should do the same thing in the case that only slices and ints are given.

- I am also not sure of the implication of defining a selection with block of more than 1 on `self._sel` and `self._mshape`. Currently, I have defined them as before, ignoring the block size.

- Most importantly: Is this a worthwhile addition?

### DIFF
--- a/examples/multiblockslice_interleave.py
+++ b/examples/multiblockslice_interleave.py
@@ -1,0 +1,36 @@
+"""
+An example of using MultiBlockSlice to interleave frames in a virtual dataset.
+"""
+import h5py
+
+
+# These files were written round-robin in blocks of 1000 frames from a single source
+# 1.h5 has 0-999, 4000-4999, ...; 2.h4 has 1000-1999, 5000-5999, ...; etc
+# The frames from each block are contiguous within each file
+# e.g. 1.h5 = [..., 998, 999, 4000, 4001, ... ]
+files = ["1.h5", "2.h5", "3.h5", "4.h5"]
+dataset_name = "data"
+dtype = "float"
+source_shape = (25000, 256, 512)
+target_shape = (100000, 256, 512)
+block_size = 1000
+
+v_layout = h5py.VirtualLayout(shape=target_shape, dtype=dtype)
+
+for file_idx, file_path in enumerate(files):
+    v_source = h5py.VirtualSource(
+        file_path, name=dataset_name, shape=source_shape, dtype=dtype
+    )
+    dataset_frames = v_source.shape[0]
+
+    # A single raw file maps to every len(files)th block of frames in the VDS
+    start = file_idx * block_size  # 0, 1000, 2000, 3000
+    count = dataset_frames // block_size  # 25
+    stride = len(files) * block_size  # 4000
+    block = block_size  # 1000
+
+    # MultiBlockSlice for frame dimension and full extent for height and width
+    v_layout[h5py.MultiBlockSlice(start, count, stride, block), :, :] = v_source
+
+with h5py.File("interleave_vds.h5", "w", libver="latest") as f:
+    f.create_virtual_dataset(dataset_name, v_layout, fillvalue=0)
--- a/h5py/_hl/selections.py
+++ b/h5py/_hl/selections.py
@@ -45,7 +45,7 @@ def select(shape, args, dsid):
     RegionReference
         Returns a Selection instance.
 
-    Indices, slices, ellipses only
+    Indices, slices, ellipses, MultiBlockSlices only
         Returns a SimpleSelection instance
 
     Indices, slices, ellipses, lists or boolean index arrays
@@ -76,7 +76,7 @@ def select(shape, args, dsid):
             return Selection(shape, spaceid=sid)
 
     for a in args:
-        if not isinstance(a, slice) and a is not Ellipsis:
+        if not isinstance(a, (slice, MultiBlockSlice)) and a is not Ellipsis:
             try:
                 int(a)
                 if isinstance(a, np.ndarray) and a.shape == (1,):
@@ -89,6 +89,71 @@ def select(shape, args, dsid):
     sel = SimpleSelection(shape)
     sel[args]
     return sel
+
+
+class MultiBlockSlice(object):
+
+    """
+        A conceptual extension of the built-in slice object to allow selections
+        using start, stride, count and block.
+
+        If given, these parameters will be passed directly to
+        H5Sselect_hyperslab. The defaults are start=0, stride=1, block=1,
+        count=length, which will select the full extent.
+
+        __init__(start, stride, count, block) => Create a new MultiBlockSlice, storing
+            any given selection parameters and using defaults for the others
+        start => The offset of the starting element of the specified hyperslab
+        stride => The number of elements between the start of one block and the next
+        count => The number of blocks to select
+        block => The number of elements in each block
+
+    """
+
+    def __init__(self, start=0, stride=1, count=None, block=1):
+        if start < 0:
+            raise ValueError("Start can't be negative")
+        if stride < 1 or (count is not None and count < 1) or block < 1:
+            raise ValueError("Stride, count and block can't be 0 or negative")
+        if block > stride:
+            raise ValueError("Blocks will overlap if block > stride")
+
+        self.start = start
+        self.stride = stride
+        self.count = count
+        self.block = block
+
+    def indices(self, length):
+        """Calculate and validate start, count, stride and block for the given length"""
+        if self.count is None:
+            # Select as many full blocks as possible without exceeding extent
+            count = (length - self.start - self.block) // self.stride + 1
+            if count < 1:
+                raise ValueError(
+                    "No full blocks can be selected using {} "
+                    "on dimension of length {}".format(self._repr(), length)
+                )
+        else:
+            count = self.count
+
+        end_index = self.start + self.block + (count - 1) * self.stride - 1
+        if end_index >= length:
+            raise ValueError(
+                "{} range ({} - {}) extends beyond maximum index ({})".format(
+                    self._repr(count), self.start, end_index, length - 1
+                ))
+
+        return self.start, count, self.stride, self.block
+
+    def _repr(self, count=None):
+        if count is None:
+            count = self.count
+        return "{}(start={}, stride={}, count={}, block={})".format(
+            self.__class__.__name__, self.start, self.stride, count, self.block
+        )
+
+    def __repr__(self):
+        return self._repr(count=None)
 
 
 class Selection(object):
@@ -247,14 +312,17 @@ class SimpleSelection(Selection):
             self._id.select_all()
             return self
 
-        start, count, step, scalar = _handle_simple(self.shape,args)
+        start, count, stride, block, scalar = _handle_simple(self.shape, args)
 
-        self._id.select_hyperslab(start, count, step)
+        self._id.select_hyperslab(start, count, stride, block)
 
-        self._sel = (start, count, step, scalar)
+        length = tuple(count * block for count, block in zip(count, block))
+        self._sel = (start, length, stride, scalar)
 
         # array shape drops dimensions where a scalar index was selected
-        self._array_shape = tuple(x for x, y in zip(count, scalar) if not y)
+        self._array_shape = tuple(
+            dim_length for dim_length, scalar in zip(length, scalar) if not scalar
+        )
 
         return self
 
@@ -411,8 +479,8 @@ class FancySelection(Selection):
 
         self._id.select_none()
         for idx, vector in enumerate(argvector):
-            start, count, step, scalar = _handle_simple(self.shape, vector)
-            self._id.select_hyperslab(start, count, step, op=h5s.SELECT_OR)
+            start, count, stride, block, scalar = _handle_simple(self.shape, vector)
+            self._id.select_hyperslab(start, count, stride, block, op=h5s.SELECT_OR)
 
         # Final shape excludes scalars, except where
         # they correspond to sequence entries
@@ -460,38 +528,48 @@ def _expand_ellipsis(args, rank):
 
     return final_args
 
+
 def _handle_simple(shape, args):
-    """ Process a "simple" selection tuple, containing only slices and
-        integer objects.  Return is a 4-tuple with tuples for start,
-        count, step, and a flag which tells if the axis is a "scalar"
-        selection (indexed by an integer).
+    """ Process a "simple" selection tuple, containing only integers, slices
+        or MultiBlockSlices.
+        Return is a 5-tuple with tuples for start, count, stride, block plus a
+        flag which tells us if the axis is a "scalar" selection (indexed by an
+        integer).
 
         If "args" is shorter than "shape", the remaining axes are fully
         selected.
     """
-    args = _expand_ellipsis(args, len(shape))
-
     start = []
     count = []
-    step  = []
+    stride = []
+    block = []
     scalar = []
 
+    args = _expand_ellipsis(args, len(shape))
+
     for arg, length in zip(args, shape):
+        _scalar = False
         if isinstance(arg, slice):
-            x,y,z = _translate_slice(arg, length)
-            s = False
+            _start, _count, _stride = _translate_slice(arg, length)
+            _block = 1
+        elif isinstance(arg, MultiBlockSlice):
+            _start, _count, _stride, _block = _translate_multi_block_slice(arg, length)
         else:
             try:
-                x,y,z = _translate_int(int(arg), length)
-                s = True
+                _start, _count, _stride = _translate_int(int(arg), length)
+                _block = 1
+                _scalar = True
             except TypeError:
                 raise TypeError('Illegal index "%s" (must be a slice or number)' % arg)
-        start.append(x)
-        count.append(y)
-        step.append(z)
-        scalar.append(s)
 
-    return tuple(start), tuple(count), tuple(step), tuple(scalar)
+        start.append(_start)
+        count.append(_count)
+        stride.append(_stride)
+        block.append(_block)
+        scalar.append(_scalar)
+
+    return tuple(start), tuple(count), tuple(stride), tuple(block), tuple(scalar)
+
 
 def _translate_int(exp, length):
     """ Given an integer index, return a 3-tuple
@@ -525,6 +603,15 @@ def _translate_slice(exp, length):
     count = 1 + (stop - start - 1) // step
 
     return start, count, step
+
+
+def _translate_multi_block_slice(exp, length):
+    """ Given a MultiBlockSlice object, return a 4-tuple
+        (start, count, stride, block) for use with the hyperslab selection
+        routines.
+    """
+    return exp.indices(length)
+
 
 def guess_shape(sid):
     """ Given a dataspace, try to deduce the shape of the selection.
--- a/h5py/tests/test_slicing.py
+++ b/h5py/tests/test_slicing.py
@@ -22,7 +22,7 @@ from .common import ut, TestCase
 
 import h5py
 from h5py import h5s, h5t, h5d
-from h5py import File
+from h5py import File, MultiBlockSlice
 
 class BaseSlicing(TestCase):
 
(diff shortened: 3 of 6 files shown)


takluyver:
Thanks, I think it's worthwhile to have a way to select multiple blocks at once like this. Out of interest, have you seen a performance difference for accessing data through per-file and per-frame mappings?

I think we should find a more explicit API than jamming tuples into Python slicing syntax, though. Perhaps a method like `.select_blocks()`? Or perhaps some of the objects from the `selections` module should become properly documented public API, and you could make a simpler selection and then extend it with a `.repeat()` method or something similar?

(I'm hoping @tacaswell & @aragilar will weigh in with their ideas or preferences here)

I'd like to avoid adding an `AdvancedSelection`, because `SimpleSelection` is already in contrast to `FancySelection`; 'fancy' and 'advanced' don't have a clear distinction. But we can work out implementation and naming once we've figured out what the API should look like.

takluyver:
Just to add: I'm thinking that whatever API we come up with to select multiple blocks should also work on a regular dataset object, returning an array with an extra dimension for the blocks. It may be less important there, because it's less of a problem to select each block individually, but there could still be cases where it's useful, and it preserves the helpful symmetry between a `Dataset` and a `VirtualSource`.

GDYendell:
@takluyver 

>  Out of interest, have you seen a performance difference for accessing data through per-file and per-frame mappings?

The performance difference in opening the dataset and accessing hyperslab `[:, :, 0]` is also quite noticeable; ~1s  for per-frame vs ~0.01s for per file. There was initially a problem with accessing large datasets created with these tiled selections, where just opening the dataset would take minutes. This has been (quietly) fixed in 1-10-2 (the writer has to be 1-10-2). 

> I think we should find a more explicit API than jamming tuples into Python slicing syntax, though. Perhaps a method like `.select_blocks()`?

The slicing API is unfortunately limiting in this way. I agree, using the tuple in square brackets is not very Pythonic. I guess I wanted to allow square brackets logic, as well as allowing use of `:` and `...` nicely alongside the 4-tuple. I think it looks really clean, but perhaps it looks too unfamiliar / confusing?

A similar option would be passing a tuple of `(stride, block)` as the `step` of the `slice` object, like this:

```
v_layout[start:count:(stride, block), :, :] = v_source
```
but, while mapping `step` to `(stride, block)` does make logical sense, mapping `stop` to `count` might not be clear.

By adding explicit methods we could have

```
v_layout.set_blocks((start, stop, stride, block), ..., v_source)
```

or possibly

```
v_layout.select_blocks((start, stop, stride, block), ...) = v_source
```

where the assignment would do the set_virtual for you under the hood.

One question I would have with this is would I then have to pass `slice(None)` to select an entire axis, like so

```
v_layout.set_hyperslab((start, stop, stride, block), slice(None), slice(None), v_source)
```

or is there a nicer way?

> Or perhaps some of the objects from the selections module should become properly documented public API, and you could make a simpler selection and then extend it with a `.repeat()` method or something similar?

Exposing selections could be work. You could have `get_selection` / `set_selection` on `VirtualSource` and `VirtualLayout` that each take the tuple selection, then have an explicit `add_source` to append a mapping instead of using assignment with square brackets.

Adding on block to a (start, count, stride) selection may be a bit confusing. Block needs to be matched correctly with stride to make sure the blocks don't overlap (and with count to make sure the last block doesn't overrun). I.e., you get this:

```
  File "h5py/h5s.pyx", line 571, in h5py.h5s.SpaceID.select_hyperslab
  H5Sselect_hyperslab(self.id, <H5S_seloper_t>op, start_array,
RuntimeError: Unable to set hyperslab selection (hyperslab blocks overlap)
```

I think it is better to be fully explicit and pass all four elements together.

> I'd like to avoid adding an `AdvancedSelection`, because `SimpleSelection` is already in contrast to `FancySelection`

I agree, it would not be clear to have both. In fact I initially thought `FancySelection` might already do what I want. If you want to make this more generic than just vds, I would lean towards extending SimpleSelection to allow block, which I think is included the HDF5 meaning of simple.

I don't have sufficient knowledge to comment on the implication of extending the change to `Dataset` as well.

GDYendell:
I have rebased on 2.9.0 (I got errors in the tests on master) to resolve the conflicts after updates to master. I have also updated the changes to just directly extend the "simple" classes and methods to support 4-tuples, removing the concept of "advanced".

I think it might be better to create a public `Hyperslab` object that could be created with N 4-tuples, defining an N-dimensional hyperslab selection. This could then be used to index the `VirtualSource` and `VirtualLayout` classes with and checked for in `_handle_simple` instead of tuples. Would that be a better solution versus slicing with tuples directly, @takluyver ?

Something like:

```
source_hyperslab = Hyperslab((0, 10, 1, 1), None, None, shape=(10, 256, 256))
target_hyperslab = Hyperslab((0, 40, 4, 1), None, None, shape=(40, 256, 256))

v_layout[target_hyperslab] = v_source[source_hyperslab]
```

where `None` would be equivalent to `:` in slicing syntax to say use the entire axis. Possibly there is a better option for this.

takluyver:
It's the 4-tuples themselves that I don't find very clear - it's not obvious which number means what. I think slicing already provides a reasonable interface to select a hyperslab, but what it's missing is the notion of repeating blocks. I see a couple of possible ways this could be added:

```python
# 1: Blocks slice objects
# conceptually an extension of slice() to allow block sizes > 1
v_source[blocks_slice(0, 12, step=4, block=2), :, :]

# 2: Selections as a first class API
sel = Selection(v_source.shape)[0:2, :, :]
sel.repeat_block(stride=4, stop=12, axis=0)
# or with a number of blocks:
sel.repeat_block(stride=4, n=3, axis=0)
v_source[sel]

# 3: Selection methods on VirtualSource & VirtualLayout objects
v_source.select_blocks(0, 12, stride=4, block=2, axis=0)
```

These examples would all select indexes 0, 1, 4, 5, 8, 9 on the first axis, if I've understood the intent correctly.

takluyver:
```
2a: Alternative way to build blocks
sel = Selection(v_source.shape)[0:12:4, :, :]
sel.fill_blocks(length=2, axis=0)
v_source[sel]
```

In example 2, you're selecting the whole of the first block, and then declaring that it should be repeated at regular intervals. In 2a, you're selecting the first element of each block, and then extending the blocks to 2 elements. Of these two styles, I think I marginally prefer 2a.

GDYendell:
I think I initially misunderstood your reservations. I like option 1. It's clear and intuitive to both hdf5 and python users. It could be used on `Dataset` objects as you mentioned before. I can amend my local branch to use this.

I don't know if `blocks_slice` is a clear name, though. It can't really be `Hyperslab`, because that implies a selection in all dimensions. Would `hyperslice` make sense?

takluyver:
I don't like the hyper- prefix for this, because that implies multidimensionality, whereas this would just be selecting in one dimension. I'm thinking of it as blocks because you can select e.g. a block of 2 at each step of 4, whereas standard slicing only lets you select every 4th element (a 'block' of 1). But there may well be a better name for this which hasn't occurred to me.

I'm also aware that I'm the only one to have commented on this PR so far; it might be a good idea to wait for more input, particularly from @tacaswell or @aragilar, before spending time implementing my suggestions. They may have different ideas.

GDYendell:
Added a Slice object to pass the four slicing arguments into in a clearer way. I think this API is quite clean. I thought the name `Slice` matched well with the current `File` object; builtin `file` -> `h5.File`, so builtin `slice` -> `h5.Slice`.

takluyver:
I'd still like the naming to be more explicit, e.g. `BlocksSlice`. I appreciate the analogy with `h5py.File`, but I think it's much more obvious what the `File` class in a module for working with HDF5 files is. Also, the `file` builtin is gone in Python 3. I really don't think it's clear how an HDF5 'slice' differs from a normal Python one.

I'm open to finding a better naming pattern than 'blocks' - I get the feeling that you don't like that. But 'hyper' isn't right, and I haven't yet thought of any other term. My rationale is that the difference between this and regular Python slices is the ability to select a block of n contiguous pieces at each stride.

Other than that, I'm happy enough with the proposed API. It would of course need some docs and tests.

GDYendell:
Yeah it's a tricky one; names are hard. I am happy to use the `Blocks` prefix if nothing better falls out. My reservation is I am not sure it explains the distinction very well, because you can use multiple blocks to define a regular contiguous slice (start=0, count=N, stride=1, block=1) or you could have only one (start=0, count=1, stride=1, block=N). Really it's the stride>1 that makes it interesting. I might contact the hdf group to see if they already have a technical definition for what we are trying to describe.

Yes I am happy to add tests and docs if we are vaguely happy with the API. I can always refactor the name afterwards. I was also thinking of adding something to `examples` to show how it could be used to interleave frames.

takluyver:
I think I see - you're coming at it from the angle of "you can already define a block with a regular Python slice, this adds multiple blocks with a stride", and I'm thinking "you can already define a strided selection with a regular Python slice, this allows a block at each stride".

Maybe something like `MultiBlockSlice` would be clearer, if a bit lengthy? A Python slice allows either a single 'block' or multiple points defined by a step, but not both. If you assume that a single entry at each step doesn't count as a 'block', at least.

Or make it very explicit, like `BlockStrideSlice`? Or ditch the slice naming for something like `StridedBlocks`? Arguably the slice naming is misleading anyway, because this uses a count rather than a stop-index like Python's slice.

takluyver:
Ping @GDYendell - do you think you'll have time to come back to this soon? Sorry to pester you after I left it alone for a few months, but we're closing in on 2.10, and at some point we'll have to postpone PRs that aren't ready to be merged.

No problem if not - it can go in 3.0, which I imagine we'll be eager to get to soon.

GDYendell:
Not a problem. I will not have a chance to work on this for the next week or two. When I can I will check the state of 2.10 and make a comment on here.

takluyver:
OK, thanks. There are currently just 5 other issues and PRs left to consider for 2.10, so if we get to the release sooner, I'll bump this one to the 3.0 milestone.

GDYendell:
@takluyver it is not clear to me where I should add/modify tests for my changes. Are there already tests for selections.py?

takluyver:
I think either `h5py.tests.old.test_slicing` or `h5py.tests.hl.test_dataset_getitem` would be good places to add some tests. There's also `h5py.tests.old.test_selections` if you want some more unit-y tests.

(I've not figured out why some of the tests are marked 'old' - they still seem useful, and I'm not aware of any plan to get rid of them)

takluyver:
I opened #1281 about reorganising the tests.

GDYendell:
Hi @takluyver . I have rebased on master refactored this branch a bit. I implemented the new slicing object as `MultiBlockSlice` under advisement from Elena of the hdf group. I think I now understand what _sel and _array_size in `SimpleSelection` need to be set to when given a `MultiBlockSlice` and I have updated it as such. I added some error checking in the `MultiBlockSlice` `__init__` and `indices` to help protect against user error, but I may be missing some other edge cases. I have also added some docs and an example. My vdsgen module now runs from this branch (with the patching of h5py objects I had in place removed) and the various convoluted system tests that live in there all pass.

I think this is close now. Let me know if you need anything else adding or any changes made.

GDYendell:
I have improved the docs and comments in the tests and example script that I could see were needed after returning to this after a few months.

h5py/_hl/selections.py
```diff
@@ -91,6 +91,59 @@ def select(shape, args, dsid):
...
+            raise ValueError("Count, block and stride must be non-zero")
+        if block > stride:
+            raise ValueError("Blocks will overlap if block > stride")
+
+        self.start = start
+        self.count = count
+        self.stride = stride
+        self.block = block
+
+    def indices(self, length):
```
scopatz, line 118: Please add a docstring here

h5py/_hl/selections.py
```diff
@@ -91,6 +91,59 @@ def select(shape, args, dsid):
...
 
+class MultiBlockSlice(object):
+
+    """
+        A conceptual extension of the built-in slice object to allow selections
+        using start, stride, count and block.
+
+        If given, these parameters will be passed directly to
+        H5Sselect_hyperslab. The defaults are start=0, stride=1, block=1,
+        count=length, which will select the full extent.
```
scopatz, line 102: Please have the parameter description here match that of the surrounding classes

GDYendell:
I have added to the doc strings as requested.

I am wondering now if `start`, `stop`, `stride` and `count` should have leading underscores. They shouldn't be changed directly - they shouldn't really even need to be read directly, only via `indices()`.

scopatz:
Yep the docstrings look better to me now. Thanks! @takluyver?

examples/multiblockslice_interleave.py
```diff
@@ -0,0 +1,36 @@
+"""
+An example of using MultiBlockSlice to interleave frames.
```
takluyver, line 2:
```suggestion
An example of using MultiBlockSlice to interleave frames in a virtual dataset.
```

h5py/_hl/selections.py
```diff
@@ -91,6 +91,68 @@ def select(shape, args, dsid):
...
+        count => The number of blocks to select
+        stride => The number of elements between the start of one block and the next
+        block => The number of elements in each block
+
+    """
+
+    def __init__(self, start=0, count=None, stride=1, block=1):
+        if start < 0:
+            raise ValueError("Start must be positive")
+        if count is not None and count < 1 or stride < 1 or block < 1:
```
takluyver, line 116:
```suggestion
        if (count is not None and count < 1) or stride < 1 or block < 1:
```

It's probably correct for the computer anyway, but I basically always want parentheses where the precedence matters, so I don't have to think about how Python evaluates it.

h5py/_hl/selections.py
```diff
@@ -91,6 +91,68 @@ def select(shape, args, dsid):
...
+            raise ValueError("Count, block and stride must be non-zero")
```
takluyver, line 117:
```suggestion
            raise ValueError("Count, block and stride can't be 0 or negative")
```

h5py/_hl/selections.py
```diff
@@ -91,6 +91,68 @@ def select(shape, args, dsid):
...
+        start => The offset of the starting element of the specified hyperslab
+        count => The number of blocks to select
+        stride => The number of elements between the start of one block and the next
+        block => The number of elements in each block
+
+    """
+
+    def __init__(self, start=0, count=None, stride=1, block=1):
+        if start < 0:
+            raise ValueError("Start must be positive")
```
takluyver, line 115:
```suggestion
            raise ValueError("Start can't be negative")
```

It can be 0, which isn't positive. ;-) This is nitpicking, but I wouldn't want to upset the mathematicians.

h5py/_hl/selections.py
```diff
@@ -91,6 +91,68 @@ def select(shape, args, dsid):
...
+        __init__(start, count, stride, block) => Create a new MultiBlockSlice, storing
+            any given selection parameters and using defaults for the others
+        start => The offset of the starting element of the specified hyperslab
+        count => The number of blocks to select
+        stride => The number of elements between the start of one block and the next
+        block => The number of elements in each block
+
+    """
+
+    def __init__(self, start=0, count=None, stride=1, block=1):
```
takluyver, line 113: Is there a reason for this ordering? In [H5Sselect_hyperslab](https://confluence.hdfgroup.org/display/HDF5/H5S_SELECT_HYPERSLAB), the order is start, stride, count, block.

h5py/_hl/selections.py
```diff
@@ -91,6 +91,68 @@ def select(shape, args, dsid):
...
+                raise ValueError(
+                    "No full blocks can be selected using {} "
+                    "on dimension of length {}".format(self._repr(), length)
+                )
+        else:
+            count = self.count
+
+        end_index = self.start + self.block + (count - 1) * self.stride - 1
+        if end_index >= length:
+            raise ValueError(
```
takluyver, line 141: This should never happen if count is calculated, right? The other error would catch it in that case. No problem to have an extra sanity check, I'm just making sure that I'm understanding the code correctly.

h5py/_hl/selections.py
```diff
@@ -91,6 +91,68 @@ def select(shape, args, dsid):
...
+                ))
+
+        return self.start, count, self.stride, self.block
+
+    def _repr(self, count=None):
+        if count is None:
+            count = self.count
+        return "{}(start={}, stride={}, count={}, block={})".format(
+            self.__class__.__name__, self.start, self.stride, count, self.block
+        )
```
takluyver, line 153: Why not use this as an actual `__repr__` method? It may be handy to inspect these objects.

h5py/tests/test_slicing.py
```diff
@@ -316,3 +316,103 @@ def test_write_noncompound(self):
...
+        super(TestMultiBlockSlice, self).setUp()
+        self.arr = np.arange(10)
+        self.dset = self.f.create_dataset('x', data=self.arr)
+
+    def test_default(self):
+        # Default selects entire dataset as one block
+        mbslice = MultiBlockSlice()
+
+        self.assertEqual((0, 10, 1, 1), mbslice.indices(10))
+        np.testing.assert_array_equal(self.arr, self.dset[mbslice])
```
takluyver, line 333:
```suggestion
        self.assertEqual(mbslice.indices(10), (0, 10, 1, 1))
        np.testing.assert_array_equal(self.dset[mbslice], self.arr)
```

It's cosmetic, but I've glanced at a few of our existing tests, and it looks like we generally stick to `(result, expected)` rather than `(expected, result)`, so I'd like to keep it consistent.

(And the same through the additions in this file)

takluyver:
Oh, and I meant to say: if you won't get to this soon, let me know and I can jump in to make the changes I'm suggesting.

h5py/_hl/selections.py
GDYendell: I think I just defined it in the same order as `handle_slice`.

h5py/_hl/selections.py
GDYendell: Yeah, this check is for when count is user defined, but it is invalid for the length of the dataset the selection is applied to (this case is hit in `test_block_overruns_extent_error`). The block could be indented into the else of the previous block, but I didn't think there was any reason to. Plus, I guess it is possible that in future the count calculation is updated and broken and it would still be caught in that case.

takluyver: Thanks for confirming. Let's leave it as is.

h5py/_hl/selections.py
takluyver: `handle_slice` isn't public, so there's no particular need to match it. I'd match the order in the HDF5 function unless we can come up with a solid reason to put them in another order.

h5py/_hl/selections.py
GDYendell: I didn't use `__repr__` because I didn't know if it was good practice to define a __repr__ that takes an argument. I wanted it to take an argument in indices because there I have the calculated count to display, rather than `None`, and I thought that would be useful. What are your thoughts?

takluyver: Well, I added a parameter to `Dataset.__getitem__`, so I'm not in a position to complain about that. :wink: But probably a nicer way to do it is to leave this method as is, and define `__repr__` as a wrapper around it.

h5py/_hl/selections.py
GDYendell: Good point. I will update the `__init__`, but leave it as is elsewhere.

GDYendell: Oh, and in `__repr__`

GDYendell:
I have updated and also rebased on master. I think it covers all of your comments.

takluyver:
I think you have, thank you - and thank you for sticking with this for the last ~18 months!

GDYendell:
Awesome. Thanks for your help with it. I hope it is useful for others in the community.

GDYendell:
@takluyver I have just noticed a mistake. I did not update the example for the re-ordering of stride and count. Also, in the tests and docs keyword args are used, so it is valid, but the wrong ordering might be confusing.

I have fixed the former with a new commit on this branch (https://github.com/dls-controls/h5py/tree/advanced-slicing). Maybe you just want to cherry pick it onto master, but let me know if your want a new PR for it.

takluyver:
Thanks, I've cherry picked that into #1526.

I think I might have spotted another issue as well while I was fiddling with this - I'm not sure that `MultiBlockSlice` will work with broadcasting (e.g. writing a 1D array to a 2D selection made with MultiBlockSlice). It may not show up when broadcasting from a scalar, because there's some special casing.

I probably won't have time to dig into this today, so if you've got any cycles to have a look at it, that would be good. One possibility is that we just disallow broadcasting to a selection made with MultiBlockSlice, as we already do for selections using fancy indexing.

PR #1123, opened 2018-10-26
Auto-detect HDF5 version on Windows

aparamon:


tacaswell:
Happy to merge this after the RC but before the final 2.9 if it is confirmed to work by someone else with a windows machine

@cgohlke will this make things better for you?

cgohlke:
AFAICT, this should not affect my builds since I set `HDF5_VERSION` and `HDF5_DIR` environment variables.

takluyver:
So, is this going in for 2.9?

If Christoph's builds and the builds on PyPI don't need it, I'd be inclined to leave it for a future release, so we're not making changes to the build script right before a release. But I'm not using h5py on Windows, so I don't have any particular stake in the decision.

aparamon:
This is intended as minor but easy-to-achieve improvement, not in any way critical for any realistic application. Please feel free to postpone merge.

Issue #1124, opened 2018-10-28
virtual datasets: support for sparse index based mappings


dnelson86:
Hello,

I've been playing with #1060 @takluyver (HDF5 1.10.4, h5py 11da4d6), it's really awesome! I've been waiting a long time for this for our use case, and it is really coming together nicely.

One aspect of my use case is that I would like to link in virtual datasets which are sparse (i.e. have values for only a select subset of entries of a dataset). In this case, the indices corresponding to each value are also saved. So I'd like to change the vds_simple example into (dummy example):

```python
    # Create source files (1.h5 to 4.h5)
    for n in range(1, 5):
        with h5py.File('{}.h5'.format(n), 'w') as f:
            d = f.create_dataset('data', (100,), 'i4')
            d[:] = np.arange(100) + n

    # Assemble virtual dataset
    layout = h5py.VirtualLayout(shape=(4, 100), dtype='i4')

    for n in range(1, 5):
        filename = "{}.h5".format(n)
        vsource = h5py.VirtualSource(filename, 'data', shape=(100,))
        inds = np.arange(100)[::-1]
        layout[n - 1,inds] = vsource

    # Add virtual dataset to output file
    with h5py.File("VDS.h5", 'w', libver='latest') as f:
        f.create_virtual_dataset('data', layout, fillvalue=-5)
        print("Virtual dataset:")
        print(f['data'][:, :10])
```

Is this within the VDS spec, and not yet implemented in h5py, or is this request outside the spec? Currently I get:

```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
/usr/local/lib/python3.6/site-packages/h5py/_hl/selections.py in _handle_simple(shape, args)
    450             try:
--> 451                 x,y,z = _translate_int(int(arg), length)
    452                 s = True

TypeError: only size-1 arrays can be converted to Python scalars

During handling of the above exception, another exception occurred:

TypeError                                 Traceback (most recent call last)
<ipython-input-19-5e8d1051e801> in <module>
----> 1 util.dataConvert.vdsTest()

/u/dnelson/python/util/dataConvert.py in vdsTest()
   1951         vsource = h5py.VirtualSource(filename, 'data', shape=(100,))
   1952         inds = np.arange(100)[::-1]
-> 1953         layout[n - 1,inds] = vsource
   1954
   1955     # Add virtual dataset to output file

[...]

PR #1126, opened 2018-10-30
Change VDS use of SimpleSelection to generic select().

dnelson86:
Addresses #1124.

Tests could use work / additions.

### DIFF
--- a/h5py/_hl/vds.py
+++ b/h5py/_hl/vds.py
@@ -13,7 +13,7 @@
 
 from copy import deepcopy as copy
 from collections import namedtuple
-from .selections import SimpleSelection
+from .selections import SimpleSelection, select
 from .. import h5s, h5
 from .. import version
 
@@ -95,7 +95,7 @@ class VirtualSource(object):
 
     def __getitem__(self, key):
         tmp = copy(self)
-        tmp.sel = SimpleSelection(self.shape)[key]
+        tmp.sel = select(self.shape, key, dsid=None)
         return tmp
 
 class VirtualLayout(object):
@@ -123,7 +123,7 @@ class VirtualLayout(object):
         self.sources = []
 
     def __setitem__(self, key, source):
-        sel = SimpleSelection(self.shape)[key]
+        sel = select(self.shape, key, source.sel.id)
         self.sources.append(VDSmap(sel.id,
                                    source.path,
                                    source.name,
--- a/h5py/_hl/vds.py
+++ b/h5py/_hl/vds.py
@@ -123,7 +123,7 @@ class VirtualLayout(object):
         self.sources = []
 
     def __setitem__(self, key, source):
-        sel = select(self.shape, key, source.sel.id)
+        sel = select(self.shape, key, dsid=None)
         self.sources.append(VDSmap(sel.id,
                                    source.path,
                                    source.name,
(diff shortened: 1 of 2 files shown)


tacaswell:
Would like @takluyver to review / merge this.

h5py/_hl/vds.py
```diff
@@ -123,7 +123,7 @@ def __init__(self, shape, dtype=None, maxshape=None):
         self.sources = []
 
     def __setitem__(self, key, source):
-        sel = SimpleSelection(self.shape)[key]
+        sel = select(self.shape, key, source.sel.id)
```
takluyver: I'm pretty sure passing `source.sel.id` here is wrong. It's only used if we're selecting with a `RegionReference` object, which I haven't used. But I think setting it to `None`, as you have for the source selection, is probably appropriate. That will just mean that we can't use a `RegionReference` for now.

takluyver:
Ping @dnelson86 - do you think you'll have time to work on this again soon?

dnelson86:
I'm not really familiar with these other functionalities of `select()`, so can't really verify expected behavior there.

h5py/tests/hl/test_vds/test_highlevel_vds.py
```diff
@@ -267,5 +267,59 @@ def test_slice_source(self):
...
+        vsource = h5.VirtualSource(filename, 'data', shape=(10,))
+        layout[inds] = vsource
+
+        outfile = osp.join(self.tmpdir, 'VDS.h5')
+
+        # Assembly virtual dataset (indexing source)
+        layout2 = h5.VirtualLayout((5,), 'i4')
+
+        inds2 = [0,1,4,5,8]
+        layout2 = vsource[inds2]
```
takluyver: This doesn't work, because you're replacing `layout2` instead of assigning into it. You need something like this:

```python
layout2[:] = vsource[inds2]
```

h5py/tests/hl/test_vds/test_highlevel_vds.py
```diff
@@ -267,5 +267,59 @@ def test_slice_source(self):
...
+
+        # Add virtual datasets to output file and close
+        with h5.File(outfile, 'w', libver='latest') as f:
+            f.create_virtual_dataset('/data', layout, fillvalue=-5)
+            f.create_virutal_dataset('/data2', layout2, fillvalue=-3)
```
takluyver: virutal -> virtual

h5py/tests/hl/test_vds/test_highlevel_vds.py
```diff
@@ -267,5 +267,59 @@ def test_slice_source(self):
...
+        assert_array_equal(data[inds], np.arange(10)*10)
+        assert_array_equal(data2, [0,10,40,50,80])
+
+        mask = np.zeros(100)
+        mask[inds] = 1
+        self.assertEqual(data[mask == 0].min(), -5)
+        self.assertEqual(data[mask == 0].max(), -5)
+
+        self.assertEqual(data2[(2,3,6,7,9)].min(), -3)
+        self.assertEqual(data2[(2,3,6,7,9)].max(), -3)
```
takluyver: I don't think these checks make sense. You've filled the whole virtual dataset for `data2` with selected values from the source, so the fill value shouldn't appear at all.

dnelson86: Yes apologies, I don't actually have a h5py development environment set up, so I am trying to write tests without actually running them.

takluyver:
I'm happy with this now. I'll give @tacaswell and @aragilar a chance to have an extra look, otherwise I'll merge it in a day or two.

PR #1127, opened 2018-11-04
Allow h5t.TypeID as dtype arg to create_dataset()

pezcore:
- This allows a user who needs low level control of a dataset's datatype to
  create the dataset with the high level `create_dataset()` interface but
  passing an instance of a low-level h5t.TypeID as the dtype argument instead
  of a numpy.dtype

### DIFF
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -37,6 +37,14 @@ optionally the data type (defaults to ``'f'``)::
     >>> dset = f.create_dataset("default", (100,))
     >>> dset = f.create_dataset("ints", (100,), dtype='i8')
 
+If you need a complex datatype, or if you perfer low-level HDF5 API, you can
+pass an instance of `h5py.h5t.TypeID` directly to :meth:`Group.create_dataset`
+as the `dtype`:
+
+    >>> my_h5t_id = h5t.create(h5t.COMPOUND, 100)
+    >>> my_h5t_id = h5t.insert(b'my_field', 40, h5t.NATIVE_DOUBLE)
+    >>> dset = f.create_dataset("default", (100,), my_h5t_id)
+
 You may initialize the dataset to an existing NumPy array::
 
     >>> arr = np.arange(100)
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -103,7 +103,10 @@ def make_new_dset(parent, shape=None, dtype=None, data=None,
                  "{} is not compatible with {}".format(chunks, shape)
         raise ValueError(errmsg)
 
-    if isinstance(dtype, Datatype):
+    if isinstance(dtype, h5t.TypeID):
+        tid = dtype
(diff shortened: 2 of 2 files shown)


aparamon:
I believe it would be better to support [h5py.Datatype](https://github.com/h5py/h5py/blob/67551907f5542cb70a4aa1f5a0103d5ff4a6a693/h5py/_hl/datatype.py#L21) than low-level `TypeID`.

pezcore:
Makes sense, and I agree, but why not both? The `make_new_dset()` implementation is already full of type checking and branches, and it's natural and intuitive that you should be able to specify the dtype directly for dataset creation. Allowing dtype specification in the form of a `h5t.TypeID` instance, seems to me to be beneficial for several reasons:

1. It allows direct usage of the much more prevalent C API (as implemented by `h5t.TypeID`) from `libhdf5`. Most HDF5 application developers will already be very familiar with this and can immediately start developing with `h5py` without needing to learn anything.
2. `h5t.TypeID` already trivially exposes what is needed by `make_new_dset()` w.r.t datatypes, hence this simple PR.
3. IIRC, the reason I wrote this patch in the first place is because I actually *needed* it, i.e. my application required fine grained control over the datatype that straight up wasn't possible to create using the high-level interface offered by `h5py`. When such control is necessary, it is important for `h5py` to support it.

The only reason I can think of **not** to allow this change is if you didn't want to eat the cost of checking if the `dtype` argument is an instance of `h5t.TypeID` every time, when you rarely expect it to be. But the cost of doing that is tremendously small compared to the rest of the `make_new_dset()` function when the low level objects have to be constructed. 

Perhaps I am missing something?

aparamon:
@pezcore Does the following work for you?
````py
f.create_dataset("default", (100,), h5py.Datatype(my_h5t_id))
````

aparamon:
@tacaswell @takluyver Proposing to reject: the desired behavior is easily achievable with existing `h5py.Datatype(type_id)` call.

takluyver:
I generally lean towards having as few ways as practical to use an API, as I think that makes it easier to document and to understand that API, and easier to change it in the future. So if wrapping a `TypeID` into a `Datatype` already does what's needed, I'm -0.5 on allowing passing `TypeID` directly.

I don't buy the argument that it's important to allow direct conversion of C code to Python, because the `File` class is already part of the high-level Pythonic API h5py adds. Code converted directly from C would use the [low level API](http://api.h5py.org/), which sticks much closer to libhdf5.

I don't feel strongly about this, though; if @tacaswell or @aragilar want it, I'm not going to argue.

pezcore:
I agree with the above mentioned points. I thought it wasn't possible to make this work with the compound data type I was attempting with originally, but I'll assume I missed something for now and close this for now. If I can formulate an actual reproducable use case that absolutely requires this, I'll bring it back up then.

takluyver:
Thanks :-)

PR #1132, opened 2018-11-12
Breakout special dtypes

### LINKED ISSUE #1118: API proposal: break up special_dtype into three functions
The [special_dtype](http://docs.h5py.org/en/2.8.0/special.html) function takes one of three keyword arguments vlen/enum/ref. It's only legal to pass one, and the meaning of the parameter value and the implementation of the function entirely depend on which keyword is used. It's effectively three different functions combined under one name.

The name 'special' also isn't very meaningful. I can see from the docs that 'special' means types which can't be directly represented in numpy, but this definition is only helpful if you know numpy very well. A string doesn't feel 'special' to me, for instance.

So I think it would be clearer to split this up into three functions: `vlen_dtype(element_dtype)`, `enum_dtype(base_type, enum_vals)` and `ref_dtype(ref_class)`. As there are only two legal types for the last one, we could further split that up into constants `ref_dtype` and `region_ref_dtype`.

Of course, the `special_dtype` function would remain, so existing code using that would continue to work. But the documentation would point people to the separate functions.

takluyver:
Closes #1118.

This adds three new pairs of functions for handling special dtypes:

* `string_dtype` and `check_string_dtype`
* `vlen_dtype` and `check_vlen_dtype`
* `enum_dtype` and `check_enum_dtype`

And two constants: `ref_dtype` and `regionref_dtype`, along with the function `check_ref_dtype`.

@aragilar I haven't yet changed the representation of string dtypes, as I think you were suggesting in #1118. I think it may make sense, especially if the changes requested in #379 go ahead, but there's more risk of breaking working code with that, so maybe it would be better to leave it for 3.0.

### DIFF
--- a/h5py/h5t.pyx
+++ b/h5py/h5t.pyx
@@ -748,15 +748,19 @@ cdef class TypeStringID(TypeID):
 
     cdef object py_dtype(self):
         # Numpy translation function for string types
-        if self.is_variable_str():
-            if self.get_cset() == H5T_CSET_ASCII:
-                return string_dtype(encoding='ascii')
-            elif self.get_cset() == H5T_CSET_UTF8:
-                return string_dtype(encoding='utf-8')
-            else:
-                raise TypeError("Unknown string encoding (value %d)" % self.get_cset())
+        if self.get_cset() == H5T_CSET_ASCII:
+            encoding = 'ascii'
+        elif self.get_cset() == H5T_CSET_UTF8:
+            encoding = 'utf-8'
+        else:
+            raise TypeError("Unknown string encoding (value %d)" % self.get_cset())
 
-        return dtype("|S" + str(self.get_size()))
+        if self.is_variable_str():
+            length = None
+        else:
+            length = self.get_size()
+
+        return string_dtype(encoding=encoding, length=length)
 
 cdef class TypeVlenID(TypeID):
 
@@ -1505,6 +1509,8 @@ cdef TypeStringID _c_string(dtype dt):
     tid = H5Tcopy(H5T_C_S1)
     H5Tset_size(tid, dt.itemsize)
     H5Tset_strpad(tid, H5T_STR_NULLPAD)
+    if dt.metadata.get('h5py_encoding') == 'utf-8':
+        H5Tset_cset(tid, H5T_CSET_UTF8)
     return TypeStringID(tid)
 
 cdef TypeCompoundID _c_complex(dtype dt):
@@ -1719,20 +1725,33 @@ def vlen_dtype(basetype):
     """
     return dtype('O', metadata={'vlen': basetype})
 
-def string_dtype(encoding='utf-8'):
-    """Make a numpy dtype for HDF5 variable-length strings
+def string_dtype(encoding='utf-8', length=None):
+    """Make a numpy dtype for HDF5 strings
 
-    encoding may be 'utf-8' or 'ascii'. If it is 'utf-8', the data should be
-    passed as Python str objects (unicode in Python 2). For 'ascii', data should
-    be passed as bytes.
+    encoding may be 'utf-8' or 'ascii'.
+
+    length may be an integer for a fixed length string dtype, or None for
+    variable length strings. String lengths for HDF5 are counted in bytes,
+    not unicode code points.
+
+    For variable length strings, the data should be passed as Python str objects
+    (unicode in Python 2) if the encoding is 'utf-8', and bytes if it is 'ascii'.
+    For fixed length strings, the data should be numpy fixed length *bytes*
+    arrays, regardless of the encoding. Fixed length unicode data is not
+    supported.
     """
-    if encoding == 'utf-8':
-        return dtype('O', metadata={'vlen': unicode})
-    elif encoding == 'ascii':
-        return dtype('O', metadata={'vlen': bytes})
-    else:
+    if encoding not in {'ascii', 'utf-8'}:
         raise ValueError("Invalid encoding (%r); 'utf-8' or 'ascii' allowed"
                          % encoding)
+
+    if isinstance(length, int):
+        # Fixed length string
+        return dtype("|S" + str(length), metadata={'h5py_encoding': encoding})
+    elif length is None:
+        vlen = unicode if (encoding == 'utf-8') else bytes
+        return dtype('O', metadata={'vlen': unicode})
+    else:
+        raise TypeError("length must be integer or None (got %r)" % length)
 
 def enum_dtype(values_dict, basetype=np.uint8):
     """Create a NumPy representation of an HDF5 enumerated type
--- a/h5py/h5t.pyx
+++ b/h5py/h5t.pyx
@@ -1741,7 +1741,7 @@ def enum_dtype(values_dict, basetype=np.uint8):
     appropriate integer base dtype large enough to hold the possible options.
     """
     dt = dtype(basetype)
-    if dt.kind not in "iu":
+    if not np.issubdtype(dt, np.integer):
         raise TypeError("Only integer types can be used as enums")
 
     return dtype(dt, metadata={'enum': values_dict})
--- a/h5py/h5t.pyx
+++ b/h5py/h5t.pyx
@@ -1798,6 +1798,44 @@ def special_dtype(**kwds):
     raise TypeError('Unknown special type "%s"' % name)
 
 
+def check_vlen_dtype(dt):
+    """If the dtype represents an HDF5 vlen, returns the Python base class.
+
+    Returns None if the dtype does not represent an HDF5 vlen.
+    """
+    return dt.metadata.get('vlen', None)
+
+def check_string_dtype(dt):
+    """If the dtype represents an HDF5 vlen string, returns the encoding.
+
+    Encodings can only be 'utf-8' or 'ascii'.
+
+    Returns None if the dtype does not represent an HDF5 vlen string.
+    """
+    vlen_kind = check_vlen_dtype(dt)
+    if vlen_kind is unicode:
+        return 'utf-8'
+    elif vlen_kind is bytes:
+        return 'ascii'
+    else:
+        return None
+
+def check_enum_dtype(dt):
+    """If the dtype represents an HDF5 enumerated type, returns the dictionary
+    mapping string names to integer values.
+
+    Returns None if the dtype does not represent an HDF5 enumerated type.
+    """
+    return dt.metadata.get('enum', None)
+
+def check_ref_dtype(dt):
+    """If the dtype represents an HDF5 reference type, returns the reference
+    class (either Reference or RegionReference).
+
+    Returns None if the dtype does not represent an HDF5 reference type.
+    """
+    return dt.metadata.get('ref', None)
+
 @with_phil
 def check_dtype(**kwds):
     """ Check a dtype for h5py special type "hint" information.  Only one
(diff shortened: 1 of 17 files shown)


takluyver:
Optimistically marking for 2.9 because I can and the tests are passing, but feel free to bump this to a later release if you like.

aparamon:
This looks nice and useful, esp. `ref_dtype` and `regionref_dtype`!
I'm curious about the naming for `check_*_dtype`: these do actually inspect, not just check the `dtype` objects (and do so in different ways). Recognizing the historical context, is it intentional to keep these uniformly named given the different, non-compatible return values?

takluyver:
I think it's a neat symmetry to have the `x_dtype` functions (or constants) to create dtypes, and `check_x_dtype` functions to see if you have such a dtype.

aragilar:
With the `string_dtype`/`check_string_dtype` I was thinking of using that for fixed length strings (#719, #988 and similar), rather than the variable length strings. So to create a variable length utf-8 string (matching that of `special_dtype(vlen=unicode)`), I guess `h5py.vlen_dtype('utf-8')` would be the recommended usage, with `h5py.vlen_dtype(h5py.string_dtype(encoding='utf-8'))`also working once we add the fixed utf-8 string support (which can be put off to a later PR).

shoyer:
I like `string_dtype('utf-8', length=None)`. `length=None` feels like a pretty reasonable sentinel value for "no fixed length".

takluyver:
I've bumped this to 2.10, because trying to land redesigned APIs just before a release sounds like a bad idea.

wavexx:
I work with questionnaire based-datasets, and the entire tooling revolves around utf-8 for obvious reasons. vlen/fixed strings are used appropriately due to space constraints. I actually gave up with h5py due to this limitation more than 6 months ago and ended up having to write my own thin C wrapper around libhdf5.

I've been revisiting h5py lately for other reasons, and I stumbled *again* onto the same issue.
I feel the handling of variable/fixed utf-8 strings should be fixed soon, breaking the API if necessary.

wavexx:
Fixed-length UTF-8 is missing, both for reading and for writing.
See #973 and #988

vasole:
Thanks. I do not what you will decide, but the proposal above seems excellent:

# Fixed length
string_dtype('utf-8',  length=10)

# Variable length:
string_dtype('utf-8', length=None)

wavexx:
The above proposal also looks good to me.

h5py/h5t.pyx
```diff
@@ -1712,6 +1712,42 @@ cpdef TypeID py_create(object dtype_in, bint logical=0, bint aligned=0):
...
+    *values_dict* maps string names to integer values. *basetype* is an
+    appropriate integer base dtype large enough to hold the possible options.
+    """
+    dt = dtype(basetype)
+    if not np.issubdtype(dt, np.integer):
+        raise TypeError("Only integer types can be used as enums")
+
+    return dtype(dt, metadata={'enum': values_dict})
+
+ref_dtype = dtype('O', metadata={'ref': Reference})
```
tacaswell: There is now one instance each of these two dtypes that are re-used, where as previously we would call `dytpe` each time.  Do we care and are `dtype` instance mutable?

shoyer: I'm pretty sure `dtype` instances are indeed mutable, for better or worse.

tacaswell:
Do we want to deprecate  `special_dtype` in the code or just leave it?

h5py/h5t.pyx
tacaswell: That seems like a bit of a foot-cannon to hand to users....

aparamon:
> **Fixed length**
> string_dtype('utf-8', length=10)
> **Variable length**
> string_dtype('utf-8', length=None)
Looks nice to me as well, with a potential development of None becoming the default *maybe* (because varlen strings are arguably more common).

h5py/h5t.pyx
takluyver: Do you want these to be functions returning new dtype objects on each call?

takluyver:
I've added the `length=` parameter to `string_dtype`, as suggested.

However, that now breaks the symmetry between `string_dtype` and `check_string_dtype` - the latter only checks for a variable-length string dtype. Do we want to:

- Make `check_string_dtype` return `(encoding, length)` for fixed- and variable-length strings, restoring the symmetry, but potentially making it less convenient to check for a variable length string (not sure how common this is?).
- Rename it to `check_vlen_string_dtype`, making the broken symmetry more clear.
- Drop `check_string_dtype` entirely - it's introduced by this pull request, so there are no compatibility requirements, and it's not used in h5py at present. I wrote it to fit the pattern of having a pair of functions to construct and check some kind of dtype.

shoyer:
We definitely need *some* way to get full string dtype information (encoding and size) out from arbitrary h5py datasets.

> * Make `check_string_dtype` return `(encoding, length)` for fixed- and variable-length strings, restoring the symmetry, but potentially making it less convenient to check for a variable length string (not sure how common this is?).

This sounds pretty reasonable to me. My guess is that most of the time, users who are using h5py aren't bothering to check the the returned dtype at all -- they just rely on reading the data into a NumPy array. But if they *do* care, then they likely care about both details. At least it's not obvious to me why `encoding` is more important than `length`.

If you wanted to make it slightly more readable/convenient, you could have `check_string_dtype` return a namedtuple with fields `encoding` and `length`.

The other option is to split this into two functions, e.g., `check_string_dtype_encoding()` (defined on both fixed size and vlen string dtypes) and `check_dtype_size()` (the later could be defined for other types, too). But I think I like the single function returning a namedtuple or `None` a bit more.

takluyver:
The dtype of the dataset for a fixed-length string will be a normal numpy fixed-length bytes dtype (e.g. `dtype('|S10')`, so there's some way to access this information even without h5py providing a function - you can check `dataset.dtype.itemsize` and `dataset.dtype.metadata['h5py_encoding']`.

That leads to another question: do we want to publicly document the dtype metadata h5py uses (in particular, string encoding) as part of its API, or keep that as an implementation detail? And is `'h5py_encoding'` a good name? The other fields h5py is using (`vlen`, `enum`, `ref`) don't have the `h5py` prefix, but I wondered if using such generic terms could lead to clashes with other software?

Also, how strict do we want to be about encoding names? So far I've only allowed the exact strings `{'utf-8', 'ascii'}`, but Python's own codec lookup (when you do `'hello'.encode(x)`) is case insensitive, normalises hyphens and underscores, and has aliases. We could normalise names through Python before checking them.

shoyer:
> The dtype of the dataset for a fixed-length string will be a normal numpy fixed-length bytes dtype (e.g. `dtype('|S10')`, so there's some way to access this information even without h5py providing a function - you can check `dataset.dtype.itemsize` and `dataset.dtype.metadata['h5py_encoding']`.

True, but fixed-length UTF-8 (if/when we add it) won't have a NumPy equivalent.

> That leads to another question: do we want to publicly document the dtype metadata h5py uses (in particular, string encoding) as part of its API, or keep that as an implementation detail?

I would slightly rather keep this an implementation detail. Long term, I'm optimistic that the NumPy dtype refactor (which is finally being started on) will let us write real NumPy dtypes that correspond to each type of h5py string.

> Also, how strict do we want to be about encoding names? So far I've only allowed the exact strings `{'utf-8', 'ascii'}`, but Python's own codec lookup (when you do `'hello'.encode(x)`) is case insensitive, normalises hyphens and underscores, and has aliases. We could normalise names through Python before checking them.

Normalization seems easy enough to do, and definitely a user-friendly choice.

h5py/h5t.pyx
tacaswell: I am not sure I have a good enough grasp of the trade-offs here. 

A function makes it more like the others and gets out of mutablity issue.

On the other hand, if dtypes are mutable then this also applies to the dtypes from numpy it's self which no one seems to worry about so we should not either and it is nice that these are constants.

shoyer: Maybe I was wrong here -- it seems that you cannot assign into the metadata dict after creating a dtype:
```
>>> d1 = np.dtype(object, metadata={'x': 1})
>>> d1.metadata['x'] = 2
TypeError: 'mappingproxy' object does not support item assignment
```

takluyver: And attempting to modify `dt.kind` and `dt.itemsize` gives me `AttributeError: readonly attribute
`. So maybe even if they're [not totally immutable](https://github.com/numpy/numpy/issues/1725), they're hard enough to modify that we needn't worry about the potential footgun.

aragilar:
I'd suggest having the metadata explicitly documented as internal to h5py, as I'm not sure whether `dtype.metadata` is explicitly documented anywhere (it's definitely not widely advertised as a feature). Also, it means if we do need to make backwards-incompatible changes to the metadata (as may need to happen if numpy is changing the structure of dtypes), we've explicitly called out not to rely on a specific internal format.

Is there a plan for what `check_vlen_dtype` will do for strings? I would suggest returning the same output as `check_string_dtype`, unless people have other suggestions?

takluyver:
I agree on documenting that the metadata is not a stable API.

I made `check_vlen_dtype(x)` so it should be equivalent to `check_dtype(vlen=x)`, i.e. returning `str` or `bytes` for a vlen string. It's a bit ugly, but it's easy to adapt existing code to it. If we wanted to make it more consistent, it could return a numpy dtype like `dtype('S1')` or `dtype('U1')`, representing the unit which can be repeated.

takluyver:
I believe this is ready for further review. I'll summarise what the PR now does, because its scope has grown a bit since it started:

1. `string_dtype()` provides a unified way to prepare an HDF5 compatible dtype for strings - fixed or variable length, with ASCII or UTF-8 encoding. This does not change what kind of Python objects can be stored, but it does give us the first high-level API to declare that fixed-length strings should be interpreted as UTF-8.
2. The functionality of `special_dtype()` is broken out into `vlen_dtype()`, `enum_dtype()`, `ref_dtype` and `regionref_dtype`, plus `string_dtype()` as already described. The ref/regionref ones are constants rather than functions, since they take no parameters; investigation in this PR suggested that dtype objects are immutable enough that this is probably workable, but it's easy to make them constructor functions if preferred.
3. The corresponding functionality of `check_dtype()` is split into `check_string_dtype()`, `check_vlen_dtype()`, `check_enum_dtype()` and `check_ref_dtype()`. These are all used in a similar way: they return the construction information if the dtype is of the kind that function checks, and None otherwise.

Naturally, the existing functions are still there. It's pretty painless to keep them working unless we have to change how the information is represented in dtypes, which I'm not proposing to do.

h5py/tests/old/test_dataset.py
```diff
@@ -746,15 +746,15 @@ class TestStrings(BaseDataset):
 
     def test_vlen_bytes(self):
```
aparamon: What about some `check_string_dtype()` (and `check_vlen_dtype()`) here and in the following tests?

aparamon:
Looks really good to me! (save the tiny nitpicks wrt tests)

h5py/tests/old/test_dataset.py
takluyver: I've just added some in `test_datatype` - does that cover what you meant?

aparamon: Not exactly (I thought about `check_*_dtype()` calls after `create_dataset()`, in these tests), but now it's definitely good enough to be merged.

shoyer:
Looks good to me, too.

h5py/tests/old/test_dataset.py
takluyver: Ah, you want to check that they roundtrip correctly through HDF5. That makes sense. I've added some tests.

aragilar:
Looks good to me too! I told appveyor to re-run the failing tests to see if the failures were down to appveyor playing up, if they pass, does anyone have a problem with merging this?

takluyver:
Thanks everyone for the reviews :-)

Issue #1051, opened 2018-06-06
h5py==2.8.0 floats cast to python.float or numpy.float64 non-deterministically (windows)


davidnewman02:
When reading a file using h5py==2.80, float values read from the file as a numpy.float64 do not cast correctly back to a python.float
i.e.
```
read_float = h5py.File(filename,'r')['test_group'].attrs['test_float']
cast_float = numpy.asscalar(read_float)
type(cast_float) # This returns either a python.float or a numpy.float64 at random
```
This works fine in h5py==2.7.1; although it throws a FutureWarning which seems tangentially related 
`FutureWarning: Conversion of the second argument of issubdtype from 'float' to 'np.floating' is deprecated. In future, it will be treated as 'np.float64 == np.dtype(float).type'`

To assist reproducing bugs, please include the following:
 * Operating System Windows10
*  h5py    2.8.0
*  HDF5    1.10.2
*  Python  3.5.2 (v3.5.2:4def2a2901a5, Jun 25 2016, 22:18:55) [MSC v.1900 64 bit (AMD64)]
*  sys.platform    win32
*  sys.maxsize     9223372036854775807
*  numpy   1.14.3

**To reproduce:**
1 - Create a minimal fast5 file
```
# create_simple_fast5.py
import h5py
filename = r"C:\test.fast5"

with h5py.File(filename,'w') as fh:
    fh.create_group('test_group')
    fh['test_group'].attrs['test_float'] = 3.7
```
2 - Read it and inspect value
``` 
# test_load_h5py.py
# Read and cast resulting numpy.float64 to python.float using np.asscalar
import h5py
import numpy
filename = r"C:\test.fast5"

with h5py.File(filename,'r') as fh:
    test_out = fh['test_group'].attrs['test_float']
    np_out = numpy.asscalar(test_out)
    print(type(test_out), type(np_out)) # The type of np_out is not deterministic
```
3- Run in a loop
```
# run_loop.py
import subprocess

for i in range(100):
    subprocess.call(["python", "test_load_h5py.py"])
```
The resulting datatype after the `numpy.asscalar` conversion is non-deterministic 
```
<class 'numpy.float64'> <class 'float'>
<class 'numpy.float64'> <class 'float'>
<class 'numpy.float64'> <class 'float'>
<class 'numpy.float64'> <class 'numpy.float64'>
<class 'numpy.float64'> <class 'float'>
<class 'numpy.float64'> <class 'float'>
[...]

PR #1134, opened 2018-12-01
Fix float type collection in `_get_available_ftypes`

jakirkham:
Fixes https://github.com/h5py/h5py/issues/1051

Makes use of NumPy's `typecodes` to get all of the floating typecodes and collect information about them into `available_ftypes` within `_get_available_ftypes`. This may make `_get_available_ftypes` ~50x faster (though this is probably unnoticeable) as it only checks known floating typecodes, types variables, and trusts NumPy's ordering. Simplifies the code a little bit as well. Most importantly this ensure the order is reliable across Python versions and OSes as the order is determined by NumPy in the `str` containing all of the floating typecodes; where, it has remained in ascending `itemsize` order at least for the last 8 years (i.e. NumPy 1.6.0). Though appears to have been there much longer and in order for that time.

ref: https://github.com/numpy/numpy/blob/v1.7.0/numpy/core/numerictypes.py#L907


jakirkham:
This is done on my end. Would be great if someone could take a look. 😄

tacaswell:
Nominally tagged for 2.9, will decide tomorrow if we should bump to 2.10 or not (and do the 2.9 tag!)

tacaswell:
Moved to 2.10 so it can get tests and stew a bit longer...

jakirkham:
Hmm...what is it we would be testing against here? Asking partially as it sounds like this function had no tests to begin with, correct?

aparamon:
@jakirkham Not a unit test for `_get_available_ftypes`, but something more high-level like reported
````py
read_float = h5py.File(filename,'r')['test_group'].attrs['test_float']
cast_float = numpy.asscalar(read_float)
type(cast_float) # This returns either a python.float or a numpy.float64 at random
````

aragilar:
The code which uses `_available_dtypes` assumes that it is ordered by "precision" (the number of bytes used to store is a good first approximation, but there are multiple formats which share the same size (x86 80 bit float on 64 bit system, double-double, ieee-quad), so byte size isn't the only factor), i.e. we want to find the smallest (in terms of byte size) numpy type which will hold the range stored in the HDF5 type (as it's both likely faster to use and requires less memory). A reasonable test could be to check that the byte size increases for each dtype in the tuple. Another one is to check that `maxexp` increases.

jakirkham:
Here's a [link]( https://github.com/numpy/numpy/blob/v1.15.4/numpy/core/numerictypes.py#L888 )  to the definition of `Float` in `typecodes`. This is ordered by precision and has a fixed set of types. When I went through all of the types this function used previously, the type code of each `dtype` mapped to one of the characters in that `str` (if it was actually a floating point type and not something else like an integral type). Also that `str` has not changed for a long time and is ordered by size. Admittedly this was a month ago and I could have forgotten things.

takluyver:
Where are we with this? Can we see a way to write a reasonable test for it? It seems like something that should go in for 2.10.

jakirkham:
Sorry. I’m not currently working on this. I’d be ok closing, merging, or handing it off to someone to complete. Thoughts?

takluyver:
My understanding from the thread above is that it's working, and the only hold up is a test. Is that right?

If so, I'd be OK either with someone else picking it up, or merging as is if we're confident enough that it won't cause regressions.

jakirkham:
That sounds like an accurate summary.

SGTM

takluyver:
I'll propose that we merge this without adding a specific test. If anyone objects, please speak up - I'll take silence as assent in a few days.

Rationale:

- It doesn't cause failures in any existing tests, including `test_float_round_tripping`, which checks that all numpy floating dtypes roundtrip correctly.
- Having read the various threads, I believe it fixes the issue as I understand it.
- It's simpler than the code it replaces.
- The problem only occurs intermittently on a platform that I don't use, so if I write a test, it's hard to be sure it's a useful one.

I'm all for thorough testing, but if we believe this code is an improvement, there's no particular point leaving it out for the lack of a test if no-one's prepared to write one.

If you disagree, please consider volunteering to write the test yourself! :wink: But even if you can't do that, make your objection known.

takluyver:
OK, there have been no objections, so I'm going to merge this.

jakirkham:
Thanks for seeing this one through, @takluyver. Hope it helps some people 🙂

PR #1135, opened 2018-12-04
DOC: start do write 2.9 release docs

tacaswell:


tacaswell:
attn @aragilar @takluyver @aparamon 

This is the last thing for the release.  I went through the git history and noted all the PRs that looked important to me, these need to be fleshed out a bit (because some of these new features are super awsome and need to be advertised!).

my notes from the last few releases (which needs to get moved into the docs and fleshed out a bit):

 - write release notes
 - bump version in
   - docs/conf.py
   - h5py/version.py
   - setup.py
 - tag (no v)
 - bump versions to include a post0
   - docs/conf.py
   - h5py/version.py
   - setup.py

Sorry I did not get farther on this today.

tacaswell:
Please feel free to push commits to this branch!

docs/whatsnew/2.9.rst
```diff
@@ -1,6 +1,44 @@
...
+
+* ``H5LTopen_file_image`` as :func:`h5py.h5f.open_file_image` (:issue:`1075`)
+* External dataset storage functions ``H5Pset_external``, ``H5Pget_external``
+  and ``H5Pget_external_count`` as methods on :class:`h5py.h5p.PropDCID`
+  (:issue:`1000`)
+
+Bugfixes
+--------
+
+* Reading float128 data on ppc64 architecture systems (:issue:`1114`)
```
takluyver: @tacaswell is this correct? You mentioned Windows in the TODO list, but I didn't see anything about that in the PR, and the changes looked like they would only affect ppc64.

docs/high/file.rst
```diff
@@ -212,43 +212,43 @@ directly from memory rather than reading the file again.  The details of a
...
-the hash table (which is determined by the ``rdcc_nslots`` parameter) is large
-enough to minimize the number of hash value collisions.  Due to the hashing
-strategy, this value should ideally be a prime number.  As a rule of thumb,
-this value should be at least 10 times the number of chunks that can fit in
-``rdcc_nbytes`` bytes. For maximum performance, this value should be set
-approximately 100 times that number of chunks. The default value is 521.
+The parameters controlling this behavior are prefixed by ``rdcc``, for *raw data
+chunk cache*.
+
+* ``rdcc_nbytes`` sets the total size (measured in bytes) of the raw data chunk
```
takluyver: I felt that a bulleted list makes it easier to see these parameters than complete prose.

docs/high/file.rst
```diff
@@ -212,43 +212,43 @@ directly from memory rather than reading the file again.  The details of a
...
+  which is called a hash table.  A chunk's hash value is simply the index into
+  the hash table of the pointer to that chunk.  While the pointer at this
+  location might instead point to a different chunk or to nothing at all, no
+  other locations in the hash table can contain a pointer to the chunk in
+  question.  Therefore, the library only has to check this one location in the
+  hash table to tell if a chunk is in cache or not.  This also means that if two
+  or more chunks share the same hash value, then only one of those chunks can be
+  in the cache at the same time.  When a chunk is brought into cache and another
+  chunk with the same hash value is already in cache, the second chunk must be
+  evicted first.  Therefore it is very important to make sure that the size of
```
takluyver: I've mostly left the wording alone for now, but I'd like to cut out some of the details about the implementation - I don't think it's necessary to understand the parameter, and it makes for a somewhat forbidding wall of text.

If other people would like to keep it, maybe a compromise would be to move the details to a separate paragraph/box, after the use of the parameter has been described.

docs/whatsnew/2.9.rst
tacaswell: The OP reports is is a problem on linux as well (and from my comments I could apparently reproduce it locally on x86_64 linux).  I think that what happened was that the fixes for ppc64 elsewhere ended up breaking float128 in other ways.  I think #1114 gets it to work correctly everywhere we need and adds round-trip tests for the platforms we test on.

Hopefully the RC will be built by the debian folks who have some more exotic platforms.

PR #1143, opened 2018-12-19
Deprecate special default default behaviour when opening file

takluyver:
From discussion in #397: introduce a global config option for the default mode when opening files, and a deprecation warning if you haven't explicitly specified a mode.

The idea is that this change would happen in a 2.x release, and then 3.0 would switch the global default to `'r'` and get rid of the special try readwrite, try readonly, try create sequence. At a later point, we could then get rid of the global config option; I think it's always clearer to pass the mode directly when you open a file.

This implementation doesn't allow any way to get *exactly* the old behaviour back without the warning. You can get close by setting mode `'a'` (try readwrite, try exclusive create). I don't think a fallback to read-only really makes sense: if you need to write data, it's best to fail fast, and if you don't need to write data, it's best to open the file read-only. If there's a corner case where the pattern is useful, someone can still write their own try/excepts to achieve it.

### DIFF
--- a/h5py/h5.pyx
+++ b/h5py/h5.pyx
@@ -60,6 +60,7 @@ cdef class H5PYConfig:
         self._t_name = b'TRUE'
         self._bytestrings = ByteStringContext()
         self._track_order = False
+        self.default_file_mode = None
 
     property complex_names:
         """ Settable 2-tuple controlling how complex numbers are saved.
--- a/h5py/h5.pyx
+++ b/h5py/h5.pyx
@@ -60,7 +60,7 @@ cdef class H5PYConfig:
         self._t_name = b'TRUE'
         self._bytestrings = ByteStringContext()
         self._track_order = False
-        self.default_file_mode = None
+        self._default_file_mode = None
 
     property complex_names:
         """ Settable 2-tuple controlling how complex numbers are saved.
@@ -148,6 +148,17 @@ cdef class H5PYConfig:
             return self._track_order
         def __set__(self, val):
             self._track_order = val
+
+    property default_file_mode:
+        """Default mode for h5py.File()"""
+        def __get__(self):
+            return self._default_file_mode
+
+        def __set__(self, val):
+            if val is None or val in {'r', 'r+', 'x', 'w-', 'w', 'a'}:
+                self._default_file_mode = val
+            else:
+                raise ValueError("Invalid mode; must be one of r, r+, w, w-, x, a or None")
 
 cdef H5PYConfig cfg = H5PYConfig()
 
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -6,6 +6,7 @@ include README.rst
 include setup_build.py
 include setup_configure.py
 include tox.ini
+include pytest.ini
 
 recursive-include docs *
 prune docs/_build
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -183,6 +183,11 @@ def make_fid(name, mode, userblock_size, fapl, fcpl=None, swmr=False):
         except IOError:
             fid = h5f.create(name, h5f.ACC_EXCL, fapl=fapl, fcpl=fcpl)
     elif mode is None:
+        warn("The default file mode will change to 'r' (read-only) in h5py 3.0. "
+             "To suppress this warning, pass the mode you need to h5py.File(), "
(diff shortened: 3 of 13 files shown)


aparamon:
@takluyver That's the step in the right direction!
From my experience, users do really hate the warnings. It's easy to suppress one now by setting `h5.get_config().default_file_mode`. However, it only will take effect in 2.10+.
Might it make sense to introduce additional intermediate step: use `h5.get_config().default_file_mode` but do not warn, in 2.10? Then, introduce the warning in 2.11.
Overall, like the change even as it is.

takluyver:
That might be necessary. @tacaswell, any preference?

takluyver:
cc @tacaswell - do you want to add a warning and the option needed to suppress it in the same release, or have the option in for a release before it starts warning?

Another possible approach could be an environment variable, like `H5PY_DEFAULT_READONLY`. Setting it to any non-empty value would enable the 'future' behaviour and suppress the warning. The advantage is that it won't cause an error on previous versions of h5py, but the downside is that they will silently ignore it instead, so you can set it without knowing whether it is respected.

tacaswell:
Sorry, just seeing this now (my December and January were very busy).

> From my experience, users do really hate the warnings.

And also ignore them!

I think adding an env is also useful (so people who have this behavior in code that they do not want to edit an suppress the warning).  The ENV is also useful for supporting more than one version as you can unconditionally set the env, old versions safely ignore it, new versions will respect it

Would have been better to get the option into 2.9, oh well :man_shrugging: .

Between this and #1177 we would have a release schedule something like:

2.9: Dec 2018
2.10 : April 2019 (adds warning + option to change default + suppress warning)
3.0 : Jan 2020 (changes default)

tacaswell:
@takluyver mind if I rebase this an current master (and force-push)?

takluyver:
Go for it, thanks for letting me know so I don't get an erroneous merge.

tacaswell:
@takluyver Done.

There is still one warning in the test suite that is subtle numpy behavior that I could not understand quickly...

takluyver:
OK. Shall I add the environment variable as well? Are you happy with the name I suggested: `H5PY_DEFAULT_READONLY`?

h5py/h5.pyx
```diff
@@ -60,6 +60,7 @@ cdef class H5PYConfig:
         self._t_name = b'TRUE'
         self._bytestrings = ByteStringContext()
         self._track_order = False
+        self.default_file_mode = None
```
tacaswell: Do we want to make this a property and validate that it is a valid mode string?

tacaswell:
Do we want to make the ENV a flag for "use future flag" or `H5PY_DEFAULT_FILE_MODE` to match what can be done via the config?

On one hand the "future flag" is simpler, on the other hand, the 'file mode' is more powerful and matches the functionality in the config.  I lean slightly towards the 'file mode' version, but think either way will be ok.

takluyver:
My rationale for making it a flag is that it's easier to remove it later that way. If people start setting `H5PY_DEFAULT_FILE_MODE=a` (for instance), and expecting that to work, then it's another breaking change to get rid of it.

I don't think it's good design to allow an ambient setting to choose the file mode  - I think it should be visible in the code where you open the file. So I see the environment variable as a temporary measure which we will drop again once the transition is made.

It's easier with the config option, because once we remove that, attempting to set it should be an error, rather than being silently ignored.

tacaswell:
@takluyver ok, I am convinced :+1: .

takluyver:
I've added the environment variable and made the config option a property with validation, as you suggested.

I've allowed setting `default_file_mode` back to None for now. I don't think this is a good idea to explicitly set, but I can imagine someone might write code like this:

```python
old_default_mode = h5py.h5.get_config().default_file_mode
h5py.h5.get_config().default_file_mode = 'r'
# ... call some code using h5py ...
h5py.h5.get_config().default_file_mode = old_default_mode
```

So the steps will be something like:

- 2.10 adds the warning, the config option and the environment variable
- 3.0 removes the `None` mode, makes `'r'` the default, and removes the environment variable
- 3.1 warns if the config option is set to anything besides `'r'`
- 3.2 errors if the config option is set to anything besides `'r'` (so it's a no-op, but convenient for writing compatible code)
- 3.3 removes the config option

We can shift the 3.1-3.3 points to later releases if those releases are happening quickly and we want to give people more time to shift.

PR #1144, opened 2018-12-20
MNT: add a .mailmap file

tacaswell:


tacaswell:
This is failing on a manifest check.  Do we want to ship the mailmap or relax that test?

aragilar:
Adding `exclude .mailmap` to MANIFEST.in should fix the test (I'm presuming there's not much point having the mailmap file in the sdist).

PR #1146, opened 2018-12-20
Set stacklevel for warnings

takluyver:
Without these, the code shown with the warning is the final line of the `warn()` call itself, which isn't very informative. This tells it to show the next level up the stack, which is the code that will likely need to be changed.


takluyver:
It seems warnings from Cython code don't need this, whether because Cython is doing something clever or because calls into compiled code work differently in Python.

PR #1155, opened 2019-01-02
Add new libver value for H5F_LIBVER_V18

ajelenak-thg:
HDF5 library 1.10.2 introduced two new constants for declaring file format version bounds: `H5F_LIBVER_V18` and `H5F_LIBVER_V110`. Since `H5F_LIBVER_LATEST` is set to `H5F_LIBVER_V110`, I decided to only expose `H5F_LIBVER_V18` as a new `libver` option value: `v18`.

This PR includes also updates to the tests and documentation.


tacaswell:
Thanks for this!

I lean a bit towards including a `'v110'` string as well so people can program a bit defensively so in the future if `latest' moves to 1.12, code that use to produce 1.10 compatible files will continue to do so with no change (presumably that is why hdfgroup included `H5F_LIBVER_V110` as a symbol?).

docs/high/file.rst
```diff
@@ -132,6 +132,15 @@ compatible.
 
 The default is "earliest".
 
+Starting with HDF5 version 1.10.2, it is possible to specify another version:
+
+    >>> f = h5py.File('name.hdf5', libver='v18')
+
+that will store objects compatible with HDF5 1.8. For example, to ensure full
```
takluyver: What happens if you do this and then try to use the new 1.10 features, e.g. creating a virtual dataset? Does the 'best effort' qualifier for `'earliest'` also apply to this,  so it will break 1.8 compatibility if you need a 1.10 feature? Or will it refuse to do anything that makes the file incompatible with 1.8? And if so, how clear is the error?

ajelenak-thg: It will do the latter and throw an error. I am not sure on the error message clarity. Will have to try.

Good overview is at https://www.hdfgroup.org/2018/04/why-should-i-care-about-the-hdf5-1-10-2-release/. More technical details are at https://www.hdfgroup.org/wp-content/uploads/2018/04/RFC-Setting-Bounds-for-Object-Creation-in-HDF5-1.10.0-Update.pdf.

ajelenak-thg:
Having `v110` as well will introduce two version bound labels with the same value (`v110` and `latest` currently) which then creates a problem when retrieving the version bounds tuple for the `File.libver` property. One could expect (`earliest`, `latest`) but get (`earliest`, `v110`). Both would be correct but perhaps confusing for users.

Maybe remembering the `libver` argument's value. If a string, report upper bound as `latest`; if a tuple, report upper bound as the tuple's second (`libver[1]`) element. Is this acceptable?

aparamon:
@ajelenak-thg Thanks for your contribution!
@tacaswell Having `v110` seems a good idea.
The problem of `v110` vs `latest` roots from the fact that in h5py those are strings while in HDF5 library these are integers. Note that `H5F_LIBVER_18 < H5F_LIBVER_LATEST` btw.
Might is be possible to re-implement `File.libver` with something like https://docs.python.org/3/library/enum.html? I'm not sure how keep current code relying on `File.libver` still working though.

ajelenak-thg:
I added `v110` as another libver bound label. Because `latest` is not any more a uniquely valued label, I also modified the code for  the `File.libver` property to report the same labels as used if creating that file. I'm interested in feedback about this approach.

Alternatively, `latest` should only be used when creating files and have `File.libver` report the actual bound labels, i.e. use `v110` instead of `latest`. This is simpler to implement than what I just committed.

takluyver:
I'd go for the simpler implementation, because the index arithmetic to get the label you want makes my head hurt. This would also better represent what HDF5 is really doing, because `H5F_LIBVER_LATEST` is now just an alias for `H5F_LIBVER_V110`.

ajelenak-thg:
Okay @takluyver, I agree. The latest changes (891d168) simplify the situation. `latest` can still be used when creating files but `File.libver` property will only use the actual bound labels, currently: `earliest`, `v18`, and `v110`.

```python
In [1]: import h5py

In [2]: f = h5py.File('foo.h5', 'w')

In [3]: f.libver
Out[3]: ('earliest', 'v110')

In [4]: f.close()

In [5]: f = h5py.File('foo.h5', 'w', libver='latest')

In [6]: f.libver
Out[6]: ('v110', 'v110')

In [7]: f.close()
```

Better?

ajelenak-thg:
> With the new additions, only one of the following important properties holds:
> 
> * [x]  `earliest` <= `v18`

@aparamon: `earliest` < `v18` (if I understand what you are trying to say)

aparamon:
@ajelenak-thg Indeed, thanks for the correction!

ajelenak-thg:
@aparamon Can you please review your request for changes? I made a commit after your request that changed how `File.libver` works.

aparamon:
@ajelenak-thg Hmm, I do not notice the commit. Did you forget to push?
In the code I see now
````py
'v18' >= 'v110'
'v110' > 'latest'
````

**Comment by ajelenak-thg** [comment]
> @ajelenak-thg Hmm, I do not notice the commit. Did you forget to push?

This commit: 891d168. And this comment explains it a bit: 
[#issuecomment-453780034](
https://github.com/h5py/h5py/pull/1155#issuecomment-453780034)

> In the code I see now
> 
> ```python
> 'v18' >= 'v110'
> 'v110' > 'latest'
> ```

Where's that?

**Comment by aparamon** [comment]
> ````py
> 'v18' >= 'v110'
> 'v110' > 'latest'
> ````
It's the property of Python strings.
HDF5 integer enums behave differently (and correctly).

**Comment by ajelenak-thg** [comment]
So you would like the new bound label names to support the same comparison as their HDF5 enum values?

**Comment by aparamon** [comment]
Definitely!

@ajelenak-thg Did you consider https://docs.python.org/3/library/enum.html? I think your patch introduces already a serious change by returning `'v110'` instead of `'latest'`; so returning something like `LibVer.V110` shouldn't introduce *much* additional compatibility headache.

**Comment by ajelenak-thg** [comment]
The quickest fix is to use `v108` instead of `v18`. Applying Python string comparison to these version bound labels never occurred to me so just went along with `v18`.

> Did you consider https://docs.python.org/3/library/enum.html?

It is available since Python 3.4. Having the same functionality in Python 2.7 would require installing https://pypi.org/project/enum34/. What other h5py maintainers think about this approach? I am fine if there's a consensus.

> I think your patch introduces already a serious change by returning 'v110' instead of 'latest' as previously

I don't see how this can be avoided. What is `latest` today may not be so tomorrow...

**Comment by aparamon** [comment]
> I don't see how this can be avoided. What is `latest` today may not be so tomorrow...

It's sub-optimal but tolerable.

File: h5py/_hl/files.py [hunk 2]
```diff
@@ -327,8 +332,8 @@ def __init__(self, name, mode=None, driver=None,
             Name of the driver to use.  Legal values are None (default,
             recommended), 'core', 'sec2', 'stdio', 'mpio'.
         libver
-            Library version bounds.  Currently only the strings 'earliest'
-            and 'latest' are defined.
+            Library version bounds.  Supported values: 'earliest', 'v18' (HDF5
```
takluyver: `v108`?

h5py/_hl/files.py
```diff
@@ -327,8 +332,8 @@ def __init__(self, name, mode=None, driver=None,
...
+            1.10.2 or later), 'v110' (HDF5 1.10.2 or later)  and 'latest'.
```
takluyver: It's a bit confusing that the version 1.8 marker is described as "1.10.2 or later". I know what you mean, but I bet people will try to 'fix' that.

Why don't we give the list, and then put a note afterwards like:

> The v18 and v110 options can only be specified on HDF5 1.10.2 or later.

h5py/_hl/files.py
ajelenak-thg: Yes, will fix that.

h5py/_hl/files.py
ajelenak-thg: Agree.

ajelenak-thg: This PR has been quiet for some time until now. I don't see any issue beyond these two minor documentation fixes that you pointed out.

takluyver:
Thanks. Can you resolve the conflicts once you've updated it as well?

@aparamon are you happy with this PR?

h5py/_hl/files.py
ajelenak-thg: Fixed in ff047339.

h5py/_hl/files.py
ajelenak-thg: Fixed in ff04733.

ajelenak-thg:
@takluyver 
> Can you resolve the conflicts once you've updated it as well?

Done.

takluyver:
Thanks. I'll leave some time for @aparamon to have another look at it.

takluyver:
Ping @aparamon - are you happy with this PR as it stands?

scopatz:
Lgtm

h5py/_hl/files.py
```diff
@@ -35,8 +35,13 @@
...
 
 
-libver_dict = {'earliest': h5f.LIBVER_EARLIEST, 'latest': h5f.LIBVER_LATEST}
+libver_dict = {'earliest': h5f.LIBVER_EARLIEST}
+if hdf5_version >= (1, 10, 2):
+    libver_dict.update({'v108': h5f.LIBVER_V18, 'v110': h5f.LIBVER_V110})
 libver_dict_r = dict((y, x) for x, y in six.iteritems(libver_dict))
+libver_dict.update({'latest': h5f.LIBVER_LATEST})
+if hdf5_version < (1, 10, 2):
+    libver_dict_r = dict((y, x) for x, y in six.iteritems(libver_dict))
```
takluyver: I've just spotted that this recreation should be unnecessary; just finish defining `libver_dict` before creating `libver_dict_r`, and then it only needs to be done once.

takluyver: Ah, no, I understand. If `LIBVER_LATEST` is an alias for `LIBVER_V110`, then we'd be relying on dict ordering, which is only consistent from Python 3.6. That's tricky. We might want a comment explaining why it's done like that.

h5py/h5p.pyx
```diff
@@ -1135,6 +1135,7 @@ cdef class PropFAID(PropInstanceID):
         Set the compatibility level for file format.  Legal values are:
 
         - h5f.LIBVER_EARLIEST
+        - h5f.LIBVER_V18 (HDF5 1.10.2 or later)
```
takluyver: Should these docstrings also mention `LIBVER_V110`?

h5py/_hl/files.py
takluyver: I've rewritten it in what I think is a clearer way, at the cost of a bit of repetition.

PR #1161, opened 2019-01-10
Replace deprecated assertion methods

takluyver:
While looking at [a CI test log](https://travis-ci.org/h5py/h5py/jobs/477786641), I noticed that there was a lot of output warning about deprecated assertion methods `assertEquals` and `assert_`. I think it's worth cleaning them up so it's easier to see other information in test logs.

- `assertEquals()` I find-and-replaced to `assertEqual()`, which is the standard name
- `assert_()` I changed by hand to various things that seemed like a good idea. In many cases I've just used a bare `assert` statement, relying on either the explicit message or on pytest's magic to give informative errors.

### DIFF
--- a/h5py/tests/common.py
+++ b/h5py/tests/common.py
@@ -105,39 +105,27 @@ class TestCase(ut.TestCase):
             message = ' (%s)' % message
 
         if np.isscalar(dset) or np.isscalar(arr):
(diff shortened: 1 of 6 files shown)


aragilar:
Cool, thanks for doing this @takluyver.

PR #1165, opened 2019-01-23
Allow list/tuple of unicode strings to be saved as datasets

woodychow:
This MR is very similar to https://github.com/h5py/h5py/pull/1032. I think list/tuple of unicode strings should be supported by dataset as well.

```py
import h5py
f = h5py.File('test.h5', 'w')
f['a'] = [u'X', u'Y']
print(repr(f['a'][...]))
```
prints
```
array([u'X', u'Y'], dtype=object)
```


aragilar:
This has be done via the new string logic covered in #1338 (noting that datasets always return bytes), hence I'm going to close this.

Issue #928, opened 2017-09-19
Provide a interface for registering filter plugins


aragilar:
Currently filters either need to be either built into HDF5, packaged as part of h5py, or the filter provider has to modify HDF5_PLUGIN_PATH, none of which are ideal, and can lead to hard-to-debug problems (see #923). There exists a function which appends to the internal path (rather than an environment variable), which h5py currently does not wrap. We should wrap this, provide fallbacks for older HDF5 versions, and provide a way of getting the current path (if possible).

See https://github.com/silx-kit/hdf5plugin/issues/14 for further discussion about what the interface should look like.

aldanor:
One way is through setuptools entrypoints. Kind of like pytest registers plugins for it, IIRC.

aldanor:
@aragilar Any thoughts/progress re: this?

vasole:
I am waiting for HDF5 1.10.2 to be out to see if one can properly deal with plugins.

In principle that version should allow us to check what plugins are available and to add new ones as needed thus solving the problems encountered in the hdf5plugin module (it has to be imported prior to import h5py, it cannot add plugins selectively...)

aparamon:
HDF5 lib now provides a set of procedures to customize plugin discovery:
https://support.hdfgroup.org/HDF5/doc/RM/RM_H5PL.html

aparamon:
@tacaswell @takluyver What do you think of providing a set of pre-built popular plugins as a separate pip-installable package, something like https://pypi.org/project/hdf5plugin?

With the new `H5PL*` procedures it seems that the restriction "import `hdf5plugin` **before** importing `h5py`" can be relieved.

Maybe even better would be to package each plugin as a separate pip package, so user has more fine-grained control over which plugins to install (note that they generally have different licenses). `h5py`'s LZF implementation could be extracted a one separate package too.

takluyver:
Are there particular plugins you have in mind? My use cases so far haven't
involved any plugins, except experimenting with the lzf filter.

vasole:
At this point in hdf5plugin we have put the plugins needed to read data from our detectors.
I have in mind to add the blosc family too.

Please consider that in order to be able to supply plugin filters usable everywhere we had to patch the direct calls of those plugins to the HDF5 library under linux and MacOS. If not, they are bound to a particular version of the HDF5 library. The price to pay is that one can only use those filters for reading under linux and MacOS. If one could ship them built with the same library as h5py then the limitation would disappear.

[...]

PR #1166, opened 2019-01-23
H5PL: Plugin Interface

aparamon:
Implementation of https://github.com/h5py/h5py/issues/928.

Please review. [MutableSequence](https://docs.python.org/3/library/collections.abc.html#collections.abc.MutableSequence) high-level interface can in principle be implemented, but do we really need it for such a niche functionality?

### DIFF
--- a/h5py/h5pl.pyx
+++ b/h5py/h5pl.pyx
@@ -42,10 +42,10 @@ IF HDF5_VERSION >= (1, 10, 1):
         cpdef size_t n
         cpdef char* buf = NULL
 
-        n = H5PLget(index, NULL, 0) + 1
-        buf = <char*>emalloc(sizeof(char)*n)
+        n = H5PLget(index, NULL, 0)
+        buf = <char*>emalloc(sizeof(char)*(n + 1))
         try:
-            H5PLget(index, buf, n)
+            H5PLget(index, buf, n + 1)
             return PyBytes_FromStringAndSize(buf, n)
         finally:
             efree(buf)
--- a/h5py/h5pl.pyx
+++ b/h5py/h5pl.pyx
@@ -11,45 +11,47 @@
     HDF5 plugin interface.
 """
 
+include "config.pxi"
 from utils cimport emalloc, efree
 
 # === C API ===================================================================
 
-cpdef append(const char* search_path):
-    """(STRING search_path)"""
-    H5PLappend(search_path)
+IF HDF5_VERSION >= (1, 10, 1):
+    cpdef append(const char* search_path):
+        """(STRING search_path)"""
+        H5PLappend(search_path)
 
-cpdef prepend(const char* search_path):
-    """(STRING search_path)"""
-    H5PLprepend(search_path)
+    cpdef prepend(const char* search_path):
+        """(STRING search_path)"""
+        H5PLprepend(search_path)
 
-cpdef replace(const char* search_path, unsigned int index):
-    """(STRING search_path, UINT index)"""
-    H5PLreplace(search_path, index)
+    cpdef replace(const char* search_path, unsigned int index):
+        """(STRING search_path, UINT index)"""
+        H5PLreplace(search_path, index)
 
-cpdef insert(const char* search_path, unsigned int index):
-    """(STRING search_path, UINT index)"""
-    H5PLinsert(search_path, index)
+    cpdef insert(const char* search_path, unsigned int index):
+        """(STRING search_path, UINT index)"""
+        H5PLinsert(search_path, index)
 
-cpdef remove(unsigned int index):
-    """(UINT index)"""
-    H5PLremove(index)
+    cpdef remove(unsigned int index):
+        """(UINT index)"""
+        H5PLremove(index)
 
-cpdef get(unsigned int index):
-    """(UINT index) => STRING"""
-    cpdef size_t n
-    cpdef char* buf = NULL
+    cpdef get(unsigned int index):
+        """(UINT index) => STRING"""
+        cpdef size_t n
+        cpdef char* buf = NULL
 
-    n = H5PLget(index, NULL, 0) + 1
-    buf = <char*>emalloc(sizeof(char)*n)
-    try:
-        H5PLget(index, buf, n)
-        return PyBytes_FromStringAndSize(buf, n)
-    finally:
-        efree(buf)
+        n = H5PLget(index, NULL, 0) + 1
+        buf = <char*>emalloc(sizeof(char)*n)
+        try:
+            H5PLget(index, buf, n)
+            return PyBytes_FromStringAndSize(buf, n)
+        finally:
+            efree(buf)
 
-cpdef size():
-    """() => UINT"""
-    cpdef unsigned int n = 0
-    H5PLsize(&n)
-    return n
+    cpdef size():
+        """() => UINT"""
+        cpdef unsigned int n = 0
+        H5PLsize(&n)
+        return n
--- a/h5py/h5pl.pyx
+++ b/h5py/h5pl.pyx
@@ -1,0 +1,55 @@
+# This file is part of h5py, a Python interface to the HDF5 library.
+#
+# http://www.h5py.org
+#
+# Copyright 2008-2019 Andrew Collette and contributors
+#
+# License:  Standard 3-clause BSD; see "license.txt" for full license terms
+#           and contributor agreement.
+
+"""
+    HDF5 plugin interface.
+"""
+
+from utils cimport emalloc, efree
+
+# === C API ===================================================================
+
+cpdef append(const char* search_path):
+    """(STRING search_path)"""
+    H5PLappend(search_path)
+
+cpdef prepend(const char* search_path):
+    """(STRING search_path)"""
+    H5PLprepend(search_path)
+
+cpdef replace(const char* search_path, unsigned int index):
+    """(STRING search_path, UINT index)"""
+    H5PLreplace(search_path, index)
+
+cpdef insert(const char* search_path, unsigned int index):
+    """(STRING search_path, UINT index)"""
+    H5PLinsert(search_path, index)
+
+cpdef remove(unsigned int index):
+    """(UINT index)"""
+    H5PLremove(index)
+
+cpdef get(unsigned int index):
+    """(UINT index) => STRING"""
+    cpdef size_t n
+    cpdef char* buf = NULL
+
+    n = H5PLget(index, NULL, 0) + 1
+    buf = <char*>emalloc(sizeof(char)*n)
+    try:
+        H5PLget(index, buf, n)
+        return PyBytes_FromStringAndSize(buf, n)
+    finally:
+        efree(buf)
+
+cpdef size():
+    """() => UINT"""
+    cpdef unsigned int n = 0
+    H5PLsize(&n)
+    return n
--- a/h5py/tests/old/test_h5pl.py
+++ b/h5py/tests/old/test_h5pl.py
@@ -32,7 +32,7 @@ def sandboxed(test):
         lines, start = inspect.getsourcelines(test)
         source = '\n'.join(lines[1:])  # remove this decorator
 
-        with tempfile.NamedTemporaryFile('w', dir='.', prefix='sandbox-', suffix='.py', delete=False) as f:
+        with tempfile.NamedTemporaryFile('w', prefix='sandbox-', suffix='.py', delete=False) as f:
             f.write('''
 try:
     import unittest2 as ut
@@ -44,13 +44,21 @@ from h5py import h5pl
 
 class {}(ut.TestCase):
 {}
-'''.format(test.__qualname__.split('.')[0], source, check=True))
+'''.format((test.__qualname__.split('.')[0]) if hasattr(test, '__qualname__')
+           else test.im_class.__name__,
+           source, check=True))
+
             fname = f.name
             try:
                 # support Windows
                 f.close()
-                subprocess.run((sys.executable, '-m', 'pytest', '-q', fname),
-                               check=True)
+
+                env = {**os.environ}
+                if 'HDF5_PLUGIN_PATH' in env:
+                    del env['HDF5_PLUGIN_PATH']
+
+                subprocess.check_call((sys.executable, '-m', 'pytest', '-q', fname),
+                                      env=env)
             finally:
                 os.remove(fname)
 
--- a/h5py/tests/old/test_h5pl.py
+++ b/h5py/tests/old/test_h5pl.py
@@ -75,34 +75,34 @@ class TestSearchPaths(ut.TestCase):
     @sandboxed
     def test_default(self):
         self.assertEqual(h5pl.size(), 1)
-        self.assertTrue(h5pl.get(0).endswith(b'hdf5/lib/plugin\x00'))
+        self.assertTrue(h5pl.get(0).endswith(b'hdf5/lib/plugin'))
 
     @sandboxed
     def test_append(self):
         h5pl.append(b'/opt/hdf5/vendor-plugin')
         self.assertEqual(h5pl.size(), 2)
-        self.assertTrue(h5pl.get(0).endswith(b'hdf5/lib/plugin\x00'))
-        self.assertEqual(h5pl.get(1), b'/opt/hdf5/vendor-plugin\x00')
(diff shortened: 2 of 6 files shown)


h5py/tests/old/test_h5pl.py
```diff
@@ -0,0 +1,56 @@
...
+    import unittest2 as ut
+except ImportError:
+    import unittest as ut
+
+import h5py
+from h5py import h5pl
+
+from ..common import TestCase
+
+@ut.skip('The tests have side effects')
```
takluyver: If we go for this, I think we should come up with a way to run the tests by default; maybe run them in a separate process? Tests that don't run by default aren't very helpful.

h5py/h5pl.pyx
```diff
@@ -0,0 +1,57 @@
...
+    HDF5 plugin interface.
+"""
+
+include "config.pxi"
+from utils cimport emalloc, efree
+
+# === C API ===================================================================
+
+IF HDF5_VERSION >= (1, 10, 1):
+    cpdef append(const char* search_path):
```
takluyver: I'm not very familiar with the low-level API, but should these have the `with_phil` decorator for locking? Check with the other maintainers before adding this;  it might not be necessary.

h5py/h5pl.pyx
```diff
@@ -0,0 +1,57 @@
...
+        """(STRING search_path)"""
+        H5PLappend(search_path)
```
takluyver: Should these have error handling, or at least a return value so that you can check if they succeeded?

(Again, I'm not that familiar with the low-level code; maybe this is somehow taken care of automatically)

h5py/tests/old/test_h5pl.py
```diff
@@ -0,0 +1,56 @@
...
+@ut.skipUnless(h5py.version.hdf5_version_tuple >= (1, 10, 1), 'HDF5 1.10.1+ required')
+class TestSearchPaths(TestCase):
+
+    def test_default(self):
+        self.assertEqual(h5pl.size(), 1)
```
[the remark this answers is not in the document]
aparamon: No, quite the contrary: it assumes the default value (see next line).
Running in the separate process, without side effects and predictable `HDF5_PLUGIN_PATH` sounds good. What's the most elegant way to do so?

h5py/tests/old/test_h5pl.py
aparamon: Tests are also documentation ;-)
But I agree; see my other comment.

h5py/h5pl.pyx
aparamon: I believe it is, due to the way `api_gen.py` works (it adds `except *` to every generated `cdef`).

h5py/tests/old/test_h5pl.py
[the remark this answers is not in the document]
takluyver: I don't know if pytest has some neat magic to make this convenient, but I would do something like this:

```python
from subprocess import run, PIPE, STDOUT

res = run([sys.executable, '-m', 'pytest', '...'], stdout=PIPE, stderr=STDOUT)
print(res.stdout)
assert res.returncode == 0
```

The `...` there would contain whatever arguments are needed to make pytest run the tests in a module it would otherwise skip.

h5py/h5pl.pyx
takluyver: OK, great, I hadn't looked at `api_gen.py`.

h5py/tests/old/test_h5pl.py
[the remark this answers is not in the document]
takluyver: PR #1003 was merged, so as far as I know, it uses pytest.

A lot of tests are still written with the unittest API; pytest can find and run these with no problem. I'm not personally enthusiastic about 'converting' existing tests to a more idiomatic pytest style - it sounds like a lot of change for very little benefit - but I'd say it's OK to write new tests assuming pytest.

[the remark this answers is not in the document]
aparamon: What do you think about the approach of https://github.com/pytest-dev/pytest-forked?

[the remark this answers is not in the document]
takluyver: Not enthusiastic! It looks like it's implemented as a command-line option which applies to all tests, whereas we want to apply this to only a specific set of tests that manipulate global state. Also, fork-without-exec is easy to mess up.

I can imagine this being a useful general utility, though - run some tests in a subprocess and integrate their results back into the main test results. Pytest itself has some machinery for testing pytest plugins which might be a useful starting point for something like this: https://docs.pytest.org/en/latest/writing_plugins.html#testing-plugins

aparamon:
Uhh, it took longer to make it run on all Pythons, but now at least Travis passes. I have no idea what's going on with AppVeyor though: https://ci.appveyor.com/project/h5py/h5py/builds/22108613/job/1aqbnhig7tw4hxr4#L237

@takluyver Could you please take another look and make your judgement whether my sandboxing implementation is in principle tolerable.

h5py/tests/old/test_h5pl.py
```diff
@@ -0,0 +1,110 @@
...
+        self.assertTrue(h5pl.get(0).endswith(b'hdf5/lib/plugin\x00'))
+
+    @sandboxed
+    def test_append(self):
+        h5pl.append(b'/opt/hdf5/vendor-plugin')
+        self.assertEqual(h5pl.size(), 2)
+        self.assertTrue(h5pl.get(0).endswith(b'hdf5/lib/plugin\x00'))
+        self.assertEqual(h5pl.get(1), b'/opt/hdf5/vendor-plugin\x00')
```
takluyver: I would generally expect a Python API to return strings without the trailing null. Does h5py's low level API include the trailing null elsewhere?

aparamon: I'm not sure, and frankly I was quite surprised HDF5 returns nulls here. [The docs](https://support.hdfgroup.org/HDF5/doc/RM/RM_H5PL.html#Plugin-Get) do not mention it (if I'm not missing something).
Being in doubt I went with *"h5py is a **thin**, pythonic wrapper"* motto.

takluyver:
For running the tests in a subprocess, I was thinking of something like using a [pytest marker](https://docs.pytest.org/en/latest/example/markers.html) along with a [pytest configuration file](https://docs.pytest.org/en/latest/customize.html#adding-default-options) to exclude these tests by default. Then run pytest in a subprocess, overriding the marker selection to run only those tests.

I haven't 100% figured this out, but I'm pretty sure it should be possible, and it saves creating temporary files for the test code.

takluyver:
The problem with Appveyor appears to be affecting my build on #1132 as well, so it looks like it's not your PR causing it.

Someone gets to have fun debugging Windows testing issues! :laughing: :disappointed:

aparamon:
Well, each of H5PL tests needs to be run in a subprocess. If we are less pedantic, these could be run independently of the first batch but without isolation; the failures would be a bit harder to localize, but pragmatically that might still be better compared to per-test sandboxing hackery.

However I see a great surprise effect for the users when they discover multiple pytest invocations are needed :-/ These tests are making me sad; maybe remove them altogether?

aparamon:
> Someone gets to have fun debugging Windows testing issues! 😆 😞
It there a way to ssh/Remote Desktop there?

h5py/tests/old/test_h5pl.py
takluyver: I'd guess it doesn't mention it because it's normal for C to store strings in n+1 bytes including the terminating null byte. Maybe Cython has some convenient functionality to trim off the null byte?

aparamon:
https://github.com/pytest-dev/pytest-forked is probably closest from what I have googled so far.

h5py/h5pl.pyx
```diff
@@ -0,0 +1,57 @@
...
+        cpdef size_t n
+        cpdef char* buf = NULL
+
+        n = H5PLget(index, NULL, 0)
+        buf = <char*>emalloc(sizeof(char)*(n + 1))
+        try:
+            H5PLget(index, buf, n + 1)
+            return PyBytes_FromStringAndSize(buf, n)
+        finally:
+            efree(buf)
```
takluyver: FWIW, here's another place in the codebase doing something similar:

https://github.com/h5py/h5py/blob/c5859441d47e58be185daf8dc812bd328a8867bb/h5py/h5p.pyx#L849-L856

I suspect that the hardcoded decode-as-utf8 there is not quite right, but it looks like one can call `bytes(buf)` to convert the buffer to Python bytes.

takluyver:
Yeah, pytest-forked is similar,  but that's for when you want to isolate every test in your test suite, because any of them might segfault. And it doesn't work on Windows. I think there must be other projects which want to isolate just a few particular tests that are doing strange things.

aparamon:
Well, I'm attending https://conf.python.ru/2019, so might be pointed to something worthy there ;-)
I agree that that the task is quite generic, but it's just hard to implement (e.g. Windows lacking fork).

takluyver:
It's not easy to implement, but I don't think it's all that hard either; it needs a bit of thinking about how to get the right data to the right place. Maybe other projects just have a similar bit of ugly infrastructure built into their test suites. I don't know where to look for it, though. Perhaps it's more niche than I thought.

aparamon:
Hit by morning inspiration, I went for something completely different. Why bother with subprocess isolation if HDF5 can be re-initialized afresh with [H5close](https://support.hdfgroup.org/HDF5/doc/RM/RM_H5.html#Library-Close)+[H5open](https://support.hdfgroup.org/HDF5/doc/RM/RM_H5.html#Library-Open) combo?
````py
class TestSearchPaths(ut.TestCase):

    def setUp(self):
        h5.close()
        # as per https://support.hdfgroup.org/HDF5/doc/Advanced/DynamicallyLoadedFilters/HDF5DynamicallyLoadedFilters.pdf,
        # in case your HDF5 setup has it different
        # (e.g. https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=826522)
        os.environ['HDF5_PLUGIN_PATH'] = os.path.expandvars('%ALLUSERSPROFILE%/hdf5/lib/plugin') \
            if platform.system() == 'Windows' else '/usr/local/hdf5/lib/plugin'
        h5.open()

    def tearDown(self):
        h5.close()
        h5.open()
````
Exposing `H5open`/`H5close` was easy; unfortunately it soon appeared that `h5py` itself does it's own initialization. And `importlib.reload(h5py)` didn't prove enough, unfortunately:
https://github.com/h5py/h5py/blob/c5859441d47e58be185daf8dc812bd328a8867bb/h5py/h5t.pyx#L211-L247
https://github.com/h5py/h5py/blob/c5859441d47e58be185daf8dc812bd328a8867bb/h5py/h5t.pyx#L1402-L1408
Even `importlib.reload(h5py.h5t)` is not enough, due to https://github.com/cython/cython/issues/2659.
The following code allowed me to pass all tests:
````py
    def tearDown(self):
        h5.close()
        h5.open()
        h5py.h5t.init()  # new function
        importlib.reload(h5py._hl.base)
        h5py._hl.base.dlapl = h5py._hl.base.default_lapl()
        h5py._hl.base.dlcpl = h5py._hl.base.default_lcpl()
        importlib.reload(h5py)
````
but I'm now not sure if it's as appealing as perceived initially :-)

Do we want to make targeted changes to allow re-initialization of `h5py` after it's loaded? Might it give some additional benefits?

**Comment by takluyver** [comment]
That doesn't sound like a promising approach, unfortunately. I don't think tests should rely on a global 'reset button' that isn't used explicitly in normal code. It's easy for oversights to mean the reset button doesn't reset everything, and there may be corner cases when it messes up debugging. The guaranteed way to get a clean, isolated process environment is to start a new process.

Let's wait and see what the other maintainers (@tacaswell @aragilar) think of this. Maybe if it's only adding low-level API functions, it's OK for the tests not to run by default. Or perhaps they have some other idea about how to handle it.

**Comment by vasole** [comment]
@aparamon 

I have been able to set the path after importing h5py and I have been able to read the data. Thanks a lot!

Is it intentional that I have to specify the path as a byte string? It is a bit confusing that when using HDF5_PLUGIN_PATH I can use a string and that when using the function I have to use bytes.

It took me some time to figure out why append was not working. I had to get rid of the non existing default directory or to use insert(0, path) that was also working.

Tested on windows, python 3.6, hdf5 1.10.5 as supplied by the HDF Group.

```python
import os
import sys
import h5py
# cleanup directories that might not exist
# other way append does not work and I have to use insert (path, 0)
# that works
delete = []
for i in range(h5py.h5pl.size()):
    dirname = h5py.h5pl.get(i)
    if not os.path.exists(dirname):
        print("directory <%s> to be deleted" % dirname)
        delete.append(i)
delete.reverse()
for i in delete:
    h5py.h5pl.remove(i)
h5py.h5pl.append(b"C:\\GIT\\hdf5plugin\\hdf5plugin\\VS2015\\x64")
#h5py.h5pl.insert(b"C:\\GIT\\hdf5plugin\\hdf5plugin\\VS2015\\x64", 0)

for i in range(h5py.h5pl.size()):
    print("i = ", i, "path = ", h5py.h5pl.get(i))

fname = "lz4.h5"
assert(os.path.exists(fname))
h5 = h5py.File(fname, "r")
data = h5["/entry/data"][:]
h5.close()
expected_shape = (50, 2167, 2070)
assert(data.shape[0] == 50)
assert(data.shape[1] == 2167)
assert(data.shape[2] == 2070)
assert(data[21, 1911, 1549] == 3141)
```

aparamon:
@vasole Thanks for the review!
I was hesitant to decode HDF5 byte-strings because I was not sure about the encoding. Maybe it's the right thing to do in some higher-level interface (based on [MutableSequence](https://docs.python.org/3/library/collections.abc.html#collections.abc.MutableSequence)), but for now I propose to explicitly `encode()`/`decode()` strings for use in `h5pl` functions.
`HDF5_PLUGIN_PATH` is environment variable which is already a decoded string in Python.

takluyver:
Ping @tacaswell @aragilar - I wanted your ideas about testing this before we merge it.

To summarise the discussion: it's tricky to test nicely, because the desired effect of these functions is global changes in the HDF5 library. Unless you can be sure you're undoing global changes properly, they can cause other tests to behave strangely, which is horrendous to debug. So I'd like to run these tests in their own subprocess so they don't interfere. But we haven't found a good way to do that.

So, do we:

1. Add the functions without tests, or with tests that aren't run by default? It's only exposing some more functions from HDF5, so maybe we don't need to test them.
2. Write normal tests, and trust that either the cleanup or the other tests are robust enough to cope with this.
3. Hack our own machinery to run these tests in a subprocess and bring the results back to the main process (the current state of this PR).
4. Use some package I'm not aware of to deal with running the tests in a separate process.

aragilar:
A variant of the first option might be a good middle ground: skip the plugin tests in the main test run, and have a separate test run for just the plugin tests (using something like https://docs.pytest.org/en/latest/example/simple.html#control-skipping-of-tests-according-to-command-line-option and an addition text env in tox). This won't stop the plugin tests interfering with each other, but at least it will localise the problem to the plugin system.

h5py/h5pl.pyx
tacaswell: the `phil` lock is mostly about making sure that we don't create / reuse / destroy c++ objects in an unsafe way (this is to support the correct ref-counting on holding c++ File objects open even if the user drops all (direct) python references to the File and only keeps Datasets).  See the code in `_objects.pyx` which is rather racey (including an issue where we discovered we have to turn off gc temporarily due to `list(of_a_dict)` having "mutation during iteration" issues).

That is a very long way of saying I don't think we need to lock with the same lock as everything else (as if the user has race conditions between adding plugins to the search path and opening files I think that is on them).

On the other hand, these should be used rarely enough that locking could not hurt.  We may also want to lock them between eachother (as I can imagine an `append` and a `remove` colliding and ending up with a "gap" in the hdf5 datastrucutres....).

scopatz:
It is a little hard to believe that we are still using unittest here

takluyver:
Can I ask you to rebase this on master and try making use of the `@insubprocess` test decorator @scopatz added in #1224?

Also, I think that the new `h5pl` module should have a corresponding rst document in [docs_api](https://github.com/h5py/h5py/tree/master/docs_api).

takluyver:
I've rebased and continued this in #1256

takluyver:
This was merged as #1256.

PR #1168, opened 2019-02-04
BLD: update cython files work with langage-level 3str

tacaswell:


tacaswell:
Well, this fails because the version of cython is to old....

tacaswell:
This needs to have the level changed to `3` and then fix all of the `char *` vs unicode issues.  Only the current version of cython actually supports the future default value of `'3str'` so we need to do the full conversion. 

I have also opened https://github.com/h5py/h5py/pull/1168 which just future proofs the status-quo and should go in a bug-fix release.

takluyver:
@tacaswell I'm assuming we're not going to bump the required Cython version for 2.10. Does it make sense to move this to the 3.0 milestone? Or is it something for further in the future?

tacaswell:
attn @kif

kif:
Noted ...
As I switched for cython language level "3str" to "3"  in #1332, it should be safe to decrease the minimal cython version needed to 0.23 in the setup.py.
I am right now testing on debian9 (py35, cython 0.25) and will update  #1332 accordingly.

takluyver:
This has been superseded by #1332.

PR #1171, opened 2019-02-06
BLD: explicitly set cython language level

tacaswell:
This is future-proofing against the next version of cython which will
change the default language level to '3str' (from the current 2)


aragilar:
What do we need to do to backport this?

takluyver:
The bot opened PR #1173 to backport it; I just merged that.

Issue #566, opened 2015-04-20
h5py files and groups could use a more informative repr


shoyer:
I've recently been playing around with h5py with an eye towards [writing netCDF4 files](https://github.com/shoyer/h5netcdf). So far it's been quite a pleasure -- thanks for your hard work!

One thing I've noticed is that the standard `__repr__` makes it rather slow to explore h5py files -- I end up calling `.items()` or `dict` on groups quite often to figure out what is inside them. This makes things less pleasant than they could be -- see [this blog post](http://matthewrocklin.com/blog/work/2014/11/19/Blaze-Datasets/) by @mrocklin for an extended discussion.

It would be nice if the default repr showed more than `<HDF5 file "test.nc" (mode r)>`. For example, by default it might show short summaries of all (direct) child groups, variables and attributes. Thoughts?

andrewcollette:
Cool blog post... thanks!  But I think we'll leave this one to third-party packages, like blaze.  A fast, compact repr() is appropriate for a library package like h5py.

mrocklin:
Alternatively to `repr` you could override `__dir__` and `__getattr__`.  This is cheap, straightforward, and fairly unambiguous.

andrewcollette:
@mrocklin... one of the design goals for h5py was to not mix the Python attribute namespace with the file object namespace.  We have a strict dictionary metaphor for groups and attributes.

Btw, there is an IPython completer for h5py... call h5py.enable_ipython_completer() and you can use TAB completion on names:

```
>>> import h5py
>>> h5py.enable_ipython_completer()
>>> f = h5py.File('myfile.hdf5')
>>> f['<TAB>
group1 group2 group2
```

mrocklin:
Fair enough.  I've appreciated the sanity of h5py in the past.

anthonytw:
Can this possibly be reconsidered? I appreciate the overall design philosophy of h5py as a streamlined library interface, and I think a modestly improved repr would not derail that goal. I have a short patch for files, groups, and key views that makes browsing much more convenient. I can submit a pull request which you are of course free to reject, but I think would be a greatly appreciated fix with no performance impact on library performance.

PR #1182, opened 2019-03-15
Update __repr__ to provide more convinient interface for browsing h5…

anthonytw:
… files interactively.

This is a short commit to improve the readability of the repr output. Despite being originally denied, I hope this can be reconsidered as it considerably improves the ability of an interactive user to browse an HDF5 container while not otherwise affecting the performance or design of the library. Minimal improvements (as seen here) I believe could go a long way.

An example of the changes in use:

```
In [1]: import h5py as h5                                                                                             

In [2]: f = h5.File('output.h5', 'r')                                                                                 

In [3]: f                                                                                                             
Out[3]: 
<HDF5 file "output.h5" (mode r)>
<KeysViewHDF5>:
  + numeric
  + waveform

In [4]: f['waveform']                                                                                                 
Out[4]: 
<HDF5 group "/waveform" (9 members)>
<KeysViewHDF5>:
  + ECG.I
  + ECG.II
  + ECG.III
  + ECG.MCL
  + ECG.V
  + ECG.aVR
  + Pleth
  + PlethT
  + Resp

In [5]: f['waveform']['ECG.I']                                                                                        
Out[5]: 
<HDF5 group "/waveform/ECG.I" (3 members)>
<KeysViewHDF5>:
  + data
  + meta
  + unit
```

#566

### DIFF
--- a/h5py/_hl/base.py
+++ b/h5py/_hl/base.py
@@ -310,7 +310,9 @@ class KeysViewHDF5(KeysView):
     def __str__(self):
         return "<KeysViewHDF5 {}>".format(list(self))
 
-    __repr__ = __str__
+    def __repr__(self):
+        return '\n'.join(["<KeysViewHDF5>:"] + ["  + " + key for key in list(self)])
+
 
 class ValuesViewHDF5(ValuesView):
 
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -461,8 +461,9 @@ class File(Group):
             filename = self.filename
             if isinstance(filename, bytes):  # Can't decode fname
                 filename = filename.decode('utf8', 'replace')
-            r = u'<HDF5 file "%s" (mode %s)>' % (os.path.basename(filename),
+            r = u'<HDF5 file "%s" (mode %s)>\n' % (os.path.basename(filename),
                                                  self.mode)
+            r += repr(self.keys())
 
         if six.PY2:
             return r.encode('utf8')
--- a/h5py/_hl/group.py
+++ b/h5py/_hl/group.py
@@ -570,7 +570,8 @@ class Group(HLObject, MutableMappingHDF5):
             namestr = (
                 u'"%s"' % self.name
             ) if self.name is not None else u"(anonymous)"
-            r = u'<HDF5 group %s (%d members)>' % (namestr, len(self))
+            r = u'<HDF5 group %s (%d members)>\n' % (namestr, len(self))
+            r += repr(self.keys())
 
         if six.PY2:
             return r.encode('utf8')


tacaswell:
This is probably better as an IPython display hook rather than over-writing the repr.

While this seems very nice for interactive work, I am very wary of adding new lines to reprs as it can have surprising down-stream consequences (like if people are currently using repr in log messages or as part of their own down-stream reprs) [1].  By putting this as a display hook we will only trigger this behavior when printed an the prompt.


[1] we recently made a change like this at my day job on some of our custom objects and reverted it almost immediately as it made the repr in any context that was not the interactive prompt completely unusable.

anthonytw:
In this case I'm using IPython but I suspect there still exist ~~savages~~ "individuals" who work interactively using the standard Python REPL, and it would be optimal if it worked there as well.

For IPython only, I could stuff the modified `repr` in a new `_repr_pretty_` function, which will certainly solve the issue at least for IPython users. I'm not sure how the display hooks you're describing work, all recent references to IPython and Jupyter output formatters seem to point to an extended set of `_repr_<format>_` functions. Maybe you can point me to some documentation on what you were thinking about?

I'd recommend (last time, I promise :-)) re-re-considering an updated `repr` though, it's a more general fix. Code that relies on the assumption of a known, consistent `repr` format, especially given that it's primarily used to format output for human consumption (and primarily in the interpreter), sounds like it would be more of an outlier than standard practice (and I probably wouldn't consider it a _good_ practice either, since that's probably more often than not an invalid assumption to make). As an end-user, I would always assume the `repr` is subject to change, possibly even based on end-user configuration I have might have no control over (e.g. modifying numeric precision in `numpy` outputs via the `set_printoptions`).

Let me know what you think, and I'll make the appropriate changes.

tacaswell:
> As an end-user, I would always assume the `repr` is subject to change, possibly even based on end-user configuration I have might have no control over (e.g. modifying numeric precision in `numpy` outputs via the `set_printoptions`).

Fully on board tat we can change it at will



> Code that relies on the assumption of a known, consistent `repr` format, especially given that it's primarily used to format output for human consumption

The concern is the length (in lines).


> Maybe you can point me to some documentation on what you were thinking about?

See https://ipython.readthedocs.io/en/stable/config/integrating.html#formatters-for-third-party-types

takluyver:
I'd agree with @tacaswell here that the repr shouldn't be multi-line. Imagine if you have h5py File objects in a list or a dict: multi-line reprs will get ugly pretty quickly.

I've already got enhanced HTML reprs with an IPython display hook for use in notebooks [in the h5glance project](https://nbviewer.jupyter.org/github/European-XFEL/h5glance/blob/master/Demo.ipynb). I'd be open to adding plain-text IPython display features there too, if you're interested.

anthonytw:
@tacaswell Okay, I'll rework a solution to register a formatter.

@takluyver That sounds nice, your solution looks better than too! A text formatter would be nice, I can configure an IPython formatter to default to using h5glance if it's available, otherwise falling back on my more basic view.

anthonytw:
Updated to PR #1188

PR #1188, opened 2019-03-19
Add IPython formatters to provide more convinient interface for browsing HDF5 files interactively

anthonytw:
This commit attempts to enable pretty printing of some HDF5 objects.

An example of the changes in use:

```
In [1]: import h5py

In [2]: f = h5py.File('output.h5', 'r')

In [3]: f
Out[3]: <HDF5 file "output.h5" (mode r)>

In [4]: h5py.enable_ipython_formatter()

In [5]: f
Out[5]:
<HDF5 file "output.h5" (mode r)>
<HDF5 group "/" (2 members)>
<KeysViewHDF5>:
  + numeric
  + waveform

In [6]: f['waveform']
Out[6]:
<HDF5 group "/waveform" (9 members)>
<KeysViewHDF5>:
  + ECG.I
  + ECG.II
  + ECG.III
  + ECG.MCL
  + ECG.V
  + ECG.aVR
  + Pleth
  + PlethT
  + Resp
```

#566

### DIFF
--- a/h5py/_ipython/__init__.py
+++ b/h5py/_ipython/__init__.py
@@ -1,0 +1,47 @@
+#+
+#
+# This file is part of h5py, a low-level Python interface to the HDF5 library.
+#
+# Contributed by Anthony Wertz
+#
+# http://h5py.alfven.org
+# License: BSD  (See LICENSE.txt for full license)
+#
+#-
+
+# pylint: disable=eval-used,protected-access
+
+"""
+    This provides an interface for enabling IPython extensions.
+"""
+
+def enable_ipython_completer():
+    """ Call this from an interactive IPython session to enable tab-completion
+    of group and attribute names.
+    """
+    from .utils import is_ipython_initialized
+    if is_ipython_initialized():
+        from . import completer
+        return completer.load_ipython_extension()
+
+    raise RuntimeError('Completer must be enabled in active ipython session')
+
+
+def enable_ipython_formatter():
+    """ Call this from an interactive IPython session to enable prettier printing
+    of h5py objects.
+    """
+    from .utils import is_ipython_initialized
+    if is_ipython_initialized():
+        from . import formatter
+        return formatter.load_ipython_extension()
+
+    raise RuntimeError('Formatter must be enabled in active ipython session')
+
+
+def enable_ipython(completer=True, formatter=True):
+    """ Call this from an interactive IPython session to enable multiple
+    convenience extensions for IPython.
+    """
+    if completer: enable_ipython_completer()
+    if formatter: enable_ipython_formatter()
--- a/h5py/__init__.py
+++ b/h5py/__init__.py
@@ -64,6 +64,9 @@ from .h5t import (special_dtype, check_dtype,
     check_vlen_dtype, check_string_dtype, check_enum_dtype, check_ref_dtype,
 )
 
+from ._ipython import (
+    enable_ipython_completer, enable_ipython_formatter, enable_ipython)
+
 from . import version
 from .version import version as __version__
 
@@ -86,29 +89,6 @@ def run_tests(args=''):
     # requirements, e.g. pytest
     from .tests import run_tests
     return run_tests(args)
-
-
-def enable_ipython_completer():
-    """ Call this from an interactive IPython session to enable tab-completion
-    of group and attribute names.
-    """
-    import sys
-    if 'IPython' in sys.modules:
-        ip_running = False
-        try:
-            from IPython.core.interactiveshell import InteractiveShell
-            ip_running = InteractiveShell.initialized()
-        except ImportError:
-            # support <ipython-0.11
-            from IPython import ipapi as _ipapi
-            ip_running = _ipapi.get() is not None
-        except Exception:
-            pass
-        if ip_running:
-            from . import ipy_completer
-            return ipy_completer.load_ipython_extension()
-
-    raise RuntimeError('Completer must be enabled in active ipython session')
 
 
 # --- Legacy API --------------------------------------------------------------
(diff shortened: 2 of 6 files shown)


h5py/_ipython/__init__.py
```diff
@@ -0,0 +1,47 @@
+#+
+#
+# This file is part of h5py, a low-level Python interface to the HDF5 library.
+#
+# Contributed by Anthony Wertz
+#
+# http://h5py.alfven.org
```
takluyver: I see this is copy-pasted from files, but the URL doesn't actually work.

aragilar:
Is this something we're planing on having, or can we put this (or equivalent functionality, if it doesn't already exist) in h5glance? I'm not going to look at rebasing this as there's no tests, and the ipython/jupyter functionally in h5py has tended to bitrot due to lack of interest (and tests).

takluyver:
I was actually working just the other day on integrating h5glance's text-based output into IPython: https://github.com/European-XFEL/h5glance/pull/21 .

I still have a slight preference for putting fancy formatting features in a separate package, rather than h5py itself. I've mentioned this elsewhere, but there are two main arguments:

- Visualisation can be surprisingly complex, and there's no obvious natural stopping point where we can say "this is in scope, that isn't". So if h5py starts doing fancy visualisations, it could become a large bulk of extra code which is never exactly done.
- Different visualisations may make sense in different cases, e.g. depending on how many nested groups your files have, or what kind of information is stored in attributes. This is easier with space for specialised visualisation tools than if h5py includes an official standard visualisation.

Of course, the counterargument is that there's value in having a single shared visualisation rather than everyone going off and doing their own thing. And I have a conflict of interest because h5py doing this would make h5glance less valuable - so I shouldn't be the one to decide.

anthonytw:
I agree with @takluyver on complicated visualizations, but I think it's reasonable (if not expected) to include a minimal set of data structure "visualization" capability in the package itself. In fact, it shouldn't even be relegated to some tucked-away portion of the code like this implementation that has to be manually loaded. And it definitely shouldn't be stuffed in a completely separate package. Both are non-intuitive (and cumbersome) workflows.

Imagine how much less useful pandas or numpy would be if viewing a DataFrame or recarray just told you there were N rows and M columns, or a numpy array that told you there were N elements, but none showed you any data, types, column names, etc. There are much more complex and useful visualizations out there that can't be accomplished with the stock pandas or numpy library directly and those get their own packages (various plotting libraries being the obvious examples), but a simple rendering of a complex (or not so complex) data type is almost universally expected to get a quick look at what you're dealing with. Telling me that an HDF5 group has 2 members (groups? datasets? attributes are not included? no names?) is not terribly useful. And the root file object (which is also a group but prints differently) doesn't tell me anything useful at all, just the file name and mode!

There's no reason listing out the member groups, attributes, and datasets shouldn't be the default behavior, I would go as far as to suggest it's what most people would expect to see when viewing any reasonably presentable data structure in Python. And the features and scope can remain pretty minimal: just print out something that's helpful and descriptive. For an HDF5 file, that's incredibly easy to define. h5glance looks neat and adds a lot of cool features (I like the command line tool with bash integration and notebook integration) and definitely counts as out-of-scope visualization. h5py should be much simpler, but more useful than it currently is.

Just my two cents! 👍

vasole:
In my opinion h5py itself should not go beyond providing a text-based representation(s) of the file structure.  There are many tools around capable of complex visualizations and trying to go beyond text-based representations within h5py might withdraw some resources from h5py itself. At most, I would consider the possibility to simplify registering user-specified formatters overwriting a default one provided by h5py. If people jump into it and there is a clear "winner" than why not to adopt it at a later stage.

For instance, for browsing HDF5 files interactively one comfortable thing is to use an external file-browser-like  tool allowing to inspect and to cut&paste or drag items into the ipython session. There are several solutions already available and it does not consume any resources from the h5py project.

aragilar:
@anthonytw The big issue is tests. Without tests the formatters will break (we've seen this with the ipython completers).

As an aside, a better comparison library would be [astropy.io.fits](https://docs.astropy.org/en/stable/io/fits/), it wraps the FITS file format (which is substantially simpler than HDF5, in many cases you would load the whole file into memory), rather than either numpy or pandas, which are in-memory objects. astropy.io.fits also provides very little information by default, instead providing a small amount of pretty printing behind some helper functions and scripts (at the level of h5dump).

anthonytw:
@vasole I agree on most points, and I think a simple list of attributes showing the structure fits into that mold. However, the issue I wanted to bring up with this (and an earlier PR) is that h5py does _not_ give a text-based representation of the file structure. I think doing so in a simple fashion (e.g., using repr of a file or group to list members) would be incredibly helpful, while being simple to implement and test.

@aragilar Agreed about tests. I initially initiated a PR with the IPython implementation to get feedback after some concerns about a repr implementation were raised.

Also, I think the comparisons with pandas and numpy are still relevant. For any single object (file, group, dataset) for which the text representation would be relevant, its immediate children (groups, datasets, attributes) or structure (e.g., dataset size and columns) should be available in memory, and that's all I propose rendering for an object view. So it's not a view of the actual data, which may or may not be loaded in memory, but of the data's organization within that file.

I think it's a bit cumbersome to figure out (interactively) what's present at an address in the HDF5 file system (first list the members and determine which are groups and which are datasets by type-checking the values, then separately list the attributes, and now you have the full picture). Granted this PR doesn't implement exactly that (I didn't list attributes or distinguish between groups and datasets as @takluyver pointed out) but I wanted to propose something to get feedback (which I see now I missed from last year) and make sure the group approved of my implementation of the ipython injection since I was not particularly familiar with that mechanism.

However, if it's not going to sit in the repr, it might be best to just drop the request. I'm not sure it even makes sense in h5glance which, at a ... glance (-_-') wouldn't affect the default views of h5py objects anyway. Another option might be to implement a method like "members" which returns a list of subgroups, datasets, and attribute as a dictionary without affecting the repr.

PR #1190, opened 2019-03-25
Expose H5DOread_chunk

vallsv:
Hi,

I try to expose `H5DOread_chunk` (https://confluence.hdfgroup.org/display/HDF5/H5D_READ_CHUNK) to h5py, in the way `H5DOwrite_chunk` was exposed. Here is the current state. It is mostly a copy-paste of `write_direct_chunk`.

- I have no idea how to read the data if the dataset was not chunked.
- The in/out `filter_mask` param is returned as a return value in a tuple.


vallsv:
My bad, `filters` was not supposed to be NULL.

vallsv:
If somebody have an idea why there is symbol error in the CI, it would be very useful.

vasole:
H5Dget_chunk_storage_size was introduced in 1.8.19 and 1.10.2 therefore the tests with 1.10.1 fail.

vallsv:
I get it... For tomorrow.

vallsv:
This PR is ok to be reviewed.
BTW, I have another PR to allow to expose it for 1.8 branches if you like.

h5py/tests/old/test_h5d_direct_chunk_write.py
```diff
@@ -32,3 +32,56 @@ def test_write_direct_chunk(self):
...
+
+
+@ut.skipUnless(h5py.version.hdf5_version_tuple >= (1, 10, 2), 'Direct Chunk Reading requires HDF5 >= 1.10.2')
+class TestReadDirectChunk(TestCase):
+    def test_read_offsets(self):
+
+        filename = self.mktemp().encode()
+        filehandle = h5py.File(filename, "w")
+
+        slice = numpy.arange(16).reshape(4, 4)
```
takluyver: Can we call this variable something else? `slice` is a Python builtin, so it confused me for a moment looking at the code.

takluyver:
This looks OK to me, but I'm not particularly well qualified to review this kind of low-level code.

h5py/tests/old/test_h5d_direct_chunk_write.py
vallsv: Oups, yes, thanks for the feedback. I fixed that on the following commits.

h5py/h5d.pyx
```diff
@@ -430,3 +431,65 @@ cdef class DatasetID(ObjectID):
...
+                    status = H5DOread_chunk(dset_id, dxpl_id, offset, &filters, data.data.as_voidptr)
+
+                if status < 0:
+                    raise TypeError("Error while reading chunk %s", offsets)
+            finally:
+                efree(offset)
+                if space_id:
+                    H5Sclose(space_id)
+
+            return bytes(data), filters
```
takluyver: Should this match the order of the parameters in the HDF5 API? i.e. filters first, then data?

takluyver:
Should we also add H5Dwrite_chunk while we're looking at this? It looks like both the functions from H5DO are deprecated.

vallsv:
> Should we also add H5Dwrite_chunk

I would be good, but the maintenance in parallel of the branch 1.8 and 1.10 make the thing difficult to manage. I would say, it would be better to fix it into another PR.

h5py/h5d.pyx
vallsv: Make sense, but the reader and writer do not use the same order. I follow the other one.
Unfortunately it will not be convenient to use, anyway.
Let me know and i update this PR.

takluyver: All the functions I could find with multiple out parameters have them in the same order for the return value. But I couldn't immediately see any others with parameters that serve for both input and output, so I don't know if there's a reason to make them an exception to the rule.

h5py/h5d.pyx
```diff
@@ -430,3 +431,65 @@ cdef class DatasetID(ObjectID):
                 efree(offset)
                 if space_id:
                     H5Sclose(space_id)
+
+    IF HDF5_VERSION >= (1, 10, 2):
+
+        def read_direct_chunk(self, offsets, filter_mask=0xFFFF, PropID dxpl=None):
+            """ (offsets, H5Z_filter_t filter_mask=[], PropID dxpl=None)
```
takluyver: I'm pretty sure the default value `[]` and the type `H5Z_filter_t` are both wrong. The [hdf5 docs](https://confluence.hdfgroup.org/display/HDF5/H5D_READ_CHUNK) say the type is uint32_t.

H5Z_filter_t is a numeric ID for a single filter. This value is a mask to select filters in a pipeline.

h5py/h5d.pyx
```diff
@@ -430,3 +431,65 @@ cdef class DatasetID(ObjectID):
                 efree(offset)
                 if space_id:
                     H5Sclose(space_id)
+
+    IF HDF5_VERSION >= (1, 10, 2):
+
+        def read_direct_chunk(self, offsets, filter_mask=0xFFFF, PropID dxpl=None):
```
takluyver: For parity with `write_direct_chunk`, I think `filter_mask` should probably default to 0 (i.e. apply all enabled filters), rather than `0xFFFF` (i.e. skip all filters).

h5py/h5d.pyx
```diff
@@ -430,3 +431,65 @@ cdef class DatasetID(ObjectID):
...
+
+            if offsets is not None:
+                if len(offsets) != rank:
+                    raise TypeError("offset length (%d) must match dataset rank (%d)" % (len(offsets), rank))
+
+            try:
+                offset = <hsize_t*>emalloc(sizeof(hsize_t)*rank)
+                convert_tuple(offsets, offset, rank)
+                status = H5Dget_chunk_storage_size(dset_id, offset, &read_chunk_nbytes)
+                if status < 0:
```
takluyver: I don't think you need the status checks; h5py should autogenerate a function in `defs.pyx` which wraps the HDF5 function to turn errors into Python exceptions.

h5py/h5d.pyx
```diff
@@ -430,3 +431,65 @@ cdef class DatasetID(ObjectID):
...
+                    raise TypeError("Error while reaching chunk storage size")
+                array.resize(data, read_chunk_nbytes)
+
+                IF HDF5_VERSION >= (1, 10, 3):
+                    status = H5Dread_chunk(dset_id, dxpl_id, offset, &filters, data.data.as_voidptr)
+                ELSE:
+                    status = H5DOread_chunk(dset_id, dxpl_id, offset, &filters, data.data.as_voidptr)
+
+                if status < 0:
```
takluyver: Ditto here.

h5py/h5d.pyx
vallsv: My default expectation is to read the raw data, which is the case by disabling all the filters. Else, we have to inject this magic `0xFFFF`, which is IMO not very nice. Or we have to create on our side a new const var. Maybe i misunderstand something.

BTW i just check the HDF5 API `H5DOwrite_chunk`, and it do not use filter type `H5Z_filter_t`, i am pretty sure it was the case before! Or it was a mistake from `h5py`? Then `H5Z_filter_t filter_mask=H5Z_FILTER_NONE` should be rewritten to `uint_t filter_mask=0x0000`. Which far way much better.

In this case, if `H5Z_FILTER_NONE` is not used in the writer, i think it is fine to use a magic `0xFFFF` on application side.

BTW do you use a resource to follow the changes of the HDF5 API?

vallsv: Oh `H5DOwrite_chunk` from `api_functions.txt` uses `uint32_t`. Then it's fine. I will update the thing as you suggest. And patch `write_direct_chunk` on the same time.

h5py/h5d.pyx
vallsv: Oh, interesting.

h5py/h5d.pyx
vallsv: Fixed

vallsv:
Here is an update for each of your comments. I also update `write_direct_chunk` with 0e06abe to match the HDF5 API. It should not break anything as it is mostly the same code, but not the same meaning.

h5py/h5d.pyx
```diff
@@ -430,3 +436,64 @@ cdef class DatasetID(ObjectID):
...
+                filters = 0xFFFF
+            else:
+                filters = int(filter_mask)
+
+            dset_id = self.id
+            dxpl_id = pdefault(dxpl)
+            space_id = H5Dget_space(self.id)
+            rank = H5Sget_simple_extent_ndims(space_id)
+
+            if offsets is not None:
```
takluyver: This 'if' should be unnecessary - people shouldn't be passing None.

h5py/h5d.pyx
```diff
@@ -430,3 +436,64 @@ cdef class DatasetID(ObjectID):
...
+        def read_direct_chunk(self, offsets, filter_mask=0x0000, PropID dxpl=None):
+            """ (offsets, uint32_t filter_mask=0x0000, PropID dxpl=None)
+
+            Reads data to a bytes array directly from a chunk at position
+            specified by the offsets argument.
+
+            The `filter_mask` is a bit field of up to 32 values. It specify
+            which filters in the pipeline are used with the chunk.
+            Using the default `0x0000` with use all the filters, while `0xFFFF`
+            will skip all the filters.
```
takluyver: I've just gone and looked at the HDF5 code, and I think `filter_mask` is actually not an input parameter for reading the chunk at all - it's just a pointer to receive the information HDF5 has recorded about filters applied to that chunk when writing. Sorry, I should have tried harder to understand this before.

h5py/h5d.pyx
```diff
@@ -399,12 +399,17 @@ cdef class DatasetID(ObjectID):
...
-            """ (offsets, bytes data, H5Z_filter_t filter_mask=H5Z_FILTER_NONE, PropID dxpl=None)
+        def write_direct_chunk(self, offsets, bytes data, filter_mask=0x0000, PropID dxpl=None):
+            """ (offsets, bytes data, uint32_t filter_mask=0x0000, PropID dxpl=None)
 
             Writes data from a bytes array (as provided e.g. by struct.pack) directly
             to a chunk at position specified by the offsets argument.
 
+            The `filter_mask` is a bit field of up to 32 values. It specify
+            which filters in the pipeline are used with the chunk.
+            Using the default `0x0000` with use all the filters, while `0xFFFF`
```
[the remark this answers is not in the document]
takluyver: Also also, now that I've delved more into what the filters parameter does, it doesn't actually *control* filters, it's just meant to record which ones have been applied to the data before this is called. So we'll need a bit of rewording to accommodate that.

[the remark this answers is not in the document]
vallsv: My understanding is it control the filter which have to be used when you have to read back the data: which filters have to be applied to the data you sent to read decompressed values.

h5py/h5d.pyx
vallsv: I don't think so cause it is an in/out argument. That's maybe a mistake on the documentation.
It needs more investigation to check with a real dataset.

takluyver: I think the docs are mistaken, or at least confusing. But the value it points to appears to be [zeroed out](https://bitbucket.hdfgroup.org/projects/HDFFV/repos/hdf5/browse/src/H5Dchunk.c?at=refs%2Ftags%2Fhdf5-1_10_5#524) before being [set to a retrieved value](https://bitbucket.hdfgroup.org/projects/HDFFV/repos/hdf5/browse/src/H5Dchunk.c?at=refs%2Ftags%2Fhdf5-1_10_5#582-583). I can't see anywhere that the value is used as input.

This also fits with the docs saying that the 'direct' read and write interfaces bypass the filter pipeline entirely.

vallsv: Thanks. It makes more sense now. I also have to rewrite the function to remove the input filter mask.

vallsv:
Here is a fix of the documentation and an a remove of `filter_mask` from the `read_direct_chunk` function.

vallsv:
But i tried to create a compressed empty dataset, feed it with uncompressed data. And i have problem to read back the chunk size. In case you have an idea.

Here is the patch https://github.com/vallsv/h5py/commit/4baf0d9fcfbbbd54f0b47163a03445adcb0b704c

```
        # Write uncompressed data
        DISABLE_ALL_FILTERS = 0x0FFFFFFF
        dataset.id.write_direct_chunk((0, 0, 0), frame.tostring(), filter_mask=DISABLE_ALL_FILTERS)
    
>       filter_mask, compressed_frame = dataset.id.read_direct_chunk((0, 0, 0))

/python3.5/site-packages/h5py/tests/old/test_h5d_direct_chunk.py:79: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   ???
E   RuntimeError: Can't get storage size of chunk (chunk storage is not allocated)

h5py/h5d.pyx:482: RuntimeError
```

h5py/h5d.pyx
takluyver: Again, I don't think you need to check for `offsets` being None. The corresponding write function doesn't.

h5py/tests/old/test_h5d_direct_chunk.py
```diff
@@ -0,0 +1,92 @@
...
+            dataset.id.write_direct_chunk((index, 0, 0), a.tostring(), filter_mask=1)
+            array[index] = a
+
+        filehandle.close()
+
+        # checking
+        filehandle = h5py.File(filename, "r")
+        for i in range(10):
+            read_data = filehandle["data"][i]
+            self.assertTrue((array[i] == read_data).all())
```
takluyver: numpy has some testing functions like `assert_array_equal`. Using these means more informative errors if the assertion ever fails.

https://docs.scipy.org/doc/numpy/reference/routines.testing.html

h5py/tests/old/test_h5d_direct_chunk.py
```diff
@@ -0,0 +1,92 @@
...
+                                            maxshape=(None, 100, 100),
+                                            chunks=(1, 100, 100),
+                                            dtype='float32')
+
+        # writing
+        array = numpy.zeros((10, 100, 100))
+        for index in range(10):
+            a = numpy.random.rand(100, 100).astype('float32')
+            dataset.id.write_direct_chunk((index, 0, 0), a.tostring(), filter_mask=1)
+            array[index] = a
```
takluyver: Isn't this overwriting the chunk you just wrote with `write_direct_chunk`?

h5py/tests/old/test_h5d_direct_chunk.py
```diff
@@ -0,0 +1,92 @@
...
+        dataset = filehandle.create_dataset("created",
+                                            shape=frame_dataset.shape,
+                                            maxshape=frame_dataset.shape,
+                                            chunks=frame_dataset.chunks,
+                                            dtype=frame_dataset.dtype,
+                                            compression="gzip",
+                                            compression_opts=9)
+
+        # copy the data
+        dataset.id.write_direct_chunk((0, 0), compressed_frame, filter_mask=filter_mask)
```
takluyver: I'd like a test which uses a dataset with compression defined and writes a chunk with the compression filter *disabled*. It should then check that:

- Accessing the relevant data normally (`ds[i]`) gives the expected data
- Reading the chunk through `read_direct_chunk` gives us the same filter_mask back.

h5py/h5d.pyx
```diff
@@ -398,12 +399,19 @@ cdef class DatasetID(ObjectID):
...
+            """ (offsets, bytes data, uint32_t filter_mask=0x00000000, PropID dxpl=None)
 
             Writes data from a bytes array (as provided e.g. by struct.pack) directly
             to a chunk at position specified by the offsets argument.
 
+            The `filter_mask` is a bit field of up to 32 values. It specify
+            which filters in the pipeline have to be disabled to decompress the
+            `data`.
+            Using the default `0x00000000` with use all the filters when the
+            data will be read, while `0xFFFFFFFF` will skip all the filters
```
takluyver: I'd still like this to be a bit clearer. Something like:

> The `write_direct_chunk` function bypasses any filters HDF5 would normally apply to written data. However, calling code may apply filters (e.g. gzip compression) itself before writing the data. `filter_mask` records which filters have been applied to this chunk, of the filter pipeline defined for that dataset. Each bit set to 1 means that the filter in the corresponding position in the pipeline was *not* applied. So the default value of 0 means that all defined filters have been applied to the data before calling this function.

And something similar for `read_direct_chunk`.

scopatz:
Also, the appveyor tests failed.

h5py/tests/old/test_h5d_direct_chunk.py
takluyver: Oh I see you already tried to do a test like this and described an error above. I'm not sure what that could mean.

The [example here](https://confluence.hdfgroup.org/display/HDF5/Using+the+Direct+Chunk+Write+Function) only uses `filter_mask = 0x00000001` for writing uncompressed data. I wonder if claiming to disable more filters than the dataset has does something strange, or if there's an edge case with setting 0xffffffff.

takluyver:
It looks like they might be failing on all recent builds, unfortunately.

takluyver:
I've opened PR #1225 to try to figure that out.

takluyver:
The tests are passing on master again, so next time you update this branch, it should be tested properly again.

h5py/tests/old/test_h5d_direct_chunk.py
vallsv: Ill try that. But i think it was not the pbm.
I guess the problem is the use of dataset which can grow.

h5py/h5d.pyx
vallsv: As you like

h5py/tests/old/test_h5d_direct_chunk.py
vallsv: Yes, but this test is not mine, and there is no problem to overwrite.

h5py/tests/old/test_h5d_direct_chunk.py
vallsv: Yes sure, it's much better with numpy test

vallsv:
Hi. Here is an update with the last requests.

I introduce back the test checking `write` then `read`, but to make it work, the file have to be reloaded. Do you think it could be a cache problem in h5py side?

takluyver:
I don't think h5py does any caching itself, but it might be cached in HDF5. Or I might be wrong about h5py.

Calling `filehandle.flush()` might work instead of closing and reopening it.

vallsv:
The flush do not improve the thing.
But I found that using
```
            dataset = None
            dataset = filehandle["frame"]
```
make it work.

takluyver:
In HDF5 terms, that probably means you're closing and reopening the dataset. I don't know whether it's meant to work without that step or not - you could try asking on the hdf forum. But I think we can go with that in the tests for now.

takluyver:
I've added this to the 2.10 milestone because I think it's basically ready to merge.

PR #1196, opened 2019-04-01
Use appveyor plugin for tox

takluyver:
Apparently the way we were setting TOXPYTHON for all environments is not recommended by the tox devs, and is probably going to raise an error in a future version of tox.

This is an attempt at making and using a plugin `tox-appveyor` to select 32-bit or 64-bit Python builds on Appveyor.


takluyver:
That seems to be working. The Appveyor tests take a long, long time when they can't use the cached build of HDF5.

I saw some warnings about missing executables on Travis, and figured out that tox was trying to run every configured Python interpreter to interpolate its site-packages dir into the config. So the later changes are working around that.

takluyver:
@aragilar - it looks like you've done most of the work for Appveyor. Could I ask you to review this? The motivation is that I think warnings we're currently triggering in tox are due to turn into errors in a future version.

aragilar:
This looks so much nicer now :) and yeah, appveyor is very slow. Thanks for chasing this up.

PR #1199, opened 2019-04-02
Create LICENSE

claweyenuk:
Added license from http://docs.h5py.org/en/stable/licenses.html

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,5 +1,4 @@
 Copyright (c) 2008 Andrew Collette and contributors
-http://h5py.alfven.org
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
--- a/LICENSE
+++ b/LICENSE
@@ -6,23 +6,23 @@ Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions are
 met:
 
-a. Redistributions of source code must retain the above copyright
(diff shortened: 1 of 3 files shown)


takluyver:
Thanks! Can we use the canonical format of the BSD 3-clause license (which for example uses 1, 2, 3 rather than a, b, c): https://choosealicense.com/licenses/bsd-3-clause/

claweyenuk:
Done (there were some other minor changes I copied from that license too (copyright owner->copyright holder, and the like)

tacaswell:
I prefer that @andrewcollette sign off on this as it is (trivially) changing the text of the license.

andrewcollette:
I don’t have a problem with it!  I would suggest, though, that the URL in the license be removed or pointed to h5py.org.

claweyenuk:
Sure, it's a quick change, and I even have the terminal open still where I made the changes.  Any preference between the different URLs?

https://www.h5py.org/
http://www.h5py.org/
http://h5py.org/

(I notice https://h5py.org/ doesn't work which I've seen with github sites before I think b/c SSL certs not matching)

andrewcollette:
Actually I’d say just remove it at this point since those details are in flux.

takluyver:
Thanks!

PR #1200, opened 2019-04-04
Low-level wrappers for new HDF5 chunk query API

ajelenak-thg:
HDF5 v1.10.5 introduced several new functions for reporting file location (offset and size) of dataset chunks. In this PR:

1. Low-level h5py wrappers for [`H5Dget_num_chunks()`](https://portal.hdfgroup.org/display/HDF5/H5D_GET_NUM_CHUNKS), [`H5Dget_chunk_info()`](https://portal.hdfgroup.org/display/HDF5/H5D_GET_CHUNK_INFO), and [`H5Dget_chunk_info_by_coord()`](https://portal.hdfgroup.org/display/HDF5/H5D_GET_CHUNK_INFO_BY_COORD).
1. A Cython extension type `h5d.StoreInfo` for holding file location information about one dataset chunk.
1. A convenience low-level function `DatasetID.get_store_info()` that provides file location information for both chunked and contiguous datasets as a list of `StoreInfo` objects.

I thought to start with this PR and then create an issue to plan how to expose this new capability at the Python level.


aparamon:
Generally LGTM, but I'd say `get_store_info()` rather belongs to high-level API (Cython wrappers are usually kept very thin).

takluyver:
Other low-level functions where the C function has multiple out parameters appear to return tuples - e.g. `PropFAID.get_cache()` and `PropFCID.get_version()`. I'd say stick with that pattern, and then discuss adding a high-level ChunkInfo class later.

And I'd agree with @aparamon that the low-level API should stay true to the HDF5 API, not try to add utilities. What if a future HDF5 version adds a different `H5Dget_store_info` function?

tacaswell:
Re-based to account for conflict with the `read_direct_chunk` addition.

tacaswell:
In in person discussion with @takluyver and @aragilar we decided to switch to using a named tuple.  This has the advantages of the custom class (named attribute access) while also giving us iteration, unpacking, and integer slicing while requiring less code!

ajelenak-thg:
@tacaswell Thanks for moving this forward. Should I assume there are no forthcoming changes to this PR? I wouldn't want to overlap with someone else's ongoing work.

ajelenak-thg:
The main change is new `Dataset.store_info()` method as a replacement for `DatasetID.get_store_info()`, as per earlier comments. The new method:

* For a continuous dataset, returns a list with one `StoreInfo` object if byte offset is not `None`, otherwise a `None`.

* For a chunked dataset, returns a list of `StoreInfo` objects for each chunk, or `None` if number of chunks is zero.

* Returns `None` for anything else.

I also added a contiguous dataset test for the new method.

I plan on adding more tests and documentation once we agree on the new method.

takluyver:
I'd suggest sticking to low-level API in this PR, and thinking about high-level APIs as a separate second step, if at all. I'm not convinced that this is the kind of information that we want to expose in the high-level API at all.

The low-level is relatively straightforward because we can largely follow the design decision of HDF5 itself, whereas with the high-level API there's a lot more scope for discussion - what should be exposed in the high-level, how do we ensure it's pythonic and clear while still accurately representing the intention of HDF5.

ajelenak-thg:
Your suggestion does have merit given the PR's title! 😄 How to go from your suggestion to the final decision? The low-level API seems basically done, just some clean up, so this PR could be merged soon.

takluyver:
Can you disentangle the high-level changes from this PR? I.e. make it so it doesn't change anything in `h5py/_hl/`? Then it should, I think, be relatively easy to review the low-level changes. And then maybe open an issue to discuss what, if anything, should be exposed in the high-level API.

ajelenak-thg:
@takluyver Ready for review.

h5py/h5d.pyx
```diff
@@ -497,3 +503,92 @@ cdef class DatasetID(ObjectID):
...
+            chunk_offset = <hsize_t*>emalloc(sizeof(hsize_t) * rank)
+            H5Dget_chunk_info(self.id, space_id, index, chunk_offset,
+                              &filter_mask, &byte_offset, &size)
+            cdef tuple cot = convert_dims(chunk_offset, <hsize_t>rank)
+            efree(chunk_offset)
+
+            if space is None:
+                H5Sclose(space_id)
+
+            return StoreInfo(index,
```
takluyver: I think we should drop `index` from the return value. For `get_chunk_info` it's an input parameter, so the caller already knows it, and for `get_chunk_info_by_coord` it's not set.

h5py/h5d.pyx
```diff
@@ -497,3 +503,92 @@ cdef class DatasetID(ObjectID):
...
+                             byte_offset if byte_offset != HADDR_UNDEF else None,
+                             size)
+
+
+        @with_phil
+        def get_chunk_info_by_coord(self, tuple chunk_offset not None):
+            """ (TUPLE chunk_offset) => StoreInfo
+
+            Retrieve information about a chunk specified by the array
+            address of one of its dataset elements.
```
takluyver: The documentation seems to say that the coordinate given needs to be the first in the dataset, not any.

h5py/h5d.pyx
```diff
@@ -497,3 +503,92 @@ cdef class DatasetID(ObjectID):
...
+
+            space_id = H5Dget_space(self.id)
+            rank = H5Sget_simple_extent_ndims(space_id)
+            H5Sclose(space_id)
+            co = <hsize_t*>emalloc(sizeof(hsize_t) * rank)
+            convert_tuple(chunk_offset, co, rank)
+            H5Dget_chunk_info_by_coord(self.id, co,
+                                       &filter_mask, &byte_offset,
+                                       &size)
+            cdef tuple cot = convert_dims(co, <hsize_t>rank)
```
takluyver: The offset is only [documented](https://confluence.hdfgroup.org/display/HDF5/H5D_GET_CHUNK_INFO_BY_COORD) as an input parameter, and looking at the HDF5 code, I think that's correct. So converting the array back to a tuple is redundant - you can just use `chunk_offset` as passed in.

scopatz:
Thanks @ajelenak-thg - please also add a news entry for this: https://regro.github.io/rever-docs/news

h5py/h5d.pyx
ajelenak-thg: Done.

h5py/h5d.pyx
ajelenak-thg: Fixed.

h5py/h5d.pyx
ajelenak-thg: Keeping `index` in the same place could help simplify the code that uses chunk storage info.

h5py/h5d.pyx
takluyver: Is there a push still to come? I was suggesting that the docstring should be tweaked slightly.

h5py/h5d.pyx
takluyver: I'm not sure what you mean.

In case it wasn't clear, I'm suggesting that we remove `index` from the `StoreInfo` named tuple. It's only used as a return value from these two methods. One of them doesn't have a value for it, and for the other it's an input to the method. So it seems more consistent not to return it.

takluyver:
Also, don't forget to add a news entry, as @scopatz suggested.

h5py/h5d.pyx
ajelenak-thg: My bad. I somehow managed to lose the new text when merging with master.

h5py/h5d.pyx
ajelenak-thg: Sorry for being terse in my reply.  Is there any damage keeping `index` in `StoreInfo`? You are correct that it is not vital but it will stay with all the other chunk storage information. Bear in mind, also, that the chunk storage API does not seem final since there is a placeholder for a dataspace which is currently unused.

takluyver: It seems like an unnecessary complication. If you get a `StoreInfo` tuple from an unknown source, you can't assume it has a useful index, because `get_chunk_info_by_coord` doesn't return one. So you effectively have two types of `StoreInfo`, one with an index, one without. And any code that gets the *with index* kind has passed the same index in, so it's not necessary.

If this was already in a released version, it would be a pain to change. But it's not, so we can iron out this kind of API wart before it's in a release.

h5py/h5d.pyx
takluyver: Thanks, the new wording looks good.

h5py/h5d.pyx
ajelenak-thg: Okay, removed `index`.

takluyver:
Thanks @ajelenak-thg for your patience on this. :-)

Issue #965, opened 2017-12-20
MPI hang during collective writes due to erroneous elision of ranks with nothing to write


chissg:
* Operating System (e.g. Windows 10, MacOS 10.11, Ubuntu 16.04.2 LTS, CentOS 7)
CentOS 7-ish (Scientific Linux 7.3)
 * Python version (e.g. 2.7, 3.5)
2.7.14
 * Where Python was acquired (e.g. system Python on MacOS or Linux, Anaconda on
   Windows)
Institutional build (https://scisoft.fnal.gov/packages/python/)
 * h5py version (e.g. 2.6)
2.7.1
* numpy version
1.13.3
 * HDF5 version (e.g. 1.8.17)
develop@4d4eb4cd5c (verified with 1.10.1)
 * The full traceback/stack trace shown (if it appears)
N/A

In current versions of HDF5, collective writes are an available feature, but there is no really compelling reason to use them. They are faster than the equivalent set of individual writes, but the work required to ensure collectivity is often not worth the trouble. With the upcoming release though, HDF5 offers the ability to use filters on parallel writes to a datasets, but only if those writes are collective. This means that, in the event a particular rank has nothing to write, _the HDF5 write command must still be made_. Unfortunately, as shown by the attached python demonstration program, h5py appears to elide NOP writes, thus causing an MPI-related hang during the collective operation.

Correspondence with an HDF5 dev (attached) assures that the underlying HDF5 system is perfectly capable of dealing with NOP writes as part of a collective operation, but they do need to be passed through.

I'd be grateful if there were some way to enhance h5py to no longer elide NOP collective write operations.

To demonstrate the problem:
```
    mpiexec -np 4 ./demonstrate_h5py_cw_issue.py test.hdf5 4 # Works
    mpiexec -np 4 ./demonstrate_h5py_cw_issue.py test.hdf5 3 # Hangs
```

```python
#!/usr/bin/env python
"""Demonstrate issue with collective writes when at least one rank has nothing to do."""

from __future__ import print_function

import h5py
import numpy as np
import string
import sys
import re
import os
import argparse

try:
    from mpi4py import MPI
    n_ranks = MPI.COMM_WORLD.size
    my_rank = MPI.COMM_WORLD.rank
    WANT_MPI = os.environ["WANT_MPI"] if "WANT_MPI" in os.environ else (n_ranks > 1)
except ImportError:
    MPI = None
    n_ranks = 1
    my_rank = 0
    WANT_MPI = False

[...]

Issue #1067, opened 2018-07-05
slowdown and 100% CPU usage for simple chunked file writing operation


Leberwurscht:
Hi,

I was trying to use h5py's chunked storage and unfortunately ran into performance issues. What I want to achieve is to efficiently write data to a 600x600x600 complex array (3.5 GB) in chunks of 50x50x50 (2 MB).

I execute the following code:

```python
from __future__ import division
import numpy as np
import time, subprocess, os

import h5py
datashape = (600,600,600)

f = h5py.File("test.h5", 'w')
test = f.create_dataset('data', datashape, np.complex, chunks=(50,50,50))

subprocess.call("sync")

t = time.time()
for i0 in xrange(0,600,50):
  for i1 in xrange(0,600,50):
    for i2 in xrange(0,600,50):
      print i0,i1,i2
      test[i0:i0+50,i1:i1+50,i2:i2+50] = 2.+0j
      # same problem if loop is avoided and test[...]=2.+0j is used
      # (loop used for better comparison with PyTables code below)
f.close()
subprocess.call("sync")
print "writing done", time.time()-t
os.remove("test.h5")
```

I expect a I/O-dominated operation, but get 100% CPU usage and the writing takes 4 minutes on a SSD drive, while writing an zero-filled file of the same size takes 35 s:

```
sync; time bash -c "dd if=/dev/zero of=test.h5 bs=2M count=1650 ; sync"
```

Even when syncing after each chunk (I'm not sure if h5py is doing this), it only takes <2min, and does not use 100% CPU:

```
sync; time bash -c "for i in \`seq 1 1650\`; do dd if=/dev/zero of=test.h5 bs=2M count=1; sync; done"
```

PyTables does not have this performance problem -- the following code runs in 36 s:

```python
from __future__ import division
import numpy as np

import time, subprocess, os
import tables

datashape = (600,600,600)

f = tables.open_file("test.h5", 'w')
test = f.create_carray(f.root, 'data', tables.ComplexAtom(16), datashape, chunkshape=(50,50,50))
subprocess.call("sync")

t = time.time()
for i0 in xrange(0,600,50):
  for i1 in xrange(0,600,50):
    for i2 in xrange(0,600,50):
      print i0,i1,i2, 50*50*50*128/8/1e6, "MB"
      test[i0:i0+50,i1:i1+50,i2:i2+50] = 2.
      # test[...] = 2. would consume huge amount of RAM
      if 0:
        f.flush()
        subprocess.call("sync")

[...]

PR #1206, opened 2019-04-13
Fix for zero length collective IO (#965)

jrs65:
Collective IO where one rank attempts to read or write a zero length slice causes the a hang in an underlying collective call (issue #965). This bit me recently, so I thought I'd attempt to fix it. This PR ensures that all ranks participate in any collective IO operation even if some operations are on zero length slices. This also includes some basic unit testing for collective IO that tests for this problem.

I've attempted to limits the effects of this fix to only happen when MPI collective mode is enabled, to do this I needed to make a slight change to the API of `Selection.broadcast` to add a parameter which tells it to return zero length selections when in collective mode.

Anyway, please let me know what other work this needs. Thanks!

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -552,7 +552,14 @@ class Dataset(HLObject):
         # Perform the dataspace selection.
         selection = sel.select(self.shape, args, dsid=self.id)
 
-        if selection.nselect == 0:
+        # If we are running in MPI mode we need to check if we are set in
+        # collective IO mode to ensure we actually perform all IO modes
+        if MPI:
+            is_collective = self._dxpl.get_dxpl_mpio() == h5fd.MPIO_COLLECTIVE
+        else:
+            is_collective = False
+
+        if selection.nselect == 0 and not is_collective:
             return numpy.ndarray(selection.mshape, dtype=new_dtype)
 
         # Up-converting to (1,) so that numpy.ndarray correctly creates
@@ -684,7 +691,14 @@ class Dataset(HLObject):
         # Perform the dataspace selection
         selection = sel.select(self.shape, args, dsid=self.id)
 
-        if selection.nselect == 0:
+        # If we are running in MPI mode we need to check if we are set in
+        # collective IO mode to ensure we actually perform all IO modes
+        if MPI:
+            is_collective = self._dxpl.get_dxpl_mpio() == h5fd.MPIO_COLLECTIVE
+        else:
+            is_collective = False
+
+        if selection.nselect == 0 and not is_collective:
             return
 
         # Broadcast scalars if necessary.
@@ -699,12 +713,13 @@ class Dataset(HLObject):
         # Perform the write, with broadcasting
         # Be careful to pad memory shape with ones to avoid HDF5 chunking
         # glitch, which kicks in for mismatched memory/file selections
+        # (note that zero length dimensions are given zero in collective mode)
         if len(mshape) < len(self.shape):
-            mshape_pad = (1,)*(len(self.shape)-len(mshape)) + mshape
+            mshape_pad = tuple([1 if l > 0 else 0 for l in selection.mshape[:-len(mshape)]]) + mshape
         else:
             mshape_pad = mshape
         mspace = h5s.create_simple(mshape_pad, (h5s.UNLIMITED,)*len(mshape_pad))
-        for fspace in selection.broadcast(mshape):
+        for fspace in selection.broadcast(mshape, collective=is_collective):
             self.id.write(mspace, fspace, val, mtype, dxpl=self._dxpl)
 
     def read_direct(self, dest, source_sel=None, dest_sel=None):
--- a/h5py/_hl/selections.py
+++ b/h5py/_hl/selections.py
@@ -175,8 +175,12 @@ class Selection(object):
         """ Shape of selection (always 1-D for this class) """
         return (self.nselect,)
 
-    def broadcast(self, target_shape):
-        """ Get an iterable for broadcasting """
+    def broadcast(self, target_shape, collective=False):
+        """ Get an iterable for broadcasting.
+
+        Setting `collective` includes zero length selections that are
+        required for collective operations.
+        """
         if np.product(target_shape) != self.nselect:
             raise TypeError("Broadcasting is not supported for point-wise selections")
         yield self._id
@@ -269,7 +273,7 @@ class SimpleSelection(Selection):
         return self
 
 
-    def broadcast(self, target_shape):
+    def broadcast(self, target_shape, collective=False):
         """ Return an iterator over target dataspaces for broadcasting.
 
         Follows the standard NumPy broadcasting rules against the current
@@ -309,7 +313,7 @@ class SimpleSelection(Selection):
         chunks = tuple(x//y for x, y in zip(count, tshape))
         nchunks = int(np.product(chunks))
 
-        if nchunks == 1:
+        if nchunks == 1 or (nchunks == 0 and collective):
             yield self._id
         else:
             sid = self._id.copy()
@@ -413,7 +417,7 @@ class FancySelection(Selection):
 
         self._mshape = tuple(x for x in mshape if x >= 0)
 
-    def broadcast(self, target_shape):
+    def broadcast(self, target_shape, collective=False):
         if not target_shape == self.mshape:
             raise TypeError("Broadcasting is not supported for complex selections")
         yield self._id
--- a/h5py/tests/hl/test_mpi_collective.py
+++ b/h5py/tests/hl/test_mpi_collective.py
@@ -1,0 +1,132 @@
+# This file is part of h5py, a Python interface to the HDF5 library.
+#
+# http://www.h5py.org
+#
+# Copyright 2008-2013 Andrew Collette and contributors
+#
+# License:  Standard 3-clause BSD; see "license.txt" for full license terms
+#           and contributor agreement.
+
+"""
+    Tests the h5py collective IO
+
+"""
+from __future__ import absolute_import
+
+import numpy as np
+import h5py
+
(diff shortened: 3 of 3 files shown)


takluyver:
This looks OK to me, but as I don't use MPI, I'd like someone who does to review it.

Notes:

- Could you clarify the comments a bit about why it needs to behave differently when doing collective operations? I think I follow it from your PR description, but I don't think I'd understand it if I was just looking at the code in a few months time.
  - Possibly also mention why we're not doing the same thing in all cases - presumably it's a performance optimisation. Have we actually measured how much it saves?
- I'd be inclined to rename the parameters when calling into the select module, something like `include_empty`. This means you can understand that code without knowing about MPI, even if you can't see *why* the parameter is there.
- Are the collective I/O tests run in any of the many CI jobs? If not, is it practical to do so? When CI runs most of the tests, it's easy to overlook the ones that it doesn't.

jrs65:
Thanks @takluyver.

That all makes sense. I will add some more inline comments to explain exactly why this needs to be done, and I'll update the parameter names.

As for why this was being done in the first place, I'm not sure. Presumably as a performance optimisation, although whether it's a worthwhile one I don't know. For this PR I wanted to keep it minimal, it seemed like it could be a subtly breaking change for a much larger pool of users to now start doing the HDF5 calls for empty selections.

Finally on the CI server front, I don't see why it couldn't be run the CI server. You just need to call your test runner for that test only with `mpirun`. I found the CI config a little impenetrable though. I'll try and look back through it again, but if you have any pointers that would be appreciated.

takluyver:
The CI is a bit complex. It's all controlled through [tox](https://tox.readthedocs.io/en/latest/), which is configured by the `tox.ini` file. That defines a number of named 'environments', i.e. test jobs. Each environment has dependencies to be installed and commands to run.

Then the Travis and Appveyor config files largely just specify which tox environment to run in each job (plus some other bits and pieces to ensure it can build against HDF5 properly.

There's an `mpi4py` factor already in the tox config which includes mpi4py in the environment. But there's not currently anything to run with mpirun, so it would need a new factor for that.

Alternatively, if there are only going to be a few tests that need this, it might be simpler for the test code itself to start mpirun as a subprocess, run a script, and assert on the expected results from the script. This could also make it practical to turn those hangs into failures, by imposing a (generous) time-limit on waiting for the subprocess.

h5py/_hl/dataset.py
```diff
@@ -552,7 +552,14 @@ def __getitem__(self, args):
         # Perform the dataspace selection.
         selection = sel.select(self.shape, args, dsid=self.id)
 
-        if selection.nselect == 0:
+        # If we are running in MPI mode we need to check if we are set in
+        # collective IO mode to ensure we actually perform all IO modes
```
takluyver: I think the comment could use a bit of extra explanation, something like:

> ...so that it doesn't get stuck waiting for an I/O operation that some workers have skipped

(IDK if 'workers' is the standard term in MPI, but I trust you'll know this)

h5py/_hl/selections.py
```diff
@@ -175,8 +175,12 @@ def mshape(self):
         """ Shape of selection (always 1-D for this class) """
         return (self.nselect,)
 
-    def broadcast(self, target_shape):
-        """ Get an iterable for broadcasting """
+    def broadcast(self, target_shape, collective=False):
+        """ Get an iterable for broadcasting.
+
+        Setting `collective` includes zero length selections that are
+        required for collective operations.
```
takluyver: Actually, looking at it some more, does the selections code need to know about collective mode at all? Shouldn't it handle empty selections correctly regardless?

I appreciate the desire to not change anything for normal I/O. But if we can avoid special-casing it as far as possible, the code is cleaner and better tested.

h5py/_hl/dataset.py
jrs65: Thanks. Worker would make sense, but I think "MPI process" is most clear, so I'll use that.

h5py/_hl/selections.py
jrs65: I agree that it would be best to avoid special cases. If you're amenable to that I'll try and go that route. Hopefully the tests will illuminate any unintended problems.

takluyver: Oh, I see. The obvious thing for a generator over an empty selection to do is to yield nothing. But that won't work in collective mode.

What happens if the different workers get different (non-zero) numbers of chunks from `.broadcast()`? Does that cause a hang because they're doing different numbers of HDF5 writes?

I feel like I'm missing something here. It shouldn't need this much code to be aware that it's doing something collectively. Maybe the answer is just to block broadcasting for collective I/O, so you have to set up the right shape of data yourself.

jrs65: > What happens if the different workers get different (non-zero) numbers of chunks from `.broadcast()`? Does that cause a hang because they're doing different numbers of HDF5 writes?

That is exactly what happens.

Empirically, just changing the clause from
```diff
-        if nchunks == 1 or (nchunks == 0 and collective):
+        if nchunks < 2:
```
does actually work fine. I also think it probably doesn't have any odd side effects because if `nchunks` was zero, then the `nselect == 0` check probably means that normal IO calls don't get there.

jrs65: @takluyver I'm starting to agree with your suggestion that we should forbid broadcasting in collective mode, or at least strongly restrict it. In theory you might want to do something like this:
```
arr = np.ones(10)
dset = fh.create_dataset('test', shape=(3, 10))
with dset.collective:
    if comm.rank == 0:
        dset[:2, :10] = arr   # 2 chunks, yields and writes twice on MPI rank=0
    else:
        dset[2:, :10] = arr  # 1 chunk, yields and writes once on MPI rank=1
```
which even with either of the changes I've tried would hang, because there's fundamentally a different number of chunks needed to be written by each. I think there's two options here:

1. Disallow broadcasting when doing MPI collective writes.
2. Before writing, communicate with all other ranks to determine how many chunks will be written, and then insert empty writes to keep the number the same across all ranks.

The latter seems like a bit of a hack.

takluyver: What happens if one MPI process throws an error? Because if we tried to disallow broadcasting, at least in the obvious way, I suspect rank 0 in your example would throw an error, while rank 1 would do its 1 chunk write and then wait for a matching write in rank 0 which isn't going to happen.

This is ugly. I'm trying to think about how best to deal with it, but I'm not sure I know enough about MPI.

I also notice that we should update the docs, because the section on collective I/O doesn't even mention with `with dset.collective` context manager:
http://docs.h5py.org/en/stable/mpi.html#collective-versus-independent-operations

jrs65: In the way I was thinking of implementing this, I don't think it would break. You probably want to insist that each ranks file and memory shapes match exactly. So rank=1 wouldn't work here because it tries to write a shape=(10,) array into a shape=(2,10) selection.

However, despite that I'd probably have each rank do it's preparation for writing, and then just before writing have all ranks communicate if they have any errors and have then *all* throw an exception about the need to broadcast. Something like:
```python

local_mpi_broadcast_error = (memshape != selshape)

# Reduce to see if *any* rank had an error
global_mpi_bcast_error = self.comm.allreduce(local_mpi_broadcast_error, op=MPI.MAX)

if local_mpi_broadcast_error:
    raise ValueError("Cannot broadcast in MPI collective mode (mem shape=%s, file selection=%s) % ....)

if global_broadcast_error:
     raise ValueError("Yikes. Another rank tried to broadcast during a write.")

# ... actually do the write now
```

A bit like that anyway. Let me change it again and send in a PR.

takluyver: Thinking about it more, I think we can do it well enough with a simple check, without exchanging messages, which means less to go wrong. It will always be possible to create a hang if you branch on `if comm.rank ...`, but we want to make it obvious that things like this are a problem:

```python
dset = fh.create_dataset('test', shape=(10,))
chunk_start = int(comm.rank / comm.size)
chunk_end = int((comm.rank + 1) / comm.size)
with dset.collective:
    dset[chunk_start:chunk_end] = comm.rank
```

At the moment, that will work (inefficiently: #1067) for 4 workers, but fail for 8, because some workers will write 12 values and some 13. But if we forced people to do the broadcasting with numpy beforehand, it would work:

```python
chunk = np.full(chunk_end - chunk_start, comm.rank)
with dset.collective:
    dset[chunk_start:chunk_end] = comm.rank
```

I think it needs to be a fairly strict check: not just that broadcast would only return one chunk, but that the dimensions of the data match those of the selection, even when those are length-1 dimensions.

takluyver:
I've had a go at implementing what I was thinking of. We'll still need to work out testing before this can be merged, though.

aragilar:
I had a look at this (in terms of testing mpi), and I'm not sure we're actually testing the codebase under MPI. I've kinda got something working (I'll push a PR once I've sorted out a few things).

@jrs65 Did you have the `test_collective_write_empty_rank` working, and on which commit id was it?

scopatz:
Is [this](https://travis-ci.org/h5py/h5py/jobs/528138387) not testing with mpi?

takluyver:
I think it only tests building with MPI enabled; when I looked before I didn't find anything using mpirun in the tests.

aragilar:
At best, travis is currently running a MPI build of HDF5 with a MPI build of h5py (HDF5_MPI isn't passed through by tox currently, so it's possible using CC=mpicc is enough). I'm not sure which tests are being skipped though, but I've nearly got a PR ready which will show which tests are being skipped (passing -rsxX to pytest), and runs the tests under mpirun (which showed that our MPI tests don't work under mpirun with more than one process). I'll try to finish the PR tonight.

scopatz:
Thanks @aragilar!

aragilar:
I think this needs rebasing to test with the latest MPI support (not sure if there's anything else holding this up)?

Issue #830, opened 2017-01-20
dimension scale creation method is in wrong (confusing) location


equaeghe:
Currently, dimension scales are created as `dataset.dims.create_scale(scale_dataset, 'scale name')`, i.e., using the `create_scale` method of the `dims` attribute of a `Dataset`. However, as far as I can tell, dimension scale creation only acts on `scale_dataset` and has nothing to do with `dataset`, so it would be more logical to have a `create_scale` method (which can take a scale name as an argument) on `Dataset` or perhaps a static method that takes a scale dataset (and a scale name) as an argument.

The current practice is confusing, as it may lead people to believe that the scale is created specifically for `dataset`, which is not the case, I think. (At least I was confused.)

The change I propose does not need to lead to an API incompatibility: the new method could be added and the old one kept (perhaps in some way deprecated or advised against).

tacaswell:
This looks like it could be converted into a static method on the `dims` proxy, however it is using an instance method to encode the name which _may_ pull information from the underlying object (but it looks like this code path does not).

Having the `create_scale` method on the `dims` proxy makes a fair amount of sense to me, it provides a name space so we don't have too many loose functions floating around.

At a minimum it should be better documented.

takluyver:
I agree with @equaeghe that it's somewhat confusing to have this method associated with an uninvolved dataset.

Instead of this:

```python
f['data'].dims.create_scale(f['x2'], 'x2 name')
```

It would conceptually align with HDF5 better to do something like this:

```python
f['x2'].make_scale('x2 name')
```

Or it could be an option when you create a dataset - this makes the intent clearer to my mind, though it's somewhat less flexible.

```python
f.create_dataset('x2', data=np.arange(10), as_scale='x2 name')
```

rmvanhees:
I looked at it in a different way: I use dimension scales as follows:

row = np.arange(0, 256, dtype='u2')                # define data for my
dimension
ds_row = fid.create_dataset('row', data=row)    # create a dataset
ds_row.dims.create_scale(fid['row'], 'some description')   # turn this
dataset into a dimension
dset = fid.create_dataset('frame', data=frame)  # create a real dataset
dset.dims[0].attach_scale(ds_row)                 # attach one of its
dimensions

I think the problem is in the third line, where I have to use fid['row']
again. Reading the previous emails I realize that I create a
dimension-scale by saying that it is its own dimension scale, which I can
then use it for other datasets. This approach works well and is also
[...]

PR #1212, opened 2019-04-26
Add ds.make_scale() method.

takluyver:
Replaces `other_ds.dims.create_scale(ds)`, which is deprecated here.

Making `ds` into a dimension scale doesn't affect or use the dimensions of `other_ds` at all, so it's confusing to expose the functionality through `other_ds.dims`.

Closes gh-830


scopatz:
Thanks @takluyver!

Issue #782, opened 2016-11-29
Out of order compound dtypes


ngoldbaum:
It looks the way h5py is using numpy's dtype constuctor is causing issues with hdf5 files that have out-of-order dtypes. This is causing issues. See #764, #765, #738, as well as @nevion's comment: https://github.com/h5py/h5py/pull/765#issuecomment-263463634. It's possible this is a numpy bug rather than an h5py bug, but h5py will still need to work around it given the existence of old versions of numpy in the wild.

aragilar:
This should be fixed in the latest master by #793.

nevion:
@aragilar bump for re-eval - for #765 specifically - #793 focuses on generating a correct h5py compound type for a given numpy dtype, but #765 focused on generating a correct dtype for a given h5py - they're complementary.  I just want to make sure, as it seemed that we caught numpy's dtype constructor barfling on out of order fields also, that we have both cases covered and tested working.

aragilar:
@nevion I've created #802 which is an implementation of the test you suggested (I thought that was already added in a previous request, but apparently not), is there anything else you think isn't covered by #793 or #802?

nevion:
one more thing before you close this out, in #802, can you make it an unpacked struct or test one of those in addition?

aragilar:
@nevion I'm confused how you can have an unpacked struct (which I understand as a struct whose elements are aligned based on CPU arch/type size, as opposed to packed where the elements are next to each other without padding, I fully accept this may be superficial and/or wrong) which is out-of-order, given that to have the struct be out-of-order you must define offsets, which seems the antithesis of unpacked struct. Maybe an example would clear up my confusion? In any case, I'm going to reopen this until we've finished and merged #802.

nevion:
Consider a couple cases (and the potential for real world mixing of them):

- a user who reorders data fields inside a struct, after writing compound type wrapper code.  That is the offsets change later on.  The API does not prohibit this and it's not a bad thing at all to preserve the order of how things were appended to the compound type's field list, but it might trip some people up, who expected monotonic offsets.  This case would frequently happen in refactors or insertion of new fields.
- a user who exports only a subset of struct fields
```
struct SensedPacket{
  int acc, usr1, usr2, diag, temp, val;
}
```
And the user only wants to export acc, and diag - for instance (forgoing compound type creation)


[...]

PR #1214, opened 2019-04-28
add struct/compound type robustness test

nevion:
as discussed in https://github.com/h5py/h5py/issues/782 , here is a unit test which might suss out past and future regressions in numpy or h5py's struct handling.  At present it detects nothing on h5py HEAD and numpy 1.16.2

It adds a multi-issue targeted unit test based on my own experiences that either h5py or numpy have faulted in handling when writing or loading compound/struct types to hdf5 files through h5py, see https://github.com/h5py/h5py/issues/782#issuecomment-267140511

* gaps - non standard layout/padding (common w/ sensor IO, partial exports of data)
* padded end - end of type != lastfield_offset + lastfield_itemsize
* out of order field offsets (in the same dtype)

### DIFF
--- a/h5py/tests/hl/test_datatype.py
+++ b/h5py/tests/hl/test_datatype.py
@@ -196,6 +196,64 @@ class TestOffsets(TestCase):
         with h5py.File(fname, 'r') as f:
             self.assertArrayEqual(f['data'], data)
 
+
+    def test_compound_robustness(self):
+        #make an out of order compound type with gaps in it, and larger itemsize than minimum
+        # Idea is to be robust to type descriptions we *could* get out of HDF5 files, from custom descriptions
+        # of types in addition to numpy's flakey history on unaligned fields with non-standard or padded layouts.
+        fields = [
+            ('f0', np.float64, 25),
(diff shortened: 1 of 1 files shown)


h5py/tests/hl/test_datatype.py
```diff
@@ -196,6 +196,64 @@ def test_aligned_data(self):
...
+        self.assertArrayEqual(data['f2'], f2)
+
+        fname = self.mktemp()
+
+        with h5py.File(fname, 'w') as fd:
+            fd.create_dataset('data', data=data)
+
+        with h5py.File(fname, 'r') as fd:
+            readback = fd['data']
+            self.assertTrue( readback.dtype == dt )
```
takluyver: I dislike `self.assertTrue` when the `assert` statement would do instead. Additionally, in this case, pytest can give smarter error message if the assertion fails.

takluyver: On reflection,  I see this is already used in the code, and this PR doesn't need to clean up our test style. Thanks!

PR #746, opened 2016-08-07
Enh add extra pt exposure function

tacaswell:
There are a couple more of these coming.

PR #1224, opened 2019-06-03
Enh add extra pt exposurec (again)

scopatz:
This builds off of #746, resolving a trivial conflict.


takluyver:
> "This needs a test" - @tacaswell

- @scopatz on #746

In fact, I'd be OK if we just decide to merge it without a test, since it's simply exposing a function from HDF5. But I couldn't resist the opportunity to do that :slightly_smiling_face: . And it would be good to have a test if someone can see a good way to write one.

This also ties into discussion on #1166 about how best to test functions which modify global state. I attempted to summarise the options in [this comment](https://github.com/h5py/h5py/pull/1166#issuecomment-487009390).

scopatz:
Right, yeah.  I'll take a look tomorrow

scopatz:
> But I couldn't resist the opportunity to do that :slightly_smiling_face: 

haha past me is the worst me, but in this case it was good, because there was a bug in the code as written. I have added a test for this and a written a test framework (slightly hacky) for testing in a subprocess.

scopatz:
> But I couldn't resist the opportunity to do that :slightly_smiling_face: 

haha past me is the worst me, but in this case it was good, because there was a bug in the code as written. I have added a test for this and a written a test framework (slightly hacky) for testing in a subprocess.

h5py/tests/common.py
```diff
@@ -165,3 +167,32 @@ def closed_tempfile(suffix='', text=None):
...
+            insub = insub.replace("/", "_")
+        defined = os.environ.get(insub, None)
+        # fork process
+        if defined:
+            return f(*args, **kwargs)
+        else:
+            os.environ[insub] = '1'
+            with closed_tempfile() as stdout:
+                with open(stdout, 'w+t') as fh:
+                    rtn = subprocess.call([sys.argv[0], curr_test],
```
takluyver: `argv[0]` is `'.../pytest'` here? Should we be using `[sys.executable, '-m', 'pytest', ...]` to ensure it's run on the same Python interpreter?

h5py/tests/common.py
```diff
@@ -165,3 +167,32 @@ def closed_tempfile(suffix='', text=None):
...
+    @wraps(f)
+    def wrapper(*args, **kwargs):
+        curr_test = os.environ.get('PYTEST_CURRENT_TEST', None)
+        if not curr_test:
+            raise ut.SkipTest("not running in pytest")
+        curr_test, _, _ = curr_test.partition(' ')
+        # get block around test name
+        insub = "IN_SUBPROCESS_" + curr_test
+        for c in "/\\,:.":
+            insub = insub.replace("/", "_")
```
takluyver: I'm guessing `"/"` should be `c`?

h5py/tests/common.py
```diff
@@ -165,3 +167,32 @@ def closed_tempfile(suffix='', text=None):
             test_file.flush()
     yield file_name
     shutil.rmtree(file_name, ignore_errors=True)
+
+
+def insubprocess(f):
+    """Runs a test in its own subprocess"""
+    @wraps(f)
+    def wrapper(*args, **kwargs):
+        curr_test = os.environ.get('PYTEST_CURRENT_TEST', None)
```
takluyver: This looks convenient, but the pytest docs explicitly say you shouldn't rely on the format:

> The contents of PYTEST_CURRENT_TEST is meant to be human readable and the actual format can be changed between releases (even bug fixes) so it shouldn’t be relied on for scripting or automation.

https://docs.pytest.org/en/latest/example/simple.html#pytest-current-test-env

It looks like the request object might be provide a way to get the necessary information: https://stackoverflow.com/a/34732269/434217 . I don't know how easily you can get that in a decorator.

h5py/tests/common.py
scopatz: yeah that seems reasonable

h5py/tests/common.py
scopatz: Ahh yes, good catch.

h5py/tests/common.py
scopatz: updated!

h5py/tests/common.py
```diff
@@ -165,3 +167,29 @@ def closed_tempfile(suffix='', text=None):
             test_file.flush()
     yield file_name
     shutil.rmtree(file_name, ignore_errors=True)
+
+
+def insubprocess(f):
+    """Runs a test in its own subprocess"""
+    @wraps(f)
+    def wrapper(request, *args, **kwargs):
+        curr_test = request.node.name
```
takluyver: It looks like this only gets the function name, unfortunately, not the module name. We could try something like `request.node.function.__module__` to get the module name. Or use just the function name with the `-k` option, which would rely on there being only one test with that name across all files.

scopatz: let's try the module trick

takluyver:
Thanks!

PR #1225, opened 2019-06-03
Fix test failures on Appveyor

takluyver:
It looks like it's failing on import, but the real error is being obscured by the test setup.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -94,7 +94,7 @@ environment:
 install:
   # We need wheel installed to build wheels
   - "%PYTHON%\\python.exe -m pip install --upgrade wheel pip setuptools"
-  - "py -3.5 -m pip install --upgrade wheel pip setuptools"
+  - "py -3.5 -m pip install --upgrade wheel pip setuptools virtualenv"
   - "py -3.5 -m pip install requests"
   - "py -3.5 ci\\appveyor\\get_hdf5.py"
   - "py -3.5 -m pip install tox codecov tox-appveyor"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -109,7 +109,7 @@ test_script:
   # Note that you must use the environment variable %PYTHON% to refer to
   # the interpreter you're using - Appveyor does not do anything special
   # to put the Python evrsion you want to use on PATH.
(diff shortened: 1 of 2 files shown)


takluyver:
Well, I was right that it's failing to import. Unfortunately error message doesn't point to an obvious problem beyond that:

```
py36-test-deps run-test: commands[2] | python -c 'import h5py; from h5py.h5py_warnings import H5pyDeprecationWarning'
ModuleNotFoundError: No module named 'h5py'
```

takluyver:
OK, I think we're getting somewhere. Older versions of tox automatically downloaded the newest pip & setuptools into newly created environments. Tox 3.10 disabled that by default, which appears to be when our builds started failing. A couple of other projects appear to have been hit by this on Appveyor (https://github.com/tox-dev/tox/issues/1302, https://github.com/tox-dev/tox/issues/1304).

I'm not entirely sure what happened, but when tox tried to install h5py with the old version of pip included in the old version of virtualenv on appveyor, it appeared to erroneously decide that h5py was already installed:

```
[2916] C:\projects\h5py$ 'C:\projects\h5py\.tox\py36-test-deps\Scripts\python.EXE' -m pip install --exists-action w '.tox\.tmp\package\1\h5py-2.9.0.post0.zip'
Processing c:\projects\h5py\.tox\.tmp\package\1\h5py-2.9.0.post0.zip
  Requirement already satisfied (use --upgrade to upgrade): h5py==2.9.0.post0 from file:///C:/projects/h5py/.tox/.tmp/package/1/h5py-2.9.0.post0.zip in c:\projects\h5py
```

Upgrading virtualenv on Appveyor (from 15.0.1 to 16.6.0) appears to fix this, presumably because the new virtualenv bundles a newer pip.

If the remaining test jobs succeed, I'll rebase this PR to clean up unwanted changes from debugging.

scopatz:
Sounds great!

takluyver:
I've done the rebase. We should obviously wait for the tests to finish, but I think this should do the trick.

takluyver:
Tests passing. I'll self-merge this, so that Appveyor tests are usable again.

PR #1228, opened 2019-06-06
ENH: Add custom tab completions for IPython.

ssanderson:
I often explore hdf5 files interactively in IPython using h5py. Defining [`_ipython_key_completions_`](https://ipython.readthedocs.io/en/stable/config/integrating.html) on `Group` allows IPython to generate tab completions for `__getitem__` lookups on groups.

Assuming this is of interest to others, one question I have is whether this belongs here or on `MappingHDF5`, which is where `keys` is actually defined. The other consumers of MappingHDF5 are DimensionManager and AttributeManager. AttributeManager seems like it would benefit from this. I'm not sure about DimensionManager (I don't think I've interacted with `.dims`).

### DIFF
--- a/h5py/_hl/base.py
+++ b/h5py/_hl/base.py
@@ -416,6 +416,14 @@ class MappingHDF5(Mapping):
             """ Get a view object on member items """
             return ItemsViewHDF5(self)
 
+    def _ipython_key_completions_(self):
+        """ Custom tab completions for __getitem__ in IPython >=5.0. """
+        keys = self.keys()
+        # keys() is lazy in Python 3.
+        if not isinstance(keys, list):
+            keys = list(keys)
+        return keys
+
 
 class MutableMappingHDF5(MappingHDF5, MutableMapping):
 
--- a/h5py/_hl/base.py
+++ b/h5py/_hl/base.py
@@ -418,11 +418,7 @@ class MappingHDF5(Mapping):
 
     def _ipython_key_completions_(self):
         """ Custom tab completions for __getitem__ in IPython >=5.0. """
-        keys = self.keys()
-        # keys() is lazy in Python 3.
-        if not isinstance(keys, list):
-            keys = list(keys)
-        return keys
+        return sorted(self.keys())
 
 
 class MutableMappingHDF5(MappingHDF5, MutableMapping):
--- a/h5py/_hl/group.py
+++ b/h5py/_hl/group.py
@@ -576,13 +576,6 @@ class Group(HLObject, MutableMappingHDF5):
             return r.encode('utf8')
         return r
 
-    def _ipython_key_completions_(self):
-        """ Custom tab completions for __getitem__ in IPython >=5.0."""
-        keys = self.keys()
-        # keys() is lazy in Python 2.
-        if not isinstance(keys, list):
-            return list(keys)
-
 
 class HardLink(object):
 
--- a/h5py/_hl/group.py
+++ b/h5py/_hl/group.py
@@ -576,6 +576,13 @@ class Group(HLObject, MutableMappingHDF5):
             return r.encode('utf8')
         return r
 
+    def _ipython_key_completions_(self):
+        """ Custom tab completions for __getitem__ in IPython >=5.0."""
+        keys = self.keys()
+        # keys() is lazy in Python 2.
+        if not isinstance(keys, list):
+            return list(keys)
+
 
 class HardLink(object):
 
--- a/h5py/tests/hl/test_completions.py
+++ b/h5py/tests/hl/test_completions.py
@@ -1,0 +1,52 @@
+from ..common import TestCase
+
+
+class TestCompletions(TestCase):
+
+    def test_group_completions(self):
+        # Test completions on top-level file.
+        g = self.f.create_group('g')
+        self.f.create_group('h')
+        self.f.create_dataset('data', [1, 2, 3])
+        self.assertEqual(
+            self.f._ipython_key_completions_(),
+            [u'data', u'g', u'h'],
+        )
+
+        self.f.create_group('data2', [1, 2, 3])
+        self.assertEqual(
+            self.f._ipython_key_completions_(),
+            [u'data', u'data2', u'g', u'h'],
+        )
+
+        # Test on subgroup.
+        g.create_dataset('g_data1', [1, 2, 3])
+        g.create_dataset('g_data2', [4, 5, 6])
+        self.assertEqual(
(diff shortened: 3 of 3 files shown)


ssanderson:
@takluyver you might be interested in this as well?

tacaswell:
How does this interact with the code in https://github.com/h5py/h5py/blob/master/h5py/ipy_completer.py ?

scopatz:
Seems fine to me otherwise.

ssanderson:
TIL about `enable_ipython_completer`! 

From a quick look, it seems like the existing completer probably gives nicer completions than this would, at the cost of requiring an explicit opt-in from the user. In particular, the handling of `f['item1/item2/it<tab>` that's supported there would be tricky using `_ipython_key_completions_`, because the interface of that method is that your object has to return the list of all completions available on the current object, which would require a (probably expensive in many cases) recursive walk of the file.

As-is, this only really gives nice behavior for single-level lookups. If you want to traverse multiple levels of the file (which is probably the case where completion is most useful), you have to bind the intermediate results and complete on them, which is suboptimal. We could do better if `_ipython_key_completions_` passed us the key being completed, since then we could rsplit on `/` and list just the keys from the relevant group (which is what the current completer does). From looking at the IPython source, they have the completion text available at the call-site for `_ipython_key_completions_`, so that might not be too bad to get working.

The good news is that if you load the existing extension, that just supercedes this, so this is still a strict upgrade (albeit a less compelling one than I'd like). That might not be worth the cost of having two ways to do completions though.

ssanderson:
So, I think the main question here is, "Is it worth having an *ok* completer that's enabled by default, or should we just have a better completer that's opt-in?" If we decide on the latter, would it be worth pursuing changes in IPython that would allow us to implement the existing completion behavior without requiring the user to opt-in?

h5py/_hl/group.py
```diff
@@ -576,6 +576,13 @@ def __repr__(self):
             return r.encode('utf8')
         return r
 
+    def _ipython_key_completions_(self):
+        """ Custom tab completions for __getitem__ in IPython >=5.0."""
+        keys = self.keys()
+        # keys() is lazy in Python 2.
```
takluyver: You mean Python *3*?

takluyver:
Given how simple this is, and the fact that I rarely load the opt-in completer, I'd say go for it.

I invariably realise I want completions when I'm mid-way through typing a line, and it's usually less effort to finish the line manually than to clear it, figure out the command to enable explicit completion, and re-type the command I wanted.

IIRC, a lot of the messiness around key completion is dealing with object reprs - the difference between `a[foo`, `a['foo`, `a["foo` etc., and we went for the list-of-objects API in part so that it's IPython's responsibility to handle those quirks. But maybe literal string keys are common enough that it's worth special-casing them for an API like `obj._ipython_key_string_completions_(prefix)`, with the eval-ed string prefix passed to h5py.

(Aside: we also gave `h5glance` completions into HDF5 files, integrated with bash and zsh, and it's really nice for complex files)

ssanderson:
> But maybe literal string keys are common enough that it's worth special-casing them for an API like obj._ipython_key_string_completions_(prefix), with the eval-ed string prefix passed to h5py.

What would eval-ed mean in this context? If I wrote something like `group[some_function()]`, would IPython eval `some_function()` to see if it returns a string? That seems a bit scary to have enabled by default. IIRC IPython already has some configuration options around whether it will eval dot-access? Would you imagine this being toggle-able in the same way?

FWIW, I think a reasonable starting point would be if `_ipython_key_completions_` passed the content of the completion text. Then if we wanted to only complete on strings, we could do so by having something like:

```python
def _ipython_key_completions_(self, text):
    if is_start_of_string(text):
        # Traverse to last group in `text` and gets its keys.
        return _get_text_completions(self, text)
    return []
```

`is_start_of_string` is probably slightly tricky to write if you want to properly handle raw and triple-quoted strings, but the alternative right now is that library code doesn't even have the option of writing that tricky code if it wants to. Maybe IPython could provide that function as a utility for clients?

takluyver:
> What would eval-ed mean in this context? If I wrote something like group[some_function()], would IPython eval some_function() to see if it returns a string?

No, sorry, that was unclear. I was thinking only of string literals. So if you typed `obj['bob\'s ke<TAB>`, IPython would try calling something like `obj._ipython_key_string_completions_("bob's ke")` - giving the object the *content* of the incomplete string rather than the literal code. This means IPython would still be responsible for dealing with the string literal syntax (different quoting styles, escaping, etc.), rather than requiring `obj` to deal with that.

takluyver:
Any way, if you want to fix the comment, I think this looks good for the near term.

ssanderson:
> Any way, if you want to fix the comment, I think this looks good for the near term.

Sounds good. Do you have an opinion as to whether `_ipython_key_completions_` belongs on `Group` or on `MappingHDF5`? It tentatively looks to me like the latter might be preferable to get completions on `group.attrs[]`

ssanderson:
Is there a place to put a whatsnew entry for this?

ssanderson:
I also just noticed I wasn't returning in the python 2 case. Going to add a quick test suite for this.

takluyver:
I don't think we have anything set up for gathering whatsnew notes. But as I'm pushing for a release, I might start putting them together.

h5py/_hl/base.py
```diff
@@ -416,6 +416,10 @@ def items(self):
             """ Get a view object on member items """
             return ItemsViewHDF5(self)
 
+    def _ipython_key_completions_(self):
+        """ Custom tab completions for __getitem__ in IPython >=5.0. """
+        return sorted(self.keys())
```
ssanderson: Empirically, these seem to come back sorted already, but I couldn't find a guarantee of that fact. Regardless, we need to force the list in py3, so this seems like a reasonable way to do that either way.

takluyver: It depends on whether you specify `track_order=True` when creating the file/group - if so, they're in creation order. Otherwise, they should be in name order.

ssanderson: Ah, ok. Does it seem reasonable to guarantee that these come back sorted? I'm not sure what IPython's display logic is for these. Do they get sorted anyway downstream?

takluyver: Looking at IPython's completion code, I think it sorts results anyway. But it's fairly convoluted, and I don't think it's an API guarantee, so sorting them seems a reasonable thing to do.

takluyver:
Thanks! One of us should remember to add this to the release notes I started in #1233.

Issue #1181, opened 2019-03-14
document installation on aarch64 + ubuntu


talregev:
When I install:
`pip3 install h5py`

I am using python 3.6 also happen in python 2.7 Ubuntu aarch64 / arm64.
I install:
`sudo apt install libhdf5-dev`

The include files is on `/usr/include/hdf5/serial` and the script lookin in `/usr/include`

the compiler looking for `libhdf5.so` but it is `libhdf5_serial.so`
and it looking for `libhdf5_hl.so` but it `libhdf5_serial_hl.so`

can you change that?

tacaswell:
If you are installing the wheels they should have hdf5 bundled with it.  I am a bit unclear how you are installing h5py.

If you are installing from source, see http://docs.h5py.org/en/stable/build.html#custom-installation for how to point the installation process at where hdf5 is on your system

Can you try with h5py 2.9?

talregev:
I install it:
pip3 install h5py
It install h5py-2.9.0
hdf5 version:
Architecture: arm64
Source: hdf5
Version: 1.10.0-patch1+docs-4
Replaces: libhdf5-serial-dev (<< 1.8.12-9~)

It download this file:


https://files.pythonhosted.org/packages/43/27/a6e7dcb8ae20a4dbf3725321058923fec262b6f7835179d78ccc8d98deec/h5py-2.9.0.tar.gz

tacaswell:
Ah, I miss-read your first message as saying you installed h5py 2.6.

Did it work if you followed the instructions for setting your hdf5 path does it work?

talregev:
for me that it work i do the following:
cp /usr/include/hdf5/serial/* /usr/include/
ln -s /usr/lib/aarch64-linux-gnu/libhdf5_serial.so /usr/lib/aarch64-linux-gnu/libhdf5.so
ln -s /usr/lib/aarch64-linux-gnu/libhdf5_serial_hl.so /usr/lib/aarch64-linux-gnu/libhdf5_hl.so

then it work.

errors:
```bash   
[...]

PR #1231, opened 2019-06-07
Require pkg-config on Unix platforms

takluyver:
We currently require the Python [pkgconfig](https://pypi.org/project/pkgconfig/) wrapper for building from source, but try to fall back to default locations if the underlying `pkg-config` command is not found. But this has caused some confusion (#1181) when the default locations aren't right.

This PR would make the `pkg-config` command required on Unix systems, unless you explicitly specify the path to an HDF5 installation. So rather than failing because it's trying to use an incorrect path, leaving you trying to figure out how to solve that, it should fail with an error that `pkg-config` is missing, which has a more obvious resolution.

The drawback is that in some cases, using the default fallback paths presumably does work. This would break these, requiring the user to install `pkg-config` and the necessary `.pc` file for hdf5. I don't know how common these cases are, or how difficult it might be to resolve them.

Closes gh-1181

### DIFF
--- a/docs/build.rst
+++ b/docs/build.rst
@@ -83,6 +83,9 @@ To install h5py from source, you need three things installed:
 * HDF5 1.8.4 or newer with development headers
 * A C compiler
 
+On Unix platforms, you also need ``pkg-config`` unless you explicitly specify
+a path for HDF5 as described in :ref:`custom_install`.
+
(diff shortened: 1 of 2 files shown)


scopatz:
This seems pretty reasonable to me. Thanks @takluyver!

Issue #1177, opened 2019-02-16
Require Python 3?


takluyver:
Quite a lot of Python projects, including core components like numpy, are [dropping support for Python 2](https://python3statement.org/) in the near future, and some already have. Should h5py join them?

The benefits are probably familiar: development is easier when you're not having to think about compatibility, tests finish quicker because there are fewer platforms to test on, we can take advantage of new language features, and static analysis tools tend to be more informative when they don't have to make sense of compatibility layers. Cython hides some of the differences, but there's a fair bit of highlevel code written in pure Python.

h5py is in a relatively good position for this; the HDF5 file format doesn't change very fast, so if people have existing code written for h5py & Python 2, in many cases they don't need to upgrade to the latest h5py.

A number of projects have also said they will backport important fixes to a 'long term support' release series supporting Python 2 for a while. While I think the term 'support' implies a kind of guarantee that volunteer-maintained projects don't really offer, this plan is a reasonable option to develop new features on Python 3 before completely stopping work on bugs affecting Python 2 users.

tacaswell:
:+1: I am fully on board with this.  I think we should do a major version bump when we do (same as mpl).

What is the oldest version of python to support?  We currently go all the way back to 3.4, I think we should jump to 3.6 (so we get f strings and ordered dicts). 

If we aim for late summer / early fall for a 3.0 release I don't think we have to worry too much about an LTS release (unless we find a truly show-stopping bug in h5py2.9) and just do a 2.9.x final patch release around the same time.

takluyver:
If we're aiming for a major version bump, should we try to get #1143 into a release soon? A major version bump is probably the only chance to change the default file open mode.

If we were releasing today, my instinct is to still support 3.5. I often use Ubuntu releases as a benchmark for when it's reasonable to drop Python versions, and Ubuntu 16.04, still the default at work, has Python 3.5. But of course, we're not releasing today, and anyone using Anaconda gets new Python versions much quicker than Ubuntu defaults.

takluyver:
Just to make that slightly cryptic comment clearer: #1143 deprecates the existing default behaviour and provides an option for people to change it. The idea is that that goes in at least one release (2.10) before the major version bump, so people have a chance to adapt code.

aparamon:
Obsoleting Python 2 is timely; let's go for it!

[...]

Issue #568, opened 2015-04-24
TypeError: No conversion path for dtype: dtype('<M8[us]')


jkhoogland:
Hi,
I was trying out some basic conversion of an yahoo option chain to hdf5 using odo and h5py.

```
from pandas.io.data import Options

aapl = Options('aapl', 'yahoo')
data = aapl.get_all_data()

from odo import odo

odo(data,'aapl.hdf5::/oc')
```

The data loads fine, but when trying to save to the hdf5 file,
it fails with the following error:

```
/Users/Jiri/anaconda/lib/python3.4/site-packages/h5py/_hl/dataset.py in make_new_dset(parent, shape, dtype, data, chunks, compression, shuffle, fletcher32, maxshape, compression_opts, fillvalue, scaleoffset, track_times)
     77         else:
     78             dtype = numpy.dtype(dtype)
---> 79         tid = h5t.py_create(dtype, logical=1)
     80 
     81     # Legacy

h5t.pyx in h5py.h5t.py_create (h5py/h5t.c:13966)()

h5t.pyx in h5py.h5t.py_create (h5py/h5t.c:13504)()

h5t.pyx in h5py.h5t._c_compound (h5py/h5t.c:12883)()

h5t.pyx in h5py.h5t.py_create (h5py/h5t.c:13858)()

TypeError: No conversion path for dtype: dtype('<M8[us]')
```

I might be missing something, but this seems like a bug, 
ie a missing conversion looking at the exception.

Is this a known issue ?
Regards
Jiri 

I m using the following packages

   package                    |            build
    ---------------------------|-------------------
    pytz-2015.2                |           py34_0
    toolz-0.7.1                |           py34_0
    python-dateutil-2.4.2      |           py34_0
    datashape-0.4.4            |       np19py34_1
    pandas-0.16.0              |       np19py34_1
    odo-0.3.1                  |       np19py34_0

tacaswell:
For reference, '<M8[us]' is a datetime64 type specification (http://docs.scipy.org/doc/numpy/reference/arrays.datetime.html#datetime-units) which seem to not be well supported by things that are not pandas.

attn @cpcloud

andrewcollette:
@tacaswell, thanks... I've not seen that dtype before.

[...]

Issue #443, opened 2014-05-09
Support for numpy.datetime64


mrocklin:
What is the status of numpy.datetime64 support in h5py?  Is there a way to make it work?  Is there a suggested contribution within h5py?

Datetimes are pretty important to me.  I was surprised when the following didn't work.

``` Python
In [1]: import h5py

In [2]: import numpy as np

In [3]: f = h5py.File('foo.hdfs', 'w')

In [4]: d = f.create_dataset('data', shape=(2, 2), dtype=np.datetime64)
TypeError: No conversion path for dtype: dtype('<M8')
```

If there is a simple answer to my question then I encourage answers on an accompanying stack-overflow post http://stackoverflow.com/questions/23570632/store-datetimes-in-hdf5-with-h5py

PR #1232, opened 2019-06-10
Add ability to register/deregister NumPy dtypes.

hameerabbasi:
Adds `datetime64` support to `h5py`.


hameerabbasi:
cc @scopatz

h5py/h5t.pyx
```diff
@@ -247,6 +247,13 @@ H5Tlock(H5PY_OBJ)
 
 PYTHON_OBJECT = lockid(H5PY_OBJ)
 
+# Datetime64
+cdef hid_t H5PY_DATETIME64 = H5Tcreate(H5T_OPAQUE, 8)
+H5Tset_tag(H5PY_DATETIME64, "NUMPY:DATETIME64")
+H5Tlock(H5PY_DATETIME64)
+
+NUMPY_DATETIME64 = lockid(H5PY_DATETIME64)
```
scopatz: I think we need more information about precision and endinaness encoded, which might mean more HT5s.  For example,  `np.datetime64('2017-11-29T17:11:44.638713')` has `dtype('<M8[us]')` and this could / would be read in incorrectly if you built a file with one native endianness and brought it to a system with a different endianness.

scopatz:
Thanks for putting this in @hameerabbasi!

hameerabbasi:
@scopatz I'm able to reproduce the error locally, but I have no idea what's causing it.

hameerabbasi:
Ahh... Right. `str`/`bytes` Python 2/3.

scopatz:
@tacaswell @takluyver - I'd like to have the be a part of v2.10, if that is OK with you?

h5py/h5t.pyx
```diff
@@ -669,6 +669,12 @@ cdef class TypeOpaqueID(TypeID):
             free(buf)
 
     cdef object py_dtype(self):
+        cdef bytes tag = self.get_tag()
+        cdef bytes dt_str
+        if tag.startswith(b"NUMPY:"):
+            dt_str = tag[len(b"NUMPY:"):]
```
scopatz: Shouldn't `len(b"NUMPY:")` just be `6`, so Cython (and the C compiler) can optimize this?

hameerabbasi: I didn't want to obfuscate it too much, but I can change it if needed. 😄

h5py/h5t.pyx
```diff
@@ -1994,3 +2004,19 @@ cpdef object py_get_vlen(object dt_in):
     warn("Deprecated; use check_vlen_dtype(dtype) instead",
         H5pyDeprecationWarning)
     return check_vlen_dtype(dt_in)
+
+def register_py_type(dtype dt_in):
```
scopatz: This should probably be `register_numpy_type`, since it specifically prepends "NUMPY:" in the tag. Also, I am happy for this to be public (so that user's can register their own custom numpy dtypes), but it will need a docstring :wink:

h5py/h5t.pyx
```diff
@@ -250,6 +249,7 @@ PYTHON_OBJECT = lockid(H5PY_OBJ)
 # Translation tables for HDF5 -> NumPy dtype conversion
 cdef dict _order_map = { H5T_ORDER_NONE: '|', H5T_ORDER_LE: '<', H5T_ORDER_BE: '>'}
 cdef dict _sign_map  = { H5T_SGN_NONE: 'u', H5T_SGN_2: 'i' }
+cdef dict _custom_dtype_map = {}
```
scopatz: Should this be called `_dtype_map`, as there is nothing necessarily custom about them?

h5py/h5t.pyx
hameerabbasi: Would it be nice to do automatic registration for users, make it transparent instead of throwing errors?

h5py/h5t.pyx
scopatz: You can add a comment on the line above if you feel it is unclear.

takluyver:
This brings up the question of what h5py's goals are. If your goal is 'store and load numpy arrays in HDF5 files', then this makes perfect sense.

For me, a key use of h5py is to exchange information with non-Python tools creating and reading HDF5 files. So making a convenient way to store information in an `OPAQUE` datatype with a special `'NUMPY:xyz'` tag feels like a step backwards. That information isn't very meaningful to anything else, and enabling the format would make it harder to switch to some more standard way of storing datetimes in the future.

I'd be more interested in a survey of how datetimes are stored in HDF5 files at large. I believe there's a deprecated H5T_TIME type, but I haven't heard about why it's deprecated.

h5py/h5t.pyx
```diff
@@ -1994,3 +2004,19 @@ cpdef object py_get_vlen(object dt_in):
     warn("Deprecated; use check_vlen_dtype(dtype) instead",
         H5pyDeprecationWarning)
     return check_vlen_dtype(dt_in)
+
+def register_py_type(dtype dt_in):
+    # Python 2/3 compat for str/bytes
+    try:
```
scopatz: This should be `if PY3:` rather than a try-except.

h5py/tests/hl/test_datatype.py
```diff
@@ -340,6 +340,16 @@ def test_fixed_ascii(self):
         assert string_info.length == 10
         assert h5py.check_vlen_dtype(dt) is None
 
+def TestDateTime(TestCase):
+    def test_datetime(self):
+        fname = self.mktemp()
+
+        arr = np.array([np.datetime64('2019-06-10')])
```
scopatz: Can you please add more tests for the other precisons?

scopatz:
Thanks @hameerabbasi! More feedback inline.

h5py/h5t.pyx
scopatz: Yeah, but that is probably beyond the scope of this PR.

hameerabbasi: Would `register_dtype` be okay?

scopatz:
>  I believe there's a deprecated H5T_TIME type, but I haven't heard about why it's deprecated.

My understanding is that `H5T_TIME` is (inefficiently) stored as [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) strings (see [here](https://support.hdfgroup.org/HDF5/doc1.6/Datatypes.html)).  One of the big problems with this is that is was incorrectly or inconsistently implemented across different platforms, as per [the deprecation notice](https://support.hdfgroup.org/HDF5/doc/RM/RM_H5T.html): 

> Unsupported datatype: The time datatype class, H5T_TIME, is not supported. If H5T_TIME is used, the resulting data will be readable and modifiable only on the originating computing platform; it will not be portable to other platforms.

In terms of what is done outside of HDF5 itself for datetimes is even lot less clear.  PyTables fails on datetimes and instead has [Time32Atom](https://www.pytables.org/usersguide/libref/declarative_classes.html?highlight=time32atom#tables.Time32Atom) and Time64Atom, which are currently incompatible with NumPy datetime dtypes. IIRC the way that these are implemented in PyTables is as attributes on the data set. I agree with Hameer that opaque tags are actually the proper way to store this kind of dtype metadata. @andrewcollette has [previously](https://github.com/h5py/h5py/issues/568#issuecomment-96699595) [said](https://github.com/h5py/h5py/issues/443#issuecomment-42704477) that people should handle this at "the application level." So because of this, I don't think that there is a consensus on how to handle datetimes in h5py or HDF5. 

So what is implemented here is more of standard and cross-platform (i.e. numpy) than `H5T_TYPE` in that [the spec is published as part of the numpy docs](https://docs.scipy.org/doc/numpy/reference/arrays.datetime.html#datetime-units). This also has the benefit of being fixed width and self-describing. I actually do think that anyone could recover information stored in `'NUMPY:{str(dtype)}` format relatively easily. It also seems unlikely that the HDF Group is going to put out a format more standard than numpy's, as they appeard to have abandoned datetime support a long time ago.

tl;dr I think this is the best we can do for the foreseeable future, and it doesn't lock any one in or out. Instead it promotes consistency when storing datetimes.

scopatz:
@hameerabbasi - you have a couple of things failing here: https://travis-ci.org/h5py/h5py/jobs/546803074

hameerabbasi:
CI is all green.

takluyver:
Personally, I'd still rather we provided the mechanism but required users to opt in by registering the dtype. We could then document how to do this, with a note that it's not a native HDF5 datetime type, so other applications might not recognise it. This would make it easier to switch to e.g. an unsigned integer instead of OPAQUE in the future.

I won't block doing the registration by default, but I would like to get input from other maintainers.

scopatz:
Getting more input sounds good to me! CC @h5py/core @h5py/team-h5py

scopatz:
Also, using some of @costrouc's nacent work on [scientific Python APIs](https://github.com/Quansight-Labs/python-api-inspect) it appears that none of the top 4k packages use [both datetime and h5py](https://python-api-inspect.aves.io/inspect-1a93586?sql=select+*+from+FunctionStats+%0D%0A++where+%22stats%22+like+%22%25datetime%25%22%0D%0A++and+%22stats%22+like+%22%25h5py%25%22%0D%0A++order+by+id%2C+namespace+). So this would likely not be breaking any existing code that is out there.  

As to the survey question, the [only thing I could find](https://python-api-inspect.aves.io/inspect-1a93586?sql=select+*+from+FunctionStats+%0D%0A++where+%22stats%22+like+%22%25datetime%25%22%0D%0A++and+%22stats%22+like+%22%25h5%25%22%0D%0A++order+by+id%2C+namespace+) out there that uses "h5" and "datetime" is the ulmo package. They are using [Python datetimes in one spot via pytables](https://github.com/ulmo-dev/ulmo/blob/master/ulmo/usgs/nwis/hdf5.py#L223). Their [`convert_datetime()` function](https://github.com/ulmo-dev/ulmo/blob/master/ulmo/util/misc.py#L48) is just a thin wrapper around pandas and seems to be stored as a string.

Hope this helps!

scopatz:
pinging this for feedback

hameerabbasi:
> `NUMPY:{str(dtype)}`

Small correction: As `int64` is a pointer to the native byte order `int64`, I opted instead to use descriptors, which include information on byte order.

vallsv:
Hi. I should not comment, but i was around, then here is my 2 cents.

This feature looks very interesting, but i agree with @takluyver.

I still don't understand why datetime type is no more part of HDF5, but storing it as a `h5py` specific format sounds a bad idea. I don't think it is the aim of `h5py`, especially if there is no warn about it.

Maybe you could choose an intermediate solution. I think `h5py` did manage complex numbers well. Transparent conversion with `numpy/python`, but stored as a `H5T_COMPOUND` type in the HDF5 file. Then each numeric datatime fields could be stored independently, or you could store it as text string containing ISO 8601 following the NeXus specification. With anyway a transparent conversion with `numpy`. This could be much better.

takluyver:
Thanks @vallsv - please do comment, this is the sort of issue where it's good to get different people's input.

I don't particularly like either of those solutions either (compound types or strings) - they sound fiddly to work with. But I think it's best not to do anything by default with datetimes, as there isn't one true answer in HDF5, and leave that up to users. I can certainly see uses for having a convenient way to store the data from numpy to use back in numpy, but if you're trying to produce NeXuS files that might be the wrong thing to do.

scopatz:
> I think h5py did manage complex numbers well. 

Yeah, so I think this is a big part of the problem. Currently, h5py supports numpy constructs for complex and boolean arrays, so the code base is clearly not strictly HDF5.  On the other hand, it doesn't support nested compound data types (which are supported by both numpy and HDF5), so it isn't fully HDF5 either. 

The "right" things to do are either to:

1. support only what HDF5 supports (rip out the complex and boolean pieces and have folks register those to use them)
2. Try to fully support numpy, making decisions for how to implement HDF5-unsupported data types as we go.

This PR as it stands it fully in the spirit of (2), but I totally understand (1).

For option (1), I wonder if it would be OK if we instead had it so that if you wanted to enable these extension types we had people import a specific module, so that they were knowingly enabling standard extensions.  For example:

```python
import h5py.resgister_bool
import h5py.resgister_complex
import h5py.resgister_datetime64
```

That way we get the benefit of having this registration code live in the h5py and come 'stock', but also enable people to register their own mechanism for handling these types, if they need them too.

For bool and complex we could go through a deperecation cycle where they would work for now with a warning, and then folks would have to register them in 3.0.

takluyver:
If I was writing h5py anew, I think I'd have opt-ins for everything which isn't a native part of HDF5. But that would be a major change, and it's hard to argue it's an improvement if code which worked before needs an extra incantation to keep working.

I'm also concerned about this addition for a couple of extra reasons that don't apply to complex or bool:

1. There are various different ways to store datetimes which may make sense in different scenarios (ticks of different size since some epoch, ISO 8601 strings, compounds).
2. Using the `OPAQUE` datatype seems to be saying 'this is some data only for numpy, nothing else is expected to work with it'. We store bool and complex with standard h5py primitives, not in an OPAQUE wrapper. I'd be more comfortable if it was e.g. a uint64 type with a documented meaning.

hameerabbasi:
I would be okay changing the registration mechanism to accept two inputs: A storage type and a NumPy type, and verifying that they’re equal before doing anything else.

And then changing `datetime64` registration to use an `int64`, as requested.

I believe that’d address your concerns, @takluyver.

scopatz:
> I'd be more comfortable if it was e.g. a uint64 type with a documented meaning

I don't really understand this, as we'd still have to store the metadata about what dtype it is *somewhere*. The way that PyTables does this sort of thing is to store an attribute on the dataset with the dtype information. This gets really hard to deal with for compund data types, where the attribute name encodes information about the data type, e.g. `PYTABLES_DTYPE_<column number>`.  The bookeeping is really unreasonable for nested compound types, e.g  ``PYTABLES_DTYPE_<column number in level 0>_<column number in level 1>`.  The really nice thing about using OPAQUE and registration here is that all of the that bookeeping goes away because the data type metadata flows along with the type.  It's much cleaner, and I feel much more general than storing metadata in the way pytables does.

I am trying to understand your concern with the `uint64` here @takluyver.  Are you worried that shuffle filters, etc won't work because HDF5 won't know about them? 

> There are various different ways to store datetimes which may make sense in different scenarios (ticks of different size since some epoch, ISO 8601 strings, compounds

While I agree that datetimes do have more variety, I have encountred issues with multiple ways to store complex in the wild.  Sepcifically the difference between whether complex is represented by 

```C
struct {
  double real;
  double imag;
} complex;
```

or

```C
struct {
  double r;
  double i;
} complex;
```

and other naming choices has caused me issues in the past. At least in theory under option (1), this is something that should be registerable so that more data sets could be read in.

> 'this is some data only for numpy, nothing else is expected to work with it'

Ahh, yeah. I interpret this a bit differently. I think of OPAQUE more as "HDF5 does not know what to do with datetimes, so these bytes adhere to the layout for this dtype." 

Sorry for the text wall.

takluyver:
> Sepcifically the difference between whether complex is represented by... [real/imag or r/i]

h5py does actually make this configurable. It's a bit clumsy because it's a global option, but you can change those names: http://docs.h5py.org/en/stable/config.html

But besides the different names, there are far fewer sensible ways to store complex numbers than datetimes.

> I am trying to understand your concern with the uint64 here @takluyver. Are you worried that shuffle filters, etc won't work because HDF5 won't know about them?

Nothing so technical - I wasn't even aware it had any implications for the filters. It's a philosophical objection to defining a data format by reference to a library that can make sense of it. If it's an integer with some metadata identifying it as a datetime and describing the units, it invites other code to make sense of it.

I was working on a proposal to hammer out a convention for representing units in HDF5 files; I must get back to that.

aragilar:
I'm -0 on this: I'm not going to object if people want to merge this, but I think date/time storage requires more thought than complex numbers (e.g. how to handle timezones). Also, I'd suggest dropping the 2.10 milestone (to avoid holding up the release).

Looking at h5t and the rest of h5py, using (reading and writing) the opaque datatype from the high-level API appears to require a fair bit of boilerplate, and could be made easier (the opaque type is always wrapped in void, which is going to cause problems with compound types). We could provide support for moving between dtypes and opaque tags, so that the conversion happens automatically (in a similar way to the compression and filters, which are automatic). Then we can have a separate module (or package) where the mapping for datetime (and timedelta and other dtypes) can be enabled (what types should be supported by that module/package is a different matter). This means that existing formats that use tags are easier to use, and we don't enforce a specific format for datetime. 

I'd suggest modifying the `register_dtype` function so that it takes a tag name and a dtype, and sets up the mapping, and adding a `unregister_dtype` to remove it. It's probably also worth creating up a context manager which allows for reading a specific tag as a type (similar to the existing `astype` context manager). Then registering/unregistering the datatime tags could be used as a test for the code.

hameerabbasi:
If we're willing to all agree on a design, I'll implement it.

@takluyver does @aragilar's solution sit well with you?

takluyver:
I think so - I haven't followed exactly what it would look like.

Maybe it should be based on using an explicit dtype to create datasets, like `f.create_dataset('data', dtype=h5py.opaque_numpy_datetime64)`.

Or maybe we could use the existing `astype` context manager:

```python
with dset.astype(np.datetime64):
    arr = dset[:]
```

(I haven't looked into how either of these could be implemented - if they require ugly hacks to provide that API I'd probably go off them)

hameerabbasi:
> (I haven't looked into how either of these could be implemented - if they require ugly hacks to provide that API I'd probably go off them)

The register and de-register are simple, although I believe @aragilar is suggesting we specify the tag as well as the `dtype` manually, instead of determining it automatically.

The context manager is tricky, and is only easy with thread-local variables or the `contextvars` package to do without any hacks (built in in >= 3.7). If thread-safety isn't a concern, then it's simple too.

takluyver:
The existing `astype()` context manager uses thread-local storage, so presumably it could still have unexpected consequences with coroutines.

hameerabbasi:
> The existing `astype()` context manager uses thread-local storage, so presumably it could still have unexpected consequences with coroutines.

Only in very tricky corner cases and extreme race conditions, I'd assume, thread-local storage works fine but isn't the best way to do things. `contextvars` is available for Python going back to 3.5 with an external package, and is built-in in Py37+.

If you're not comfortable with a (temporary) dependency, I'd suggest skipping the context manager for now, given it was optional in the original suggestion.

takluyver:
I'm OK with skipping it - it was just an idea I was putting out there.

hameerabbasi:
Is there a simple way to delete an HDF5 type?

scopatz:
> h5py does actually make this configurable. It's a bit clumsy because it's a global option, but you can change those names: http://docs.h5py.org/en/stable/config.html

Cool! I didn't know that. That is helpful for reading.  

> It's a philosophical objection to defining a data format by reference to a library that can make sense of it.

Yeah, I agree. I guess there are a couple of things going on here.  One is that this (accidentally) ended up being a relitigation of numpy-as-library vs numpy-as-specification. I am probably too used to thinking of numpy as a spec, so I assumed people can hunt down the numpy spec and reimplement it if they need to. That might be asking too much. The second is that this is a "purity-beats-practicatlity" argument.  I am fine with that, but I do also believe that if we go down this route we should take this opputunity to fix up the np.complex and np.bool implementations to use the register/deregister tools in time for 2.10 with a deprecation warning around current behavior, so that we can fully deprecate on 3.0. A major version but is a relatively rare oppurtunity to fix this kind of thing.  @aragilar - this deprecation cycle is exactly why I think this discussion needs to happen before 2.10.

> [@aragilar is] suggesting we specify the tag as well as the dtype manually, instead of determining it automatically

I believe that the tag name can be automatically generated by default, (e.g. `"NUMPY:" + repr(dtype)`) but if someone wants to supply their own, that's fine too and that value will overide the default.

> Is there a simple way to delete an HDF5 type?

In HDF5, you [close the data type](https://support.hdfgroup.org/HDF5/doc/RM/RM_H5T.html#Datatype-Close).  And `H5Tclose()` is exposed by h5py, see https://github.com/h5py/h5py/blob/master/h5py/h5t.pyx#L1258

hameerabbasi:
I updated this PR to be in line with @aragilar's design.

h5py/h5t.pyx
```diff
@@ -1994,3 +2004,71 @@ cpdef object py_get_vlen(object dt_in):
     warn("Deprecated; use check_vlen_dtype(dtype) instead",
         H5pyDeprecationWarning)
     return check_vlen_dtype(dt_in)
+
+
+def register_dtype(bytes tag, dtype dt_in):
```
scopatz: I think that the tag should have a default value that is generated from the dtype name.  Thoughts here @takluyver @aragilar

scopatz:
Thanks for implementing the register/deregister functions @hameerabbasi! Other than the default tag issue, this looks good to me.

takluyver:
I'm open to having a default tag name, so long as we clearly document what registering numpy dtypes achieves. I'd also consider blocking registration of the dtypes which h5py already supports.

I lean towards letting complex and bool be grandfathered in to allow storing them by default without extra complexity. I know it's not entirely consistent while I'm arguing that we shouldn't allow the same for datetime, but my thinking is:

1. I might not allow these by default if I was writing h5py from scratch, but I don't think it's worth the disruption of taking them out now, even in a new major version.
2. Both bool and complex have what seems like a very obvious underlying storage format (i.e. 0/1 for bool, (real, imaginary) for complex).

hameerabbasi:
If this is approved, I'll move on to docs. 😄 Is anything required beyond the docstrings?

scopatz:
> Is anything required beyond the docstrings?

Yeah, I think we need some user docs here for the website, including example use cases (such as datetimes).  I also think the `'NUMPY:' + dtype.descr[0][1].encode()` part needs to be in the documentation somewhere so that it is clear how the default tags are generated and how they can be used by external codes.

Also, it would be nice if you could update the "whatsnew" file to include a little blurb about this.

scopatz:
@takluyver - I don't like the grandfathering in of these types, but I like breaking existing databases even less. I think a separate PR / discussion could be about moving bool & complex to use the registration mechanism, implement the existing APIs with them. No need to get into that here though.

hameerabbasi:
Building locally, there seems to be no `Datatype` page in the docs at all and no page for the main namespace methods. Where would be the best place to put these docs?

scopatz:
I would say either add it to "Datasets" or make a new Datatypes page.  I lean towards starting a new Datatypes page, but if you find it more expediant to add to datasets, that's fine with me.

aragilar:
This looks good to me (the context manager idea was more of a convenience than anything), it might be worth stating in the docs that the backwards/forwards compatibility degree is limited to that given by numpy (which is pretty good, but doesn't mean numpy won't change things in the future) or whoever provides the dtype.

RE complex/bool: I'm not sure either bool or complex would work via the opaque type.
h5py and pytables currently store bools differently (no idea about what other projects do), having another way of storing them is probably going to be even more confusing.
With complex numbers, we actually want to use HDF5 floating point datatype, it handles the complexity of what format floats are in much better than trying to come up with a scheme in the tag representing all the existing metadata we need.

scopatz:
>  it might be worth stating in the docs that the backwards/forwards compatibility degree is limited to that given by numpy 

Great point!  Should the tag embed the numpy version by default?

> having another way of storing them is probably going to be even more confusing.

For me this is less about storing and more about being able to correctly read datasets produced by other tools.  A user could register a way of reading bool and complex that are different than h5py's default.

hameerabbasi:
I’ve taken my first stab at documenting. Does anything else need to be done here?

docs/faq.rst
```diff
@@ -27,6 +27,7 @@ Array                               Any supported type
 Enumeration                         Any NumPy integer type                          Read/write as integers
 References                          Region and object
 Variable length array               Any supported type                              See :ref:`Special Types <vlen>`
+Any other NumPy dtype               Map to opaque type with specific tag            You must register types manually
 =========================           ============================================    ======================
```
scopatz: I thin the last set of `====` need to span the text in the table, because this is rst.

scopatz: Generally speaking, listtable is a better construct here, but you don't need to change that, Hameer.

docs/high/datatype.rst
```diff
@@ -0,0 +1,38 @@
+.. _datatype:
+
+Datatypes
+=========
+
+Usually, :class:`Datatype` objects are created for you on the fly and you do not
+have to worry about them. However, in certain cases, namely when no HDF5
+equivalent exists for a given :class:`dtype`, you must register the :class:`dtype`
+manually for use with H5Py.::
```
scopatz: I think h5py should be lowercase here

scopatz:
A couple of minor non-essential copy-edits.  @takluyver - does this look good to you?

scopatz:
And @aragilar - any other thoughts?

docs/faq.rst
```diff
@@ -27,7 +27,8 @@ Array                               Any supported type
 Enumeration                         Any NumPy integer type                          Read/write as integers
 References                          Region and object
 Variable length array               Any supported type                              See :ref:`Special Types <vlen>`
-=========================           ============================================    ======================
+Any other NumPy dtype               Map to opaque type with specific tag            You must register types manually
+=========================           ============================================    ================================
```
aragilar: I wouldn't add this here, I'd add a section in the unsupported types noting that users can use the opaque type to store other data and can tag it (and put a comment in the opaque type line referring to that section).

docs/high/datatype.rst
```diff
@@ -0,0 +1,38 @@
...
+=========
+
+Usually, :class:`Datatype` objects are created for you on the fly and you do not
+have to worry about them. However, in certain cases, namely when no HDF5
+equivalent exists for a given :class:`dtype`, you must register the :class:`dtype`
+manually for use with h5py.::
+
+    arr = np.array([np.datetime64('2019-06-30')])
+    h5py.register_dtype(arr.dtype)
+    dset = f.create_dataset("datetimes", data=arr)
```
aragilar: I think adding some commentary about the tradeoffs/assumptions when using opaque types to store data (e.g. harder to read vs exact representation of type). My main concern is we're going to get lots of bugs filed from people who assumed that using `h5py.register_dtype` absolves them about thinking how their encoding their data (c.f. pickle being used for long term storage).

aragilar:
I'm fine with this. I've added some suggestions in the docs, so that we warn ahead of time that using opaque isn't like using one of the other datatypes, so that people don't treat it like pickle and try to push everything through it.

hameerabbasi:
I believe this addresses everyone's concerns. :)

docs/faq.rst
```diff
@@ -37,6 +37,8 @@ Type                                Status
 HDF5 "time" type
 NumPy "U" strings                   No HDF5 equivalent
 NumPy generic "O"                   Not planned
+Any other NumPy dtype               Map to opaque type with specific tag, You
+                                    must register types manually
```
takluyver: I'd like this to cross-reference `register_dtype` for people to find more information.

docs/high/datatype.rst
```diff
@@ -0,0 +1,47 @@
...
+    be readable insofar as the ``dtype`` is binary-compatible across old
+    versions. In general, opaque datatypes are very sensitive to how you may
+    decide to encode your data.
+
+Reference
+=========
+
+.. class:: Datatype(identifier)
+
+    Datatype objects are typically created via :func:`register_dtype` function,
```
takluyver: This isn't quite right - `register_dtype` doesn't return a `Datatype` object, or even the lower-level `TypeID`.

docs/high/datatype.rst
```diff
@@ -0,0 +1,47 @@
...
+
+    Opaque types with this tag will be mapped to NumPy types in the same way.
+
+    The default tag is generated via the code:
+    ``b"NUMPY:" + dt_in.descr[0][1].encode()``.
+
+.. function:: deregister_dtype(object obj)
+
+    Deregister a dtype/tag from the NumPy-tag mapping, along with the.
+    corresponding tag/dtype.
```
takluyver: I think the description of these should ultimately go together with the other `h5t` functions documented in `special.rst`. But combining them is a bigger job which needn't be done as part of this PR - I also want to avoid calling them 'special', because that's not really a helpful term.

h5py/h5t.pyx
```diff
@@ -1994,3 +2004,80 @@ cpdef object py_get_vlen(object dt_in):
...
+
+
+def _deregister_tag(bytes tag):
+    """ (bytes tag)
+
+    Deregister a tag from the NumPy-tag mapping, along with the.
+    corresponding dtype.
+    """
+    if tag not in _tag_map:
+        raise RuntimeError("This tag is not registered.")
```
takluyver: I wonder if this and the corresponding 'dtype not registered' error should be KeyError, rather than the very generic RuntimeError. I haven't thought much about the exception semantics, though.

hameerabbasi: From the [docs](https://docs.python.org/3/library/exceptions.html#KeyError):

> Raised when a mapping (dictionary) key is not found in the set of existing keys.

Here, it would be confusing as the user doesn't see the map directly. How about `ValueError`?

h5py/tests/hl/test_datatype.py
```diff
@@ -340,6 +340,37 @@ def test_fixed_ascii(self):
...
+                    arr = (np.array([0], dtype=np.int64).view(dtype=dt)
+                        if dt_kind == 'M8'
+                        else np.array([np.timedelta64(500, dt_unit[1:-1])], dtype=dt))
+
+                    with h5py.File(fname, 'w') as f:
+                        dset = f.create_dataset("default", data=arr)
+                        self.assertArrayEqual(arr, dset)
+                        self.assertEqual(arr.dtype, dset.dtype)
+                        # So both branches are tested.
+                        h5py.deregister_dtype(dt)
```
takluyver: At a minimum, we should put the deregister in a `finally` block so it still happens if something goes wrong during the test.

But this also highlights that global mutable state is always a tricky thing. I'm still inclined to postpone this for a future release so we can think about how best to design it rather than rushing to get it in before 2.10.

hameerabbasi: I tend to agree here... My concern is Py2 compat. I can use thread_local if needed.

hameerabbasi:
Waiting for responses on the rest before proceeding.

h5py/tests/hl/test_datatype.py
hameerabbasi: I put it in `finally`.

scopatz: I really would like for this to go into 2.10.  If we think more effort and discussion is needed to be done to make this acceptable for 2.10, @hameerabbasi and I are happy to go through that effort.

I agree that a context manager would be nice for this, so we wouldn't have to worry about using try-finallys everywhere.

Maybe we could use the thread_local behavior on Python 3 and then not use that for Python 2, with a warning in the docs that the context manager might behave slightly differently on Python 2 and explain why.  Presumably, we aren't planning on supporting Python 2 forever, so this issue will go away in the relatively near future.

takluyver: The plan is to drop Python 2 for 3.0 - see issue #1177. That will probably be the next release after 2.10.

Rather than some (thread-)global registry, would it make sense to require manually creating wrapper objects in a similar pattern to strings, vlens and enums? Something like:

```python
dt = h5py.opaque_dtype(arr.dtype)
dset = f.create_dataset('timestamps', data=arr, dtype=dt)
```

hameerabbasi: Would we be able to put such a type inside a compound type?

takluyver: I imagine it would be possible, but perhaps it's a bit ugly if you have to define the compound structure again for h5py. On the other hand, it would make it explicit how the opaque data type is being used.

takluyver:
I think this is now the last thing really holding up the 2.10 release, assuming someone with upload permission has time to make a release. The other PRs with the 2.10 milestone haven't been actively worked on recently, so I don't think anyone will object to bumping them to a later release.

I'm still not entirely happy with the global registry of tagged opaque types. But my suggested alternative is a sketch, not implemented or even fully thought through, and I'm unlikely to have time to flesh it out in the next week or two.

So: someone please make a decision on this. I lean towards leaving it out, but that's an easy position as I don't have any use for it at present. I'll accept if the decision is to land it - sometimes it's more important to have working code than keep arguing about the design.

aragilar:
@takluyver I like the `h5py.opaque_dtype` idea, but I'm not sure how that would work with reading (as we'd need to map from tag to dtype), which is where I'd imagine we'd need some kind of global registry (which admittedly was the use case I thought of when suggesting making the opaque type more accessible)?

My general impression from this PR (and other issues such as #1258) is that people are fine with some complexity when writing (because you want to specify how things are encoded), but want things to "just work" when reading (which we mostly have with the most common types)? While I'd be happy to put this in 2.10, if we can come up with a better framework (for opaque types, but also for other HDF5 types such as the H5T_NATIVE_B8), maybe we should put this off?

takluyver:
I think I'd agree that reading could just work - there's no real ambiguity about what's the right thing to do. I don't think it even needs a global registry, if the numpy dtype is fully recorded in the tag. This could even be extended to writing into an existing dataset which has this type.

So perhaps the only point where I'd like to require an explicit definition is when creating a new dataset - and I think the `opaque_dtype` idea works for this.

hameerabbasi:
Do you have an idea of how it'd be done in code? Just create an `h5t` and pass it in or would there be further changes needed?

hameerabbasi:
Is there a way to get a `Datatype` object from a `Dataset` object?

hameerabbasi:
The issue I'm running into is that in various places in the code, the `Datatype` is converted into a `np.dtype` and passed back into `py_create` and then either I have to make a decision: either let the error be raised or somehow detect that it came from the right dtype... Which needs a global registry.

The third option is to fix that up, but for that, I need an answer to my previous comment.

takluyver:
`dset.id.get_type()` should get you the low-level `TypeID` object, I think. The high-level `Datatype` class is just a thin wrapper around a `TypeID`.

takluyver:
PS if you're having a go at the `opaque_dtype` idea, you might want to do it in a separate PR so it's easy to come back to this one if we discover problems with that. Up to you, though!

scopatz:
Hi All, @tacaswell and I had a discussion about this during SciPy 2019 last week.  The outcome of this was basically that I am OK with moving this to 3.0 if we are also OK with doing a 3.0 release relatively soon after the 2.10 release.  I am happy to help push on both of these. 

If we do that, we'll have more lattitude w.r.t. to the thread safe context managers and dataclasses.

@takluyver @aragilar - how do you feel about this proposal?

takluyver:
It's certainly my preference to bump this to a later release - it means we can get 2.10 out sooner, and it gives us more time to work out the best way of doing this.

I'm happy in principle to aim for 3.0 relatively soon, but:

1. 2.10 deprecates something we plan to change in 3.0 (how files are opened if you don't specify a mode). A deprecation cycle only makes sense if people have time to see the warnings and adapt their code before the change. This is part of why I want to get 2.10 out. So 'relatively soon' would probably still be more like a few months than a few weeks. We could do 2.11 if you want to get the feature out sooner, but splitting developer attention between two upcoming releases might slow everything down.
2. I can't make a firm commitment to spend time on it. I have several trips coming up, and then my contract ends at the end of September - I'm optimistic that it will be renewed, but it's not guaranteed, and my interest in HDF5 may wane rapidly if I'm not using it at work.

scopatz:
> So 'relatively soon' would probably still be more like a few months than a few weeks.

Yeah, I was thinking more on the few weeks time-scale.  I think it is fine for folks to have to pin to the <3.0 if they are relying on those features.

takluyver:
If you're thinking a few weeks, let's call it 2.11. The past few 2.x feature releases have been separated by at least a few months each, and often a year. 3.0 is going to be a chance to do some cleanup and breaking changes which we wouldn't in a 2.x release - I hope we won't be too eager with that, but there are people pushing for changes to string handling, for instance. So I don't think it's realistic to aim for 3.0 in a few weeks.

scopatz:
Well, the hope is that we could use some of the Python 3 features to make these features more robust.  Maybe I misunderstood, but wouldn't 2.11 still be Python 2 compatible.  If that is the case, then I don't see the reason perform a 2.11, and would rather this stuff was worked into 2.10 in some way.

takluyver:
It doesn't seem plausible to me that we'll go from 2.10 to 3.0 in a few weeks, when each 2.x release has taken at least 6 months.

I'd be open to having 2.11 require Python 3, if other people agree on that - we can mark it so that `pip install h5py` on Python 2 still gets 2.10, so it needn't break things for people, and it's rapidly becoming common to require Python 3, so I don't think that needs to align with version 3.0 of h5py.

scopatz:
Right, so if we think the next release is going to take 6 months, then it is more important to get this in for 2.10.

I'd like to add [rever](https://regro.github.io/rever-docs/) support to h5py so that we can speed up the release cycle.

takluyver:
I think 2.11 could be quicker than that if we had a bit of a push for it. I imagine that for 3.0 people will want to spend some more time deciding what it's worth changing in a major release.

Nice tool! I started writing a release automation tool ('modeltee'), but I've not found time to take it beyond a prototype. Making releases more streamlined certainly seems like a good idea - can you open a separate issue for that?

scopatz:
> I think 2.11 could be quicker than that if we had a bit of a push for it. 

OK! That makes sense.

> can you open a separate issue for that?

See #1270

scopatz:
So it seems as though the current plan is to:

* release 2.10 without this
* release 2.11 with this, Python 3 only, and related updates to PR and source
* add rever support

Any objections / thoughts on this plan?

takluyver:
If that's an ordered list, rever support can happen sooner - I assume that's relatively easy for you to add once you know how a release is done (which I'm not the person to tell you - I think @tacaswell did the last couple of releases).

Apart from that, that matches my understanding.

scopatz:
Yeah, sorry, unordered list. I am planning on working on rever support this week, but any input from @tacaswell would be awesome!

takluyver:
Superseded by #1262 and #1339.

PR #1238, opened 2019-06-15
Add recent PRs to whatsnew for 2.10

takluyver:


takluyver:
I'm going to merge this myself, as it's a straightforward docs change, and I want to check that the RTD integration is working while I'm still thinking about it.

takluyver:
Yup, the webhook worked, so the docs are up to date again.

Issue #1237, opened 2019-06-14
AttributeError on import when building from source


hameerabbasi:
To assist reproducing bugs, please include the following:
 * Operating System: macOS 10.14
 * Python version: 3.7.2
 * Where Python was acquired: Anaconda
 * h5py version (e.g. 2.6): f5463ad632510a39bd04f895a60784985c8ddbfa
 * HDF5 version (e.g. 1.8.17): 1.10
 * The full traceback/stack trace shown (if it appears):

```
>>> import h5py
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/hameerabbasi/Quansight/h5py/h5py/__init__.py", line 49, in <module>
    from ._hl.files import (
  File "/Users/hameerabbasi/Quansight/h5py/h5py/_hl/files.py", line 41, in <module>
    libver_dict.update({'v108': h5f.LIBVER_V18, 'v110': h5f.LIBVER_V110})
AttributeError: module 'h5py.h5f' has no attribute 'LIBVER_V18'
```

tacaswell:
can you run a `git clean -xfd` and try re-building?

hameerabbasi:
I already did that, I forgot to mention that in the issue description. But, I just double-checked and it does happen.

hameerabbasi:
As an update: Removing and re-creating my conda environment has no effect.

tacaswell:
I can not reproduce this locally (and CI is passing so we can not reproduce it their either).  How are you installing h5py?  From the paths it looks like `pip install -e .`?  Did you ever do a `pip install --user` or have a conda also installed version of h5py giving you out-of-date files?

hameerabbasi:
I'm doing the following:

```
conda create -n h5py-dev python=3 numpy pytest hdf5 pkg-config
conda activate h5py-dev
pip install -e .
```

hameerabbasi:
`--user` is redundant in `conda`, nevertheless, I did follow your instructions and it still didn't work.

tacaswell:
Try not doing editable and remember to `git clean -xfd`.

Following your instructions exactly I get a much worse issue:

```
jupiter@13:03  ➤  python
Python 3.7.3 (default, Mar 27 2019, 22:11:17) 
[GCC 7.3.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import h5py
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
[...]

PR #1239, opened 2019-06-16
Build & configure cleanups, extra info

takluyver:
I think it would be useful for debugging build issues (e.g. #1237) to know what library path it tries to load to autodetect the version number.

I also spotted a couple of minor cleanups, which shouldn't make any material difference.


scopatz:
Thanks @takluyver!

PR #1241, opened 2019-06-17
Don't catch exceptions when trying to autodetect HDF5 version in build

takluyver:
If it can't find libhdf5, or can't get the version, let the build bail out rather than trying to continue.

This is part of a general move to make builds fail fast. Before wheels & conda were in widespread use, many more people would have had to install h5py from source, so there was probably more pressure for the build system to smooth over corner cases and try to build anyway. Now that people who just want the package installed can mostly get pre-built binaries, it's probably better to do less guesswork and error out if there are things the build system can't work out.

### DIFF
--- a/setup_configure.py
+++ b/setup_configure.py
@@ -157,12 +157,8 @@ class configure(Command):
         if self.hdf5_version is None:
             self.hdf5_version = oldsettings.get('env_hdf5_version')
         if self.hdf5_version is None:
-            try:
-                self.hdf5_version = autodetect_version(self.hdf5)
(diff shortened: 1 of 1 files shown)


takluyver:
N.B. This is for discussion - it seems like a good move to me,  but I don't know if there's some case where the fallback is still important.

scopatz:
From the discussion in #1237, this seems like it should be merged in.  Also, now that 1.10.x has been out for a while, it seems doubly odd to assume 1.8.4.  Thanks @takluyver!

tacaswell:
Late to the party, but :+1: to failing fast at build time if we don't know what version of hdf5 to use.

PR #1242, opened 2019-06-17
Move check for HDF5 build & runtime versions earlier

takluyver:
If the HDF5 version we're loaded against does not match the version we were built against, we can get errors loading various submodules, as it e.g. tries to load symbols that aren't there.

We have a check to warn about this, but if there's an error before execution gets to it, the warning can't be shown. So this moves it nearer the start of the module, where it will hopefully show an informative error message in more cases.

From discussion on #1237.

### DIFF
--- a/h5py/__init__.py
+++ b/h5py/__init__.py
@@ -31,6 +31,16 @@ except ImportError:
     else:
         raise
 
+from . import version
+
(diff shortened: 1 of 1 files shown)


scopatz:
Thanks @takluyver!

Issue #1244, opened 2019-06-19
[ppc64le] test suite failing on power9


kif:
As suggested by  @takluyver, here are the results of the test-suite running on IBM power9 computer running Ubuntu 18.04. The python core is 3.6, all modules have been pip installed in a venv (and recompiled as no binary wheels exist yet on this platform) and HDF5 has been compile from sources in version 1.10.5.

Most failing tests are related in a way or another to "float128" which have just been deactivated in  #1243. They should probably be marked as "xfailed".

```
(venv-kieffer_system) test@power9:~/HDF5/h5py/build/lib.linux-ppc64le-3.6$ python
Python 3.6.7 (default, Oct 22 2018, 11:32:17) 
[GCC 8.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import h5py
>>> h5py.run_tests()
=================================================================================== test session starts ===================================================================================
platform linux -- Python 3.6.7, pytest-4.4.2, py-1.8.0, pluggy-0.11.0
rootdir: /home/test/HDF5/h5py, inifile: pytest.ini
collected 537 items                                                                                                                                                                       

h5py/tests/hl/test_attribute_create.py ...                                                                                                                                          [  0%]
h5py/tests/hl/test_completions.py ..                                                                                                                                                [  0%]
h5py/tests/hl/test_dataset_getitem.py .................................................x....................................                                                        [ 16%]
h5py/tests/hl/test_dataset_swmr.py ssss..........                                                                                                                                   [ 19%]
h5py/tests/hl/test_datatype.py .........F.....ssss                                                                                                                                  [ 23%]
h5py/tests/hl/test_deprecation.py .                                                                                                                                                 [ 23%]
h5py/tests/hl/test_dims_dimensionproxy.py .                                                                                                                                         [ 23%]
h5py/tests/hl/test_file.py .................                                                                                                                                        [ 26%]
h5py/tests/hl/test_filters.py .s.                                                                                                                                                   [ 27%]
h5py/tests/hl/test_threads.py ..                                                                                                                                                    [ 27%]
h5py/tests/hl/test_vds/test_highlevel_vds.py ......                                                                                                                                 [ 28%]
h5py/tests/hl/test_vds/test_lowlevel_vds.py ...                                                                                                                                     [ 29%]
h5py/tests/hl/test_vds/test_virtual_source.py ................                                                                                                                      [ 32%]
[...]

PR #1246, opened 2019-06-20
skip tests that use float128 on ppc64le ...

kif:
one test is still failing. Investigation is needed here.

### DIFF
--- a/h5py/tests/old/test_dataset.py
+++ b/h5py/tests/old/test_dataset.py
@@ -26,7 +26,7 @@ import numpy as np
 
 import platform
 
-import pytest 
+import pytest
 
 from ..common import ut, TestCase
 from h5py import File, Group, Dataset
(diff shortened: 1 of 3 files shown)


kif:
close #1244 

Beside this the >f8 attribute becomes a f16 and fails ...

h5py/tests/old/test_dataset.py
```diff
@@ -91,6 +95,8 @@ def test_missing_shape(self):
     def test_long_double(self):
         """ Confirm that the default dtype is float """
         dset = self.f.create_dataset('foo', (63,), dtype=np.longdouble)
+        if platform.machine() in ['ppc64le']:
+            pytest.xfail("Storage of long double deactivated on %s"%platform.machine())
```
scopatz: Should this be a skip, rather than a fail?

kif:
On Thu, 20 Jun 2019 08:48:40 -0700
Anthony Scopatz <notifications@github.com> wrote:

> Should this be a skip, rather than a fail?

This is up to you ...

takluyver:
I'd say xfail if this is failing now but we hope to fix it and re-enable the tests. Skipping it implies that it can't work, or we don't aim to support it.

scopatz:
Makes sense! It looks like there is one failing style issue https://travis-ci.org/h5py/h5py/jobs/548242859#L323

scopatz:
Thanks @kif!

PR #1247, opened 2019-06-21
Clarify docs on core driver

takluyver:
Prompted by a question on the mailing list. I checked the [source code of the 'core' driver](https://bitbucket.hdfgroup.org/projects/HDFFV/repos/hdf5/browse/src/H5FDcore.c) to confirm my hunch: this is not memory mapping. It is designed to read the entire file into RAM and manipulate it there.

Also fixed some formatting with the keywords - see the current docs:
http://docs.h5py.org/en/stable/high/file.html#file-drivers

### DIFF
--- a/docs/high/file.rst
+++ b/docs/high/file.rst
@@ -57,15 +57,17 @@ of supported drivers and their options:
         Buffered I/O using functions from stdio.h.
 
     'core'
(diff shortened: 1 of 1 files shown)


scopatz:
Thanks @takluyver!

Issue #1245, opened 2019-06-19
Segmentation fault when writing empty VLEN data


colingavin:
The following segfaults for me on macOS (10.12.6) with python 3.7.3 from pyenv.
```
import h5py
import numpy as np

print(h5py.version.info)
d = np.core.records.fromarrays([[], []], names='a,b', formats='|V16,O')
print('array: ', repr(d))
f = h5py.File('tmp.h5', 'w')
f.create_dataset('test', data=d, dtype=[('a', '|V16'), ('b', h5py.special_dtype(vlen=np.float_))])
```
The output is:
```
Summary of the h5py configuration
---------------------------------

h5py    2.9.0
HDF5    1.10.4
Python  3.7.3 (default, Mar 27 2019, 09:23:32) 
[Clang 9.0.0 (clang-900.0.39.2)]
sys.platform    darwin
sys.maxsize     9223372036854775807
numpy   1.16.4

array:  rec.array([],
          dtype=[('a', 'V16'), ('b', 'O')])
Segmentation fault: 11
```
Note that replacing `d` with the following does not crash and writes valid data:
```
b = np.tile(None, 1)
b[0] = np.array([1.0])
d = np.core.records.fromarrays([[np.array(b'\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0', dtype='|V16')], b], names='a,b', formats='|V16,O')
```
There does not appear to be anything special about the length of the byte string, but the crash goes away if it is removed entirely (so the dtype is `['b', 'O']`).

scopatz:
There probably isn't a check for zero-sized data and so it is trying to deref the array location.  PRs welcome!

colingavin:
A little more research: the crash is at https://github.com/h5py/h5py/blob/7d50c06ea42c9238a2659f501fae4816ad4c4fd2/h5py/_conv.pyx#L745
in `ndarray2vlen` when the number of elements `nl` is zero. It appears that in this case the `void* buf_i` being passed in is not a valid `PyObject**`.

This call is originating from https://github.com/h5py/h5py/blob/7d50c06ea42c9238a2659f501fae4816ad4c4fd2/h5py/_proxy.pyx#L163
I don't know the internals of h5py well enough to know whether the problem is the allocation of `conv_buf` at the original call site, or whether the zero-length case should be handled in the conversion function.

scopatz:
I think handling it in the conv.pyx might be the right way to go here.

takluyver:
Closed by #1253.

PR #1248, opened 2019-06-22
Fix segfault when writing empty VLEN (issue #1245)

colingavin:
As suggested in the discussion on the issue (https://github.com/h5py/h5py/issues/1245#issuecomment-504447512), this adds a special case to the conversion function to handle zero-length conversions which would otherwise cause a bad memory access.

### DIFF
--- a/h5py/_conv.pyx
+++ b/h5py/_conv.pyx
@@ -740,6 +740,11 @@ cdef herr_t ndarray2vlen(hid_t src_id, hid_t dst_id, H5T_cdata_t *cdata,
 
     elif command == H5T_CONV_CONV:
 
+        # If there are no elements to convert, pdata will not point to
+        # a valid PyObject*, so bail here to prevent accessing the dtype below
+        if nl == 0:
+            return 0
+
         # need to pass element dtype to converter
         memcpy(&pdata_elem, pdata, sizeof(pdata_elem))
         supertype = py_create((<np.ndarray> pdata_elem).dtype)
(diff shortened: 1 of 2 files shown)


takluyver:
Unfortunately the Windows CI is failing. I can't see how the issue could be with this change, so it might be unlucky timing. But we'll need to figure that issue out so it can run the tests.

takluyver:
Ah, it's because you branched off 2.9.x instead of master. Can you redo the PR against master? That's how we normally work for development, and we're no longer planning to do a 2.9.1 release anyway.

takluyver:
I see the docs say to start from a stable branch for bugfixes; I think that the docs are out of date on this. Sorry about that. I'll submit a PR to change them.

takluyver:
Opened #1249

takluyver:
@colingavin do you have time to redo this against master? If not, let me know and I (or someone else) might find time to cherry-pick the commits for you.

scopatz:
Does this just need a merge against master?

takluyver:
I'd rather cherry-pick the commits, otherwise we're going to get all the backport commits from the 2.9.x branch.

scopatz:
Ahh yeah, makes sense.

colingavin:
Yes, I started cherry-picking to master yesterday, but I ran into issues getting the unit test suite to run. In particular pytest complained about `invalid module name: 'h5py.h5py_warnings'` when running `python setup.py test`. Is the testing setup different in master vs. 2.9.x?

colingavin:
Oh, I see @scopatz did the cherry-pick in a new PR (thanks!), so I'll close this one.

takluyver:
I think that error means it's failing to import h5py, but it hides the real error. Not immediately sure if that could be a change on master, but I did have to solve a similar issue on Appveyor recently which was somehow triggered by a new version of tox - #1225.

PR #1249, opened 2019-06-24
Document that we always develop against master

takluyver:
I also updated the description of CI. On recent builds, Travis has taken just over 20 minutes wall-clock time, and Appveyor 30-35.


scopatz:
Thanks!

PR #1255, opened 2019-06-27
Use mpirun when testing MPI

aragilar:
This provides more freedom to specify HDF5 locations (which I needed for testing the MPI builds), and runs the tests under MPI (apart from those known to not work, e.g. those using subprocess). I fixed the existing MPI tests, and added some docs about how to test with MPI.

This depends on https://pypi.org/project/pytest-mpi/, which I created because it seems that the `conftest.py` must the at the root level for cli args to work (see https://stackoverflow.com/questions/52447767/using-pytest-addoptions-in-a-non-root-conftest-py), I'm happy to move it under the h5py org and/or give people access (on github, pypi and other services) if people think that's appropriate (I plan on adding docs and tests, happy to restructure it also, I basically used an existing pytest plugin I wrote as a template).

Before this gets merged, can people run this on their systems (and if report any lockups, the tests should run very fast, so if you're waiting a while, there's probably a problem).

#1206 will need to be rebased on this (let me know if there are more features we should expose with pytest-mpi).

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -354,3 +354,59 @@ In the Cython code, these show up as "preprocessor" defines ``MPI`` and
 
 High-level code can check the version of the HDF5 library, or check to see if
 the method is present on ``FileID`` objects.
+
+
+Testing MPI-only features/code
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Typically to run code under MPI, ``mpirun`` must be used to start the MPI
+processes. Similarly, tests using MPI features (such as collective IO), must
+also be run under ``mpirun``. h5py uses pytest markers (specifically
+``pytest.mark.mpi`` and other markers from
+`pytest-mpi <https://pytest-mpi.readthedocs.io>`_) to specify which tests
+require usage of ``mpirun``, and will handle skipping the tests as needed. A
+simple example of how to do this is::
+
+   @pytest.mark.mpi
+   def test_mpi_feature():
+      import mpi4py
+      # test the MPI feature
+
+To run these tests, you'll need to:
+
+1. Have ``tox`` installed (e.g. via ``pip install tox``)
+2. Have HDF5 built with MPI as per :ref:`build_mpi`
+
+Then running::
+
+   $ CC='mpicc' HDF5_MPI=ON tox -e py37-test-deps-mpi4py
+
+should run the tests. You may need to pass ``HDF5_DIR`` depending on the
+location of the HDF5 with MPI support. You can choose which python version to
+build against by changing py37 (e.g. py36 runs python 3.6, this is a tox
+feature), and test with the minimum version requirements by using ``mindeps``
+rather than ``deps``.
+
+If you get an error similar to::
+
+   There are not enough slots available in the system to satisfy the 4 slots
+   that were requested by the application:
+     python
+
+   Either request fewer slots for your application, or make more slots available
+   for use.
+
+then you need to reduce the number of MPI processes you are asking MPI to use.
+If you have already reduced the number of processes requested (or are running
+the default number which is 2), you will need to look up the documentation for
+your MPI implementation for handling this error. On OpenMPI (which is usually
+the default MPI implementation on most systems), running::
+
+    $ export OMPI_MCA_rmaps_base_oversubscribe=1
+
+will instruct OpenMPI to allow more MPI processes than available cores on your
+system.
+
+If you need to pass additional environment variables to your MPI implementation,
+add these variables to the `passenv` setting in the `tox.ini`, and send us a PR
+with that change noting the MPI implementation.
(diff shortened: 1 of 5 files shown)


takluyver:
I don't have time to look at this in detail right now, but this sounds excellent - having a decent testing infrastructure for MPI will make things like #1206 much simpler to address.

For the plugin: I think it's fine to add it to the h5py org if you'd like to maintain it collaboratively. But we can also depend on your project if you'd rather keep it separate.

aragilar:
I've switched back to using the system HDF5 (on azure now, rather than travis). I won't do any more experimentation on this branch, I'll leave getting a newly-built version of HDF5 to a different PR/branch.

docs/contributing.rst
```diff
@@ -354,3 +354,44 @@ In the Cython code, these show up as "preprocessor" defines ``MPI`` and
...
+1. Have ``tox`` installed (e.g. via ``pip install tox``)
+2. Have HDF5 built with MPI as per :ref:`build_mpi`
+
+Then running::
+
+   $ CC='mpicc' HDF5_MPI=ON tox -e py37-test-deps-mpi4py
+
+should run the tests. You may need to pass ``HDF5_DIR`` depending on the
+location of the HDF5 with MPI support. You can choose which python version to
+build against by changing py37 (e.g. py27 runs python 2.7, this is a tox
```
takluyver: We might want to use e.g. py36 as the example, since we don't support Python 2.7 any more.

docs/contributing.rst
```diff
@@ -354,3 +354,44 @@ In the Cython code, these show up as "preprocessor" defines ``MPI`` and
...
+If you get an error similar to::
+
+   There are not enough slots available in the system to satisfy the 4 slots
+   that were requested by the application:
+     python
+
+   Either request fewer slots for your application, or make more slots available
+   for use.
+
+then you need to reduce the number of MPI processes you are asking MPI to use.
```
takluyver: How do you control this? What's the default?

setup_configure.py
```diff
@@ -86,12 +89,18 @@ class configure(Command):
 
     user_options = [('hdf5=', 'h', 'Custom path to HDF5'),
                     ('hdf5-version=', '5', 'HDF5 version "X.Y.Z"'),
+                    ('hdf5-includedir=', 'i', 'path to HDF5 headers'),
+                    ('hdf5-libdir=', 'l', 'path to HDF5 library'),
+                    ('hdf5-pkgconfig-name=', 'p', 'name of HDF5 pkgconfig file'),
```
takluyver: It looks like the added config options are no longer required for the MPI testing. Can I suggest that we pull these additions out as a separate PR, and do the minimum necessary in this one to effectively test with MPI?

h5py/tests/test_filters.py
```diff
@@ -61,6 +62,7 @@ def test_wr_scaleoffset_fletcher32(self):
                                   )
 
 
+@pytest.mark.mpi_break
```
takluyver: For clarity, I might suggest calling the marker `mpi_skip` or `mpi_xfail` - you might even allow both to distinguish "this shouldn't work" vs "this should work but doesn't".

docs/contributing.rst
takluyver: I see in tox.ini that the default is 2. I imagine reducing this doesn't make much sense - testing with 1 process is going to miss things. Maybe we should briefly mention how to make more slots available.

RizzerOnGitHub: The OpenMPI flag 

    export OMPI_MCA_rmaps_base_oversubscribe=1

or `mpirun -oversubscribe` might be useful in this context.

docs/contributing.rst
aragilar: Yep, that's probably wise.

docs/contributing.rst
aragilar: I've added a section on this mentioning the OMPI env var, I think only OpenMPI has this issue (MPICH apparently doesn't, I've got no idea how to setup MS-MPI and I'm not sure there are any other maintained MPI systems).

setup_configure.py
aragilar: Done, now in #1414.

h5py/tests/test_filters.py
aragilar: I've release a new version of pytest-mpi which adds `mpi_skip` and `mpi_fail`, which should be more understandable.

takluyver:
It looks like the tests crashed immediately while trying to import h5py, but then sat until they timed out after an hour. Before making the tests work, can we ensure that it stops correctly on failure?

aragilar:
I've dropped the CI changes from this, so hopefully the tests now pass (leaving the current MPI tests on travis). I'll look at expanding the CI factors the MPI tests run on in a different PR.

takluyver:
The test changes here look good to me; thanks.

PR #1256, opened 2019-06-29
Expose H5PL low-level API

takluyver:
This is a continuation of #1166. I've updated it to use the `@insubprocess` decorator @scopatz added, fixed the tests running in a conda environment (conda-forge configures a non-default plugin directory when building HDF5), and added some docs for the new functions.

Remaining questions:

1. Currently the path APIs accept and return bytes. Although this is technically correct for Unix paths, it's more common in Python to work with paths as strings. We're a bit inconsistent elsewhere - the low-level API for an external link returns bytes, but getting the source of a virtual dataset returns str. I'm leaning towards staying with bytes in the low-level API.
2. Do we want to do any locking? On #1166, @tacaswell said they shouldn't need the phil lock, but there might be reasons to have a separate lock for the plugin search path:

>  We may also want to lock them between eachother (as I can imagine an append and a remove colliding and ending up with a "gap" in the hdf5 datastrucutres....).

I'm tentatively marking this for 2.10, but I don't mind if we bump it to a later release.

### DIFF
--- a/docs_api/h5pl.rst
+++ b/docs_api/h5pl.rst
@@ -1,0 +1,5 @@
+Module H5PL
+===========
+
+.. automodule:: h5py.h5pl
+    :members:
--- a/docs_api/index.rst
+++ b/docs_api/index.rst
@@ -35,6 +35,7 @@ Contents
     h5l
     h5o
     h5p
+    h5pl
     h5r
     h5s
     h5t
--- a/h5py/__init__.py
+++ b/h5py/__init__.py
@@ -52,7 +52,7 @@ _register_lzf()
 
 # --- Public API --------------------------------------------------------------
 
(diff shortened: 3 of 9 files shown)


scopatz:
This looks good to me! Thanks @takluyver!

> there might be reasons to have a separate lock for the plugin search path

Does HDF5 have a lock of its own around this?

takluyver:
I can't see locking in a quick look at the HDF5 source code, though it's possible that I'm overlooking it in a macro or something. Maybe we should just document for now that it's very likely not thread-safe.

scopatz:
> Maybe we should just document for now that it's very likely not thread-safe.

Seems reasonable for now

scopatz:
Thanks @takluyver!

PR #1259, opened 2019-07-02
Only install numpy and cython when required

### LINKED ISSUE #1257: pip download h5py fails on Python 2.7
Trying to `pip download h5py` fails on Python 2.7. This is due to `setup.py` adding a setup-time requirement for numpy. This has now become a problem because numpy 1.17.0rc1 (Python 3 only) was released yesterday.

 * Operating System: RHEL6
 * Python version: 2.7
 * Where Python was acquired: near system python
 * h5py version: 2.9
 * HDF5 version: N/A

```
$ pip download h5py==2.9.0 --no-binary=h5py --no-cache-dir -d base_tmp/ --isolated             

Collecting h5py==2.9.0
  File was already downloaded base_tmp/h5py-2.9.0.tar.gz
    ERROR: Complete output from command python setup.py egg_info:
    ERROR: Traceback (most recent call last):
      File "/tmp/pip-download-otzrwv/h5py/setup.py", line 168, in <module>
        cmdclass = CMDCLASS,
      File "python2.7/site-packages/setuptools/__init__.py", line 144, in setup
        _install_setup_requires(attrs)
      File "python2.7/site-packages/setuptools/__init__.py", line 139, in _install_setup_requires
        dist.fetch_build_eggs(dist.setup_requires)
      File "python2.7/site-packages/setuptools/dist.py", line 717, in fetch_build_eggs
        replace_conflicting=True,
(..)
      File "/scratch/wra62962/files/temp/python27/lib/python2.7/site-packages/setuptools/sandbox.py", line 45, in _execfile
        exec(code, globals, locals)
      File "/tmp/easy_install-Zo5gcH/numpy-1.17.0rc1/setup.py", line 31, in <module>
    
    RuntimeError: Python version >= 3.5 required.
    ----------------------------------------
ERROR: Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-download-otzrwv/h5py/
```

Code causing issue:
https://github.com/h5py/h5py/blob/b887dacf18c10d262cebec4b98d2554ca50cb4f8/setup.py#L39-L43

`pip` calls `setup.py`with parameter `egg_info`.
I would suggest to either
* add `egg_info` to the white list resulting in `use_setup_requires=False`, or (preferably imo)
* changing the logic to `use_setup_requires=False` to be the default and setting it to `True` only in those cases where you explicitly want it to be set.

Anthchirp:
Change from a negative list to a positive parameter list, as to exclude
odd setup.py parameters such as 'egg_info', which is used when
downloading dependencies via pip. Resolves #1257.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,11 +36,8 @@ SETUP_REQUIRES = RUN_REQUIRES + [NUMPY_DEP, 'Cython>=0.23', 'pkgconfig']
 
 # Needed to avoid trying to install numpy/cython on pythons which the latest
 # versions don't support
-if ("sdist" in sys.argv and "bdist_wheel" not in sys.argv and
-    "install" not in sys.argv) or "check" in sys.argv:
-    use_setup_requires = False
-else:
-    use_setup_requires = True
+use_setup_requires = any(parameter in sys.argv for parameter in
+    ("bdist_wheel", "build", "install", "test"))
 
 
 # --- Custom Distutils commands -----------------------------------------------
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ SETUP_REQUIRES = RUN_REQUIRES + [NUMPY_DEP, 'Cython>=0.23', 'pkgconfig']
 # Needed to avoid trying to install numpy/cython on pythons which the latest
 # versions don't support
 use_setup_requires = any(parameter in sys.argv for parameter in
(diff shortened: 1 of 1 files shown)


aragilar:
`configure` also requires the setup deps, but otherwise this looks good to me.

takluyver:
Thanks!

PR #1260, opened 2019-07-03
Modify interface for external storage

stanwest:
This pull request implements some suggestions for changes to the external storage interface. The significant changes are as follows:

-   Accepts fewer structures of the `external` argument to `Group.create_dataset`. The code here excludes the currently accepted forms
    
    *   list containing *name*, [*offset*, [*size*]],
    *   list containing *name1*, *name2*, … (which `H5Pset_external` rejects anyway), and
    *   list containing tuples such as `(name,)` and `(name, offset)` that lack the offset or size.
    
    It continues to accept the forms
    
    *   *name* and
    *   list of `(name, offset, size)` tuples.

-   Accepts more types within some parts of the `external` argument:

    *   The top-level container may be any iterable instead of only a list.
    *   The *name* may now be not only a `str` but also a `bytes` or `os.PathLike` object.
    *   The *offset* and *size* may be numpy integers as well as Python `int`s.

    However, this code requires a tuple as the container for *name*–*offset*–*size* triplets.

-   In documentation and exception messages, the PR refers to the first tuple element (the name or path of an external file) consistently as "name", like HDF's `H5Pset_external` and `h5p.set_external`, instead of "file" or "filename".

The older commits contain small fixes to code or documentation and could be applied independently of this pull request.

### DIFF
--- a/h5py/tests/test_dataset.py
+++ b/h5py/tests/test_dataset.py
@@ -600,12 +600,12 @@ class TestExternal(BaseDataset):
         self.f.create_dataset('foo', (6, 100),
                               external=pathlib.Path(self.mktemp()))
 
-    def test_multi(self):
-        """ External argument may contain multiple tuples """
+    def test_iter_multi(self):
+        """ External argument may be an iterable of multiple tuples """
 
         ext_file = self.mktemp()
         N = 100
-        external = [(ext_file, x * 1000, 1000) for x in range(N)]
+        external = iter((ext_file, x * 1000, 1000) for x in range(N))
         dset = self.f.create_dataset('poo', (6, 100), external=external)
         assert len(dset.external) == N
 
--- a/h5py/tests/test_dataset.py
+++ b/h5py/tests/test_dataset.py
@@ -591,9 +591,8 @@ class TestExternal(BaseDataset):
         ext_file = self.mktemp()
 
         self.f.create_dataset('foo', shape, external=ext_file)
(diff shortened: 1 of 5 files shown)


takluyver:
Thanks, this sounds sensible. As it could technically break some code (although I think that's unlikely), I suggest we land it for h5py 3.0.

scopatz:
Please add a news entry, as per the instructions at https://regro.github.io/rever-docs/news

stanwest:
Okay; I've allocated the changes among the "New features" and "Deprecations" sections.

h5py/tests/test_dataset.py
```diff
@@ -579,43 +583,50 @@ def test_external(self):
...
-    def test_external_other(self):
-        """ Test other forms of external lists """
+    def test_name_str(self):
+        """ External argument may be a file name str only """
 
-        shape = (6, 100)
-        ext_file = self.mktemp()
+        self.f.create_dataset('foo', (6, 100), external=self.mktemp())
+
+    @ut.skipIf(pathlib is None, "pathlib module not installed")
```
takluyver: We don't need to handle pathlib being absent now - the minimum Python version we're supporting is 3.6, so we can assume pathlib is always there.

takluyver:
Thanks @stanwest

stanwest:
Sure, and thank you for your helpful input along the way.

Issue #181, opened 2012-12-27
Error in derferencing references stored in datasets


andrewcollette:
_Original author: satyakis...@gmail.com (June 03, 2011 08:42:40)_

```python
f = h5py.File(r'C:\Users\skishore\Desktop\test.h5')
f.create_group('test')
d = f.create_dataset('test_dset',(1,),dtype=h5py.special_dtype(ref=h5py.Reference),maxshape=(None,))
d[0] = f['test'].ref
group_ref = d[0]
f[group_ref]
```

this generates an error like 
Traceback (most recent call last):
  File &quot;&lt;interactive input&gt;&quot;, line 1, in &lt;module&gt;
  File &quot;C:\Python27\lib\site-packages\h5py_hl\group.py&quot;, line 121, in **getitem**
    oid = h5r.dereference(name, self.id)
  File &quot;h5r.pyx&quot;, line 67, in h5py.h5r.dereference (h5py\h5r.c:1203)
TypeError: Argument 'ref' has incorrect type (expected h5py.h5r.Reference, got numpy.object_)

operating system: Windows 7 enterprise
Python version  : 2.7.1
h5py version    : 1.4.0 beta2

_Original issue: http://code.google.com/p/h5py/issues/detail?id=181_

PR #1262, opened 2019-07-09
Add ability to manually specify opaque types mapped to NumPy.

hameerabbasi:
Add create_opaque (see #1232 for discussion).

cc @scopatz @takluyver @aragilar


h5py/h5t.pyx
```diff
@@ -1647,12 +1652,19 @@ cpdef TypeID py_create(object dtype_in, bint logical=0, bint aligned=0):
         of kind "O" representing a string, it would return an HDF5 variable-
         length string type.
     """
-    cdef dtype dt = dtype(dtype_in)
-    cdef char kind = dt.kind
+    cdef dtype dt
+    cdef char kind
 
-    aligned = getattr(dtype_in, "isalignedstruct", aligned)
+    if hasattr(dtype_in, 'id'):
```
takluyver: `id` is quite generic - if we're checking for `Datatype`, it's probably better to do that explicitly with `isinstance()`.

h5py/h5d.pyx
```diff
@@ -208,7 +208,7 @@ cdef class DatasetID(ObjectID):
         cdef int oldflags
 
         if mtype is None:
-            mtype = py_create(arr_obj.dtype)
+            mtype = arr_obj.id.get_type()
```
takluyver: I think this is the line that's causing the test failure - `arr_obj` is (or can be) a Numpy array.

I think we'd need a separate branch here, something like:

```python
ftype = self.get_type()
if isinstance(ftype, TypeOpaqueID) and ftype.get_tag() == b'NUMPY:' + arr_obj.dtype.descr[0][1].encode():
    mtype = ftype
```

(Of course, if `mtype` is passed, that should still override this check)

h5py/h5t.pyx
hameerabbasi: That means that low-level will have a dependency on high-level, which I'd like to avoid.

h5py/h5d.pyx
hameerabbasi: This didn't work. I opted for `isinstance(ftype, TypeID)` instead, but there were still test failures... I'd like to avoid introducing special cases.

h5py/h5t.pyx
takluyver: This involves a logical dependency on the high-level object anyway - it has to know that such an object might exist, and how to get the low-level object from it. Using `isinstance()` just makes it explicit that what's going on.

I agree with the idea of keeping the high-level API out of the low-level. But in that case, this function shouldn't accept high-level objects at all - it should accept TypeID, not Datatype.

hameerabbasi: Another option is to find out where `Datatype` is being passed in and use `TypeID` instead.

h5py/h5d.pyx
takluyver: Were they different test failures?

In any case, `arr_obj` being a numpy array is not a special case - that's the normal case that the function needs to handle.

h5py/h5t.pyx
takluyver: Possibly the easiest way to do that is to let `create_opaque` return a TypeID. The high-level `Datatype` class is only described as being used for named types, and it doesn't really do anything extra on top of TypeID.

Or you could find all the places that the high-level API could pass down a Datatype to the low-level and extract the TypeID instead. That would be a more technically clean way of doing it, but also more work.

hameerabbasi:
I got writing working; but reading now segfaults.

h5py/_hl/dataset.py
```diff
@@ -505,7 +505,7 @@ def __getitem__(self, args):
             # This is necessary because in the case of array types, NumPy
             # discards the array information at the top level.
             new_dtype = readtime_dtype(self.id.dtype, names)
-        mtype = h5t.py_create(new_dtype)
+        mtype = self.id.get_type()
```
takluyver: I suspect this is too simple. It's ignoring what `readtime_dtype` does for the code that was there before, and it isn't affected by `astype`, which the code there before was.

takluyver: I think what `readtime_dtype` does is relevant to `test_compound_vlen_bool`, which is where it's segfaulting in the test run I looked at.

hameerabbasi: Do you have a way around this? I can't seem to find one.

takluyver: I don't fully follow the reasons (see notes below), but I think we need to accept that this code path converts the h5py type to a numpy dtype and back.

If we want to allow reading and writing datasets with a datetime dtype, but not creating those datasets by assigning a numpy array directly, `py_create` (converts dtype to HDF5 TypeID) needs to know the difference between those cases. There's an existing `logical=` flag which seems to be true when creating a dataset and false when reading/writing an existing one, but I'm not sure if it makes sense to reuse that or declare a separate one.

I'd really like more input from other people about how this should work. Maybe this `opaque_dtype` design is more trouble than it's worth. I'll ask on the list.

-----

**Notes from diving into this code**

I don't really understand the comment about 'numpy discards the array information at the top level'. It dates back over 10 years (5a8d34a1a002acddfc1b31c5526a6e3d4de26704), and does not refer to `readtime_dtype`, so presumably it's about why we need to convert the HDF5 type to a numpy dtype and back.

`readtime_dtype` is from only 8 years ago (d8490c808334a5f873982240ac6944ea92e93c0b). The commit message refers to issue 181, but I don't know if that's the same #181 that is now in Github.

`mtype` gets passed to `DatasetID.read()`, and used by HDF5 for type conversion. So it has to be an HDF5 description of the data type of the numpy array we're reading *into*, which may differ from the HDF5 type of the data on disk.

takluyver: I've asked the list for more input: https://groups.google.com/forum/#!topic/h5py/BLIVHygU2GU

h5py/h5t.pyx
scopatz: Yeah, almost everything in HDF5 has an `id` so this is too generic to disambiguate

scopatz: Maybe this clause could be moved below the current `elif`

takluyver:
I've rebased this and tried to continue it in #1339.

scopatz:
:+1: CC @hameerabbasi

takluyver:
Closing as superseded by #1339

PR #1263, opened 2019-07-10
Fix: raise ValueError when auto-chunking a scalar dataset

mpaolini:
instead of crashing with 

```
----> 7     outp.create_dataset('events', dtype='S1000', chunks=True)


~/devel/feedstock/ep2019/virtual/lib/python3.6/site-packages/h5py/_hl/group.py in create_dataset(self, name, shape, dtype, data, **kwds)
    134 
    135         with phil:
--> 136             dsid = dataset.make_new_dset(self, shape, dtype, data, **kwds)
    137             dset = dataset.Dataset(dsid)
    138             if name is not None:

~/devel/feedstock/ep2019/virtual/lib/python3.6/site-packages/h5py/_hl/dataset.py in make_new_dset(parent, shape, dtype, data, chunks, compression, shuffle, fletcher32, maxshape, compression_opts, fillvalue, scaleoffset, track_times, external, track_order, dcpl)
    138         dcpl or h5p.create(h5p.DATASET_CREATE), shape, dtype,
    139         chunks, compression, compression_opts, shuffle, fletcher32,
--> 140         maxshape, scaleoffset, external)
    141 
    142     if fillvalue is not None:

~/devel/feedstock/ep2019/virtual/lib/python3.6/site-packages/h5py/_hl/filters.py in fill_dcpl(plist, shape, dtype, chunks, compression, compression_opts, shuffle, fletcher32, maxshape, scaleoffset, external)
    204     (chunks is None and any((shuffle, fletcher32, compression, maxshape,
    205                              scaleoffset is not None))):
--> 206         chunks = guess_chunk(shape, maxshape, dtype.itemsize)
    207 
    208     if maxshape is True:

~/devel/feedstock/ep2019/virtual/lib/python3.6/site-packages/h5py/_hl/filters.py in guess_chunk(shape, maxshape, typesize)
    303 
    304     # For unlimited dimensions we have to guess 1024
--> 305     shape = tuple((x if x!=0 else 1024) for i, x in enumerate(shape))
    306 
    307     ndims = len(shape)

TypeError: 'NoneType' object is not iterable
```

### DIFF
--- a/h5py/_hl/filters.py
+++ b/h5py/_hl/filters.py
@@ -298,6 +298,9 @@ def guess_chunk(shape, maxshape, typesize):
     """
     # pylint: disable=unused-argument
 
+    if shape is None:
+        raise ValueError("Chunks not allowed for empty datasets.")
+
     # For unlimited dimensions we have to guess 1024
     shape = tuple((x if x!=0 else 1024) for i, x in enumerate(shape))
 
--- a/h5py/tests/old/test_dataset.py
+++ b/h5py/tests/old/test_dataset.py
@@ -236,6 +236,12 @@ class TestCreateChunked(BaseDataset):
         dset = self.f.create_dataset('foo', shape=(3,), dtype='S100000000', chunks=True)
         self.assertEqual(dset.chunks, (1,))
 
+    def test_auto_chunks_no_shape(self):
+        """ Auto-chunking of empty datasets not allowed"""
+        with pytest.raises(ValueError) as err:
+            self.f.create_dataset('foo', dtype='S100', chunks=True)
+        assert 'empty' in str(err.value)
+
 
 class TestCreateFillvalue(BaseDataset):
 


takluyver:
I think you're actually creating an [empty dataset](http://docs.h5py.org/en/stable/high/dataset.html#creating-and-reading-empty-or-null-datasets-and-attributes) (no data), rather than a scalar dataset (which has a single value). There's already code checking this for scalar datasets:

https://github.com/h5py/h5py/blob/596748d52c351258c851bb56c8df1c25d3673110/h5py/_hl/filters.py#L113-L119

I think the checks for scalar and empty datasets should probably be unified, as they presumably have similar restrictions.

mpaolini:
@takluyver good point, fixed!

takluyver:
I think we should do this as part of the existing check for scalar datasets that I pointed out; they're probably subject to the same restrictions, so it makes sense to have one set of code doing the checks and throwing the errors.

scopatz:
Agreed with @takluyver

mpaolini:
now that I read this again I think the problem was I didn't pass the `shape=()` as per the docs:

```
        shape
            Dataset shape.  Use "()" for scalar datasets.  Required if "data"
            isn't provided.
```

the problem is the sanity check is not enforced by the top level function `create_dataset` , that just passes a `shape=None` down to the called functions

I am not sure if this needs a fix at all: documentation is right. Maybe it is better to raise a ValueError in `create_dataset` ?

takluyver:
I think `shape=()` indicates a scalar dataset (zero dimensions, one piece of data), and `shape=None` indicates an empty dataset (zero dimensions, no data).

mpaolini:
the docs say `shape` is required in case `data` is not provided

shouldn't we just raise a validation error in here when shape is `None` and data is not provided?

takluyver:
[Another part of the docs](http://docs.h5py.org/en/stable/high/dataset.html#creating-and-reading-empty-or-null-datasets-and-attributes) shows creating an empty dataset by passing neither data nor shape. So for now, I think we have to assume that people are doing that and it needs to keep working. We could redesign it for 3.0 if makes things especially awkward.

takluyver:
I've rebased this and made the change I was suggesting (to integrate it with existing checks) in #1369

PR #1265, opened 2019-07-15
Make AttributeManager.create consistent with __setitem__

### LINKED ISSUE #1057: difference in behavior when writing unicode string to attrs (setter vs .create API)
Hi. 
I'm aware that there are many issues already open about writing lists of unicode strings to h5py attributes. But I found that the same problem exists when I try to write a single unicode string using the `.create` API, while it works fine when I write the string using the setter API.

This works:
`h5f["toto"].attrs["toto"] = u"toto"`

This causes an error:
`h5f["toto"].attrs.create("toto", data=u"toto")`
`TypeError: No conversion path for dtype: dtype('<U4')`

Is there a reason for this difference in behavior? Or would it be possible to also add the `if dtype is None: dtype = base.guess_dtype(value)` line to the `create` method, like it is done in `__setitem__` ? (https://github.com/h5py/h5py/blob/master/h5py/_hl/attrs.py#L102)

h5py    2.8.0
HDF5    1.10.2
Python  2.7.9 (default, Jun 29 2016, 13:08:31) 
[GCC 4.9.2]
sys.platform    linux2
numpy   1.14.5

t20100:
This PR moves the call to `base.guess_dtype` from `AttributeManager.__setitem__` to the beginning of `AttributeManager.create` so that both ways handle the `dtype` in a consistent way.

This closes #1057

### DIFF
--- a/docs/whatsnew/2.10.rst
+++ b/docs/whatsnew/2.10.rst
@@ -76,6 +76,7 @@ Bugfixes
   systems (:issue:`1235`).
 - Documented that nested compound types are not currently supported
   (:issue:`1236`).
+- Fixed attribute ``create`` method to be consistent with ``__setattr__`` (:issue:`1265`).
 
 Building h5py
 -------------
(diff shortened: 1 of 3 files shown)


scopatz:
This looks good to me.  Can you please add something to the "whatisnew" column?

t20100:
I updated whatsnew for 2.10.

scopatz:
Thanks!

PR #1267, opened 2019-07-15
Fix python setup.py test

### LINKED ISSUE #1266: python setup.py test not working
Unless I missed something, I cannot run the tests as in the [doc](http://docs.h5py.org/en/stable/contributing.html#adding-a-function-from-the-hdf5-c-api) with `python setup.py test` because `pytest` tries to import `h5py` from the source directory rather than from `build/lib-*`.

h5py    2.9.0.post0
HDF5    1.10.5
Python  3.7.2 (default, Feb 26 2019, 14:37:24) 
[GCC 4.9.2]
sys.platform    linux
numpy   1.16.3

t20100:
This PR changes the current working directory in `setup.py` before calling `h5py.run_tests` in order to load `h5py` from `build/lib-*` during the test.
Changing `sys.path` is not enough since tests are run as a subprocess.
`python -m pytest` (used by `run_tests`) adds the current working directory to `sys.path` which makes it works.

closes #1266

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -70,12 +70,17 @@ class test(Command):
         buildobj.run()
 
         oldpath = sys.path
+        oldcwd = os.getcwd()
+        build_lib_dir = op.abspath(buildobj.build_lib)
         try:
-            sys.path = [op.abspath(buildobj.build_lib)] + oldpath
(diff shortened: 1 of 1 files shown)


scopatz:
This seems reasonable. Thanks

PR #1278, opened 2019-08-01
Get low-level AttrID from high-level attributes interface

### LINKED ISSUE #797: h5a.AttrID for an attribute through the high-level API
First, thanks a lot for also offering a low-level API in addition to the high-level one.

In my case I'm making a small validation tool which validates a directory/file based storage format of ours. Among the things the tool has to do is to verify some constraints on HDF5 files (data type, dimensionality, size of datasets and attributes). In doing this, it's important that I can query the exact HDF5 type of a dataset or attribute, since we have tools in C/C++ that also must read these files, and they are more rigid than h5py, so the tool can't just inspect the numpy dtype and call it a day, but must know the underlying type.

For datasets I was able to use `data_set.id.get_type()` to get the low-level type for a high-level `Dataset`, and compare that to constants in `h5t`. But for attributes, I still haven't found a public API way to go from an attribute on a high-level `Dataset`, to a `AttrID` on which I could call `get_type` to get the low-level type.

I may have missed something, but if it's not possible, it would be great if it could be added. Perhaps as a new type of query method on `AttributeManager`?

takluyver:
With a new public method:

```python
aid = ds.attrs.get_id('a')
aid.get_type()  # example AttrID method
```

Closes #797.

### DIFF
--- a/docs/high/attr.rst
+++ b/docs/high/attr.rst
@@ -91,6 +91,11 @@ Reference
 
         Retrieve `name`, or `default` if no such attribute exists.
 
+    .. method:: get_id(name)
+
+       Get the low-level :class:`AttrID <low:h5py.h5a.AttrID>` for the named
+       attribute.
+
     .. method:: create(name, data, shape=None, dtype=None)
 
         Create a new attribute, with control over the shape and type.  Any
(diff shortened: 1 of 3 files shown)


scopatz:
Thanks @takluyver!

Issue #1281, opened 2019-08-12
Move/rename h5py.tests.old


takluyver:
PR #484 (in 2014) moved all the tests existing up to that point into a subpackage called `old`. I feel this naming is confusing:

- It implies that those tests are deprecated, when they still seem to test many things which aren't tested anywhere else, and I'm not aware of any plan to migrate away from them (please correct me if there is one).
- There are often two places to look for existing tests, or where to add new tests. E.g. tests for slicing datasets are in both `h5py.tests.old.test_slicing` and `h5py.tests.hl.test_dataset_getitem`.

I propose moving all the tests modules into a flat namespace in `h5py.tests`. Or if we need some more structure, organise them more descriptively (e.g. high-level vs low-level).

aragilar:
I'd think this is a great idea! I'd suggest keeping the VDS tests in their own directory, since they're currently split out nicely anyway, but I think keeping it mostly flat should make things simpler.

takluyver:
Done in #1283

PR #1283, opened 2019-08-14
Flatten h5py.tests package

takluyver:
See #1281. As @aragilar suggested, I've left the virtual dataset tests in their separate subpackage, but removed the distinction between `hl` and `old`.

Two module names existed in both hl and old:

- `test_file`: The copy from hl is called `test_file2` for now. Uninspired but effective.
- `test_datatype`: only the one from old actually tested the `Datatype` class, so the other is renamed to `test_dtype`.


takluyver:
I've checked one run (the first to finish) on each of Travis & Appveyor to see if any tests were missed after the move. They both show the same number of tests passed/skipped/xfail/warnings as for the latest commit on master, so I think it's OK.

takluyver:
Merging on one approval and no objections :-)

Issue #1285, opened 2019-08-17
MPI tests fail at test_close_multiple_mpio_driver


RizzerOnGitHub:
When h5py is built with mpi, run_tests() fail:
    
    mpirun -n 2 python3 -c "import h5py.tests; h5py.tests.run_tests(verbosity=2)"
    test_close (h5py.tests.old.test_file.TestClose)
    Close file via .close method ... ok
    test_close_multiple_default_driver (h5py.tests.old.test_file.TestClose) ... ok
    test_close_multiple_mpio_driver (h5py.tests.old.test_file.TestClose)
    MPIO driver and options ... ok
    test_close (h5py.tests.old.test_file.TestClose)
    Close file via .close method ... ok
    test_close_multiple_default_driver (h5py.tests.old.test_file.TestClose) ... ok
    test_close_multiple_mpio_driver (h5py.tests.old.test_file.TestClose)
    MPIO driver and options ... mca_fbtl_posix_pwritev: error in writev:Bad file descriptor

The log shows that the first processor closed the mpio file successfully, while the second processor emitted `error in writev:Bad file descriptor`.

I expect this is caused by the [double `f.close()`](https://github.com/h5py/h5py/blob/e3e0a06735038b7de9561f5df5c37a70783e15d9/h5py/tests/old/test_file.py#L590) in `test_close_multiple_mpio_driver `.  Is there a reason for running `f.close()` twice?  It doesn't seem to be doing the mpio test any good.

This is testing on Debian, 
- Linux amd64 4.19.37-6
- python3-h5py 2.9.0-5
- libhdf5-openmpi-103 1.10.4+repack-10
- libopenmpi3 3.1.3-11

Summary of the h5py configuration
---------------------------------

h5py    2.9.0
HDF5    1.10.4
Python  3.7.4 (default, Jul 11 2019, 10:43:21) 
[GCC 8.3.0]
sys.platform    linux
sys.maxsize     9223372036854775807
numpy   1.16.2

RizzerOnGitHub:
The double `f.close()` is beside the point.  The error comes from the use of `mktemp()`.  It hasn't been parallelized, so a different filename is given to the h5py File on each process. But with mpio used, the File object expects the same filename, so it emits the error.

mpi4py gives a pattern to use at https://bitbucket.org/mpi4py/mpi4py/src/master/test/test_io.py

Here is a patch.  For good measure, the patch replaces mktemp (which is deprecated, possibly unsafe) with mkstemp.

[...]

PR #1289, opened 2019-08-25
use MPI-aware temp files for mpio tests

RizzerOnGitHub:
MPI (driver=mpio) file tests currently hang since mktemp() is used to
generate a temporary file which is different for each process. mpio
files must have the same name for all processes.

This patch ensures only one temporary file is generated for mpio tests
and used by all processes.  Pattern adapted from mpi4py tests, see
https://bitbucket.org/mpi4py/mpi4py/src/master/test/test_io.py

Closes: #1285


scopatz:
Thanks!

PR #1296, opened 2019-09-07
Require Python >= 3.5

takluyver:
As discussed in #1177, we're looking to drop Python 2 support now that 2.10 is out. I'm in favour of doing so for 2.11 - I know it seems like a breaking change, but the metadata should mean that most Python 2 users automatically get the last compatible version (`pip install h5py` is smart enough to work this out).

While we're at it, Python 3.4 is also end-of-life, so we may as well drop it. I'd be open to dropping 3.5 as well, but this PR doesn't yet.

Pulling out compatibility code can come later - this just updates the package metadata and trims the CI jobs we no longer need.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,25 +30,6 @@ matrix:
     # needed to work around https://github.com/travis-ci/travis-ci/issues/4794
     # TODO: We should see if we can replace this with installing python via pyenv # based on toxenv
 
-    # Python 3.5
-    - python: 3.5
-      env:
-      - TOXENV=py35-test-deps
-    - python: 3.5
-      env:
-      - TOXENV=py35-test-mindeps
-    - python: 3.5
-      env:
-      - TOXENV=py35-test-deps
-        TOX_TESTENV_PASSENV=LANG LC_ALL
-        LANG=C
-        LC_ALL=C
-    - python: 3.5
-      env:
-      - TOXENV=py35-test-deps
-        HDF5_VERSION=1.10.1
-        HDF5_DIR=$HDF5_CACHE_DIR/$HDF5_VERSION
-
     # Python 3.6
     - python: 3.6
       env:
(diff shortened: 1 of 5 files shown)


tacaswell:
I think we should follow NEP 29 and drop 3.5 as well. (https://numpy.org/neps/nep-0029-deprecation_policy.html)

takluyver:
That sounds reasonable to me - I'll make that change in a day or two if no-one objects.

aragilar:
I think dropping 3.5 is fine (latest Ubuntu LTS, which is usually a good barometer of where things are, has 3.6 as the default py3). Did we want to make a corresponding increase in the minimum numpy and cython versions (the NEP calls out numpy, but nothing about cython)?

tacaswell:
cython is only a compile time dependency so I think it is less worrying.

I know that for 3.8 the minimum cython is going to be very recent (as older versions do not generate c-code that compiles due to change to some of python internal data structures).

takluyver:
- I have increased the minimum supported version to 3.6 - we seem to have a consensus in favour of that.
- I added Python 3.7 to the Appveyor tests, which only went up to 3.6 before.
- The coverage drop is expected, because there's Python 2 specific code which is no longer being executed. I think that should be resolved in a separate PR.
- I don't have a preference on numpy support. The minimum version currently supported for Python 3.6 is 1.12, whereas if we strictly followed NEP 29 we would currently be supporting numpy >= 1.14. Is there any code we could simplify, or any workarounds we could eliminate with a newer numpy? If not, maybe just leave it until we drop 3.6 as well.
- The minimum supported Cython is now 0.23, from 2015. I'm fine with increasing this whenever we think it's useful.

scopatz:
This LGTM!

tacaswell:
Sigh, failed to edit the commit message...

Issue #1300, opened 2019-09-10
Document release process


takluyver:
#1290 added release automation with Rever, but this doesn't (yet?) include building wheels, which is done from [this repository](https://github.com/MacPython/h5py-wheels) for Linux & Mac, and by Christoph Gohlke for Windows.

@tacaswell also mentioned that the release process is based on https://matplotlib.org/devdocs/devel/release_guide.html#all-releases

takluyver:
Also, how are the API docs (http://api.h5py.org/) updated? I've just noticed that they don't yet have the H5PL module added in 2.10?

scopatz:
> but this doesn't (yet?) include building wheels, which is done from this repository for Linux & Mac, and by Christoph Gohlke for Windows.

We can and should add this to our rever configuration.  It just wasn't fully clear to me how to do this.  Is this commit basically all that needs to happen? (CC @tacaswell)

https://github.com/MacPython/h5py-wheels/commit/48d6411e4ca000ac3674e3a6261e884abf10649c

> Also, how are the API docs (http://api.h5py.org/) updated? I've just noticed that they don't yet have the H5PL module added in 2.10?

Yeah, not sure about that either.  If we know how I can add it to rever too.

takluyver:
>  Is this commit basically all that needs to happen?

I think in principle it's that, then wait for the wheels to be created, download them from http://wheels.scipy.org/ (or the [https equivalent](https://3f23b170c54c2533c070-1c8a9b3114517dc5fe17b7c3f8c63a43.ssl.cf2.rackcdn.com/) to be safe), and upload them to PyPI. Presumably with some changes to the MacPython repository, it could upload directly to PyPI - maybe using one of the new API tokens.

It looks like it needs other changes not infrequently, but if the build/upload fails we can always do those manually and get a wheel once it's fixed. It's only a problem if it succeeds in uploading a bad wheel.

scopatz:
OK! We can definitely have rever push a change into that repo, or open up a PR to it.  Which do we want to happen?

tacaswell:
I vote for push directly (to a new branch named for the minor series).  Either it builds or it doesn't!

If we are going to automate this with rever, I would prefer it pushes the commit and then downloads the built wheels to be pushed later.  If you push the sdist and the wheels to pypi in one shot you avoid the windows where you get weird build issues.

I have asked Andrew about the API docs over email, I will dig that out...

[...]

Issue #1299, opened 2019-09-07
No PGP signature on 2.10.0 release ?


ArchangeGabriel:
Normally we used to get the .asc file at https://files.pythonhosted.org/packages/source/h/h5py/h5py-2.10.0.tar.gz.asc, and but it is not available for this release. Is that expected?

tacaswell:
attn @scopatz

scopatz:
Sorry, this was missed.  @tacaswell - how are these normally generated?

tacaswell:
I use the `--sign` flag on twine when uploading the sdist.  I normally only sign the sdist (and not the wheels) because that is what I generate on hardware I control.  I think the asc file can be uploaded after the fact.

scopatz:
Ok! Thanks!

scopatz:
I think I need to do this on the machine I released from

scopatz:
I am a little confused here.  The asc files don't seem to be uploaded for previous versions: https://pypi.org/project/h5py/2.9.0/#files  How can I verify that an asc file gets pushed?

ArchangeGabriel:
Oh, I see, there is a new release manager. Hi @scopatz ! Don’t be sorry, I can guess it’s hard to get everything right the first time. ;)

Regarding your last question: they are never displayed by the UI at PyPi indeed. You have to try to download it using an URL like the one in my first post. For instance this works: https://files.pythonhosted.org/packages/source/h/h5py/h5py-2.9.0.tar.gz.asc. But the 2.10 one seems to still be missing.

scopatz:
Ahh Ok! Great.  So this should definitely work now with future releases that are done on rever v0.4.2+.  I'll try to upload the v2.10.0 sig file later tonight

takluyver:
@ArchangeGabriel : Are you actually verifying a signature from a specific key you trust? Or are you just verifying that the package has a valid signature from some key?

ArchangeGabriel:
Until now I was verifying with @tacaswell key(s) since he was the one signing the release. I don’t have definitive trust in it (e.g. I did not met in person with him to check he was owning the key), but an “accumulated” one: the releases always matched his key(s), they are the ones available at https://github.com/tacaswell.gpg, etc.

scopatz:
That's pretty much my opinion of release signing. We can reduce this to non-blocking if we want.

tacaswell:
I have been signing both h5py and Matplotlib for a few years now on the theory that it does not hurt anything.  

[...]

PR #1301, opened 2019-09-10
initial release guide

scopatz:
for #1300 
Also adds package signing for PyPI on rever v0.4.2+

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -70,5 +70,6 @@ Meta-info about the h5py project
 
     whatsnew/index
     contributing
+    release_guide
     faq
     licenses
--- a/docs/release_guide.rst
+++ b/docs/release_guide.rst
@@ -42,5 +42,3 @@ You probably want to make sure (with ``git tag``) that the new version
 number is available. If any release activities fail while running this
 command, you may safely re-run this command. You can also safely undo
 previously run activities. Please see the rever docs for more details.
-
-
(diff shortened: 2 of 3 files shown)


takluyver:
Thanks!

Should we discuss package signing? I think we either need to commit to taking it seriously - i.e. establishing, documenting and maintaining a set of keys which people should trust to make h5py releases - or scrap package signing. If we're not doing it properly, it's a complete waste of time, and worse, it gives an impression of security that doesn't exist.

takluyver:
I don't bother with package signing on any other project I release, so that's definitely my preference. My experience is that people only complain when a project that previously did sign releases stops, not when a project never started signing, so I don't think it's really important.

scopatz:
this passes CI now

aragilar:
The only thing preventing this from be merged seems to be the addition of signing to rever (which is discussed in #1299), I suggest leaving the signing (or non-signing) to a different PR.

scopatz:
@aragilar - would you like me to remove that one line then?

aragilar:
@scopatz Yep, we can argue about signing vs not signing in #1299, no need to leave this waiting for that. Also, can rever upload to test.pypi.org and/or do dry runs of releases, as that might be worth adding to the CI, so that we can quickly catch issues with the release process?

scopatz:
@aragilar done!

> Also, can rever upload to test.pypi.org and/or do dry runs of releases

That is a great idea! There is probably a way to do it, already, but adding that to the `rever check` for the pypi activity would be ideal.  Can you open an issue on rever for this please?

PR #1304, opened 2019-09-12
TST: try adding ppc64le

tacaswell:


scopatz:
Seems as though this fails :(

tacaswell:
Which is good because that is one of the failures reported in #1244 so at least it is consistent!

tacaswell:
Replaced by #1319 which also includes a fix!

PR #1306, opened 2019-09-15
DOC: minor changes to the documentation

tacaswell:
I suspect the last commit will be the only contentious one.


takluyver:
I think with the release automation, the WIP release notes are meant to be in `news/` - @scopatz, can you confirm exactly what we need there?

scopatz:
Yeah, you should follow the instructions here: https://regro.github.io/rever-docs/news

docs/whatsnew/3.0.rst
```diff
@@ -0,0 +1,12 @@
+What's new in h5py 3.0
+======================
```
scopatz: rever will create this file at release time, so no need to add it here

scopatz: (ie please remove it)

docs/whatsnew/index.rst
```diff
@@ -8,6 +8,7 @@ These document the changes between minor (or major) versions of h5py.
 
 .. toctree::
 
+    3.0
```
scopatz: Likewise, rever will add this link automatically, so please remove it

scopatz:
That all seems reasonable to me

takluyver:
I've just fixed a typo which would probably have caused an error when the release notes were built into docs.

We already have a CI job that builds docs - could we set that up to pull together release notes using rever first, so any invalid markup in the release notes is flagged automatically?

tacaswell:
@aragilar Thanks for cleaning up after me!  I've been super busy/sick since I got home.

@takluyver I agree, getting CI warnings about invalid rst as close to when it goes in is important.

Issue #1302, opened 2019-09-10
Benchmarking suite


takluyver:
h5py does complex operations for which people might well care about performance. It would therefore be good to create some benchmarks, using a tool like [asv](https://asv.readthedocs.io/en/stable/). This would be a way to detect any performance regressions, and to see the benefit of any performance improvements.

I'd expect it to cover different patterns of reading and writing data, e.g. with different layout and compression options, and using different kinds of broadcasting. The examples directory and issues about performance could be starting points.

[I'm opening this now as something people could work on at the h5py code camp next week]

kif:
I did not get very far with *asv*. anyway there is a need to a benchmark suite.

The first thing to do is to define a certain number of operation to perform:
* group creation
* dataset creation (many datatypes)
* attribute creation
* reading
* sliced reading 
[...]
I can start writing some of them. How long should it take in total ?
OK for using the timeit module ?
Then it could be interesting to evaluate the nogil path.

takluyver:
Maybe we can collaborate today on getting asv working. I think the list of things you mentioned is a good starting point - we'll probably want quite a few different cases for reading & writing data in different patterns e.g.:

- chunked vs contiguous layouts on disk
- slicing across data directions
- giving a list of indices
- writing with broadcasting
- type conversion

takluyver:
#1302 adds the configuration and a first benchmark. I'll leave this open as a meta issue for adding further benchmarks.

Adding benchmarks is something relatively straightforward that new contributors could do. See the [asv docs on writing benchmarks](https://asv.readthedocs.io/en/stable/writing_benchmarks.html). It might be useful to suggest what you plan to write a benchmark for first.

PR #1317, opened 2019-09-19
More efficient handling of scalar assignments.

vasole:
Related to #1067

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -688,11 +688,24 @@ class Dataset(HLObject):
             return
 
         # Broadcast scalars if necessary.
+        # In order to avoid slow broadcasting filling the destination by
+        # the scalar value, we create an intermediate array of the same
+        # size as the destination buffer provided that size is reasonable.
+        # We assume as reasonable a size smaller or equal as the used dataset
+        # chunk size if any.
+        # In case of dealing with a non-chunked destination dataset or with
+        # a selection whose size is larger than the dataset chunk size we fall
+        # back to using an intermediate array of size equal to the last dimension
+        # of the destination buffer.
+        # The reasoning behind is that it makes sense to assume the creator of
+        # the dataset used an appropriate chunk size according the available
+        # memory. In any case, if we cannot afford to create an intermediate
+        # array of the same size as the dataset chunk size, the user program has
+        # little hope to go much further. Solves h5py isue #1067
         if mshape == () and selection.mshape != ():
             if self.dtype.subdtype is not None:
                 raise TypeError("Scalar broadcasting is not supported for array dtypes")
             if self.chunks and (numpy.prod(self.chunks) >= numpy.prod(selection.mshape)):
-                # TODO should one use the destination dtype?
                 val2 = numpy.empty(selection.mshape, dtype=val.dtype)
             else:
                 val2 = numpy.empty(selection.mshape[-1], dtype=val.dtype)
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -691,7 +691,11 @@ class Dataset(HLObject):
         if mshape == () and selection.mshape != ():
             if self.dtype.subdtype is not None:
                 raise TypeError("Scalar broadcasting is not supported for array dtypes")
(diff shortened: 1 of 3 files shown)


takluyver:
This makes sense - presumably a chunk should always be a suitable size to exist in memory. Can we have a comment by this code explaining what's going on.

This will also hopefully make a good example for a benchmarking suite (#1302) - we should be able to have a benchmark for a write like this and show it becoming much more efficient.

vasole:
I have made the changes in line with the existing code, although in my opinion the dtype of the intermediate array should be the one of the destination and not the one of the source in order to avoid unnecessary multiple casting.

takluyver:
Is there any possibility that numpy casting could behave differently to HDF5 casting?

tacaswell:
I think this is a perfect application for the `iter_chunks` application that @jreadey was working on at the end of the day.  We probably want to special case this writing to be chunk aligned etc if the file system is chunked.

If I am reading this correctly, if I make a (10k, 10k, 10k) but chunked data set and set it to a scalar, this code will now try to allocate the full array to set it?

vasole:
If your chunk size is (10k, 10k, 10k) yes. If your chunk size is anything smaller like (1, 10k, 10k)  then it will revert to the previous behavior. The assumption is that we can allocate a single chunk in memory.

vasole:
The iter_chunks approach targets a more ambitious goal not just limited to scalars.

vasole:
@takluyver 

h5py and numpy use the same casting rules. My caveats are different: 

- in case of broadcasting the casting is made multiple times
- we can allocate more memory than absolutely necessary (for instance a selection in a boolean destination array assigned to 1 or 0)

@tacaswell 

I guess even when we'll iterate over chunks, we'll still need to allocate one chunk that is the maximum this PR contemplates.

jreadey:
I have a preliminary PR for the chunk iterator here: https://github.com/h5py/h5py/pull/1333.  
Does anything need to be changed for it to work with this PR?

takluyver:
@jreadey this PR doesn't depend on the chunk iterator. Once both are landed, we can do another PR to get the improvement for all chunked datasets.

I think the design of the chunk iterator in your PR will broadly work for that. Though we might want a convenient way to get the shape of each chunk - we can use `ds.chunks` for a central chunk shape, but it could be wrong for chunks at the edge of a selection.

It also comes back to my question about whether chunk iteration should be allowed for contiguous datasets. It would be easy to write something like this assuming a chunk can always fit in RAM, test it only on chunked datasets, and then get a MemoryError when you try to write to a contiguous dataset.

scopatz:
Thanks @vasole - Please also add a news entry, as per the instructions at https://regro.github.io/rever-docs/news.html

h5py/_hl/dataset.py
```diff
@@ -688,10 +688,28 @@ def __setitem__(self, args, val):
...
+        # memory. In any case, if we cannot afford to create an intermediate
+        # array of the same size as the dataset chunk size, the user program has
+        # little hope to go much further. Solves h5py isue #1067
         if mshape == () and selection.mshape != ():
             if self.dtype.subdtype is not None:
                 raise TypeError("Scalar broadcasting is not supported for array dtypes")
-            val2 = numpy.empty(selection.mshape[-1], dtype=val.dtype)
+            if self.chunks and (numpy.prod(self.chunks, dtype=numpy.float) >= \
+                                numpy.prod(selection.mshape, dtype=numpy.float)):
+                val2 = numpy.empty(selection.mshape, dtype=val.dtype)
```
tacaswell: This looks like it will allocate up to the size of the selection not the size of one chunk?

vasole: @tacaswell 

No, only if the selection is smaller than the chunk size it will allocate the size of the selection.

The product of the chunk dimensions has to be bigger or equal to the product of the selection dimensions. At most we'll allocate the number of elements of one chunk.

vasole:
@scopatz 

Hopefully I have filled the news properly. I'm not sure if I had to delete the non-filled entries.

scopatz:
@vasole - looks right to me!

takluyver:
Thanks @vasole!

PR #1324, opened 2019-09-19
CI: Add azure pipelines ci

aragilar:
Assuming the rebased version passes, this should be ready to merge. The commented out sections currently do not pass (for a number of reasons as noted below), but it's probably worth having a framework that works rather than waiting on those final fixes (and this adds macOS to the systems we are running a CI on)
Some comments:
* On some OSes/setups newer HDF5 fails, so we'll need to look into that
* Building HDF5 from source is failing on macOS for some reason, those tests are commented out, but we should work out why we fail to build on macOS
* Caching isn't set up yet (we may want to wait until the feature stabilises)
* I haven't added every test on travis yet, so don't stop testing on travis (I'll create a PR once this is merged which drop unneeded travis tests)
* This is currently only running 64-bit tests, I haven't worked out where exactly to specify bit-size yet, so don't stop testing on appveyor (I'll also create a PR to drop unneeded appveyor tests)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,228 @@
+# from matplotlib's azure setup
+
+variables:
+  HDF5_CACHE_DIR: $(Pipeline.Workspace)/cache/hdf5
+  PIP_CACHE_DIR: $(Pipeline.Workspace)/cache/pip
+  CCACHE_DIR: $(Pipeline.Workspace)/cache/ccache
+
+jobs:
+
+- job: 'ubuntu1604'
+  pool:
+    vmImage: ubuntu-16.04
+  strategy:
+    matrix:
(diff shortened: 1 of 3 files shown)


takluyver:
"35 successful checks" - Nice :-D

takluyver:
It turns out that with that many jobs, even Azure quickly gets bogged down once I've merged a few PRs. I think we should look at dropping some of the combinations, using some kind of caching so we're not rebuilding stable versions of HDF5 every time, and/or turning on the [batching option](https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml#batching-ci-builds) so multiple merges in quick succession don't all get run separately.

PR #1328, opened 2019-09-19
Added 'value attribute' to list of attributes

Helveg:
There was no mention of the value attribute. While this attribute can be usefull in many use cases where there's a difference between the data in the dataset, and a HDF5Dataset object

```
print(dset)
> <HDF5 dataset "dset": shape (), type "|S3130">
```
```
print(dset.value)
> b'\x80\x03csklearn.neighbors.kd_tree\nnewObj\nq\x00csklearn.neighbors.kd_tree...
```

### DIFF
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -15,6 +15,7 @@ NumPy operations like slicing, along with a variety of descriptive attributes:
   - **shape** attribute
   - **size** attribute
   - **dtype** attribute
+  - **value** attribute
 
 h5py supports most NumPy dtypes, and uses the same character codes (e.g.
 ``'f'``, ``'i8'``) and dtype machinery as


Helveg:
Oh nevermind, didn't see the deprecation warning! So then the `dset[()]` should be added somewhere in the docs

takluyver:
Yup, value is deprecated. The [reading and writing datasets](http://docs.h5py.org/en/latest/high/dataset.html#reading-writing-data) section describes various slicing possibilities. Though maybe it's worth adding a note that `[()]` or `[...]` can get all the data from any dataset.

tacaswell:
@Helveg Where would you have expected to find this information?

Helveg:
I intuitively went looking for some sort of Dataset object page, I feel
like blow "Creating datasets" there should be a section on manipulating
datasets, as I don't think that page even shows how to put data in a
dataset ;p

tacaswell:
@Helveg Could you put in a PR to https://github.com/h5py/h5py/blob/master/docs/high/dataset.rst with changes that would have made this clear to you in the first reading?

takluyver:
Closing in favour of #1330.

PR #1330, opened 2019-09-20
Added an explanation on manipulating datasets

Helveg:
As discussed in https://github.com/h5py/h5py/pull/1328

### DIFF
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -48,6 +48,25 @@ they will override ``data.shape`` and ``data.dtype``.  It's required that
 in ``data.shape``, and that (2) it's possible to cast ``data.dtype`` to
 the requested ``dtype``.
 
+Manipulating datasets
+---------------------
+
+The data in a dataset can be accessed by indexing or slicing and it can be manipulated by assigning new values to indices or slices.
+
+    >>> dset = f.create_dataset("my_dataset", data=[1,2,3,4,5])
+    >>> some_element = dset[3] # 4
+    >>> some_slice = dset[2:4] # [3, 4, 5]
+    >>> full_slice = dset[:] # [1, 2, 3, 4, 5]
+
+If you're dealing with a scalar dataspace or another datatype that does not support indexing or slicing, you can still recover the full data by using ``()`` as an index:
+
+    >>> dset = f.create_dataset("my_dataset", data=numpy.string_("A very compatible string."))
+    >>> my_string = dset[()]
+    
+The ``()`` index can also be used to assign a different data object to the dataset, provided it has the same shape or the dataset is resizable:
+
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks! I think this is substantially overlapping with the *Reading and writing data* section further down the page. We should enhance that and move it further up the page (before all the discussion of how HDF5 actually stores data), rather than adding another section, IMO.

Do you want to have a go at this?

Helveg:
Yea no, you're right, the "Reading and writing" section is exactly what I was looking for. It's just such basic information and what comes before it is so technical and specific that you assume it's not there. I'm not familiar with the syntax used in the docs but I think that if you just move up that section the document will be more clear

takluyver:
:+1: do you want to move the section up? The CI should check that the docs still build OK, though it's slow at the moment because we're doing a sprint (Hello from Grenoble!), so there are lots of changes happening.

If not, we'll leave this PR open as a reminder and someone here can work on the docs.

Helveg:
Sure!

Helveg:
@takluyver How do you make links to other sections with this doc format? I'd like to add a reference to creating empty datasets from "creating datasets" and to "fancy indexing" from "reading and writing data" :)

docs/high/dataset.rst
```diff
@@ -303,18 +314,6 @@ list of points to select, so be careful when using it with large masks::
    Selecting using an empty list is now allowed.
    This returns an array with length 0 in the relevant dimension.
 
-.. _dataset_iter:
```
takluyver: Can you preserve this line above the header? This makes a label for cross-linking.

To make internal links to a label, you'd write:

```rst
:ref:`dataset_iter` (using the text from the heading)
:ref:`iterating over datasets <dataset_iter>` (overriding it with custom text)
```

Helveg:
Ok I can get to preserve that, but it gave an error in the build of docs. I
changed the line below it from a section to a subsection and there were no
other references to it.

How would you add that line and avoid the error? I think the error said
that it required a caption (?) if it wasn't followed by a section header.

docs/high/dataset.rst
Helveg: Ok I can try to preserve that, but it gave an error in the build of docs. I changed the line below it from a section to a subsection and there were no other references to it. How would you add that line and avoid the error? I think the error said that it required a caption (?) if it wasn't followed by a section header.

takluyver: It should be OK if it's just above the header, like the other ones. The heading level (underline style) shouldn't matter.

You can try building the docs locally with `make html` if you have a local checkout. Or I can try to have a look at some point.

takluyver:
Thanks!

PR #1331, opened 2019-09-20
Dataset doc: add warning about __setitem function

### LINKED ISSUE #1277: Unexpected behaviour when indexing multiple dimensions
- OS: Linux Mint 19.2
- Python: 3.6.8
- h5py: 2.9.0

In numpy indexing via array[i,j] and array[i][j] are equivalent:
```
>>> a = np.random.random(size=3)
>>> b = np.zeros(shape=(2,2,3))
>>> b[0,0] = a
>>> b[0][0]
array([0.86413749, 0.64801094, 0.90344875])
>>> b[1][1] = a
>>> b[1,1]
array([0.86413749, 0.64801094, 0.90344875])

```
This behaviour is different when using h5py
```
>>> f = h5py.File("test_file")
>>> dset = f.create_dataset("test",(2,2,3))
>>> dset[0,0] = a
>>> dset[0,0]
array([0.8641375 , 0.64801097, 0.90344876], dtype=float32)
>>> dset[1][1] = a
>>> dset[1][1]
array([0., 0., 0.], dtype=float32)
```
It is noteworthy, that there is only a difference when using the notation for storing data, not for retrieving:
```
>>> dset[0][0]
array([0.8641375 , 0.64801097, 0.90344876], dtype=float32)
```

I found this quite unexpected, especially since the dataset simply remains unchanged without telling the user that no data is saved. This might be the correct behaviour but maybe some sort of notification would be beneficial.

payno:
/close #1277

### DIFF
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -103,34 +103,33 @@ safe to use with very large target selections.  It is supported for the above
    Currently h5py does not support nested compound types, see :issue:`1197` for
    more information.
 
+Multiple indexing
+~~~~~~~~~~~~~~~~~
 
-.. warning::
-    Dataset '[]' operator returns new numpy.ndarray.
-    Users chaining '[]' operators can find the following behavior surprising:
+Indexing a dataset once loads a numpy array into memory.
+If you try to index it twice to write data, you may be surprised that nothing
+seems to have happened:
 
    >>> f = h5py.File('my_hdf5_file.h5', 'w')
-   >>> dset = f.create_dataset("test",(2,2))
-   >>> dset[0][1] = 3.0
+   >>> dset = f.create_dataset("test", (2, 2))
+   >>> dset[0][1] = 3.0  # No effect!
    >>> print(dset[0][1])
-   >>> 0.0
+   0.0
 
-   dset[0] returns an in-memory array. This mean that the set item operation (``[1]``) will be operate on in-memory array, not the on-disk dataset
-   This is equivalent to:
+The assignment above only modifies the loaded array. It's equivalent to this:
 
-   >>> new_array = dset[1]
+   >>> new_array = dset[0]
    >>> new_array[1] = 3.0
    >>> print(new_array[1])
-   >>> 3.0
+   3.0
    >>> print(dset[0][1])
-   >>> 0.0
+   0.0
 
-   In comparaison calling directly the Dataset setter will give you:
+To write to the dataset, combine the indexes in a single step:
 
-   >>> dset[0,1] = 3.0
-   >>> print(dset[0,1])
-   >>> 3.0
-
-
+   >>> dset[0, 1] = 3.0
+   >>> print(dset[0, 1])
+   3.0
 
 .. _dataset_iter:
 
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -115,7 +115,7 @@ safe to use with very large target selections.  It is supported for the above
    >>> print(dset[0][1])
    >>> 0
 
-   dset[0] returns an in memory array. This mean that the set item operation ('[1]') will be operate in this in memory array.
+   dset[0] returns an in-memory array. This mean that the set item operation (``[1]``) will be operate on in-memory array, not the on-disk dataset
    This is equivalent to:
 
    >>> new_array = dset[1]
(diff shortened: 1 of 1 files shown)


payno:
I can also set the numpy array 'writable' flag to False (and remove this warning) If you like but I am not sure there is a consensus on this ?

tacaswell:
I don't think we want to set the mode to non-writable by default as there are good reasons for wanting to be able to mutate the extracted numpy array (for example to feed into routines that do in-place operations for efficiency reasons).

docs/high/dataset.rst
```diff
@@ -261,6 +261,34 @@ safe to use with very large target selections.  It is supported for the above
    more information.
 
 
+.. warning::
+    Dataset item getter returns new numpy.ndarray. Some mind find the following behavior 'unexpected':
```
tacaswell:
```suggestion
    Dataset item getter returns new numpy.ndarray. Some might find the following behavior surprising:

File: docs/high/dataset.rst [hunk 2]
```diff
@@ -261,6 +261,34 @@ safe to use with very large target selections.  It is supported for the above
...
+
+   >>> a = 1
+   >>> b = np.zeros(shape=(2,2))
+   >>> f = h5py.File(file_path, 'w') as f
+   >>> dset = f.create_dataset("test",(2,2))
+   >>> dset[0][1] = a
+   >>> print(dset[0][1])
+   >>> 0
+
+   dset[0] returns an in memory array. This mean that the set item operation ('[1]') will be operate in this in memory array.
```
tacaswell:
```suggestion
   dset[0] returns an in-memory array. This mean that the set item operation (``[1]``) will be operate on in-memory array, not the on-disk dataset
```

docs/high/dataset.rst
```diff
@@ -261,6 +261,34 @@ safe to use with very large target selections.  It is supported for the above
    more information.
 
 
+.. warning::
+    Dataset item getter returns new numpy.ndarray. Some might find the following behavior surprising:
```
takluyver: This doesn't quite read right - can you expand it a bit? I'm happy to have a go as a native english speaker if you prefer.

docs/high/dataset.rst
```diff
@@ -261,6 +261,34 @@ safe to use with very large target selections.  It is supported for the above
...
+
+   >>> a = 1
+   >>> b = np.zeros(shape=(2,2))
```
takluyver: b is unused? And I think it's clearer for the examples to use a literal number `1` rather than a variable `a`.

docs/high/dataset.rst
```diff
@@ -261,6 +261,34 @@ safe to use with very large target selections.  It is supported for the above
...
+   >>> f = h5py.File(file_path, 'w') as f
+   >>> dset = f.create_dataset("test",(2,2))
+   >>> dset[0][1] = a
+   >>> print(dset[0][1])
+   >>> 0
+
+   dset[0] returns an in-memory array. This mean that the set item operation (``[1]``) will be operate on in-memory array, not the on-disk dataset
+   This is equivalent to:
+
+   >>> new_array = dset[1]
```
takluyver: You switch from using coordinates `0, 1` to `1, 1` at this point.

I think it would be clearest to use two different coordinates, and a third number for the value you're setting.

takluyver:
I've rebased this, adjusted the wording a bit, and converted it to a subheading, because it will be useful to be able to link to this specific section.

docs/high/dataset.rst
```diff
@@ -103,6 +103,34 @@ safe to use with very large target selections.  It is supported for the above
    Currently h5py does not support nested compound types, see :issue:`1197` for
    more information.
 
+Multiple indexing
+~~~~~~~~~~~~~~~~~
+
+Indexing a dataset once loads a numpy array into memory.
+If you try to index it twice to write data, you may be surprised that nothing
+seems to have happened:
```
scopatz: Shouldn't these end in a double colon (`::`)

takluyver: Doesn't seem to be needed - I think Spinx recognises the `>>>` markers.

scopatz: Ahh ok

scopatz:
Thanks @payno!

PR #1332, opened 2019-09-20
Move cython code from Python2 syntax to Python3

kif:
Smooth transition from Cython language from 2->3
Initially, it was a attempt to do some more "nogil" action but they gave all been undone.

All test are now passing


kif:
While moving to Cython "3str" I found another "inconstancy" as there are many 
`from numpy cimport blabla`
Which now cimports stuff from numpy, not from the h5py/numpy.pxd

I believe the content of `h5py/numpy.pxd` should be replaced with the one provided by numpy

kif:
Hi Guys,
I think you can try to review. The PR is not as ambitious as initially targeted.
Cheers,
Jerome

api_gen.py
```diff
@@ -184,9 +199,9 @@ def wrapif(condition, code):
     def write_raw_sig(self):
         """ Write out "cdef extern"-style definition for an HDF5 function """
 
-        raw_sig = "{0.code} {0.fname}({0.sig}) except *\n".format(self.line)
+        raw_sig = "{0.code} {0.fname}({0.sig}) nogil except *\n".format(self.line)
```
takluyver: Is this accidentally leftover? I'd rather any changes that affect what the code does were separate from the more mechanical updating of the code.

docs/high/dataset.rst
```diff
@@ -48,6 +50,71 @@ they will override ``data.shape`` and ``data.dtype``.  It's required that
 in ``data.shape``, and that (2) it's possible to cast ``data.dtype`` to
 the requested ``dtype``.
 
+.. _dataset_slicing:
+
+Reading & writing data
+----------------------
```
takluyver: Possibly it needs a rebase so these changes which are already in master don't show up as part of the PR.

scopatz:
Thanks @kif! Please also add a news entry for this: https://regro.github.io/rever-docs/news

api_gen.py
kif: Hi Thomas,
No, it is not "accidentally" left-over. I can ensure you it causes no harm as the libhdf5 does not manipulate python object (except within callback but this should be handled by the callback).
Nogil function can be called with or without the gil; but only nogil functions can be called with the gil released.

I consider in a second PR to annotate every single function of the libhdf5 which does not use any (python-) callback to call them within a nogil section. Lots of manual reading, little code.

The last part and probably the most complicated is to duplicate all functions in _conv with a nogil variant to be used as callback in nogil sections, probably re-acquireing the gil.

docs/high/dataset.rst
kif: Done

h5py/_conv.pyx
```diff
@@ -1,5 +1,5 @@
 # cython: profile=False
-
+# cython: language_level=3
```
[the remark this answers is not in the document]
kif: Indeed, it is no more needed. I can remove it: it a matter of minutes to do. 
On the other hand the "Python2" remains the default until Cython3 comes out, so I would recommend leaving them until next year.

api_gen.py
takluyver: OK, sorry. I thought you had intended to drop all the nogil bits from this and just make the syntax changes.

I don't feel that I understand well enough how `nogil` interacts with Cython code, so I'll let someone else review that.

kif:
close #1168

stdeb.cfg
```diff
@@ -0,0 +1,7 @@
+[DEFAULT]
+Package: python-h5py      
+XS-Python-Version: >= 2.6
+Maintainer: Jerome Kieffer <jerome.kieffer@esrf.fr>
+Build-Depends: python-dev
+Recommends: python-numpy
```
takluyver: This doesn't seem related to the rest of the PR. And I'd rather leave Debian packaging out of the repo entirely.

aragilar:
This looks fine to me, though I don't really feel I understand the gil well enough (especially in relation to other C-extensions like numpy) to say whether we should be dropping it when calling into HDF5. Cython apparently has the option of specifying whether to drop the gil at compile time (https://cython.readthedocs.io/en/latest/src/userguide/external_C_code.html#conditional-acquiring-releasing-the-gil), I guess I'd be ok with using something like that to allow expert users (such as @kif) to understand how the gil affects their code, but to leave the default as not dropping the gil. Would that be acceptable to everyone?

kif:
Hi James,
Thanks for pointing me to the conditional GIL. It is actually useful in the case I am investigating.
Anyway, disabling the GIL will be part of another PR.
Cheers, 
Jerome

api_gen.py
takluyver: I'd be happy to merge this quickly except for this change, which would need me to spend time reading up on Cython. Is it necessary for the rest of the PR? If not, can I suggest that we revert this bit and do nogil work in a separate PR from the syntax changes.

kif: As you wish

h5py/api_types_ext.pxd
```diff
@@ -14,40 +14,20 @@ include 'config.pxi'
...
-  char *strcpy(char *dest, char *src)
-  char *strncpy(char *dest, char *src, size_t n)
-  int strcmp(char *s1, char *s2)
-  char *strdup(char *s)
-  void *memcpy(void *dest, void *src, size_t n)
-  void *memset(void *s, int c, size_t n)
-
-cdef extern from "time.h":
-  ctypedef int time_t
+from numpy cimport int8_t, uint8_t, int16_t, uint16_t, int32_t, uint32_t, int64_t, uint64_t
```
takluyver: Just to check, does it matter whether this is `from numpy` or `from .numpy`?

kif:
On Tue, 01 Oct 2019 03:11:16 -0700

> Just to check, does it matter whether this is `from numpy` or `from .numpy`?

Hi Thomas,

Unfortunately, yes it does matter. One did work for me, the other not.

I did not have the time to investigate but there may be smth fishy hidden there.
There is no more any technical reason to keep numpy.pxd but some code
needs probably to be updated to make it run with modern numpy.

It is worth opening an issue on this topic.

-- 
Jérôme

setup.py
```diff
@@ -32,7 +32,7 @@
 # these are required to build h5py
 # RUN_REQUIRES is included as setup.py test needs RUN_REQUIRES for testing
 # RUN_REQUIRES can be removed when setup.py test is removed
-SETUP_REQUIRES = RUN_REQUIRES + [NUMPY_DEP, 'Cython>=0.23', 'pkgconfig']
+SETUP_REQUIRES = RUN_REQUIRES + [NUMPY_DEP, 'Cython>=0.25', 'pkgconfig']
```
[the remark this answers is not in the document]
kif: I tested on various flavors of cython ... the code is compatible down to 0.21 at least (debian8, py34) if the language_level is 3 as an int and not as a str.

takluyver:
Thanks for your patience on this!

PR #1333, opened 2019-09-20
Added chunk iteration class

jreadey:
This PR adds a ChunkIterator class.  The idea is to make it convenient to iterate over the data space while efficiently reading or writing to an HDF5 file.  I've got some todo's below, but wanted to get some initial feedback

Example usage:

```
  f = h5py.File("myfile.h5", "w")
  dset = f.create_dataset("dset", (100,), chunks=(32,))
  for s in ChunkIterator(dset): 
          arr = dset[s]
          # do something with arr
```
Notes:
  * You can supply a selection to the constructor to iterate over just the chunks in a given sub-region
  * For non-chunked datasets, the iterator just gives one selection covering the entire dataset
  * An ValueError is thrown in a scalar or null space dataset is used
  * Only works with PY3

ToDo:
   * Update docs
   * Add test code
   * Use to make broadcasting more efficient


takluyver:
API-wise: I'd favour a method like `dataset.iter_chunks()` rather than exposing a new class constructor.

I also lean towards only allowing this on chunked datasets (not contiguous), though I know @tacaswell disagrees with me. My rationale is that a really obvious use case is to read/write each full chunk to/from memory - which should be reasonable for a chunked dataset, but can easily not be for the single 'chunk' of a contiguous dataset.

If we do keep it for contiguous datasets, however, I think we should allow if for scalar datasets: just yield `()` as a valid indexing expression.

jreadey:
I've created a iter_chunks method as per @takluyver's comment.   Also have it setup now to throw TypeError is a continuous dataset is used.
And have added tests!  

Seems like we just need to come to agreement on whether to handle continuous datasets or not.

scopatz:
I agree with @takluyver - at the very least there should be an `iter_chunks() convience method`

scopatz:
Please also add a news entry, as per the instructions at https://regro.github.io/rever-docs/news.html

takluyver:
This looks pretty good to me.

I think the key thing to move forwards is to make a decision about whether it should work with datasets that aren't stored in chunked format. I've seemingly persuaded you to implement my preference for now, but @tacaswell had the opposite preference and I don't know what other maintainers think (@aragilar @scopatz @aparamon ).

As I understand it, the argument in favour of allowing `ds.iter_chunks()` for non-chunked data is that contiguous data is effectively a single big chunk, and you shouldn't have to know about the storage format to use it. My argument against is that chunked storage implies that a chunk can reasonably be held in memory, and I think people would write code assuming that. A chunk is written and read as a unit, whereas you can read/write parts of a contiguous dataset.

I'm not asking for *votes* - I'd like a discussion leading to a consensus. But if my view is clearly the minority, I'll concede.

jreadey:
How about we use filters.guess_chunks to come up with a chunk layout for contiguous datasets?  This won't actually change the chunk layout, but will let the iterator process reasonably sized pieces of memory (currently set to max 1MB).

aragilar:
I'd be inclined not to support non-chunked datasets, as that's simpler to reason about - `dset.iter_chunks` works where `dset.chunks is not None` rather than trying to handle scalar or empty datasets (or virtual ones). I'm curious how this interacts with virtual datasets, can those be chunked? From the docs it looks like not, and those are likely to be larger than other datasets, so that's a situation where `iter_chunks` could be problematic.

h5py/_hl/dataset.py
```diff
@@ -242,6 +242,86 @@ def __exit__(self, *args):
             # pylint: disable=protected-access
             self._dset._dxpl.set_dxpl_mpio(h5fd.MPIO_INDEPENDENT)
 
+class ChunkIterator(object):
+    """
+    Class to iterate through list of chunks of a given dataset
+    # Note: Only works with PY3 since as is doesn't have a "next = __next__" redef
```
takluyver: We can lose this note, as h5py 3.0 is going to require Python 3 anyway (but thanks for making it clear!)

h5py/_hl/dataset.py
```diff
@@ -242,6 +242,86 @@ def __exit__(self, *args):
...
+    def __iter__(self):
+        return self
+
+    def __next__(self):
+
+        def get_ret(item):
+            if len(item) == 1:
+                return item[0]
+            else:
+                return tuple(item)
```
takluyver: I'd say let's return a tuple in all cases - a 1-tuple is equally valid for slicing, and it's easy to unpack if someone knows it's a 1D dataset and wants a slice:

```python
for sl, in ds.iter_chunks():
    pass
```

h5py/_hl/dataset.py
```diff
@@ -497,6 +577,14 @@ def __iter__(self):
             yield self[i]
 
     @with_phil
+    def iter_chunks(self, sel=None):
+        """ Return chunk iterator.  TypeError if the dataset is not chunked.
+            If sel is provided, only chunks that intersect the selected region
+            will bee reeturned, otherwise all chunks in the dataset will be
```
takluyver: A few things I think the docstring should make clearer:

- What does it yield (tuples of slices)
- `sel` should be a slice or a tuple of slices, such as you can get using `numpy.s_`.
- If you pass a selection, you only get back the intersecting parts of chunks, not the whole of any chunk that intersects the selection.
- This only works for chunked datasets (for now, anyway)

h5py/tests/test_dataset.py
```diff
@@ -704,6 +704,46 @@ def test_track_times(self):
...
+            chunk_count += 1
+        self.assertEqual(chunk_count, 8)
+
+        chunk_count = 0
+        for s in dset.iter_chunks(np.s_[48:52,40:50]):
+            self.assertEqual(len(s), 2)
+            self.assertEqual(s[0], slice(48,52,1))
+            self.assertEqual(s[1], slice(40,50,1))
+            chunk_count += 1
+        self.assertEqual(chunk_count, 1)
```
takluyver: I think tests like this can be written more neatly by converting the iterator to a list, something like this:

```python
assert list(dset.iter_chunks(np.s_[48:52, 40:50])) == [
  (slice(48, 52, 1), slice(40, 50, 1))
]
```

pytest should give useful information if an assertion like this fails.

takluyver:
> I'm curious how this interacts with virtual datasets, can those be chunked?

To the best of my knowledge, no: virtual is another storage format, an alternative to contiguous and chunked datasets.

(I did read something suggesting that a dataset could combine virtual parts with data stored in it, but I think that must have been an early design idea before the implementation went in another direction)

h5py/_hl/dataset.py
```diff
@@ -242,6 +242,86 @@ def __exit__(self, *args):
...
+                self._sel = (source_sel,)
+            else:
+                self._sel = source_sel
+        if len(self._sel) != rank:
+            raise ValueError("Invalid selection")
+        self._chunk_index = []
+        for dim in range(rank):
+            s = self._sel[dim]
+            if s.start < 0 or s.stop > self._shape[dim] or s.stop <= s.start:
+                raise ValueError("Invalid selection")
```
scopatz: It would be nice if the error message for these two "Invalid selection" cases were more specific about why they are invalid

scopatz:
FWIW, I like iter_chunks

aragilar:
Have we agreed that the current API is fine, or do we want to make further modifications?

takluyver:
I wouldn't say we're 100% agreed, but I figure that if we move forwards with only supporting chunked datasets for now, we can continue to discuss adding support for other types. It's harder to go the other way, though.

I'm happy enough with the way the API works, but I'd like the details I mentioned above to be addressed. I can find time to make those changes if necessary.

h5py/_hl/dataset.py
jreadey: I've updated to always return tuples.
Also added rst doc.

h5py/_hl/dataset.py
jreadey: updated.

h5py/tests/test_dataset.py
jreadey: updated.

h5py/_hl/dataset.py
jreadey: I've expanded the error message.

jreadey:
I've made changes that I think deal with all the issues we want to address this time around.

docs/high/dataset.rst
```diff
@@ -180,6 +180,12 @@ shape for you::
 Auto-chunking is also enabled when using compression or ``maxshape``, etc.,
 if a chunk shape is not manually specified.
 
+The chunk_iter method returns an iterator that can be used to perform chunk by chunk
+reads or writes::
+
+    >>> for s in dset.chunk_iter():
+    >>>     arr = dset[s]  # get numpy array for chunk
+
```
takluyver: We should also add the method to the reference part of this doc.

h5py/_hl/dataset.py
```diff
@@ -242,6 +242,82 @@ def __exit__(self, *args):
...
+            chunk_end = self._chunk_index[dim] * self._layout[dim]
+            if chunk_end < s.stop:
+                # we still have room to extend along this dimensions
+                return get_ret(slices)
+
+            if dim > 0:
+                # reset to the start and continue iterating with higher dimension
+                self._chunk_index[dim] = 0
+            dim -= 1
+        return get_ret(slices)
```
takluyver: We can get rid of the inline function now and just `return tuple(slices)`.

h5py/_hl/dataset.py
```diff
@@ -242,6 +242,82 @@ def __exit__(self, *args):
...
+            if isinstance(source_sel, slice):
+                self._sel = (source_sel,)
+            else:
+                self._sel = source_sel
+        if len(self._sel) != rank:
+            raise ValueError("Invalid selection - selection region must have same rank as dataset")
+        self._chunk_index = []
+        for dim in range(rank):
+            s = self._sel[dim]
+            if s.start < 0 or s.stop > self._shape[dim] or s.stop <= s.start:
```
takluyver: I think we'll ultimately want to make this behave more like Python slicing does more broadly - at least allowing negative indices (`-x` equivalent to `self._shape[dim] - x`). But that needn't be in this PR - it's fine to have the basic idea working first and fix these details later.

takluyver:
As the remaining things I spotted were all minor, I decided to make the changes myself rather than asking you to make further changes. I hope you don't feel I'm treading on your toes.

I'll merge this in about a day unless someone objects or wants more time to look at it.

jreadey:
That's quite all right, thanks @takluyver

PR #1334, opened 2019-09-20
Allow dset.astype() to be indexed, not only used as a context manager

takluyver:
```python
with dset.astype(np.float32):
    arr = dset[:]

# Becomes:
arr = dset.astype(np.float32)[:]
```

The context manager still works, and you can also get an `as` object to slice. But I'd push people towards using the stateless version without a context manager.

`__getitem__` with optional keyword argument is... debatable. We can change that if preferred. :wink:

### DIFF
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -399,14 +399,20 @@ Reference
 
     .. method:: astype(dtype)
 
-        Return a context manager allowing you to read data as a particular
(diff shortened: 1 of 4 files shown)


scopatz:
Thanks @takluyver - Please add a news entry, as per the instructions here: https://regro.github.io/rever-docs/news :wink: Looks good otherwise!

scopatz:
Thanks!

PR #1335, opened 2019-09-20
Hdf5 1.11.6 (pre-release 1.12.0)

pierlauro:
Hi everyone,
as previously discussed in person with @tacaswell & others, I am presenting the PR even though  some tests are failing in order to collectively solve them.

The pull request is mainly about updated APIs (with backward compatibility) as functions/types have been aligned with HDF5's develop (v1.11.6) and information about the upcoming version [have been included](https://github.com/h5py/h5py/commit/06d8194fc49584ca3b87b679fb9903bac0824ee8).
Since some HDF5's function names have been changed, the defs parser there [gives now](https://github.com/h5py/h5py/commit/cc3e6ef3a1d03605d3cb912288fa2d0d42ed2dbc) the possibility to add `min` and `max` version for each function.

Some test are badly failing (segmentation faults) because of the following `free` calls (if removed tests are passing - but obviously the memory is not freed) : [a](https://github.com/pierlauro/h5py/blob/cbebf05b130929b50d87013fb4f6b399e8c6239d/h5py/h5t.pyx#L1108), [b](https://github.com/pierlauro/h5py/blob/cbebf05b130929b50d87013fb4f6b399e8c6239d/h5py/_proxy.pyx#L196), [c](https://github.com/pierlauro/h5py/blob/cbebf05b130929b50d87013fb4f6b399e8c6239d/h5py/_proxy.pyx#L211), [d](https://github.com/pierlauro/h5py/blob/cbebf05b130929b50d87013fb4f6b399e8c6239d/h5py/_proxy.pyx#L228). I guess the problem is due to "recent" changes in the HDF5 repo to the malloc functions called by `H5Tget_member_name`.


pierlauro:
After further inspection: the `free` calls are ending up in segmentation fault because the pointers returned by `H5Tget_member_name` are not directly referring memory zones allocated with a `malloc`.


```
Pointer
returned
by malloc
|
|
↓
[_|_|_|_|_|_|_|_|_|_|_|_|_|_|_|_]
↑                 ↑ 
|                 |
|                 |
VALID        INVALID
FREE           FREE
```

More in detail: `H5Tget_member_name` [calls](https://github.com/live-clones/hdf5/blob/93e8bbd9f47d6a2eace95f2ed3b2dd4b6336940a/src/H5Tfields.c#L185) `H5MM_xstrdup` that [calls](https://github.com/live-clones/hdf5/blob/93e8bbd9f47d6a2eace95f2ed3b2dd4b6336940a/src/H5MM.c#L465) `H5MM_malloc` that [returns](https://github.com/live-clones/hdf5/blob/93e8bbd9f47d6a2eace95f2ed3b2dd4b6336940a/src/H5MM.c#L317) a pointer referring a memory zone subsequent to the one [originally returned](https://github.com/live-clones/hdf5/blob/93e8bbd9f47d6a2eace95f2ed3b2dd4b6336940a/src/H5MM.c#L292) by the malloc.

There is a [`H5MM_xfree`](https://github.com/live-clones/hdf5/blob/93e8bbd9f47d6a2eace95f2ed3b2dd4b6336940a/src/H5MM.c#L527) function that can be used to free blocks allocated by [`H5MM_malloc`](https://github.com/live-clones/hdf5/blob/93e8bbd9f47d6a2eace95f2ed3b2dd4b6336940a/src/H5MM.c#L267). Is it convenient to include such an "internal" function in h5py? I don't see any other way to manually deallocate memory.

For the future, it would be preferable to automatically replace `free` calls with `H5MM_xfree` when needed. For this purpose, I guess it should be asked to the HDF Group to provide some way to know which kind of pointer is passed.

scopatz:
Thanks so much @pierlauro! Please also add a news entry, as per the instructions at https://regro.github.io/rever-docs/news.html

pierlauro:
> Thanks so much @pierlauro! Please also add a news entry, as per the instructions at https://regro.github.io/rever-docs/news.html

Done, thanks.

Maybe it should also be considered to not run the CI when code is not changed in PR commits (as in my last commit, just adding a `rst` file).

aragilar:
I think https://portal.hdfgroup.org/display/HDF5/H5_FREE_MEMORY may be what we want to use rather than H5MM functions, as that's what documented at https://portal.hdfgroup.org/display/HDF5/H5T_GET_MEMBER_NAME. H5MM_malloc is called internally by h5_free_memory, so if cython free works on older HDF5, then having a switch between using free where h5_free_memory is available, and using h5_free_memory otherwise is probably safer.

pierlauro:
> I think https://portal.hdfgroup.org/display/HDF5/H5_FREE_MEMORY may be what we want to use rather than H5MM functions, as that's what documented at https://portal.hdfgroup.org/display/HDF5/H5T_GET_MEMBER_NAME. H5MM_malloc is called internally by h5_free_memory, so if cython free works on older HDF5, then having a switch between using free where h5_free_memory is available, and using h5_free_memory otherwise is probably safer.

Right, moreover `H5free_memory` is included in the public headers while `H5MM` functions are not. Just pushed the change.

aragilar:
@pierlauro As we support HDF5 1.8.4 onwards, you need to add a switch because `H5free_memory` only appears in 1.8.13. Otherwise I think this is ready to merge.

aragilar:
Thanks @pierlauro!

PR #1339, opened 2019-09-22
Allow storing datetime & timedelta data in HDF5 opaque types

takluyver:
This builds on #1262, aiming to allow storing numpy types for which there is no equivalent HDF5 type as opaque data.

The change here is that it uses dtype metadata to mark a dtype which should be stored as opaque data, like we already do for strings, enums and vlen data. Because we're building on an existing idea in h5py, it requires less tricky changes than #1262 did.

Open question: what dtypes should be disallowed for this mechanism? It currently only blocks object dtypes. I suspect we probably want to block size 0 as well (numpy uses that for data size to be determined when an array is created). Do we also want to block storing opaque data for types which HDF5 natively supports? Or leave that to the user?


scopatz:
CC @hameerabbasi - also one of us should add a news entry ;)

takluyver:
This is true, but:

1. I don't think it's important: the dtype metadata is for interacting with h5py, so I don't think it needs to be preserved through numpy's file format. I'd even suggest that it's better not to preserve it, because you have to explicitly use `opaque_dtype` to transfer data from a numpy file to HDF5.
2. h5py already uses dtype metadata for string types, vlen data and enums, so even if the limitations aren't ideal, they're consistent.
3. I don't know of any way to do something similar that would be roundtripped through numpy files.

scopatz:
Who should merge this, since we all touched it? :wink:

hameerabbasi:
Compound dtypes certainly won’t work and those should be blocked at a minimum.

I agree with defaults being blocked too.

takluyver:
I haven't yet blocked all the dtypes that work by default, because I can't see a neat way to do it.

Part of me wonders if we should only allow this for whitelisted dtypes, starting with datetime64& timedelta64. It would be a totally arbitrary restriction, but if we're unsure it's always easier to lift a restriction later than to add one.

hameerabbasi:
> Part of me wonders if we should only allow this for whitelisted dtypes, starting with datetime64& timedelta64. It would be a totally arbitrary restriction, but if we're unsure it's always easier to lift a restriction later than to add one.

I would be against that. Things like Quarternions and others exist.

takluyver:
I've never worked with quaternions, but it seems like it would be better to represent them as compound data, like h5py already does for complex numbers. This shouldn't require h5py to do anything - people who need it can hopefully make that work themselves.

This is kind of why I think limiting it to certain types might be a good idea. It might be like pickle/shelve - an attractive way to avoid thinking about how your data is stored, which may bite you later on. I prefer to solve for real use cases, not try to make things work for every imaginable scenario.

hameerabbasi:
It would if it could be transparent translation from `dtype` to `H5T`. But this isn't transparent.

Even datetimes/timedeltas were supported, but not in a transparent fashion, only when viewed as `int64`.

aragilar:
I'd be inclined not to block types, as the main users would be those who use custom dtypes (apart from datetime users), but I get the appeal.

I feel that adding a sentence or two discouraging the use of the automatically tagged opaque types, and directing people to write their own wrappers via metadata (pytables does this from what I can see, and I wrote [h5preserve](https://github.com/h5preserve/h5preserve) which does a similar thing based on my use case), which for any format that builds upon HDF5 and wants to be multilanguage or easily interoperable is far better than relying on auto-generated tags (the pickle comparison is very apt I think).

h5py/h5t.pyx
```diff
@@ -1793,10 +1793,12 @@ def opaque_dtype(np_dtype):
     the opaque data as datetimes.
     """
     dt = dtype(np_dtype)
-    if np.issubdtype(dt, np.object_):
+    if dt.hasobject:
         raise TypeError("Cannot store numpy object arrays as opaque data")
     if dt.names is not None:
         raise TypeError("Cannot store numpy structured arrays as opaque data")
+    if dt.subdtype is not None:
```
hameerabbasi: Is there a reason this is blocked?

takluyver:
OK, I've added a note to the docs about it.

Investigating dtypes some more, I found that sub-arrays are another case that won't work with the current code. We could broaden support later, but as it's not the main aim of this, I just wanted to block the cases that wouldn't work. Specifically, `dt.descr` loses the structure information, and just describes the field as a void pointer:

```
In [60]: d                                                                                               
Out[60]: dtype(('<u4', (5,)))

In [61]: d.descr                                                                                         
Out[61]: [('', '|V20')]

In [62]: d.subdtype                                                                                      
Out[62]: (dtype('uint32'), (5,))
```

h5py/h5t.pyx
takluyver: I realised it wouldn't work as expected with subarrays - see my other comment.

scopatz:
LGTM too

takluyver:
I'll merge this in a couple more days unless anyone objects or wants more time to review it.

takluyver:
Merging based on approval from @scopatz and @hameerabbasi, and a lack of objections.

PR #1342, opened 2019-09-23
ENH: Make H5pyDeprecationWarning also a DeprecationWarning

yarikoptic:
This should allow for easier filtering etc.

This is a blind, not tested change

### DIFF
--- a/h5py/h5py_warnings.py
+++ b/h5py/h5py_warnings.py
@@ -18,7 +18,7 @@ class H5pyWarning(UserWarning):
     pass
 
 
-class H5pyDeprecationWarning(H5pyWarning):
+class H5pyDeprecationWarning(H5pyWarning,DeprecationWarning):
     pass
 
 


takluyver:
It would be good to check how this affects the visibility of warnings with the default filters. I guess that the reason not to use DeprecationWarning in the first place was to make the warnings more visible.

scopatz:
Thanks @yarikoptic - please also add a news entry for this: https://regro.github.io/rever-docs/news

aragilar:
Yep, it's explicitly not subclassing as the default filters hide deprecation warnings (most projects I've seen either don't subclass from `DeprecationWarning`, or switched away from subclassing when the filters changed around about the py27 release).

takluyver:
I have tested, and subclassing from DeprecationWarning does mean warnings are hidden by default. This was softened a bit in Python 3.7, so DeprecationWarnings now show if they're triggered by code in the `__main__` module (i.e. the script being run). But they're still invisible if they come from any other code.

Here's my test code, if anyone else wants to experiment:

```python
from warnings import warn

class H5pywarning(UserWarning):
    pass

class H5pydepwarning1(H5pywarning):
    pass

class H5pydepwarning2(H5pywarning, DeprecationWarning):
    pass

warn("user", UserWarning)
warn("depr", DeprecationWarning)
warn("h5py", H5pywarning)
warn("h5py-dep1", H5pydepwarning1)
warn("h5py-dep2", H5pydepwarning2)
```

I'm therefore going to close this, as I think it would hide warnings we want people to see. Sorry about that; I agree it's not ideal.

PR #1345, opened 2019-09-24
Fix getting maxshape on empty datasets

takluyver:
I'd expect `ds.maxshape` to return None, the same as `ds.shape`. It currently throws an error as it tries to iterate over None.


scopatz:
Thanks!

PR #1348, opened 2019-09-26
Change default file mode to 'r' (read-only)

### LINKED ISSUE #397: make default mode 'r'

takluyver:
Closes #397

To briefly recap some of the reasons discussed there:

- Code which lacks permission to write to a file should fail when it tries to open the file with a suitable mode, not later when it tries to modify a file which had fallen back to read-only
- Making a typo trying to open an existing file should get you an error, not a new file
- Opening HDF5 files for writing locks them so no-one else can open them
- Opening a file read/write may [break access to virtual datasets](https://forum.hdfgroup.org/t/virtual-dataset-in-read-write-file-missing-data-from-read-only-file/5647)
- Read-only by default matches Python's built-in `open()` function

### DIFF
--- a/docs/high/file.rst
+++ b/docs/high/file.rst
@@ -22,13 +22,16 @@ use.  However, there is obviously no concept of "text" vs "binary" mode.
 The file name may be a byte string or unicode string. Valid modes are:
 
     ========  ================================================
-     r        Readonly, file must exist
+     r        Readonly, file must exist (default)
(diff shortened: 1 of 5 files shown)


scopatz:
Thanks!

PR #1353, opened 2019-09-26
Fix h5py._errors.unsilence_errors

### LINKED ISSUE #1349: h5py._errors.unsilence_errors has no effect
As it is, I couldn't switch on the output of errors of the HDF5 library in order to do some debugging.
The `api_gen` script is adding some `H5Eset_auto` in each call to HDF5, and this prevents `h5py._errors.unsilence_errors()` to work as expected.

To reproduce:
```
import h5py
h5py._errors.unsilence_errors() 
f = h5py.File('test.h5', 'w')  # Create a new empty file
f['aa']  # Access an object that does not exists
```

raises:
```
KeyError                                  Traceback (most recent call last)
<ipython-input-6-b4fee186affc> in <module>
----> 1 f['aaa']

~/venvs/py3/lib/python3.7/site-packages/h5py/_hl/group.py in __getitem__(self, name)
    262                 raise ValueError("Invalid HDF5 object reference")
    263         else:
--> 264             oid = h5o.open(self.id, self._e(name), lapl=self._lapl)
    265 
    266         otype = h5i.get_type(oid)

h5py/h5o.pyx in h5py.h5o.open()

KeyError: "Unable to open object (object 'aa' doesn't exist)"
```

While it should also print:
```
HDF5-DIAG: Error detected in HDF5 (1.10.5) thread 0:
  #000: H5O.c line 120 in H5Oopen(): unable to open object
    major: Object header
    minor: Can't open object
  #001: H5Oint.c line 625 in H5O_open_name(): object not found
    major: Object header
    minor: Object not found
  #002: H5Gloc.c line 422 in H5G_loc_find(): can't find object
    major: Symbol table
    minor: Object not found
  #003: H5Gtraverse.c line 851 in H5G_traverse(): internal path traversal failed
    major: Symbol table
    minor: Object not found
  #004: H5Gtraverse.c line 627 in H5G__traverse_real(): traversal operator failed
    major: Symbol table
    minor: Callback failed
  #005: H5Gloc.c line 378 in H5G__loc_find_cb(): object 'aa' doesn't exist
    major: Symbol table
    minor: Object not found
```

h5py    2.10.0
HDF5    1.10.5
Python  3.7.3 (default, Apr  3 2019, 05:39:12) 
[GCC 8.3.0]
sys.platform    linux
numpy   1.17.1

t20100:
This PR proposes to remove calls to `H5Eset_auto` before all calls to the HDF5 library make `h5py._errors.unsilence_errors` work.
This was apparently added to avoid HDF5 to output on stderr from threads, but it seems to be no more the case (Tests now check that).

Tests were updated to use pytest fixtures and check the output on stdout/stderr.
I renamed the file because main goal of the already existing tests were testing the output on stderr more than threading.

closes #1349


takluyver:
Looks like it's something that changed in HDF5 1.10 - the tests that are failing seem to be the ones which use HDF5 1.8. I think we still need to support the older version at the moment.

t20100:
I guess so, I though I've checked with HDF5 1.8.21 before submitting.
i'll check if I find a nice way to make it work in all case.
In the meantime, I have what I need to debug.

takluyver:
Coming back to this, I see two possible solutions:

1. We could make the `H5Eset_auto` calls conditional in the Cython code: the HDF5 version is known at compile time, so we can include them only on older versions of HDF5.
2. We could conditionally skip the tests, and say that if you want to use new h5py with an old HDF5 and multiple threads, it's up to you to call `silence_errors()` in each thread.

t20100:
Hi, I was also thinking that the callback could be stored somewhere in h5py and restore before each call rather than setting it to NULL.

This does not prevent for checking the version at compile time because that would avoid quite a few unnecessary calls.

takluyver:
If simply not setting it before each call does the right thing on recent versions of hdf5, let's focus on that for starters. Fewer moving parts is always better. :-)

t20100:
I dig a bit more into it.
The issue is not related to the version of HDF5 library but to the `--enable-threadsafe` compilation option:

- Without `--enable-threadsafe`, there is a single error stack, so this PR works fine.
- With `--enable-threadsafe`, there is one error stack per thread that gets initialized at thread creation to the default (i.e., call `H5Eprint`).
  This option is used on Ubuntu Xenial used in CI
(see debian/rules in http://archive.ubuntu.com/ubuntu/pool/universe/h/hdf5/hdf5_1.8.16+docs-4ubuntu1.1.debian.tar.xz)

So the call to `H5Eset_auto` is needed before each call to an HDF5 function (else, a way to init error stack management at thread creation and to access the error stack of all threads is needed..).
Yet, we can provide something else than `NULL, NULL` to `H5Eset_auto` depending on `silence/unsilence`.

takluyver:
I wonder if we can detect at compile time whether the threadsafe option was used?

I'm also interested in reducing the overhead of HDF5 library calls (see discussion on #1391). The `H5Eset_auto` call doesn't seem to be too bad, but if we can avoid doing an extra call on every HDF5 library call, that would be good. And adding an extra check of some other state variable isn't ideal either.

takluyver:
I went to ask about this in the HDF forum, and Andrew already made a topic with word-for-word the exact title I was typing. 10 years ago! https://forum.hdfgroup.org/t/h5eset-auto-for-all-threads/1501/4

t20100:
Let's not wait another 10 years to fix it...

`H5Eset_auto` is not doing much (basically storing the function and data pointers in a struct), but I completely agree to avoid useless calls.

In HDF5 headers, `H5_HAVE_THREADSAFE` macro is defined when the library is built with `--enable-threadsafe`.
I'll give it a try.

t20100:
CI fails because the added tests raises an issue in `h5py/utils.pyx` which should be solved by PR #1406.

Accessing `H5_HAVE_THREADSAFE` from cython at runtime should be possible provided that a bit of C macro is written in h5py (that could be placed in `api_compat.h`). Do you want me to propose something like this?
That would allow to skip the call to set the HDF5 logging in case the HDF5 lib is not built with the thread safe option.

takluyver:
Is that something that can be determined at compile time, or only at runtime? If we can compile the calls out entirely when HDF5 is built without the threadsafe option, I think it's worth doing that. If we need a runtime check of some other global state, perhaps it's not worth it.

t20100:
That's for determining it at runtime.

AFAIK, to determine it at compile time, it needs to be provided to `cythonize` through `compile_time_env` or as a `DEF` as in `config.pxi`, so it means either providing a build option (as for MPI) or some sort of analysis of the HDF5 headers during the setup (which is probably not worth it).

takluyver:
OK, probably not worth it then. I'll see if we can push hdf group to provide a way to set the error handler for new threads.

takluyver:
Closing and reopening to rerun the tests now that #1406 is merged.

h5py/tests/test_errors.py
```diff
@@ -0,0 +1,81 @@
...
+def test_unsilence_errors(tmp_path, capfd):
+    """Chech that HDF5 errors can be mute/unmute from h5py"""
+    filename = str((tmp_path / 'test.h5').resolve())
+
+    # Unmute HDF5 errors
+    h5py._errors.unsilence_errors()
+    _access_not_existing_object(filename)
+    captured = capfd.readouterr()
+    assert captured.err != ''
+    assert captured.out == ''
```
takluyver: As `unsilence_errors` modifies global state, can you wrap everything between *unsilence* and *silence* in a try/finally, so that an error here doesn't leave the state changed for other tests.

h5py/tests/test_errors.py
```diff
@@ -0,0 +1,81 @@
...
+    with h5py.File(filename, 'w') as newfile:
+        try:
+            doesnt_exist = newfile['doesnt_exist'].value
+        except KeyError:
+            pass
+
+
+def test_unsilence_errors(tmp_path, capfd):
+    """Chech that HDF5 errors can be mute/unmute from h5py"""
+    filename = str((tmp_path / 'test.h5').resolve())
```
takluyver: passing pathlib objects to `h5py.File` should work,  I think.

takluyver:
Other than those two comments, LGTM

t20100:
Comments addressed and rebased for testing locally.

takluyver:
Thanks, LGTM.

I'll leave this overnight to give maintainers on other continents a chance to look at it, and assuming there are no objections or test failures, I'll merge it tomorrow.

PR #1357, opened 2019-09-27
CI: Add caching to azure

aragilar:
This tells azure to cache pip's wheel cache, our HDF5 builds and ccache. I suspect we'll need to change this as caching on azure evolves, but lets go with this for now.

### DIFF
--- a/ci/azure-pipelines-steps.yml
+++ b/ci/azure-pipelines-steps.yml
@@ -34,6 +34,20 @@ steps:
   #    prerelease: true
   #  condition: and(succeeded(), eq(variables['python.version'], 'Pre'))
 
+# apparently azure's caching fails when nothing is added to the cache...
+- script: |
+    python -c "import os; os.makedirs(r'$(PIP_CACHE_DIR)', exist_ok=True); open(r'$(PIP_CACHE_DIR)/temp', 'w')"
+  displayName: 'Fix pip cache'
+
+- script: |
+    python -c "import os; os.makedirs(r'$(HDF5_CACHE_DIR)', exist_ok=True); open(r'$(HDF5_CACHE_DIR)/temp', 'w')"
+  displayName: 'Fix HDF5 cache'
+
+- script: |
+    python -c "import os; os.makedirs(r'$(CCACHE_DIR)', exist_ok=True); open(r'$(CCACHE_DIR)/temp', 'w')"
+  displayName: 'Fix ccache cache'
+
+
 - ${{ if eq(parameters.installer, 'nuget') }}:
   - task: NuGetToolInstaller@0
     displayName: 'Use latest available Nuget'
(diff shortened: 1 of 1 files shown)


ci/azure-pipelines-steps.yml
```diff
@@ -3,6 +3,24 @@ parameters:
...
+- task: CacheBeta@0
+  displayName: pip cache
+  inputs:
+    key: pip | $(Agent.OS)
+    path: $(PIP_CACHE_DIR)
+
+- task: CacheBeta@0
+  displayName: HDF5 cache
+  inputs:
+    key: HDF5 | $(Agent.OS)
```
takluyver: I haven't looked into Azure caching, but should the cache key also include the HDF5 version?

takluyver: Also maybe the OS version? If we have tests on Ubuntu 16.04 and 18.04, HDF5 compiled on one wouldn't necessarily work on the other.

The bit I don't know is whether the cache is already specific to each separate pipeline - if it is, this doesn't matter.

takluyver:
Investigating:

- [Caches are scoped by branch & pipeline](https://docs.microsoft.com/en-us/azure/devops/pipelines/caching/?view=azure-devops#cache-isolation-and-security). If I understand correctly, we have multiple jobs within a single pipeline, so we do need to use the cache key where we need to isolate different jobs.
- The [Agent.OS variable](https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#agent-variables) is very coarse - it basically only gives us Linux/Darwin/Windows_NT. I can't see an obvious way to distinguish e.g. Ubuntu 16.04 vs 18.04. We could maybe use `System.JobName` as part of the key, though it's not clear if that differs for the matrix entries within one job definition.
- You [can't change the contents of a cache](https://docs.microsoft.com/en-us/azure/devops/pipelines/caching/?view=azure-devops#using-the-cache-task) once it's been created. So rather than creating empty files to satisfy Azure, I think we should set a [condition](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/conditions?view=azure-devops&tabs=yaml) on the cache task so it only runs when there will be something for it to store.

I'd suggest that we start with just caching the HDF5 library for the jobs that build it - this would already save a lot of time, and getting one kind of cache working is less to think about than three.

takluyver:
I've opened #1371 to try to do what I think it needs for caching HDF5.

takluyver:
I think this got marked as merged because I had based #1371 on this branch, and @scopatz merged that. But I undid some of the changes @aragilar made here, so it's not fully in effect.

Having played with Azure caching, I'm not sure that it works for pip or ccache: both of those want a cache directory they can update piecewise as individual packages/source files change. Azure pipelines doesn't let you update the cache unless the cache key has changed.

PR #1361, opened 2019-09-30
Special dtypes API changed in 2.10, not 2.9

takluyver:
I had hoped to get #1132 in for 2.9, but it actually landed in 2.10, and I missed this.


scopatz:
Thanks!

Issue #1343, opened 2019-09-23
Set up automatic building of low-level API docs


takluyver:
The docs at http://api.h5py.org/ are outdated, and not accessible over HTTPS. They're currently hosted where Andrew manually set them up some time ago. We should remedy this.

Our usual go-to service for hosting docs is readthedocs.org, but it's probably not a good answer here. The API docs need h5py to be importable, which which means getting libhdf5 installed with development headers and compiling h5py. RTD is not really designed for this kind of thing, as far as I know - but if someone 

- [x] Pick somewhere to host these docs (github pages?)
- [x] Pick a CI service to build them (Azure? See #1324)
- [x] Set up building and publishing the docs automatically when h5py master changes
- [x] Get the DNS updated to point to the new location
- [ ] ~~Stretch goal: automatically generate a separate URL for each tag, 'stable' referencing the latest tag, and master (like RTD does)~~

scopatz:
My recomendation for this is doctr (https://drdoctr.github.io/), which lets you build and manage github pages on travis.  It is precisely for the extension module use case.

takluyver:
Yes! Someone mentioned that during the code camp, but I had forgotten the name.

takluyver:
Before we ask Andrew to change the DNS for this, a question. I've set up doctr to publish versioned docs, e.g. `/master/h5p.html` or `/v3.0.0/h5p.html`. This would break existing links to the unversioned URLs (`http://api.h5py.org/h5p.html`).

Do we:

1. Stick to unversioned docs built from master, and use Sphinx versionadded/versionchanged directives to document changes between versions.
2. "Redirect" the unversioned URLs to `/master/x` or `/stable/x`. We can't do proper HTTP redirects on Github pages (as far as I know), so this would be some combination of html meta-refresh, rel=canonical links and Javascript.

Versioned docs feel like a step forwards. But on the other hand, I suspect most people use the default docs they land on anyway, and the versionadded/versionchanged directives are still useful, so maybe it's not worth the trouble. Versioned docs can also be problematic if an old version ranks highly in search results, and people don't realise it's outdated.

I think I've now talked myself into preferring option 1. But I'd like to get other opinions before I change it.

takluyver:
Ping @tacaswell @aragilar @scopatz - opinions on this? I'm leaning towards sticking with a single, unversioned set of docs built from master, to avoid breaking URLs.

aragilar:
[...]

PR #1362, opened 2019-09-30
Publish API docs from Travis to Github pages

takluyver:
Addresses #1343.

You can see the resulting docs at http://h5py-api-tmp.kluyver.me.uk/doctr/ . Obviously we should move this to a nicer domain like api.h5py.org once we're happy with it.

- API docs are built on Travis, with HDF5 1.10.5 (using the default 1.8.x version from Ubuntu, various functions can't be exposed).
- Once we merge it, http://h5py-api-tmp.kluyver.me.uk/master/ should be created, and each release will get a `/v$TAG/` folder. I suggest we manually create a symlink or a redirect for `stable`.
- PRs and other branches we create won't be deployed, but it will still build the API docs on Travis, so we can catch if any problems are introduced.
- Built docs are pushed into the [gh-pages branch](https://github.com/h5py/h5py/tree/gh-pages) of this repo. We could also configure it to use a separate repository.


scopatz:
This LGTM!

scopatz:
Thanks!

PR #1366, opened 2019-10-01
Remove long-deprecated dtype functions

takluyver:
These have been marked as deprecated as far back as the git history goes,
and have issued visible warnings since 2.8 (gh-926).


scopatz:
Thanks!

PR #1369, opened 2019-10-02
Better error message when chunks are requested for empty dataset

takluyver:
Supersedes #1263.


scopatz:
This probably also needs a news entry. Thanks @takluyver!

scopatz:
Thanks!

PR #1370, opened 2019-10-02
Remove deprecated h5py.highlevel module

takluyver:
The high-level interface is now exposed in the top-level h5py module.

This was marked as deprecated in the docstring from 2.6 (gh-579), and issued deprecation warnings from 2.9 (gh-1047).


takluyver:
Would people be OK with setting a threshold for codecov to mark a PR as failing? E.g. allow coverage to drop by up to 0.5%? It's a bit misleading that pull requests like this get a red cross because they've simply removed code that was previously covered.

scopatz:
> Would people be OK with setting a threshold for codecov to mark a PR as failing? E.g. allow coverage to drop by up to 0.5%?

Yes absolutely.  I feel like codecov is usually a red herring anyway.  It adds a lot of noise

scopatz:
This probably needs a news entry. Thanks @takluyver!

PR #1371, opened 2019-10-02
WIP: Azure caching

takluyver:
Experimenting, based on #1357

### DIFF
--- a/ci/azure-pipelines-steps.yml
+++ b/ci/azure-pipelines-steps.yml
@@ -34,6 +34,20 @@ steps:
   #    prerelease: true
   #  condition: and(succeeded(), eq(variables['python.version'], 'Pre'))
 
+# apparently azure's caching fails when nothing is added to the cache...
+- script: |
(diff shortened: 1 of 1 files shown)


takluyver:
This seems to have worked: the last commit (with cache populated) ran in 11m 27s, whereas the one before that took 26m 57s.

On Linux, the 'ensure HDF5' step took <1s. On Windows, it took slightly longer, because of installing requests, but still not a long time. Building h5py is now the slowest step.

scopatz:
Thanks @takluyver!

aragilar:
Thanks for sorting this @takluyver, I had misread the cache docs, I though they were append-only, I guess we wait until the cache service matures before adding pip or ccache (though I'm not sure ccache will speed things up, maybe the h5py builds)?

takluyver:
I think there could be some value in configuring ccache just within the
jobs, even without persisting its cache. I've opened a separate issue about
that.

We should keep an eye on the caching system as it matures, but as it stands
I don't think we can usefully cache pip or ccache data.

PR #1376, opened 2019-10-03
Drop unneeded travis tests

aragilar:
Similar to #1375, remove unneeded tests from travis. The tests remaining are the MPI tests, which I'll drop in #1255, the ppc64le test, and the api doc builder.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,8 +29,8 @@ matrix:
     - python: 3.7
       env:
       - TOXENV=py37-test-deps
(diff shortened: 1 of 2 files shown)


takluyver:
As with #1375, I'll leave this for a couple of days to give other maintainers a chance to see it.

For instance, I'd be open to arguments that if we're maintaining config for multiple CI services anyway, we should have some overlap between the configurations they test as an extra sanity check.

PR #1377, opened 2019-10-04
Publish api docs built from master

takluyver:
`--branch-whitelist` overrides the default of publishing docs from master. Removing it will publish docs from the master branch.


aragilar:
Thanks for setting this up @takluyver!

PR #1378, opened 2019-10-04
Remove deprecated File.fid attribute

takluyver:
`File.id` works instead. This has been described as "backwards compatibility" in the docstring since 2.0, and issued deprecation warnings since 2.9 (gh-1047).


aragilar:
The only failure is a very small coverage change, and given we're had this deprecated forever, I'm going to merge this. Thanks @takluyver!

PR #1380, opened 2019-10-06
CI: Add testing of HDF5 develop branch

aragilar:
This adds the develop branch of HDF5 to travis (with `allow_failures`). We could add more branches, but I'm guessing what we really need is some way of notifying the HDF5 group on failure of these and similar tests (do we know what CI system they are using?).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,8 @@ addons:
   apt:
     packages:
       - libhdf5-serial-dev
+      - dh-autoreconf
+      - libtool-bin
 
 cache:
   directories:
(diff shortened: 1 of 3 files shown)


aragilar:
This looks like it's found a bug, somewhere we're returning `v112` where we should be returning `latest`.

takluyver:
I feel like `allow_failures` jobs aren't super useful, because we so rarely look at them.

Would it be better to have a separate repo/branch which runs the tests on a schedule (weekly?) with h5py master and the HDF5 develop branch? Then it could send email notifications on failure, without marking all PRs as failing if something gets broken. Hopefully there would also be a button to manually trigger a test run when you think you've fixed a problem.

takluyver:
Azure has [scheduled triggers](https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml#scheduled-triggers), so that plan also wouldn't be tied to Travis.

aragilar:
@takluyver Maybe I missing something, but I can't see any way of filtering jobs on azure (e.g. only run these job at this time). We could use circle ci, and leave it solely for running a weekly test between h5py and HDF5 (testing against develop and each of the series branches), and there's less conditional logic inside the azure config. I'm happy to remove the travis version, I went with putting it on travis because I find it easier and faster to get something working on travis than azure.

takluyver:
I was just thinking of having a separate repo or branch with its own CI configuration for scheduled tests.

aragilar:
I'm going to close this, creating a repository which does daily rebuilds against the in-dev HDF5 is probably a better way forward (as suggested by @takluyver), and will mean that we can also test against released versions of h5py also. I won't delete the branch, in case people want to use the code as a basis for the new repository (hopefully future James will notice this message and not delete it by mistake...).

PR #1381, opened 2019-10-06
CI: Have codecov collect on all platforms

### LINKED ISSUE #1358: Azure not uploading to codecov
Not high priority yet, but if we're going to drop some jobs on Travis & Appveyor in favour of Azure, we should make sure it's uploading coverage information properly. Currently it gives:

```
Error: Missing repository upload token
```

The token itself probably can't be secret if we want it to work for pull requests, but that shouldn't be a big problem:
https://community.codecov.io/t/security-issues-regarding-upload-token/429/6

aragilar:
This adds enforcement of codecov success (which if it's a problem I'll drop, but hopefully not), and combines the coverage explicitly (so hopefully codecov will just work on all platforms now...).

Closes #1358 (now).

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,2 +1,3 @@
 [run]
 omit=*/tests/*
+parallel=true
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,8 @@ MANIFEST
 *~
 .pytest_cache
 .coverage
+.coverage_dir
+coverage.xml
 
 # Rever
 rever/
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,7 @@ matrix:
       - TOXENV=py37-test-deps
       - HDF5_VERSION=1.10.5
       - HDF5_DIR=$HDF5_CACHE_DIR/$HDF5_VERSION
+      - H5PY_ENFORCE_COVERAGE=yes
       os: linux-ppc64le
 
     # Builds the API docs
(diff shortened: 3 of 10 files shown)


aragilar:
Assuming this passes on all systems, this should be ready to merge. Closes #1358. This also fixes appveyor, so our overall overage should go up (given we should have coverage on Windows systems).

takluyver:
Is it feasible to just call `coverage combine ...` from the CI configuration before invoking `codecov`, rather than maintaining our own wrapper script to upload coverage?

takluyver:
Also, while we're looking at coverage stuff: would you be OK with setting a threshold, e.g. 0.5%, in [codecov config](https://docs.codecov.io/docs/commit-status)? If I understand correctly, this would let coverage drop by that much without giving the PR a red cross. It's been annoying me that even tiny coverage drops (e.g. from removing some tested code) make it look like CI has failed.

aragilar:
Codecov seems to have real issues finding the files on Windows (which is why I wrote that script in python to be cross platform), which is why I wrote that script. I suspect there's some issue with how the python codecov script handles file paths internally, and given we'd need to wait for any patch to be merged (there's still a patch for collecting data on azure waiting to be merged on their system - https://github.com/codecov/codecov-python/pull/194), I'd rather keep the script until codecov properly supports azure with windows. I did think about moving to coveralls (in the hope of having something "just work") but it looks like their azure support is no better (and their python support worse)...

I'll add the codecov.yml increasing the threshold, it annoys me also.

takluyver:
OK, fair enough. Is it possible that we're waiting for a fix I already got merged there to be released? https://github.com/codecov/codecov-python/pull/202

aragilar:
Yeah, it looks like we're waiting on your fix to codecov (I notice the coverage upload worked this time, even though the changes should have made no difference...).

PR #1382, opened 2019-10-07
Set environment variables for ccache under pip

takluyver:
I'm not 100% sure this can't break anything - I'd like someone else who's familiar with ccache to think about it.

Tox creates an sdist of h5py, and then uses `pip` to install that into test environments. Pip unpacks the sdist into a temporary directory to build it. This messes up ccache with the default options, because the path forms part of its cache key. Each build is done from a new temporary directory, so the path is different, and ccache ignores its cache.

`CCACHE_BASEDIR` tells ccache to treat absolute paths with this prefix as relative paths without it, so files in the build directory will have matching paths wherever that is. This can't be set by the user before calling tox/pip, because you don't know where the build directory will be. `CCACHE_NOHASHDIR` tells it to also ignore the CWD, which pip sets to the build directory.

On my system, rerunning `tox -e py37-test-deps` goes from ~2m 15s without ccache to ~35s once the cache is populated. Set `CC="ccache gcc"` to use ccache.

### DIFF
--- a/setup_build.py
+++ b/setup_build.py
@@ -156,6 +156,13 @@ class h5py_build_ext(build_ext):
         from Cython.Build import cythonize
         import numpy
 
+        # This allows ccache to recognise the files when pip builds in a temp
+        # directory. It speeds up repeatedly running tests through tox with
+        # ccache configured (CC="ccache gcc"). It should have no effect if
+        # ccache is not in use.
+        os.environ['CCACHE_BASEDIR'] = op.dirname(op.abspath(__file__))
+        os.environ['CCACHE_NOHASHDIR'] = '1'
+
         # Provides all of our build options
         config = self.distribution.get_command_obj('configure')
         config.run()


mdboom:
This all seems reasonable to me, but could break things if someone is relying on the default behavior -- I guess it mostly matters for dynamic linking.  Would it work to set these (if possible) from the `tox` config, so it only takes effect when actually using `tox`?

takluyver:
Sorry, can you clarify under what situation it could break things? I don't follow what you're saying about the default behaviour. This shouldn't affect anything unless you're using ccache to compile.

This isn't only relevant for use with tox, though that's an obvious example. I couldn't set these environment variables directly from tox, because it doesn't know the build directory, but if necessary I could set another environment variable, and check that inside the build code.

mdboom:
Many people (myself included) have ccache enabled by default at a system level.  

The scenario I was thinking of is you have two checkouts of `h5py` in different directories, and you build each.  This could break debugging, since it might look up the source file from the wrong directory if it reused a pre-compiled object from the other tree.  This is enough of a corner case that I guess you could probably ignore it.

Also, I hadn't thought through that it's setting it to the build directory of each checkout, not a single global directory for everything, so that should be good enough...

takluyver:
IIUC, it hashes the input files (and any included files), so it should only reuse objects if the source being compiled is exactly the same? So you'd get the wrong file path, but still the right source code. I know basically nothing about debugging C code, though.

takluyver:
Thanks @mdboom for looking at this. :+1:

tacaswell:
Thanks @mdboom !

PR #1386, opened 2019-10-08
Remove deprecated Dataset.value attribute

takluyver:
This has been deprecated for years (bae29c53af3922d46243d5a222ab5bda0df0cb22), but the warning wasn't working, even if you set the warning filters to show DeprecationWarning, until it was fixed in 2.9 (#1047).

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -795,14 +795,6 @@ class Dataset(HLObject):
             )
         return r
 
-    def __dir__(self):
-        names = set(super().__dir__())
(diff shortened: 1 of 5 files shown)


takluyver:
I think this concludes my push to remove deprecated features. The only thing left using H5pyDeprecationWarning is `ds.dims.create_scale()`, which I deprecated in 2.10 - I'd leave that in place for a bit longer.

aragilar:
The coverage drop is small, so I'm going to go ahead and merge this. Thanks for all the clean up @takluyver .

PR #1387, opened 2019-10-08
Clean up unnecessary fallback imports

takluyver:
`collections.abc` should work on all supported Python versions now.


takluyver:
Merging this myself,  as it's a straightforward change and the tests are passing.

PR #1388, opened 2019-10-09
Integrate check-manifest checker into pre-commit

takluyver:
This shouldn't make a whole lot of difference to the CI, but it's one more check a human can conveniently run with a single `tox -e pre-commit` command.

While playing with it, it flagged that the sdist was including the generated `_hdf5.pxd` file, so I've changed the manifest to exclude that.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,3 +4,8 @@ repos:
     hooks:
     -   id: end-of-file-fixer
     -   id: trailing-whitespace
+
+-   repo: https://github.com/mgedmin/check-manifest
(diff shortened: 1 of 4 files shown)


aragilar:
I'm slightly concerned check-manifest didn't pick that up earlier, thanks @takluyver for catching that.

takluyver:
I guess it can't pick up generated files until they exist, so running it in a clean environment in CI doesn't catch them. I guess we could have a CI job that runs a bunch of things that might generate files and then does the check-manifest test. Though that's an argument for not bundling it into pre-commit.

Issue #1315, opened 2019-09-19
[feature request] Add public API to get the path / library that h5py is using


tacaswell:
This will make writing plugin wheels easier. attn @vasole

kif:
The place to store that is in h5py.h5 ...

takluyver:
Another option we discussed was to expose pointers to the C functions from libhdf5 so plugins can use those.

@vasole - can you list the HDF5 functions you need for the compression side, so we can judge the feasibility of different ways to get them?

vasole:
Yes, the idea is that the actual HDF5 library functions could be replaced by proxys configurable via a pointer to the actual HDF5 functions. That way the decompression could be guaranteed without having to patch the original code of the plugins while compression could still be enabled.

For the bitshuffle+LZ4 filter the functions are

1. H5Zregister
2. H5Epush1
3. H5Pget_filter_by_id2

For Blosc they seem to be (untested)

1. H5Zregister
2. H5Epush
3. H5Pget_filter_by_id
4. H5Tget_size
5. H5Tget_super
6. H5Tclose
7. H5Pmodify_filter

t20100:
For getting pointers to the C function API, using `dlopen`/`dlsym` on a Python extension of h5py (e.g., `h5z`) seems to give access to the HDF5 library symbols, e.g., from Python:

```
import _ctypes, os
h = _ctypes.dlopen(h5py.h5z.__file__, os.RTLD_LOCAL)
_ctypes.dlsym(h, 'H5Epush1')
```

t20100:
We've refactored the `hdf5plugin` project (https://github.com/silx-kit/hdf5plugin) without the need from `h5py` to provide access to C pointers.
For Linux and macOS, using `dlopen` on a `h5py` module (e.g. `h5py.h5z`) gives access to the symbol of the HDF5 library, while on Windows the symbols of the HDF5 library are available for linking once the HDF5 library is loaded.

As this issue was opened as part of discussing `hdf5plugin`, I think it can be closed.

vasole:
We intend to make a new release, but some feedback using the wheels available via the link below (or trying to build from source) would be appreciated.

https://silx.gitlab-pages.esrf.fr/bob/hdf5plugin/

takluyver:
Thanks all, nice work on figuring out how to handle that.

Slightly off-topic for this issue, but I got an idea looking at your readme. Would you be interested if h5py 3.0 could allow you to get rid of the `**kwargs` thing?

```python
[...]

PR #1389, opened 2019-10-09
Nicer API for specifying a filter and options

takluyver:
This is a rough draft of an idea I mentioned on #1315. The idea is that from the user perspective, something like this would be possible:

```python
f.create_dataset('foo', data=numpy.arange(100),
    compression=hdf5plugin.Blosc(level=9, shuffle='byte'))
```

For hdf5plugin, the code would look something like this:

```python
from h5py.filters import FilterRefBase

class Blosc(FilterRefBase):
    filter_id = 32001
    def __init__(self, level, shuffle='bit', compression='blosclz'):
        self.filter_options = (0, 0, 0, 0, level, shuffle, compression)
```

### DIFF
--- a/h5py/_hl/filters.py
+++ b/h5py/_hl/filters.py
@@ -37,7 +37,7 @@
     encode
         Tuple of available filter names for encoding
 """
-
+from collections.abc import Mapping
 import operator
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


t20100:
That sounds good to me, we'll update hdf5plugin to be compatible with it.

takluyver:
I saw that you'd like to use the same object for `**kwargs` style usage with the existing h5py API. I'm not sure how easy that is with this design. Have you investigated it at all?

takluyver:
OK, I got curious. I think the last commit should allow for using it as **kwargs as well. Of course, you'd have to have a fallback copy of that class to use with existing versions of h5py that don't provide it.

Alternatively, we could make it so that the user uses a property to get the **kwargs:

```python
f.create_dataset('foo', data=numpy.arange(100),
    **hdf5plugin.Blosc(level=9, shuffle='byte').kw)
```

That's slightly more typing, but slightly less magic.

t20100:
Great, I was thinking of doing it on our side, but it is best if it is in `h5py`.

I would go for:
```
f.create_dataset('foo', data=numpy.arange(100),
    **hdf5plugin.Blosc(level=9, shuffle='byte'))
```

That's probably what people will write for a little while for 2.x, 3.x compatibility... and probably until you eventually deprecate `compression_opts`.

aragilar:
I think having the plugin being a mapping is probably the nicer API. Does anyone have any objections to this being merged?

kif:
I like this PR. Makes sense.

scopatz:
LGTM

Issue #1367, opened 2019-10-01
Clean up cimport of numpy


kif:
Numpy exposes (for already some times now) a cython C-API based on PXD files which can be c-imported in the cython code.
The numpy.pxd file present in h5py is a "duplicate" but some definition differs. It would be useful to reduce the maintenance burden to rely directly on the one provided by numpy

takluyver:
Thanks @kif :-)

kif:
So I started to hack into it. There is no PR but I could open it soon, as usual, depending on the local BW available.
I found PyArray_Scalar was defined and exposed in h5py but commented out in the cython/numpy export with the given explantation:
```
    # Functions taking dtype and returning object/ndarray are disabled
    # for now as they steal dtype references. I'm conservative and disable
    # more than is probably needed until it can be checked further.
```
Removing it breaks h5py :(

takluyver:
That's strange, because `PyArray_Scalar` doesn't seem to be used anywhere in h5py. Can you identify how h5py is broken without it?

If we do need it, we can obviously open an issue on numpy to ask that it be added to the Cython interface. But we should check first whether we do we need it.

kif:
On Wed, 02 Oct 2019 07:26:38 -0700
Thomas Kluyver <notifications@github.com> wrote:

> That's strange, because `PyArray_Scalar` doesn't seem to be used anywhere in h5py. Can you identify how h5py is broken without it?

Maybe it is smth else ...
 
> If we do need it, we can obviously open an issue on numpy to ask that it be added to the Cython interface. But we should check first whether we do we need it.

I found some definitions in the numpy.pxd provided by cython differs
from the one in h5py: the import_array function at least which returns
an int in cython and nothing in h5py.

It shouldn't be an issue but this change causes many problems. Still investigating

kif:
I found something else  ...
The definition of ndarray and dtypes differ from h5py and cython when it comes to the name of the cdef-attributes. 
For example h5py defines `dtype.elsize` while cython defines `dtype.itemsize`, the same for ndarray.nd or ndarray.ndim and ndarray.shape vs ndarray.dimensions. 

This brings again quite a bit of changes... The cython code didn't change neither in the 7 last years. Maybe it is the time to synchronize with them?

takluyver:
[...]

PR #1390, opened 2019-10-10
Cimport numpy from cython instead of maintaining a local binding

kif:
Some more work is probably needed but it will require deeper hack. This is hence the superficial layer.
Related to #1367

### DIFF
--- a/h5py/_conv.pyx
+++ b/h5py/_conv.pyx
@@ -20,31 +20,26 @@ from libc.stdlib cimport realloc
 from .utils cimport emalloc, efree
 cfg = get_config()
 
-# Initialization
-from numpy cimport _import_array, npy_intp, dtype as np_dtype, ndarray as np_ndarray,\
-        NPY_WRITEABLE, NPY_C_CONTIGUOUS, NPY_OWNDATA
+
+# Initialization of numpy
+from numpy cimport _import_array, npy_intp, dtype as np_dtype, ndarray as np_ndarray
+from numpy cimport NPY_WRITEABLE, NPY_C_CONTIGUOUS, NPY_OWNDATA
 _import_array()
 
-# Minimal interface for Python objects immune to Cython refcounting
+from cpython.object cimport PyObject, PyTypeObject
+
+# It would be nice to be able to replace PyObject* with object and use the pxd provided by cython
 cdef extern from "Python.h":
-
     # From Cython declarations
-    ctypedef int PyTypeObject
-    ctypedef struct PyObject:
-        Py_ssize_t ob_refcnt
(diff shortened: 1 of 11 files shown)


aragilar:
@kif This will need a rebase. What's stopping us from dropping the rest of the Python.h stuff?

kif:
On Sun, 13 Oct 2019 00:51:57 -0700
James Tocknell <notifications@github.com> wrote:

> @kif This will need a rebase. 

OK, I'll do it.

> What's stopping us from dropping the rest of the Python.h stuff?

Thomas K. wanted multiple small PR which are easier to review for you.
This is why I stopped here (for this first PR)
Other are likely coming after that.

aragilar:
@kif Ah, cool, that makes sense. If there's anything cython doesn't expose, it's probably worth checking with them if there's a reason, in case there are APIs we shouldn't be using.

kif:
After the rebase, I have a seg-fault on the tests :(

Concerning non exposed function, the numpy/__init__.pxd states that those functions are commented out because they are stealing dtype. but I did not yet fully understand the meaning of:
``` # Functions taking dtype and returning object/ndarray are disabled
    # for now as they steal dtype references. I'm conservative and disable
    # more than is probably needed until it can be checked further.
``` commented by Dag from cython.

takluyver:
Thanks @kif, this looks like a good step forwards. I don't fully grasp all of the Cython stuff, but the tests on CI seem to be passing (the failure shown is a problem with uploading coverage data - #1398).

Did you fix the segfault you were seeing?

kif:
I believe the bug is in the azure pipeline ... codecov passed directly but not via azure where it tries to upload some data.

kif:
I can confirm the seg-fault was an issue on my rebase. I believe it is safe for merging.

takluyver:
Thanks @kif. I'll give it a day or so in case the other maintainers around the world want to take a look, then I'm happy to merge it.

tacaswell:
Do we need to keep numpy.pxd around for back-compatibility reasons?

kif:
On Mon, 14 Oct 2019 08:50:35 -0700
Thomas A Caswell <notifications@github.com> wrote:

> Do we need to keep numpy.pxd around for back-compatibility reasons?

For now yes ... but we are working on this. It will be part of another
which may be more intruisive.

h5t still relies on the dtype class defined in h5py and cannot use the
one from cython. I suspect there is a mix between import and cimport.
Work in ongoing but with Thomas K. we agreed to split this large
refactoring in smaller part to create an adiabatic pathway (i.e.
reversible).

Cheers,

Jerome

Issue #1336, opened 2019-09-20
Crash when accessing vlen of strings


gavinmacaulay:
h5py    2.9.0
HDF5    1.10.4
Python  3.6.9 |Anaconda, Inc.| (default, Jul 30 2019, 14:00:49) [MSC v.1915 64 bit (AMD64)]
sys.platform    win32
sys.maxsize     9223372036854775807
numpy   1.16.4

I get a crash when trying to access a vlen of strings variable. I have a minimal example, defined via a netcdf4 .cdl file:

```
netcdf error {
	types:
		string(*) name_t;
	dimensions:
		n = 2;
	variables:				
		name_t names(n);
	data:
		names = {"name1.1", "name1.2"}, {"name2"};
}
```
(I can provide the resulting hdf5 file if necessary) 

Which when accessed via this code:
```
import h5py
with h5py.File('error.nc', 'r') as f:
    names = f['names']
    for n in names:
        print(n)       
```

crashes the python interpreter on the print(n) line. I think that this is probably the same problem as reported on the h5py google groups list:  https://groups.google.com/forum/#!topic/h5py/mi3ndCigBEA

The example file is read fine by other hdf5 readers (h5dump, HDFView, Matlab, etc).

takluyver:
I agree it's similar to the mailing list post from last month. Possibly also to #623. The key appears to be having a datatype which has vlen arrays of vlen strings. I've simplified an example from the mailing list thread - this reproduces the issue for me:

```python
import h5py
import numpy as np

# Create file with dataset containing vlen arrays of vlen strings
with h5py.File("h5py_string_test2.h5", "w") as h:
    vlen_str = h5py.special_dtype(vlen=str)
    vlen_vlen_str = h5py.special_dtype(vlen=vlen_str)

    ds = h.create_dataset('/com', (2,), dtype=vlen_vlen_str)
    ds[0] = (np.array(["a", "b", "c"], dtype=vlen_vlen_str))
    ds[1] = (np.array(["d", "e", "f","g"], dtype=vlen_vlen_str))

# At this point, h5py_string_test2.h5 appears to be a valid
# file, i.e. can be h5dump'd, and the contents look fine

# Read the file
with h5py.File("h5py_string_test2.h5", "r") as h:
    ds = h["com"]
    print("Shape of ds: ", ds.shape)
    print("Dtype:", ds.dtype)
    print("Dtype metadata:", ds.dtype.metadata)
[...]

PR #1393, opened 2019-10-11
fix crash when accessing vlen of string #1336

egachen:
in conv_vlen2str, a variable bkg_obj0 is not used, but it tries to call Py_XDECREF on it

### DIFF
--- a/h5py/_conv.pyx
+++ b/h5py/_conv.pyx
@@ -17,7 +17,7 @@ from .h5 import get_config
 from .h5r cimport Reference, RegionReference, hobj_ref_t, hdset_reg_ref_t
 from .h5t cimport H5PY_OBJ, typewrap, py_create, TypeID
 from . cimport numpy as np
-from libc.stdlib cimport realloc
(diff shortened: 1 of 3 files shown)


takluyver:
Thanks! I haven't reviewed the changes in detail yet, but it would be good if you could also add a test that would fail (or just crash) without the changes, and passes with them.

aragilar:
@egachen Thanks for fixing this!

takluyver:
Thanks @egachen. Am I right in thinking that the variable `bkg_obj` in this function is now also unused, and can be removed?

egachen:
> Thanks @egachen. Am I right in thinking that the variable `bkg_obj` in this function is now also unused, and can be removed?

yes

takluyver:
Thanks. If you've got a moment once #1390 is merged (it will probably conflict with that), a tiny PR would be welcome. Or I can get to it at some point.

egachen:
created a PR: https://github.com/h5py/h5py/pull/1399

PR #1394, opened 2019-10-11
Avoid equality comparison where array might be used

takluyver:
Running the tests, I see a warning like this:

```
tests/test_dataset_getitem.py::Test1DZeroFloat::test_mask
  /home/takluyver/miniconda3/envs/h5py-dev/lib/python3.7/site-packages/h5py/_hl/dataset.py:535: DeprecationWarning: The truth value of an empty array is ambiguous. Returning False, but in future this will result in an error. Use `array.size > 0` to check that an array is not empty.
    if args == (Ellipsis,) or args == tuple():
```

The deprecation was apparently [introduced in numpy 1.14](https://numpy.org/devdocs/release/1.14.0-notes.html#deprecations).

This change uses `is` rather than `==` to check for Ellipsis, which avoids getting into array comparison, so the warning goes away. It's slightly ugly, but I can't think of a neater way to fix it.

### DIFF
--- a/ci/upload_coverage.py
+++ b/ci/upload_coverage.py
@@ -8,8 +8,7 @@ from os import chdir, listdir, environ
 from pathlib import Path
 import platform
 from pprint import pprint
-import shlex
-from subprocess import run, PIPE
+from subprocess import run, PIPE, CalledProcessError
 import sys
 
 THIS_FILE = Path(__file__)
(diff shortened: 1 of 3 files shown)


takluyver:
@aragilar - I made a couple of adjustments to the `upload_coverage.py` script as it had errored in one of the Appveyor test runs.

aragilar:
@takluyver Yeah, that was a silly typo on my part. I'm confused how this avoids the numpy warning, I'm guessing it down to not comparing with a non-empty tuple?

aragilar:
Ah, it's the tuple wrapping (that avoids the array comparison). Thanks @takluyver for this.

takluyver:
I think the important thing is using `is` to check for `Ellipsis` - but to do that, we have to unwrap it from the tuple.

PR #1395, opened 2019-10-13
Cleanup tests

aragilar:
This cleans up the tests a bit, some of the changes are:
 * The imports should be in a more standard format, so we can look for duplicate tests and generally reorganise them so they make more sense (I've merged the two test_file files for example).
 * There were quite a few unneeded subclasses, which I've removed.
 * Modernised some things, such as using super or context managers, which should reduce code duplication.
 * ~~Added coverage of tests (so we can see if the CIs are running all the tests)~~
 * Other test related things, e.g. dropping unneeded imports (we should probably run flake8 and add it to tox, so that avoid that in the future).


takluyver:
This sounds good, and I'll look more at it later, but just to mention: my immediate reaction to including tests in coverage measurements is that I'd rather not.

h5py/tests/test_dtype.py
```diff
@@ -109,44 +115,45 @@ def a(items):
...
 
         # Save to HDF5 file
-        with h5py.File(fname, "w") as Fid:
-            Fid.create_dataset("Data", data=A, dtype='f2')
+        self.f.create_dataset("Data", data=A, dtype='f2')
+        fname = self.f.filename
 
-        with h5py.File(fname, "r") as Fid:
-            B = Fid["Data"][:]
+        with File(fname, "r") as f:
```
takluyver: Don't we need to close the writeable file handle before opening it again for reading?

takluyver:
I think the cleanup is good overall, but I have a minor disagreement. I don't think using `from h5py import ...` everywhere is necessarily an improvement. It's slower to write new tests if you have to add imports, and sometimes things like `h5py.version` make the meaning more obvious (vs just `version`). As I rough rule of thumb, I'd stick with `h5py.blah` until blah is used maybe 3-5 times in the file - but biased in favour of more if I want to make its origin clearer in the code.

I'm also still against including tests in coverage measurements by default. Maybe there's some way to report them without including them in the summary measurement? But I'm not sure how to effectively summarise the coverage information so that we'd actually notice if some CI job isn't running a test it should be.

h5py/tests/test_dtype.py
aragilar: Yep, I'll fix that.

aragilar:
With the replacement of `import h5py` with `from h5py import`, I was going to switch to using relative imports, but I wanted to avoid making too many changes in the one PR (`from h5py import version as h5py_version` is probably what I should have gone with, rather than plain `version`). I don't think the imports should be changing that often, as adding imports (to the existing ones) probably implies the test is in the wrong file.

I'm happy to drop the tests being added to the coverage be default (I'm going to try to experiment with https://docs.codecov.io/docs/flags and see if that allows us to cut out the tests by default, but still be able to check what tests we're actually running).

kif:
I agree with ThomasK about the "from h5py import (many things)" which makes the tests a bit less readable to me.
Technically I would even prefer replacing "from h5py import" with "from .. import" as I noticed some nasty mixing between local tests and the installed version (or maybe was it the installed tests with the dev' version of the library)

scopatz:
There are conflicts here now

takluyver:
This at least needs the conflicts resolved. And then I think we need some agreement about the imports.

I prefer having explicit `h5py` in the test imports to relative imports with `..`, and I can't see how it makes any technical difference, but that's only a minor objection, so if there's some situation where the relative imports do work, I'm OK with that.

kif:
With the latest patches (atexist(unregister_converters)) it should be possible to have code coverage on the cython part now.

aragilar:
I think the main thing that people objected to was the moving from `import h5py` to `from h5py import X`? I think starting from scratch might be easier, and doing the changes over multiple PRs would also be a good idea.

takluyver:
Yup, breaking it up into smaller pieces sounds like a good plan.

The import changes were what I was objecting to - I find `h5py.File` clearer than `File` or `h5py_File` (that's explicit, but would make me wonder what's going on that it can't be `h5py.File`). I don't know if that's exactly a consensus, but the other cleanups sound like stuff I'd agree with quite happily.

takluyver:
I'm going to close this since it's been sitting around with no changes for a couple of years and now has a bunch of merge conflicts. We can reopen if you particularly want to clear up the conflicts, but I imagine it would be as easy now to start a new branch.

aragilar:
Yeah, I think starting from scratch is better at this point :smile: (and I should go through my other long-running PRs and close them/clean them up too).

PR #1396, opened 2019-10-13
ENH: Ensure that sys.path has the current directory

aragilar:
I think I wrote this because there was something about pip not including the directory containing the setup.py to the path (I think there was a patch to setuptools or something which is why I didn't get around to submitting this). In any case, if that change were to happen, this readds that dir so we can import the configure and build files.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,6 +19,11 @@ import sys
 import os
 import os.path as op
 
+# Newer packaging standards may recommend removing the current dir from the
+# path, add it back if needed.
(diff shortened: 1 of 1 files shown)


takluyver:
Yes - maybe you ran into https://github.com/pypa/pip/issues/6163 before it got resolved. It caused enough problems that I personally suspect it will never be changed by default, but people were talking about it, and it doesn't hurt to add it in there.

PR #1399, opened 2019-10-15
Remove unsed bkg_obj from conv_vlen2str

egachen:
Remove unsed bkg_obj from conv_vlen2str


takluyver:
Thanks! The CI failure is unrelated (#1398).

PR #1401, opened 2019-10-15
Don't fail CI on random codecov errors

takluyver:
Closes gh-1398


takluyver:
Merging because the random failures are often enough to be annoying. We can easily revert this if we figure out a better solution.

Issue #1391, opened 2019-10-10
Caching seems to have no effect on read times


ivirshup:
I can't seem to get any changes in performance for slicing into a 1d dataset axis by modifying caching arguments, even though reads are contiguous. In contrast, reading chunks exactly gives much better performance. Here's an example:

```python
# Setup
import h5py
import numpy as np

x = np.random.randint(0, 1000, int(1e7))
indices = np.sort(np.random.choice(int(1e7), int(1e4), replace=False))

with h5py.File("test.h5", "w") as f:
    f.create_dataset("x", data=x, chunks=True)
```

Now I'll try and read in the array in way that should mostly be hitting the cache. I'll show you the general schema of the benchmark, followed by just how the file was opened and timings for brevity

```python
f = h5py.File("test.h5", "r")
dset = f["x"]

# In it's own cell:
%%timeit
for i in range(len(indices) - 1):
    s = slice(indices[i], indices[i+1])
    dset[s]
# Timing results:
# 1.21 s ± 6.71 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

f.close()
```

With a large cache:

```python
f = h5py.File("test.h5", "r", rdcc_nbytes=100 * (1024 ** 2), rdcc_nslots=50000)
# Timing results:
# 1.17 s ± 25.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

With no cache:

```python
f = h5py.File("test.h5", "r", rdcc_nbytes=0)
# Timing results
# 1.14 s ± 12 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Setting `rdcc_w0` to something, while using a large cache:

```python
f = h5py.File("test.h5", "r", rdcc_nbytes=100 * (1024 ** 2), rdcc_nslots=50000, rdcc_w0=.5)
# Timing results
# 1.16 s ± 37.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

These timings all fluctuate a bit, but are all pretty similar. By contrast, I'll just read in entire chunks:

```python
%%timeit
cs = dset.chunks[0]
ts = dset.shape[0]
slice_gen = (slice(i*cs, min((i+1)*cs, ts)) for i in range(ts // cs + 1))
for s in slice_gen:
    dset[s]
# Timing results:
# 136 ms ± 955 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

[...]

PR #1402, opened 2019-10-15
Cache some properties on Dataset

takluyver:
In #1391, we noticed that repeated small reads from the same dataset are surprisingly slow. Profiling showed that quite a bit of time was wasted re-fetching and calculating things which don't change from read to read.

First, the headline figures:

- Simple `ds[0:10]` reads are about 2x faster (1.2 s -> 0.55 s)
- `read_direct()` about 25% faster (0.98 s -> 0.75 s)

Times are for 10k reads of 10 numbers each from a 1D integer dataset. [Details in this comment](https://github.com/h5py/h5py/issues/1391#issuecomment-542080099), but the timings given there are done with a different method.

The fastest I've got with Python code calling the low-level API directly is 0.22 s.

-----

For properties which can't change (like whether a dataset is empty), I've used a standard [@cached_property](https://github.com/pydanny/cached-property) decorator. They'll be calculated once on first access, then saved.

Dataset shape and size can change, either if the file is writeable (through `H5Dset_extent`) or if it's being read in SWMR mode (when `H5Drefresh` is called). I've taken a compromise position: 

* For SWMR mode the `ds.refresh()` call also clears the cache. This assumes that people keeping a `Dataset` object around will use its refresh method, not the low-level API.
* If the file is writeable, it doesn't use the cache at all, so it can't get outdated if you modify it with the low-level API.

The code could be simpler if we cached shape & size for writeable files as well - but it would mean that after manipulating a dataset with the low-level API you need to call `.refresh()` or get a new high-level Dataset object.

### DIFF
--- a/h5py/_hl/base.py
+++ b/h5py/_hl/base.py
@@ -418,3 +418,15 @@ class Empty(object):
 
     def __repr__(self):
         return "Empty(dtype={0!r})".format(self.dtype)
+
+
+def product(nums):
+    """Calculate a numeric product
+
+    For small amounts of data (e.g. shape tuples), this simple code is much
+    faster than calling numpy.prod().
+    """
+    prod = 1
+    for n in nums:
+        prod *= n
+    return prod
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -278,10 +278,20 @@ class Dataset(HLObject):
         return self.id.rank
 
     @property
-    @with_phil
     def shape(self):
         """Numpy-style shape tuple giving dataset dimensions"""
-        return self.id.shape
+        if 'shape' in self._cache_props:
+            return self._cache_props['shape']
+
+        with phil:
+            shape = self.id.shape
+
+        # If the file is read-only, cache the shape to speed-up future uses.
+        # This cache is invalidated by .refresh() when using SWMR.
+        if self._readonly:
+            self._cache_props['shape'] = shape
+        return shape
+
     @shape.setter
     @with_phil
     def shape(self, shape):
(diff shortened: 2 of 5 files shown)


takluyver:
Another tweak: using a simple Python function in place of `numpy.product()` in `SimpleSelection.broadcast()`.

- `ds[0:10]` reading unchanged
- `read_direct()` gets about 30% faster again (0.75 s -> 0.52 s). This is almost a 2x speedup from before this PR.

-----

With small amounts of data, `numpy.product()` is about an order of magnitude slower than the hand-coded `product()` function I added. E.g. with a tuple of 2 numbers, numpy takes ~8 µs, and the custom function about 200 ns.

Of course, this only works for small data, and possibly only for data that's not already in an array. But in various places we have one number per dimension, like this, and [HDF5 only allows up to 32 dimensions](https://support.hdfgroup.org/HDF5/faq/limits.html). Even if that limit were removed, I suspect there are very few use cases which need more dimensions.

h5py/_hl/dataset.py
```diff
@@ -384,8 +404,17 @@ def fillvalue(self):
...
-    def __init__(self, bind):
+    def _extent_type(self):
+        return self.id.get_space().get_simple_extent_type()
+
+    @cached_property
+    def _is_empty(self):
+        return self._extent_type == h5s.NULL
+
+    @with_phil
+    def __init__(self, bind, readonly=False):
```
tacaswell: make `readonly` kwarg only?

setup.py
```diff
@@ -32,7 +32,7 @@
 NUMPY_DEP = 'numpy>=1.7'
 
 # these are required to use h5py
-RUN_REQUIRES = [NUMPY_DEP,]
+RUN_REQUIRES = [NUMPY_DEP, "cached-property"]
```
tacaswell: Does this interact nicely with py3.8?

h5py/_hl/base.py
```diff
@@ -418,3 +418,15 @@ def __eq__(self, other):
 
     def __repr__(self):
         return "Empty(dtype={0!r})".format(self.dtype)
+
+
+def product(nums):
```
tacaswell: do we want to make this `_product` to protect it a bit from users?

h5py/_hl/dataset.py
takluyver: Good idea, done.

setup.py
takluyver: No idea! Is there something you're aware of that might break it?

h5py/_hl/base.py
takluyver: I figure that anything you have to import from `h5py._hl` is effectively private anyway - the public API is exposed on the top-level `h5py` module. But I'm happy to add a `_` if you prefer.

takluyver:
I'll merge this in another day or two unless anyone objects or wants more time to review it.

takluyver:
Merging on the strength of @tacaswell's approval and lack of objections. There's still time to make changes before release.

PR #1403, opened 2019-10-15
Add asv configuration & a benchmark

takluyver:
In connection with #1402, I thought it would be a good time to try setting up asv, a tool for tracking performance across different commits, to catch performance regressions. This is the basic settings its quickstart tool produces, with one h5py benchmark. We could alternatively set it up in a separate repository.

I was able to run this (though annoyingly, I have to whitelist my branch in the config to see the results):

```
asv run master..dataset-props-cache
asv publish
asv preview
```

Screenshot from 2019-10-15 14-46-06


aragilar:
I've pushed a commit fixing the manifest issue and apparently the eof pre-commit test was failing also, which I've also fixed.

takluyver:
Thanks.

takluyver:
I think the one key question to address before merge is whether this belongs as part of this repo, or in a separate repo alongside this. asv supports both options.

Other than that, it's just some extra files which shouldn't affect anything unless you install and use asv.

scopatz:
I am fine with these files living here

takluyver:
I'll merge this in a couple of days unless anyone wants more time to look at it.

PR #1407, opened 2019-10-15
Try to document reference dtypes more clearly

takluyver:
I noticed this could be clearer from discussion in #1360.


aragilar:
Yeah, this reads much better, thanks @takluyver!

PR #1411, opened 2019-10-17
Allow get_vfd_handle() for any file driver

takluyver:
I guess at some point in the past, only sec2 supported calling this, but looking at the [docs](https://confluence.hdfgroup.org/display/HDF5/H5F_GET_VFD_HANDLE) and the source code now, multiple drivers support it.

### DIFF
--- a/h5py/h5f.pyx
+++ b/h5py/h5f.pyx
@@ -386,17 +386,20 @@ cdef class FileID(GroupID):
 
 
     @with_phil
-    def get_vfd_handle(self):
-        """ () => INT
+    def get_vfd_handle(self, fapl=None):
+        """ (PropFAID) => INT
 
         Retrieve the file handle used by the virtual file driver.
 
-        This method is only functional when the the SEC2 driver is used.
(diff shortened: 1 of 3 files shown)


h5py/h5f.pyx
```diff
@@ -386,17 +386,20 @@ cdef class FileID(GroupID):
...
 
         Retrieve the file handle used by the virtual file driver.
 
-        This method is only functional when the the SEC2 driver is used.
+        This may not be supported for all file drivers, and the meaning of the
+        return value may depend on the file driver.
+
+        The 'family' and 'multi' drivers access multiple files, and a file
+        access property list (fapl) can be used to indicate which to access,
+        with H5Pset_family_offset or H5Pset_multi_type.
```
takluyver: Neither of these are exposed even in the low-level API, so this note is mostly to explain why the `fapl` parameter exists. I don't think either driver is commonly used - h5py doesn't even expose the multi driver.

aragilar:
This looks fine to me, merging, thanks @takluyver.

PR #1412, opened 2019-10-18
Enable nogil

### LINKED ISSUE #1314: [Feature request] Is it possible to have the low level binding "nogil"
Seams possible ...
Pro: let cython find-out the gil requesting code.
Con: possible interference between *phil* and *gil* waiting each other (i.e. dead-lock).

kif:
It is now possible to call certain hdf5 function without the GIL
For this, just append the nogil argument in api-function.txt and rebuild

Conversion callback (in H5T) are re-acquiring the GIL as needed.

Those conversion callbacks are also unregistered at exit (used to cause a seg-fault)

close #1314
built on top of #1406 (review only after the merge)


takluyver:
The diff here still shows the changes from #1406 - I think you'll probably need to rebase on master to fix that. You may want to wait until #1353 lands - that's probably nearly ready, and it was cause merge conflicts with this as well.

h5py/__init__.py
```diff
@@ -38,11 +38,12 @@
...
     ))
 
-
 _errors.silence_errors()
 
-from ._conv import register_converters as _register_converters
+from ._conv import register_converters as _register_converters, \
+                   unregister_converters as  _unregister_converters
 _register_converters()
+atexit.register(_unregister_converters)
```
takluyver: Is this related to the rest of the PR?

setup_build.py
```diff
@@ -195,12 +196,14 @@ def run(self):
...
 DEF COMPLEX256_SUPPORT = %(complex256_support)s
+DEF FREE_GIL = %(free_gil)s
 """
                 s %= {
                     'mpi': bool(config.mpi),
                     'mpi4py_v2': bool(v2),
                     'version': tuple(int(x) for x in config.hdf5_version.split('.')),
-                    'complex256_support': hasattr(numpy, 'complex256')
+                    'complex256_support': hasattr(numpy, 'complex256'),
+                    'free_gil': FREE_GIL
```
takluyver: This config option doesn't seem to be used anywhere.

h5py/_conv.pyx
```diff
@@ -762,22 +762,22 @@ cdef int conv_ndarray2vlen(void* ipt,
...
+    nbytes = len * max(outtype.get_size(), intype.get_size())
 
-    if outtype.get_size() > intype.get_size():
-        data = emalloc(outtype.get_size() * len)
-    else:
-        data = emalloc(intype.get_size() * len)
-    memcpy(data, ndarray.data, intype.get_size() * len)
-    H5Tconvert(intype.id, outtype.id, len, data, NULL, H5P_DEFAULT)
+    #with nogil:
+    if 1:
```
takluyver: Leftover?

h5py/_conv.pyx
```diff
@@ -762,22 +762,22 @@ cdef int conv_ndarray2vlen(void* ipt,
...
+        data = emalloc(nbytes)
+        memcpy(data, ndarray.data, nbytes)
```
takluyver: I think if `outtype.get_size() > intype.get_size()`, this could read past the end of `ndarray.data`.

h5py/__init__.py
kif: In theory not. In practice Python seg-faults at output without this patch.
Actually h5py was not unregistering the conversion filters from hdf5 at exit. Everything was there except it was not done.
We found out that once this "nogil" feature is activated, the unregistration is called after the gil itself is destroyed, hence the seg-fault.

h5py/_conv.pyx
kif: this is why we took the max of the two a couple of lines before

h5py/_conv.pyx
kif: indeed. 
I was not sure to make the call to emalloc gil-free or not. It is a subject which is worth discussing I believe. What is your opinion ?
Shall we focus on making the  actual IO gil-free ?

setup_build.py
kif: Indeed. James suggested to use conditional system wide gil deactivation based on constants, which looked like a cool idea.
Unfortunately this is not compatible with the "with gil" function declaration that automatically re-acquires the gil when needed.

h5py/_conv.pyx
takluyver: Right, you allocate a buffer to the max of the two. But if the output data is larger than the input data, then copying the input data needs to use the input data size, not the output data size.

h5py/_conv.pyx
takluyver: Do you know how big a chunk of data HDF5 prefers to pass to converters? I guess that the bigger the chunks are, the more it makes sense to release the GIL here.

h5py/_conv.pyx
kif: right. corrected

h5py/_conv.pyx
kif: Vlen are often used to store strings of comments ... so the answer is probably "not much".

takluyver:
That also makes me think that `nogil` in `api_functions.txt` is misleading - Cython uses that to mean 'doesn't need the GIL', whereas you're using it to mean 'release the GIL'.

kif:
Hi Thomas,
You are right concerning the pyx annotation for "nogil", "with gil" and "with nogil".
My very first attempt was following the path you are suggesting and lead to various hang and seg-faults. As I discovered this was related to those call-back function, I decided to address this problem first. I did not find (easily) in the hdf5 doc if a function call was triggering (or not) any call-back.

Technically this PR does "almost nothing" except putting the mechanism in place to release the gil and testing it on the type-conversion call-back functions (which are known to be more problematic than the other functions).

setup_build.py
kif: I removed this part as it is not used

t20100:
Early test of this PR + `H5Dread` without gil with a Qt GUI reading a 32 GB dataset in a thread while displaying it as it loads: The nogil is a game changer, it makes the GUI react smoothly to user interaction while it is not without this patch.

takluyver:
Nice! I'll do some review.

I assume you've spotted that the CI is complaining about some trailing whitespace.

api_gen.py
```diff
@@ -238,7 +240,24 @@ def write_cython_imp(self):
             raise ValueError("Return code <<%s>> unknown" % self.line.code)
 
         # Have to use except * because Cython can't handle special types here
-        imp = """\
+        if self.line.nogil:
+            imp = """\
+cdef {0.code} {0.fname}({0.sig}) except *:
+    cdef {0.code} r
+    with nogil:
+        _hdf5.H5Eset_auto(NULL, NULL)
```
takluyver: Should this be using `set_default_error_handler`? I don't think we provide an API to set the handler to a Python function, so it should always be NULL or `H5Eprint`, and thus safe to use with nogil.

h5py/utils.pyx
```diff
@@ -24,10 +24,13 @@ cdef inline void* emalloc(size_t size) except? NULL:
     """Wrapper for malloc(size) with the following behavior:
 
     1. Always returns NULL for emalloc(0)
-    2. Raises MemoryError for emalloc(size<0) and returns NULL
+    2. Raises MemoryError for emalloc(size<0) and returns NULL **IMPOSSIBLE as size>=0**
```
takluyver: I agree. Let's remove this line from the docstring.

kif: Done

kif:
> A few minor comments, but I think this looks good. Do you want to add `nogil` to more functions by default? I guess the read & write functions would be good starting points.

This is what we tested with Thomas @t20100 this morning and it clearly works in his GUI. 
I agree H5Dread/write are the first candidates and we checked it works as expected. I stay on the safe side, I intended to unlock the GIL function by function. Let's start with the most problematic: the dataset-read/write.

> I'm trusting that you've added `with gil` to the functions that need it. Do you think it's worth trying to write some sort of stress test that starts several threads all trying to exercise the nogil pathways in parallel, to give us a chance of picking up if something can erroneously call back into Python without acquiring the GIL?

Making this is test is rather complicated and I remember this has been a painful journey (debugging 81 threads for a dead-lock took me 6 weeks the first time in 2012):
https://groups.google.com/forum/#!searchin/h5py/kieffer/h5py/B3tT82FaAsA/CSW0LfSqrVcJ

kif:
I did not manage to write a multi-threaded reader.
The error is:
```  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "/users/kieffer/.venv/py37/lib/python3.7/site-packages/h5py/_hl/dataset.py", line 530, in __getitem__
    if self._fast_read_ok and (new_dtype is None) and (self._local.astype is None):
AttributeError: '_thread._local' object has no attribute 'astype'
```
I doubt it makes much sense by the way, so for me it makes sense to prevent this !

kif:
Apparently, enabling nogil on the H5Dread crashes the test... two options:
* Like this it works and lays the basement for enabling subsequent functions which have to be tested (my initial idea, this PR is already big enough)
* Wait for H5Dread and H5Dwrite to be fixed but this makes the PR even larger ...

api_gen.py
kif: No idea, I just reproduced what was present.

api_gen.py
```diff
@@ -35,6 +35,7 @@ class Line(object):
 
         Exists to provide the following attributes:
 
+        nogil:      Bool indicating if the function is pure C (not calling any Python callback)
```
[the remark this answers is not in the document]
kif: corrected

kif:
Just to give you an idea, to enable H5Dread as nogil, all its possible callbacks should re-acquire the GIL. For now I found h5py.h5fd.H5FD_fileobj_read as one of its callback but I would not be surprised if there are a few other. Going for the first option is probably safer.

Here is an example of traceback:
```
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7974535 in __GI_abort () at abort.c:79
#2  0x0000000000482e40 in fatal_error (prefix=prefix@entry=0x0, 
    msg=msg@entry=0x76d3d8 "PyThreadState_Get: no current thread", status=status@entry=-1) at ../Python/pylifecycle.c:2179
#3  0x0000000000482eb4 in Py_FatalError (msg=msg@entry=0x76d3d8 "PyThreadState_Get: no current thread")
    at ../Python/pylifecycle.c:2189
#4  0x00000000004835f2 in PyThreadState_Get () at ../Python/pystate.c:709
#5  0x00007ffff349e1fa in __Pyx_PyObject_Call (func=<type at remote 0x7ffff34c26c0>, arg=((96,), 1, b'B', 'c'), 
    kw={'allocate_buffer': False}) at /home/jerome/workspace/h5py/h5py/h5fd.c:22100
#6  0x00007ffff34a48a7 in __pyx_array_new (__pyx_v_shape=__pyx_v_shape@entry=(96,), 
    __pyx_v_itemsize=__pyx_v_itemsize@entry=1, __pyx_v_format=__pyx_v_format@entry=0x7ffff7240320 "B", 
    __pyx_v_mode=__pyx_v_mode@entry=0x7ffff34b9720 "c", __pyx_v_buf=__pyx_v_buf@entry=0x10018d0 "")
    at /home/jerome/workspace/h5py/h5py/h5fd.c:8042
#7  0x00007ffff34b3837 in __pyx_f_4h5py_4h5fd_H5FD_fileobj_read (__pyx_v_f=0xf77940, __pyx_v_type=<optimized out>, 
    __pyx_v_dxpl=<optimized out>, __pyx_v_addr=<optimized out>, __pyx_v_size=96, __pyx_v_buf=0x10018d0)
    at /home/jerome/workspace/h5py/h5py/h5fd.c:3392
#8  0x00007ffff62406f1 in H5FD_read () from /usr/lib/x86_64-linux-gnu/libhdf5_serial.so.103
#9  0x00007ffff621f87c in H5F__accum_read () from /usr/lib/x86_64-linux-gnu/libhdf5_serial.so.103
#10 0x00007ffff631bf9a in H5PB_read () from /usr/lib/x86_64-linux-gnu/libhdf5_serial.so.103
#11 0x00007ffff622ac4b in H5F_block_read () from /usr/lib/x86_64-linux-gnu/libhdf5_serial.so.103
#12 0x00007ffff61dfd1e in ?? () from /usr/lib/x86_64-linux-gnu/libhdf5_serial.so.103
#13 0x00007ffff63d3d12 in H5VM_opvv () from /usr/lib/x86_64-linux-gnu/libhdf5_serial.so.103
#14 0x00007ffff61deca3 in ?? () from /usr/lib/x86_64-linux-gnu/libhdf5_serial.so.103
#15 0x00007ffff61f961a in ?? () from /usr/lib/x86_64-linux-gnu/libhdf5_serial.so.103
#16 0x00007ffff61f9e36 in H5D__select_read () from /usr/lib/x86_64-linux-gnu/libhdf5_serial.so.103
#17 0x00007ffff61dfb60 in H5D__contig_read () from /usr/lib/x86_64-linux-gnu/libhdf5_serial.so.103
#18 0x00007ffff61f2e0e in H5D__read () from /usr/lib/x86_64-linux-gnu/libhdf5_serial.so.103
#19 0x00007ffff61f3255 in H5Dread () from /usr/lib/x86_64-linux-gnu/libhdf5_serial.so.103
#20 0x00007fffdfe258a6 in __pyx_f_4h5py_4defs_H5Dread (__pyx_v_dset_id=360287970189639681, 
    __pyx_v_mem_type_id=216172782113784185, __pyx_v_mem_space_id=288230376151711749, 
    __pyx_v_file_space_id=288230376151711748, __pyx_v_plist_id=0, __pyx_v_buf=0x10018d0)
    at /home/jerome/workspace/h5py/h5py/defs.c:6702
```

takluyver:
Yes, I'm OK with merging these foundations as is and enabling nogil for reading & writing later. The AttributeError should be a simple enough fix - that was my mistake.

Probably all of the `H5FD_fileobj_*` should be marked `with gil`, as the `fileobj` driver is entirely based on HDF5 calling back into Python code. Other than that and the conversion functions, I don't think there should be many callbacks leading to Python code; I don't think we expose an interface to register Python functions in the filter pipeline.

Elsewhere in the code, there are several iteration interfaces which work with callbacks (search for `cb_` to find them). We should keep those in mind if we ever want to use `nogil` more widely, but I don't know if it would be worth releasing and reacquiring the GIL for those operations.

api_gen.py
takluyver: @t20100 changed what was present in #1353, though.

takluyver:
I'm happy with this - if you're OK with the couple of changes I just pushed, I think we can merge it.

kif:
On Fri, 29 Nov 2019 03:44:24 -0800

> I'm happy with this - if you're OK with the couple of changes I just pushed, I think we can merge it.

Yep, it is OK for me ...
I moved already forward to make the call-backs of H5Dread/write re-acquiring the gil.

takluyver:
Great, let's merge this and await your new PR :-)

PR #1413, opened 2019-10-19
Publish unversioned API docs

takluyver:
See discussion on #1343

Publishing API docs at versioned URLs would be nice, but it would break existing links, and people might find old versions as the top results on search engines. This can probably all be avoided with careful use of html meta redirects and `rel="canonical"` links, but that's an extra maintenance burden. So the simple solution is to stick to publishing docs at unversioned URLs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ matrix:
       script:
         - tox
         - pip install doctr
-        - doctr deploy --built-docs docs_api/_build/html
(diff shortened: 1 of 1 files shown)


takluyver:
I'll merge this change if there are no objections in the next couple of days, and then (assuming it builds correctly from master), ask Andrew to update the DNS for api.h5py.org.

PR #1414, opened 2019-10-20
ENH: Add options for additional HDF5 layouts and builds

aragilar:
I've split this out as per @takluyver's request in #1255. I've add some docs and the news file (we should probably set up something which checks that a news file is included, similar to the bot running on the python org).


takluyver:
Obvious question: do we need it?

I think I understand the need for specifying the pkgconfig name, e.g. if your packaging system distinguishes different configurations of hdf5, like `hdf5-openmpi`.

Do we know of use cases that need the lib and include dirs specified separately? Or are we adding them just in case?

takluyver:
If we do add all of them, I think we might need to think a bit more about how the different options interact if you specify multiple of them. I haven't tried to work out what should happen, but I suspect the logic currently here is not quite right.

kif:
Thomas, to your question, I would answer "yes, we need it". When doing packaging it is sometimes a nightmare to setup all options in the correct way, hence having environment variables it added value for packagers.

takluyver:
Specifically, you would benefit from being able to set the hdf5 library directory and include directory independently, rather than setting the parent 'hdf5 directory' and deriving those from it?

docs/build.rst
```diff
@@ -175,13 +175,16 @@ when installing via ``pip``, as you don't have direct access to setup.py::
...
-======================= =========================== ===========================
+=========================== ====================================== ===========================
+Option                      Via setup.py                           Via environment variable
+=========================== ====================================== ===========================
+Custom path to HDF5         ``--hdf5=/path/to/hdf5``               ``HDF5_DIR=/path/to/hdf5``
+Force HDF5 version          ``--hdf5-version=X.Y.Z``               ``HDF5_VERSION=X.Y.Z``
+Enable MPI mode             ``--mpi``                              ``HDF5_MPI=ON``
+Custom HDF5 include path    ``--hdf5-includedir=/path/to/headers`` ``HDF5_INCLUDEDIR=/path/to/headers``
+Custom HDF5 library path    ``--hdf5-libdir=/path/to/lib``         ``HDF5_LIBDIR=/path/to/lib``
+Custom HDF5 pkg-config name ``--hdf5-pkgconfig-name=hdf5``         ``HDF5_PKGCONFIG_NAME=hdf5``
```
takluyver, line 186: I think we should define how priority works if you specify multiple overlapping options - e.g. presumably the specific library & include paths have priority over the more general HDF5 dir.

Alternatively, we could adapt the configuration code to expect mutually exclusive options: either you specify the HDF5 dir, or the include & lib dirs, or the pkg-config name. Then it would error if it found options from different sets.

setup_configure.py
```diff
@@ -180,14 +224,14 @@ def run(self):
         print('*' * 80)
```
takluyver, line 224: Should the include & lib dirs be added to this diagnostic output?

baagaard-usgs:
The addition of specifying the include and lib directories is ver useful for cases (Linux) in which package managers often install parallel HDF5 into lower level directories like `/usr/include/hdf5/openmpi` and `/usr/lib/x86_64-linux-gnu/hdf5/openmpi`.

Specifying the HDF5 directory should be mutually exclusive with specifying the include and lib directories. I would give precedence to HDF5 directory (if it works then you shouldn't need to specify the include and lib directories).

In printing the config, I would print the path to the HDF5 include and lib directories (even in the case when the HDF5 directory is specified).

docs/build.rst
scopatz: Agreed

takluyver:
I've dug into this; the changes ended up a bit more involved than I was expecting.

- Environment variables & command line options are now mutually exclusive - it's an error to specify the same option both ways. The environment variables are specific to h5py, so I reckon that it's confusing if you think you're specifying something but it's actually being overridden.
- You can specify where to find HDF5 in one of three ways: separate lib/include dirs, a common prefix dir, or a name to look up in pkg-config. Specifying more than one is an error again.
- I reworked the logic a bit of how it saves config options for future use and detects when a rebuild is required.

I've made big enough changes to this that I'd like someone else to have a look at it before it's merged.

takluyver:
@aragilar, as you started this PR, would you have time to look over my changes to it?

aragilar:
This looks much cleaner to me @takluyver, thanks for finishing this off. I think that erroring in the cases where there could be confusion is probably best. I've tested this via tox with the serial and openmpi variants of HDF5 on my desktop, using the environment variables to configure the target library, and I didn't encounter any issues. Given this is a rewrite from my original PR, I think it's OK if I do the merge :).

takluyver:
Thanks! :-)

PR #1415, opened 2019-10-20
TST,CI: Add tests to coverage, with split on codecov

aragilar:
This is a test to see if we can have codecov report the coverage of the tests separately from the main code.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,2 @@
 [run]
-omit=*/tests/*
 parallel=true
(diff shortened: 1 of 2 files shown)


takluyver:
The separation doesn't seem to be working - as far as I can see, there's just one set of coverage numbers for everything. Do you want to try anything else with it, or shall we close this PR as a dead end?

aragilar:
It looks like there has to be multiple uploads to perform the splitting unfortunately, so yep, this is a dead end :(

PR #1421, opened 2019-10-24
Remove unused _RegionProxy class

takluyver:
The copy in h5py._hl.base is the one that's used.


takluyver:
Self-merging this one since it's just removing unused code.

PR #1423, opened 2019-10-25
Cython cleanup

takluyver:
This shouldn't change functionality at all, but it gets rid of various warnings in Pycharm.

It was prompted by the fact that Pycharm doesn't recognise the `for i from 0<=i<n` syntax - @kif started clearing this up in other PRs, and this removes all the remaining instances. Cython [automatically optimises](https://cython.readthedocs.io/en/latest/src/userguide/pyrex_differences.html#automatic-range-conversion) statements like `for i in range(...)` to C for-loops wherever possible. I've checked several of these with `cython -a` to see that they're benefitting from that.

I manually checked each import I was removing, since sometimes PyCharm erroneously highlights them as unused.

### DIFF
--- a/h5py/_conv.pyx
+++ b/h5py/_conv.pyx
@@ -91,7 +91,7 @@ cdef herr_t generic_converter(hid_t src_id, hid_t dst_id, H5T_cdata_t *cdata,
                         bkg + (i*bkg_stride),           # backing buffer
                         cdata[0].priv)                  # conversion context
             else:
-                for i from nl>i>=0:
+                for i in range(nl-1, -1, -1):
                     op( buf + (i*sizes[0].src_size),
                         buf + (i*sizes[0].dst_size),
                         bkg + (i*bkg_stride),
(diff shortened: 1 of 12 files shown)


scopatz:
This looks totally reasonable to me!

kif:
On Fri, 25 Oct 2019 01:31:54 -0700

> It was prompted by the fact that Pycharm doesn&#39;t recognise the `for i from 0&lt;=i&lt;n` syntax - @kif started clearing this up in other PRs, and this removes all the remaining instances. Cython [automatically optimises](https://cython.readthedocs.io/en/latest/src/userguide/pyrex_differences.html#automatic-range-conversion) statements like `for i in range(...)` to C for-loops wherever possible. I&#39;ve checked several of these with `cython -a` to see that they&#39;re benefitting from that.

I can confirm that cython translates `for i in range(a):` into `for (i=0; i<a; i++){}` which has no side effects (beside being slightly faster).
 
> I manually checked each import I was removing, since sometimes PyCharm erroneously highlights them as unused.

Most IDE are unable to parse properly cython code ... pydev (within eclipse) is not much better.

I don't rely on then when coding cython.

PR #1424, opened 2019-10-27
TST,CI: Run rever check on CI to reduce release friction

aragilar:
This adds running `rever check` to the list of things we run on the CI. It looks like the later tests try to modify this repo, so I'm not including them in the check. I'm not sure if there's some automated way of updating the .authors.yml, so feel free to fix it if necessary. I was looking for some way of checking whether the news files would combine fine and that the rst was valid, and checking whether a news file had been added, but I couldn't find an option to do that.

@scopatz Any suggestions about tasks that should be run, and if there's a way to check the news files/changelog rst output?

### DIFF
--- a/.authors.yml
+++ b/.authors.yml
@@ -611,3 +611,21 @@
   num_commits: 1
   first_commit: 2017-03-22 04:40:26
   github: felixonmars
+- name: Julia Garriga Ferrer
+  email: julia.garriga@esrf.fr
+  aliases:
+    - juliagarriga
+  alternate_emails:
+    - juliiaa28@gmail.com
+- name: Henri Payno
+  email: henri.payno@esrf.fr
+  aliases:
+    - payno
+  alternate_emails:
+    - payno@linazimov.esrf.fr
+    - payno@users.noreply.github.com
+    - henri.payno@gmail.com
+- name: Jerome Kieffer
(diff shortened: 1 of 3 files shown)


takluyver:
Thanks!

I suspect that the best way to validate the news entries is to run the necessary rever command to combine them before building the docs. I started trying to work out the commands for that, but got distracted onto other topics.

scopatz:
Thanks @aragilar! 

And yeah, I think that the check command in rever should just be extended to do a validation check as well.  Can you open an issue/PR for this, please? It is the best way to go for everyone. https://github.com/regro/rever

takluyver:
@scopatz - this is waiting on some input from you:

> Any suggestions about tasks that should be run, and if there's a way to check the news files/changelog rst output?

scopatz:
Sorry for not being super clear.  I think the activities listed here are the right ones, and adding different activities won't help much.  

I think the activity checks, such as for news and changelog could be improved in the ways mentioned.  However, I prefer to see that done as part of rever itself. If anyone is interested in contributing in that way, it would be great!

takluyver:
Thanks :-)

I think it would also be useful to assemble the changelog before building the docs, so we check that they're not breaking the docs build. Is that easy to achieve as well?

PR #1425, opened 2019-10-27
Add an example of storing units in HDF5

DBerke:
This example script demonstrates how to store unit information attached to an array of numbers using HDF5 in such a way that the original data can be recovered upon reading the file. It uses the Unyt package (https://github.com/yt-project/unyt) for handling units, but presumably similar methods would work for people using astropy.units or Pint. It relies on Unyt being able to parse its own string-ified versions of the units attached correctly, but as it should hopefully create such strings consistently it should work for the majority of use cases.

### DIFF
--- a/examples/store_and_retrieve_units_example.py
+++ b/examples/store_and_retrieve_units_example.py
@@ -1,0 +1,40 @@
+"""
+Author: Daniel Berke, berke.daniel@gmail.com
+Date: October 27, 2019
+Requirements: h5py>=2.10.0, unyt>=v2.4.0
+Notes: This short example script shows how to save unit information attached
+to a `unyt_array` using `attrs` in HDF5, and recover it upon reading the file.
+It uses the Unyt package (https://github.com/yt-project/unyt) because that's 
+what I'm familiar with, but presumably similar options exist for Pint and
+astropy.units. 
+"""
+
+import h5py
+import tempfile
+import unyt as u
+
+# Set up a temporary file for this example.
+tf = tempfile.TemporaryFile()
+f = h5py.File(tf, 'a')
+
+# Create some mock data with moderately complicated units (this is the 
+# dimensional representation of Joules of energy).
+test_data = [1, 2, 3, 4, 5] * u.kg * ( u.m / u.s ) ** 2
+print(test_data.units)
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks! It looks like while the examples aren't run in the tests, they are linted, and it's failing on a trailing whitespace check. Can you fix that? If you do clone the repository locally, you can run `tox -e pre-commit` to do the linting checks manually.

scopatz:
LGTM too!

ngoldbaum:
Cool! Glad you found `unyt` helpful.

One thing this won't help with is custom unit names (which come up a lot when working with modeling data), which is why we go to the trouble of serializing unit registries in `unyt`'s built-in support for serializing unit data into HDF5. For physical real-world unit names this will definitely work, in fact it's how we support loading units produced by `Pint` and `astropy.units` in `unyt`.

DBerke:
> Thanks! It looks like while the examples aren't run in the tests, they are linted, and it's failing on a trailing whitespace check. Can you fix that? If you do clone the repository locally, you can run `tox -e pre-commit` to do the linting checks manually.

@takluyver I'll try to get to that this weekend when I have time. Normally I'm pretty good about trailing whitespace, I'll have to see where some slipped past... :)

> Cool! Glad you found `unyt` helpful.
> 
> One thing this won't help with is custom unit names (which come up a lot when working with modeling data), which is why we go to the trouble of serializing unit registries in `unyt`'s built-in support for serializing unit data into HDF5. For physical real-world unit names this will definitely work, in fact it's how we support loading units produced by `Pint` and `astropy.units` in `unyt`.

@ngoldbaum Oh, wow, somehow I completely missed that functionality and it's just what I need—I may just switch over from my current implementation. I'm working with real astronomical data so it hasn't been a problem so far, at least. And yes, `unyt` has been _immensely_ helpful to me—I just wish I'd discovered it sooner than a year into my PhD!

DBerke:
OK, it was a single trailing space in a comment. No wonder I couldn't find it before! Looks like it passes now.

takluyver:
Thanks! Automated linting in CI can be a bit nitpicky.

DBerke:
Great! Glad to contribute. :)

Issue #993, opened 2018-01-30
Negative indices skip increasing order requirement


shoyer:
```
import h5py
import numpy as np

f = h5py.File('/tmp/test.h5', 'w')
f.create_dataset('data', data=np.arange(3))
f['data'][[2, 1]]
```
This raises:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
~/conda/envs/xarray-py36/lib/python3.6/site-packages/h5py/_hl/selections.py in select(shape, args, dsid)
     84             try:
---> 85                 int(a)
     86                 if isinstance(a, np.ndarray) and a.shape == (1,):

TypeError: int() argument must be a string, a bytes-like object or a number, not 'list'

During handling of the above exception, another exception occurred:

TypeError                                 Traceback (most recent call last)
<ipython-input-18-3a0bcbaae636> in <module>()
----> 1 f['data'][[2, 1]]

h5py/_objects.pyx in h5py._objects.with_phil.wrapper (/Users/ilan/minonda/conda-bld/h5py_1496887972496/work/h5py/_objects.c:2846)()

h5py/_objects.pyx in h5py._objects.with_phil.wrapper (/Users/ilan/minonda/conda-bld/h5py_1496887972496/work/h5py/_objects.c:2804)()

~/conda/envs/xarray-py36/lib/python3.6/site-packages/h5py/_hl/dataset.py in __getitem__(self, args)
    472
    473         # Perform the dataspace selection.
--> 474         selection = sel.select(self.shape, args, dsid=self.id)
    475
    476         if selection.nselect == 0:

~/conda/envs/xarray-py36/lib/python3.6/site-packages/h5py/_hl/selections.py in select(shape, args, dsid)
     88             except Exception:
     89                 sel = FancySelection(shape)
---> 90                 sel[args]
     91                 return sel
     92

~/conda/envs/xarray-py36/lib/python3.6/site-packages/h5py/_hl/selections.py in __getitem__(self, args)
    359                     adjacent = zip(list_arg[:-1], list_arg[1:])
    360                     if any(fst >= snd for fst, snd in adjacent):
--> 361                         raise TypeError("Indexing elements must be in increasing order")
    362
    363         if len(sequenceargs) > 1:

TypeError: Indexing elements must be in increasing order
```

However, I can convert this requirement by using -1 to index from the end instead:
```
f['data'][[-1, 1]]
# array([1, 2])
```

Summary of the h5py configuration
---------------------------------

[...]

PR #1426, opened 2019-10-28
Refactor selection code

takluyver:
I'm not 100% convinced by this, but I've been working on it on and off for a few days, and I'd like a reality check.

This is aimed at read performance, building on my work in #1402. I noticed that after that PR, about 1/3 of the time taken for small reads was in the selection code (15 µs per small read). Taking a small slice of a numpy array takes about 300 ns, so it seems like there's room for improvement. My thinking was to prioritise the 'normal' cases - standard dataset indexing and slicing - over more advanced things.

**The numbers**: this speeds up selection by something like 30-40%, and small reads by about 10% (~55 µs -> 49 µs).

As far as I can tell, this comes from two main sources:

- Avoiding repeated iteration over the indexing arguments - even though there aren't many, Python's iterator protocol has some overhead.
- Avoiding unnecessary HDF5 calls - don't call `select_all()` if it's not needed, and reuse & modify a cached SpaceID for repeated reads from the same dataset (with the same cache rules as for shape and size, from PR #1402).

I don't think this actually introduces a lot of extra complexity, but it does make the code longer and probably uglier. And the performance gains, while not nothing, are not the kind of massive leap I was hoping for. So maybe the best thing is to consider this a learning exercise and hope I come up with something better in the future.

----

asv graph (#1403):

Screenshot from 2019-10-28 14-24-21


takluyver:
Although, in its favour: I think I fixed #993 along the way. But that's feasible without the rest of the changes if we don't merge it.

scopatz:
Selection code is always pretty verbose.  I am totally in favor of this. It defintiely seems worth it to me

takluyver:
I now prefer my approach in #1428 - whereas this PR tries to speed up the fully general read pathway implemented largely in Python, that one makes a fast path in Cython for typical reads, and can be much faster. Maybe it's possible to expand that to the point where we don't need the hyperslab-selection code in Python anyway.

takluyver:
Closing this, as #1428 achieved a much bigger speedup. I'm hoping to look at using that code in a greater range of situations (without compromising its performance).

PR #1428, opened 2019-10-30
Faster read using Cython code

takluyver:
My latest experiment with small read performance. This has been much more successful - I can make many small reads from the same dataset about 10x faster than with the high-level API.

This is how to try it out:

```python
from h5py._reader import Reader

rdr = Reader(ds.id)
arr = rdr.read((slice(0, 10),))
```

10k small reads takes about 50-60 ms reusing the `Reader` object and about 225 ms recreating it each time, versus about 600 ms with existing high-level APIs. A plain C example can do the reads in about 10 ms, so we're within an order of magnitude.

(**Update**: When the high-level slicing API uses a cached Reader object, the same reads take about 70 ms. So there's a bit of overhead involved for the nice API, but it's still much faster than before.)

My thinking is that `Dataset.__getitem__()` would automatically used a cached `Reader` object if certain conditions were met. It would be cached like shape & size - only for readonly files, and invalidated by a SWMR refresh. At present, it works for:

- Datasets with at least 1 dimension
- Data types exactly matching numpy builtin types
- Selections using only integers, slices and Ellipsis

This is just a prototype, so we could make this class more flexible. But the aim is to be fast for common cases - HDF5's strength is in storing arrays of numbers, so I think it's fine to have some limitations and let other cases (like scalar data spaces or vlen strings) be handled by slower code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ h5py/_conv.c
 h5py/_proxy.c
 h5py/_objects.c
 h5py/_errors.c
+h5py/_reader.c
 h5py/config.pxi
 h5py/_hdf5.*
 h5py/*.dll
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -19,7 +19,7 @@ from threading import local
 
 import numpy
 
-from .. import h5, h5s, h5t, h5r, h5d, h5p, h5fd, h5ds
+from .. import h5, h5s, h5t, h5r, h5d, h5p, h5fd, h5ds, _reader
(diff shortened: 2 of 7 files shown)


takluyver:
I think this is ready for initial review. I've plumbed it into the high-level reading code, and all the tests are passing.

scopatz:
This looks awesome! Can you add a news entry, please?

kif:
I'll give it a try for benchmarking ...

kif:
I got 3x speed up in reading. Congratulations.

takluyver:
Thanks :-)

PR #1431, opened 2019-11-01
Named datasets not created as anonymous - closes #1422

### LINKED ISSUE #1422: Dataset creation (anonymous + hard link)
Hi everyone,
I've noticed that every dataset is created as anonymous and then its name is assigned as a hard link.

To be more clear: [make_new_dset](https://github.com/h5py/h5py/blob/b60bc5bb419bbd3fdd200e6f97d054b5b52399d6/h5py/_hl/dataset.py#L164) is creating anonymous datasets by always passing a `None` name and - right after - a new hard link creation [gets triggered](https://github.com/h5py/h5py/blob/b60bc5bb419bbd3fdd200e6f97d054b5b52399d6/h5py/_hl/group.py#L140). 

The same behavior is replicated for virtual datasets. 

What's the reason behind the current design choice? I am asking because it causes problems when dealing with some virtual object layers that are currently not supporting links.

I trivially solved the problem but - before making a pull request - I would like to know if I am missing some detail.

Thank you

pierlauro:
This PR is solving the problem highlighted in #1422 .

Since the changes are simple and internal - as the modified functions are never directly called by h5py users - I did not include any changelog entry.

### DIFF
--- a/h5py/_hl/group.py
+++ b/h5py/_hl/group.py
@@ -134,7 +134,7 @@ class Group(HLObject, MutableMappingHDF5):
             kwds['track_order'] = h5.get_config().track_order
 
         with phil:
-            dsid = dataset.make_new_dset(self, shape, dtype, data, name **kwds)
+            dsid = dataset.make_new_dset(self, shape, dtype, data, name, **kwds)
             dset = dataset.Dataset(dsid)
             return dset
 
--- a/h5py/_hl/group.py
+++ b/h5py/_hl/group.py
@@ -167,10 +167,12 @@ class Group(HLObject, MutableMappingHDF5):
                 if name:
                     name = str.encode(name)
                 dsid = dataset.make_new_virtual_dset(self, layout.shape,
-                         sources=sources, dtype=layout.dtype, name=name,
+                         sources=sources, dtype=layout.dtype,
                          maxshape=layout.maxshape, fillvalue=fillvalue)
 
                 dset = dataset.Dataset(dsid)
+                if name is not None:
+                    self[name] = dset
 
             return dset
 
--- a/h5py/_hl/group.py
+++ b/h5py/_hl/group.py
@@ -142,7 +142,7 @@ class Group(HLObject, MutableMappingHDF5):
                     h5objects = h5objects[:-1]
 
                     for new_group in h5objects:
-                        group = self.get(new_group) or group.create_group(new_group)
+                        group = group.get(new_group) or group.create_group(new_group)
 
                 name = self._e(name)
 
--- a/h5py/_hl/group.py
+++ b/h5py/_hl/group.py
@@ -163,12 +163,10 @@ class Group(HLObject, MutableMappingHDF5):
 
             with phil:
                 dsid = dataset.make_new_virtual_dset(self, layout.shape,
-                         sources=sources, dtype=layout.dtype,
+                         sources=sources, dtype=layout.dtype, name=name,
                          maxshape=layout.maxshape, fillvalue=fillvalue)
 
                 dset = dataset.Dataset(dsid)
-                if name is not None:
-                    self[name] = dset
 
             return dset
 
--- a/h5py/_hl/group.py
+++ b/h5py/_hl/group.py
@@ -134,9 +134,19 @@ class Group(HLObject, MutableMappingHDF5):
             kwds['track_order'] = h5.get_config().track_order
 
         with phil:
+            group = self
             if name:
-                   name = str.encode(name)
-            dsid = dataset.make_new_dset(self, shape, dtype, data, name, **kwds)
+                if '/' in name:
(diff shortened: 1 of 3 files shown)


pierlauro:
There seem to be some problems on creating datasets in not existent groups, strangely doesn't give me the same errors on the latest develop HDF5. I will work on that.

kif:
Overall, it looks good to me. I don't know the rules concerning the change in the API and the level of "pubilcity" of those functions.

h5py/_hl/group.py
```diff
@@ -134,10 +134,20 @@ def create_dataset(self, name, shape=None, dtype=None, data=None, **kwds):
...
-            dsid = dataset.make_new_dset(self, shape, dtype, data, **kwds)
+            group = self
+            if name:
+                if '/' in name:
+                    h5objects = [obj for obj in name.split('/') if len(obj)]
+                    name = h5objects[-1]
+                    h5objects = h5objects[:-1]
+
+                    for new_group in h5objects:
+                        group = self.get(new_group) or group.create_group(new_group)
```
takluyver: Should this be `group.get(new_group)` (instead of `self...`)?

h5py/_hl/group.py
```diff
@@ -134,10 +134,20 @@ def create_dataset(self, name, shape=None, dtype=None, data=None, **kwds):
...
+
+                name = str.encode(name)
```
takluyver: I think we always do encoding with `self.e_(name)`. It will do the same in this case, but the consistency is useful if we ever need to change how that works.

h5py/_hl/group.py
takluyver: Also, I think you can use `self.require_group()` to create a group if it doesn't exist. That also does the iteration for nested levels, so you don't need the loop.

takluyver:
I think it's OK to change those APIs, especially for 3.0 - `make_new_dset` et al are only available by importing from `h5py._hl`, and they're not documented.

I'm just slightly nervous about this because I don't know why the anonymous creation was written in the first place, so I don't know if something breaks by removing it. It feels a bit like the kind of situation where we fix things for one use case by breaking it for another one.

h5py/_hl/group.py
pierlauro: > Should this be `group.get(new_group)` (instead of `self...`)?

Totally right, just fixed.


> Also, I think you can use `self.require_group()` to create a group if it doesn't exist. That also does the iteration for nested levels, so you don't need the loop.

Does it work also for absolute paths or is it just creating subgroups under the group the method is invoked on? The tests are failing if I use  `require_group`

takluyver: > Does it work also for absolute paths or is it just creating subgroups under the group the method is invoked on?

Looking at the code I think it should work, but maybe something isn't handling absolute paths properly. If it's not easy to fix, we can leave it with the iteration for this PR and open an issue to investigate it.

pierlauro:
> I'm just slightly nervous about this because I don't know why the anonymous creation was written in the first place, so I don't know if something breaks by removing it. It feels a bit like the kind of situation where we fix things for one use case by breaking it for another one.

The behavior comes from a [very old commit](https://github.com/h5py/h5py/blob/bd96ddd8204eccbd06aab76034c992a13e7ee5a3/h5py/_hl/group.py#L48)  titled `More work on encoding`. I guess it was just a stylistic choice to don't manually encode `name` when calling `make_new_dset` by delegating the encoding to `h5o.link` that was doing it by itself.

h5py/_hl/group.py
pierlauro: I can check it when I will have time. The easiest way to check is to write some tests for ` require_group`.  I would propose to close this PR and I will work on it separately in an issue assigned to me.

takluyver: I think you changed to `group.get()` below, but there's still a `self.get()` on this line.

takluyver:
This looks ok to me other than the remaining `self.get()` bit. It would also be good to have a test that would have failed on those - i.e. creating a dataset 3 levels down: `f.create_dataset('foo/bar/baz')`.

Can you open an issue about `require_dataset`, since you found the problem with that.

pierlauro:
> I think you changed to `group.get()` below, but there's still a `self.get()` on this line.

Fixed and local tests passing. The [CI pipeline](https://github.com/h5py/h5py/pull/1431/checks?check_run_id=292355314) seems to have a configuration problem though.

takluyver:
@aragilar if I don't get to it (which seems likely), can you investigate what's going on with Azure pipelines here. It says:

> The pipeline is not valid. Could not get the latest source version for repository h5py/h5py hosted on https://github.com/ using ref 0f3f16de16577f43a2e151b1d9fc9bf95b964f4b. GitHub reported the error, "Bad credentials"

takluyver:
Reopening to try to rerun the tests.

PR #1440, opened 2019-11-25
Unregister converters

kif:
This prevents seg-fault at exit where Python code may be called after
the interpreter is already half way destroyed, usually in threaded environment.

It is a simpler PR extracted from the "nogil" part I am working on ...

### DIFF
--- a/.authors.yml
+++ b/.authors.yml
@@ -611,3 +611,5 @@
   num_commits: 1
   first_commit: 2017-03-22 04:40:26
   github: felixonmars
+- name: Pierlauro Sciarelli
+  email: foss@pstux.dev
--- a/h5py/__init__.py
+++ b/h5py/__init__.py
@@ -13,6 +13,7 @@
 """
 
 from warnings import warn as _warn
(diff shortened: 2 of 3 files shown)


takluyver:
I see we've both tried to fix the failing rever check. There's a second email address; feel free to copy my change in #1441, or merge that and rebase this one if you prefer.

takluyver:
CC @scopatz: it's unfortunate that when rever checks author emails and finds two missing (for the same name), the error message only seems to be mention one.

takluyver:
I remembered that the best way to report bugs is on a project's issue tracker. :wink: https://github.com/regro/rever/issues/207

takluyver:
Thanks, LGTM.

Wording nitpick for the changelog: starting with "this prevents segfaults..." makes sense for a commit message or a PR description, where it's clear what the "this" is, but I think it will read strangely in the release notes, when it's disconnected from the code change. I'd start with "Prevent segfaults..." instead.

We can fix that at release time, though.

PR #1442, opened 2019-11-25
Introduction to what 'Parallel HDF5' is

takluyver:
This is the best of my understanding, at least. In particular, I want to make it clear that you don't need 'Parallel HDF5' if you just want several processes to read data from the same file.

### DIFF
--- a/docs/mpi.rst
+++ b/docs/mpi.rst
@@ -3,9 +3,14 @@
 Parallel HDF5
 =============
 
-Starting with version 2.2.0, h5py includes support for Parallel HDF5.  This
-is the "native" way to use HDF5 in a parallel computing environment.
+Read-only parallel access to HDF5 files works with no special preparation:
+each process should open the file independently and read data normally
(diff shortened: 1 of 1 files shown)


kif:
Hi Thomas,

I had a quick look at this documentation file and it still shows the "autotools" way of building HDF5 while the HDFgroup is now promoting "cmake". This part needs probably to be upgraded as well.

As I don't use parellel HDF5, so I have only a limited know-how on all this subject.

Cheers,

J.

aragilar:
This seems pretty reasonable to me, merging.

@kif It looks like the installation docs for the parallel version (`release_docs/INSTALL_parallel`) still seems to use autotools in the examples, it looks like cmake is recommended on Windows, but I haven't seen anything about it being preferred on other systems, or with a parallel install (there's logic in the cmake build to do a parallel install, but I'd be wary of recommending installing the parallel version using cmake if it isn't not widely tested/used).

Issue #492, opened 2014-10-17
Assignment through dimension-reducing slice is very slow


bmerry:
For some reason, `ds[..., i] = data` is much slower than `ds[..., i : i + 1] = data[..., np.newaxis]`, where `data` is a numpy array, even though exactly the same data is transferred between the same locations.

```
Summary of the h5py configuration
---------------------------------

h5py    2.3.1
HDF5    1.8.11
Python  2.7.6 (default, Mar 22 2014, 22:59:56) 
[GCC 4.8.2]
sys.platform    linux2
sys.maxsize     9223372036854775807
numpy   1.9.0
```

Here is sample code, based on stripping down the code I was writing. It writes a 512MiB file, and takes about 2.5s when run with `--fast` but 16s when run without.

``` python
#!/usr/bin/env python
import h5py
import numpy as np
import argparse

dim = (128, 2048, 512)

def create_file(filename, cache_entries, cache_size):
    """Creates a file with a user-specified cache size"""
    propfaid = h5py.h5p.create(h5py.h5p.FILE_ACCESS)
    settings = list(propfaid.get_cache())
    settings[1] = cache_entries
    settings[2] = cache_size
    propfaid.set_cache(*settings)
    return h5py.File(h5py.h5f.create(filename, fapl=propfaid), 'w', libver='latest')

parser = argparse.ArgumentParser()
parser.add_argument('--fast', action='store_true')
args = parser.parse_args()
data = np.zeros(dim[0:2], dtype=np.float32)
outf = create_file('slow.hdf5', 524287, 512 * 1024**2)
ds = outf.create_dataset("data", dim, dtype=np.float32, chunks=(1, dim[1], 64))
for i in range(dim[2]):
    if args.fast:
        ds[..., i:i+1] = data[..., np.newaxis]
    else:
        ds[..., i] = data
```

takluyver:
There is a similar bug with reading, although my recent changes in #1428 avoid it in the most common scenarios.

I think I have roughly understood the cause of this, related to this bit of code in the read & write pathways:

https://github.com/h5py/h5py/blob/eede14665c46bd774124bdec925b7318c5e0fe81/h5py/_hl/dataset.py#L614-L618

[...]

PR #1444, opened 2019-11-26
Attempt to fix performance problems reading/writing with integer indexing

takluyver:
See e.g. #492

When you use an integer index for an inner dimension, I think h5py is constructing a mismatched shape, because it always adds the padding dimensions at the beginning of the shape. In the example from #492:

- Dataset shape: (128, 2048, 512)
- Indexing: `dset[..., i]`
- Selection shape (HDF5): (128, 2048, 1)
- Selection shape (Python): (128, 2048)
- Padded selection shape (`mshape_pad`): (1, 128, 2048)

I guess that trying to convert from (128, 2048, 1) to (1, 128, 2048) triggers some pathological behaviour in HDF5. From the comments, it looked like padding the shape was an attempt to work around that, but always adding the dimensions at the start of the shape isn't always right. We also found a case where reading with an integer index was ~60x slower than with a slice - but this is partly avoided on master by my fast-path reader (#1428).

This PR tries to better match the array dimensions to dataset dimensions for mspace.

- For `__getitem__`, we preserve a copy of the selected shape without dropping dimensions with a scalar selection. The only hard bit is naming: I've gone for `mshape` for the shape preserving all dimensions, and `array_shape` for the shape with scalar dimensions dropped.
- For `__setitem__`, we can't just use the selection shape, because we may be broadcasting a smaller array. So the `selection.expand_shape()` method, refactored out of `selection.broadcast()`, matches the array dimensions up to the selection dimensions.

Testing the example in #492, this does seem to make writing with `[..., i]` as fast as `[..., i:i+1]`.

Struggling to understand the `broadcast()` method, I realised that the `target_shape` parameter actually describes the shape of the *source* data which is to be broadcast to the selection. So I've also renamed that.

### DIFF
--- a/benchmarks/benchmarks.py
+++ b/benchmarks/benchmarks.py
@@ -26,3 +26,32 @@ class TimeSuite:
         ds = self.f['a']
         for i in range(10000):
             arr = ds[i * 10:(i + 1) * 10]
+
+class WritingTimeSuite:
+    """Based on example in GitHub issue 492:
+    https://github.com/h5py/h5py/issues/492
+    """
+    def setup(self):
+        self._td = TemporaryDirectory()
+        path = osp.join(self._td.name, 'test.h5')
+        self.f = h5py.File(path, 'w')
+        self.shape = shape = (128, 1024, 512)
+        self.f.create_dataset(
+            'a', shape=shape, dtype=np.float32, chunks=(1, shape[1], 64)
+        )
+
+    def teardown(self):
+        self.f.close()
+        self._td.cleanup()
+
+    def time_write_index_last_axis(self):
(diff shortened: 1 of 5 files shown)


scopatz:
LGTM!

takluyver:
I've added a benchmark based on the example from #492. Here are the results in miniature:

Screenshot from 2019-12-03 11-57-41

The left hand side shows the affected case (`ds[..., i] = `), and the right hand side shows the workaround that was already quick (`ds[..., i:i+1] =`). The y scales are not the same - they are about the same after the change.

takluyver:
I'll merge this in a couple of days unless anyone objects or wants more time to review it.

PR #1451, opened 2019-11-29
Don't assume that thread-local variable astype exists

takluyver:
It's only guaranteed to exist in the thread that created the dataset object.

@kif, this should fix the AttributeError you mentioned in this comment:
https://github.com/h5py/h5py/pull/1412#issuecomment-559566637

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -616,7 +616,10 @@ class Dataset(HLObject):
         """
         args = args if isinstance(args, tuple) else (args,)
 
-        if self._fast_read_ok and (new_dtype is None) and (self._local.astype is None):
+        if new_dtype is None:
+            new_dtype = getattr(self._local, 'astype', None)
+
+        if self._fast_read_ok and (new_dtype is None):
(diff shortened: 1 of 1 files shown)


kif:
Hi Thomas,
I tested you code and it works in a multi-threaded environment. No speed-up is seen but that was expected (it is 20% slower with threads)

takluyver:
Yup, that's what I'd expect too. As I understand it, HDF5 itself serialises access from multiple threads, so no matter what we do in h5py, threading isn't going to give a speed up. We're just aiming to ensure it doesn't crash.

(Of course, as in @t20100's test with your branch, you can speed things up if one thread is accessing HDF5 while another thread is doing something unrelated)

kif:
> (Of course, as in @t20100's test with your branch, you can speed things up if one thread is accessing HDF5 while another thread is doing something unrelated)

I believe this is our main usecase. I just quickly benchmarked and got 6-7 GB/s from h5py and 12 GB/s using hdparm -T (read from cache). That's reasonable.

PR #1452, opened 2019-11-29
Remove unused wrapper for H5Eset_auto

takluyver:
The autogenerated wrapper function is not needed - the C functions for error handling are exposed directly in `h5py._errors`. To avoid confusion, I've left a comment in the function listing explaining that.

cc @kif


kif:
OK for me.

PR #1453, opened 2019-11-29
Nogil for dataset access (H5D)

kif:
Accessing to datasets has been benchmarked the most time consuming part in h5py.
This PR releases the GIL for all H5D function calls and ensures all (?) callbacks are re-acquiring the GIL as needed.
Even if all tests are passing, this does not mean it is bug free due to the deep changes this implies.


h5py/api_functions.txt
```diff
@@ -80,43 +80,43 @@ hdf5:
...
+  hid_t     H5Dopen(hid_t file_id, char *name) nogil
+  hid_t     H5Dopen2(hid_t loc_id, char *name, hid_t dapl_id ) nogil
+  herr_t    H5Dclose(hid_t dset_id) nogil
 
-  hid_t     H5Dget_space(hid_t dset_id)
-  herr_t    H5Dget_space_status(hid_t dset_id, H5D_space_status_t *status)
-  hid_t     H5Dget_type(hid_t dset_id)
+  hid_t     H5Dget_space(hid_t dset_id) nogil
+  herr_t    H5Dget_space_status(hid_t dset_id, H5D_space_status_t *status) nogil
+  hid_t     H5Dget_type(hid_t dset_id) nogil
```
takluyver: I'm not sure whether it makes sense to release the GIL for all of these little operations, which must do a tiny amount of I/O if any (I don't know if they access data already in memory). The [Cython docs](https://cython.readthedocs.io/en/latest/src/userguide/external_C_code.html#acquiring-and-releasing-the-gil) say:

> Note that acquiring the GIL is a blocking thread-synchronising operation, and therefore potentially costly. It might not be worth releasing the GIL for minor calculations. Usually, I/O operations and substantial computations in parallel code will benefit from it.

Maybe it makes sense to only release it for the operations that can do significant amounts of I/O, like H5Dread & H5Dwrite?

takluyver:
As I mentioned on the previous PR, I think it would be good to have some kind of stress test that tries to use h5py through several threads at once, to give us some chance of spotting if there's some issue with thread safety.

It doesn't need to be part of this PR, and it doesn't need to run on every CI job, e.g. if it takes longer than the normal tests.

kif:
Hi Thomas,
I have a stress-test which requires 0.25G of RAM and 4G of storage. I know that CI virtual instances have rather limited capabilities. Do you have an idea of what is an acceptable minimum? 
Is 2 threads, 512M ram 4G storage small enough ? We could artificially increase the number of threads, it will just be serialized hence remove the usefulness of the test.
I consider putting it in the benchmark part you recently added.

takluyver:
I'd say add the code and record how to run it, but don't run it as part of the standard test suite (either by where you put it, or skip it by default). Though it looks like [the limits on Travis CI](https://docs.travis-ci.com/user/reference/overview/) would allow running it there.

h5py/api_functions.txt
```diff
@@ -80,12 +80,12 @@ hdf5:
...
-  hid_t     H5Dcreate_anon(hid_t file_id, hid_t type_id, hid_t space_id, hid_t plist_id, hid_t dapl_id)
+  hid_t     H5Dcreate2(hid_t loc_id, char *name, hid_t type_id, hid_t space_id, hid_t lcpl_id, hid_t dcpl_id, hid_t dapl_id) nogil
+  hid_t     H5Dcreate_anon(hid_t file_id, hid_t type_id, hid_t space_id, hid_t plist_id, hid_t dapl_id) nogil
 
-  hid_t     H5Dopen(hid_t file_id, char *name)
-  hid_t     H5Dopen2(hid_t loc_id, char *name, hid_t dapl_id )
-  herr_t    H5Dclose(hid_t dset_id)
+  hid_t     H5Dopen(hid_t file_id, char *name) nogil
+  hid_t     H5Dopen2(hid_t loc_id, char *name, hid_t dapl_id ) nogil
+  herr_t    H5Dclose(hid_t dset_id) nogil
```
takluyver: I'd be inclined to keep the GIL for things like open/close, iterate, etc. They're presumably only a tiny bit of I/O before it goes back to Python code. Maybe you know more about this, though?

h5py/_errors.pyx
```diff
@@ -84,7 +84,7 @@ cdef struct err_data_t:
     H5E_error_t err
     int n
 
-cdef herr_t walk_cb(int n, H5E_error_t *desc, void *e):
+cdef herr_t walk_cb(int n, H5E_error_t *desc, void *e) with gil:
```
takluyver: I don't think this one actually does need the GIL - it's entirely manipulating a cdef-ed struct. It's also only ever called in `set_exception()`, which is called after we've reobtained the GIL. Probably it doesn't add much overhead, but for the sake of clarity, I'd prefer we only add `with gil` on functions that are likely to need it.

benchmarks/benchmark_slicing.py
```diff
@@ -0,0 +1,184 @@
...
+        logger.info("Time for %s-threaded reading %sx%s slices: %.3fs fps: %.3f "%(nthreads, self.ndim, nb_read, t1-t0, self.ndim*nb_read/(t1-t0)) +
+                    "Uncompressed data read speed %.3f MB/s"%(self.ndim*nb_read*self.needed_memory/(t1-t0)/1e6))
+        return t1 - t0
+
+
+if __name__ == "__main__":
+    logging.basicConfig(level=logging.INFO)
+    #cmp = hdf5plugin.Blosc(cname='lz4', clevel=5, shuffle=2)
+    #cmp = hdf5plugin.Bitshuffle()
+    #print(dict(cmp))
```
scopatz: Please remove commented out code

h5py/api_functions.txt
scopatz: It makes sense for a lot of this to be `nogil` to me, fwiw

h5py/_errors.pyx
kif: There are indeed a few callback functions which could be declared "nogil". I have been considered this option to flag them as "safe for C"

h5py/api_functions.txt
kif: I had a quick look and those function are mainly error-handling and ref-counting. It is probably not worth releasing the GIL for them.

benchmarks/benchmark_slicing.py
```diff
@@ -0,0 +1,182 @@
...
+    def __init__(self, queue_in, queue_out, quit_event):
+        Thread.__init__(self)
+        self._queue_in = queue_in
+        self._queue_out = queue_out
+        self._quit_event = quit_event
+        self.daemon = True
+        self.start()
+
+    def run(self):
+        while not self._quit_event.is_set():
```
takluyver: Perhaps I'm missing something, but it looks like you don't need a quit event, because the poison pill breaks out of the loop by itself.

h5py/_proxy.pyx
```diff
@@ -280,8 +279,7 @@ cdef herr_t h5py_scatter_cb(void* elem, hid_t type_id, unsigned ndim,
     return 0
 
 cdef herr_t h5py_gather_cb(void* elem, hid_t type_id, unsigned ndim,
-                const hsize_t *point, void *operator_data) except -1:
-
+                const hsize_t *point, void *operator_data) except -1 with gil:
```
takluyver: I think these scatter/gather ones don't need the gil either.

benchmarks/benchmark_slicing.py
kif: "belt and suspenders"

takluyver: Fair enough!

takluyver:
LGTM, assuming CI passes. Is there anything else you want to look at before we merge this?

kif:
Did anybody try the benchmark provided ? does it make sense to you ?
On my side, the work is done. You can merge when CI passes.

takluyver:
Do these numbers look like you'd expect?

```
$ python3 benchmark_slicing.py 
INFO:__main__:Saving data in /home/takluyver/Code/h5py/benchmarks/tmpjc2fo6ao/benchmark_slicing.h5
INFO:__main__:Total size: 1024^3 volume size: 4.295GB, Needed memory: 0.268GB
INFO:__main__:Compression: 1.000 time 33.005s uncompressed data saving speed 130.130 MB/s effective write speed  130.137 MB/s 
INFO:__main__:Time for reading 3x64 slices: 102.907s fps: 1.866 Uncompressed data read speed 500.837 MB/s
INFO:__main__:Time for 4-threaded reading 3x64 slices: 76.816s fps: 2.499 Uncompressed data read speed 670.948 MB/s
```

(Ignore the path - I had to tweak the code because my laptop has a 3.9 GB partition for /tmp)

takluyver:
I realised while writing another benchmark that asv has a default timeout of 60 seconds, and I think it wants to run the benchmark function at least twice within that time. At least on my laptop, this benchmark runs into the timeout.

[It's possible](https://asv.readthedocs.io/en/stable/benchmarks.html) to increase the timeout, but I think it might be better to shrink the example down to take more like 10-20 seconds. asv makes it easy to run the benchmark across a range of commits, and that's going to be more practical if benchmarks complete relatively quickly.

kif:
On Tue, 03 Dec 2019 03:06:32 -0800

> Do these numbers look like you'd expect?
> 
> ```
> $ python3 benchmark_slicing.py 
> INFO:__main__:Saving data in /home/takluyver/Code/h5py/benchmarks/tmpjc2fo6ao/benchmark_slicing.h5
> INFO:__main__:Total size: 1024^3 volume size: 4.295GB, Needed memory: 0.268GB
> INFO:__main__:Compression: 1.000 time 33.005s uncompressed data saving speed 130.130 MB/s effective write speed  130.137 MB/s 
> INFO:__main__:Time for reading 3x64 slices: 102.907s fps: 1.866 Uncompressed data read speed 500.837 MB/s
> INFO:__main__:Time for 4-threaded reading 3x64 slices: 76.816s fps: 2.499 Uncompressed data read speed 670.948 MB/s
> ```

I typically get:

```
INFO:__main__:Total size: 1024^3volume size: 4.295GB, Needed memory: 0.268GB
INFO:__main__:Compression: 1.000 time 5.783s uncompressed data saving speed 742.665 MB/s effective write speed  742.706 MB/s 
INFO:__main__:Time for reading 3x64 slices: 7.346s fps: 26.135 Uncompressed data read speed 7015.552 MB/s
INFO:__main__:Time for 12-threaded reading 3x64 slices: 9.138s fps: 21.011 Uncompressed data read speed 5640.049 MB/s
```
on a fairly recent workstation (Xeon E5-1650 v4) with a SATA-SSD (should not mater as we stay in RAM). 
For information, single threaded memory bandwidth is 12101.61 MB/sec on my computer, so there is still a factor 16 to win.

Funny to see that threading helps in your case

kif:
On Tue, 03 Dec 2019 04:26:38 -0800

> I realised while writing another benchmark that asv has a default timeout of 60 seconds, and I think it wants to run the benchmark function at least twice within that time. At least on my laptop, this benchmark runs into the timeout.
> 
> [It's possible](https://asv.readthedocs.io/en/stable/benchmarks.html) to increase the timeout, but I think it might be better to shrink the example down to take more like 10-20 seconds. asv makes it easy to run the benchmark across a range of commits, and that's going to be more practical if benchmarks complete relatively quickly.

This comes down to the question I asked about the hardware underneath ... for me it runs in 20s

I would recommend you not to perform benchmarking/profiling on a laptop
due to their aggressive frequency scaling on the one hand and their
tendency to overheat/throttle on the other.

I don't know what you have access locally or remotely in Hambourg. 
Here in Grenoble, we have a few dedicated computers for this kind of
tasks. We could probably open you an account as part of the Panosc
project.

takluyver:
Fair enough. I can reserve a cluster machine at EuXFEL, that's not a problem. I just prefer to work on my laptop where possible. Apart from anything else, I always have to mess around with conda to get it to build correctly.

takluyver:
On the cluster machine, the threaded benchmark segfaults. I tried enabling faulthandler, but it doesn't seem to play nicely with multithreading:

```
INFO:__main__:Saving data in /tmp/tmpuqxuy0iy/benchmark_slicing.h5
INFO:__main__:Total size: 1024^3 volume size: 4.295GB, Needed memory: 0.268GB
INFO:__main__:Compression: 1.000 time 10.241s uncompressed data saving speed 419.407 MB/s effective write speed  419.430 MB/s 
INFO:__main__:Time for reading 3x64 slices: 10.106s fps: 18.998 Uncompressed data read speed 5099.756 MB/s
Fatal Python error: Segmentation fault

Thread 0x00002b4acd98f700 (most recent call first):
  File "/usr/lib64/python3.6/threading.py", line 295 in wait
  File "/usr/lib64/python3.6/queue.py", line 164 in get
  File "benchmzsh: segmentation fault  .asv/env/541b6fbd95fdb429e84621bc981fbd19/bin/python3 
```

This is currently built with HDF5 1.8.12, which might be causing the issue. I didn't manage to get it to build in a conda env with a newer HDF5 yet. :-/

takluyver:
I've got it working for a single run `asv run --quick` with HDF5 1.10.4, but it fails when asv repeats the benchmark, as it does by default.

asv calls `teardown()` and `setup()` between consecutive runs. teardown() is deleting the temporary directory, which setup() doesn't recreate, so it fails on the second setup(). Can we fix this so that it works with asv?

kif:
On Tue, 03 Dec 2019 06:14:21 -0800

> Thread 0x00002b4acd98f700 (most recent call first):
>   File "/usr/lib64/python3.6/threading.py", line 295 in wait
>   File "/usr/lib64/python3.6/queue.py", line 164 in get
>   File "benchmzsh: segmentation fault  .asv/env/541b6fbd95fdb429e84621bc981fbd19/bin/python3 
> ```
> 
> This is currently built with HDF5 1.8.12, which might be causing the issue. I didn't manage to get it to build in a conda env with a newer HDF5 yet. :-/

do you have the opportunity to test it in GDB ?

gdb python 
run benchmark_slicing.py

then "bt" to print the stack

takluyver:
OK, I've done that. To be clear, this is only with the old system version of HDF5 - I didn't see the segfault with the newer version.

```
Program received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x2aab36b53700 (LWP 16707)]
0x00002aab41a38392 in H5FL_reg_malloc () from /lib64/libhdf5.so.8
Missing separate debuginfos, use: debuginfo-install bzip2-libs-1.0.6-13.el7.x86_64 glibc-2.17-292.el7.x86_64 hdf5-1.8.12-11.el7.x86_64 keyutils-libs-1.5.8-3.el7.x86_64 krb5-libs-1.15.1-37.el7_7.2.x86_64 libaec-1.0.4-1.el7.x86_64 libcom_err-1.42.9-16.el7.x86_64 libffi-3.0.13-18.el7.x86_64 libselinux-2.5-14.1.el7.x86_64 libuuid-2.23.2-61.el7.x86_64 openssl-libs-1.0.2k-19.el7.x86_64 pcre-8.32-17.el7.x86_64 python3-libs-3.6.8-10.el7.x86_64 xz-libs-5.2.2-1.el7.x86_64 zlib-1.2.7-18.el7.x86_64
(gdb) bt
#0  0x00002aab41a38392 in H5FL_reg_malloc () from /lib64/libhdf5.so.8
#1  0x00002aab41b54eb9 in H5S_hyper_make_spans () from /lib64/libhdf5.so.8
#2  0x00002aab41b55932 in H5S_generate_hyperslab () from /lib64/libhdf5.so.8
#3  0x00002aab41b57311 in H5S_select_hyperslab () from /lib64/libhdf5.so.8
#4  0x00002aab419cb985 in H5D__create_chunk_file_map_hyper ()
   from /lib64/libhdf5.so.8
#5  0x00002aab419ca5e0 in H5D__chunk_io_init () from /lib64/libhdf5.so.8
#6  0x00002aab419eafee in H5D__read () from /lib64/libhdf5.so.8
#7  0x00002aab419e99a4 in H5Dread () from /lib64/libhdf5.so.8
#8  0x00002aab427a8a87 in __pyx_f_4h5py_4defs_H5Dread (
    __pyx_v_dset_id=83886080, __pyx_v_mem_type_id=50332062, 
    __pyx_v_mem_space_id=67109060, __pyx_v_file_space_id=67109059, 
    __pyx_v_plist_id=0, __pyx_v_buf=0x2aab480020e0)
    at /home/kluyvert/Code/h5py/h5py/defs.c:6668
#9  0x00002aab45e7fb07 in __pyx_pf_4h5py_7_reader_6Reader_4read (
    __pyx_v_args=<optimized out>, __pyx_v_self=0x2aaab71d5198)
    at /home/kluyvert/Code/h5py/h5py/_reader.c:4009
#10 __pyx_pw_4h5py_7_reader_6Reader_5read (__pyx_v_self=0x2aaab71d5198, 
    __pyx_v_args=<optimized out>)
    at /home/kluyvert/Code/h5py/h5py/_reader.c:3940
#11 0x00002aaaaadb50be in _PyCFunction_FastCallDict ()
   from /lib64/libpython3.6m.so.1.0
#12 0x00002aaaaae20a3f in call_function () from /lib64/libpython3.6m.so.1.0
---Type <return> to continue, or q <return> to quit---
#13 0x00002aaaaae150a7 in _PyEval_EvalFrameDefault ()
   from /lib64/libpython3.6m.so.1.0
#14 0x00002aaaaae1fa7c in _PyEval_EvalCodeWithName ()
   from /lib64/libpython3.6m.so.1.0
#15 0x00002aaaaae21d4e in _PyFunction_FastCallDict ()
   from /lib64/libpython3.6m.so.1.0
#16 0x00002aaaaad776be in _PyObject_FastCallDict ()
   from /lib64/libpython3.6m.so.1.0
#17 0x00002aaaaad777d1 in _PyObject_Call_Prepend ()
   from /lib64/libpython3.6m.so.1.0
#18 0x00002aaaaad77443 in PyObject_Call () from /lib64/libpython3.6m.so.1.0
#19 0x00002aaaaadca789 in call_method () from /lib64/libpython3.6m.so.1.0
#20 0x00002aaaaae1571a in _PyEval_EvalFrameDefault ()
   from /lib64/libpython3.6m.so.1.0
#21 0x00002aaaaae2082a in fast_function () from /lib64/libpython3.6m.so.1.0
#22 0x00002aaaaae20b63 in call_function () from /lib64/libpython3.6m.so.1.0
#23 0x00002aaaaae150a7 in _PyEval_EvalFrameDefault ()
   from /lib64/libpython3.6m.so.1.0
#24 0x00002aaaaae2082a in fast_function () from /lib64/libpython3.6m.so.1.0
#25 0x00002aaaaae20b63 in call_function () from /lib64/libpython3.6m.so.1.0
#26 0x00002aaaaae150a7 in _PyEval_EvalFrameDefault ()
   from /lib64/libpython3.6m.so.1.0
#27 0x00002aaaaae2082a in fast_function () from /lib64/libpython3.6m.so.1.0
---Type <return> to continue, or q <return> to quit---
#28 0x00002aaaaae20b63 in call_function () from /lib64/libpython3.6m.so.1.0
#29 0x00002aaaaae150a7 in _PyEval_EvalFrameDefault ()
   from /lib64/libpython3.6m.so.1.0
#30 0x00002aaaaae21c5a in _PyFunction_FastCallDict ()
   from /lib64/libpython3.6m.so.1.0
#31 0x00002aaaaad776be in _PyObject_FastCallDict ()
   from /lib64/libpython3.6m.so.1.0
#32 0x00002aaaaad777d1 in _PyObject_Call_Prepend ()
   from /lib64/libpython3.6m.so.1.0
#33 0x00002aaaaad77443 in PyObject_Call () from /lib64/libpython3.6m.so.1.0
#34 0x00002aaaaaeb2722 in t_bootstrap () from /lib64/libpython3.6m.so.1.0
#35 0x00002aaaaaeafbd4 in pythread_wrapper () from /lib64/libpython3.6m.so.1.0
#36 0x00002aaaab1fee65 in start_thread () from /lib64/libpthread.so.0
#37 0x00002aaaabc1a88d in clone () from /lib64/libc.so.6
```

kif:
I just installed HDF5 1.8.4 + h5py on centos5 and managed to reproduce the crash
```
#0  0x00002aaabbead6e4 in H5S_hyper_intersect_block () from /scisoft/users/kieffer/workspace/h5py/HDF5-1.8.4/lib/libhdf5.so.6
#1  0x00002aaabbda876c in H5D_chunk_io_init () from /scisoft/users/kieffer/workspace/h5py/HDF5-1.8.4/lib/libhdf5.so.6
#2  0x00002aaabbdb7375 in H5Dread () from /scisoft/users/kieffer/workspace/h5py/HDF5-1.8.4/lib/libhdf5.so.6
#3  0x00002aaabc855538 in __pyx_f_4h5py_4defs_H5Dread (__pyx_v_dset_id=83886082, __pyx_v_mem_type_id=50332062, __pyx_v_mem_space_id=67109194, __pyx_v_file_space_id=67109191, __pyx_v_plist_id=0, 
    __pyx_v_buf=0x14329b0) at /mntdirect/_scisoft/users/kieffer/workspace/h5py/h5py/defs.c:6670
```
@epourmal Do you know about this ? I took the HDF5 1.8.4 provided by the HDFgroup (you).
One cannot guarantee there is no bug in the compiler (gcc-4.1) itself.

I have those threads:
1. waiting for the GIL
5. H5Dread -> H5D_chunk_read -> H5D_chunk_lock ->H5F_block_read -> H5FD_read->H5FD_sec2_read->read () from /lib64/libpthread.so.0
6. H5Dread -> H5D_chunk_io_init
7. H5Dread -> H5D_chunk_io_init
8. H5Dread -> H5D_chunk_io_init -> H5S_hyper_intersect_block

Thomas, shouldn't those calls to H5D be protected by the phil at least ?

kif:
@takluyver: most threads (I did not check all) are accessing H5D via _reader.Reader.read and there is no phil protection on this path. 
I suspect HDF5 1.10 is now shipped thread-save but for 1.8 h5py should have its own lock to avoid this kind of crashed.

takluyver:
In most places, both the high-level API and the low-level API use `phil` - as I understand it, the low-level API has to have it because external code can call that directly, while the high-level API uses it for performance reasons - you don't want it acquiring and releasing the lock repeatedly in a tight loop.

`Reader` is meant to be private API, so I figured I could rely on the high-level API to do the locking, and minimise overhead by not interacting with it from `Reader`. But now I see that the one high-level method that *doesn't* use phil is... `__getitem__`. :disappointed:

takluyver:
That should certainly be fixed, but it can be a separate PR. I'd like to have the added benchmarks working with asv (i.e. allowing repeated setup/teardown) in this PR, though.

takluyver:
I've opened #1456 to sort out the locking. I've been able to run this benchmark with the older HDF5 version after making that change.

takluyver:
Thanks, that seems to be working.

chrisjbillington:
Just want to add my thanks for this! h5py holding the GIL has prevented us from doing performance optimisations by offloading I/O to threads and has made GUIs have poor responsiveness when we do. So I'm very happy to see this. Thanks @kif for the hard work.

PR #1457, opened 2019-12-04
Simplify low-level error handling

takluyver:
This makes several related changes to how errors are handled in the lowest level wrappers around HDF5 functions.

1. Don't use `except *` in the 'extern' declarations of raw HDF5 functions. Removing this avoids Cython checking `PyErr_Occurred()` before our generated error-checking. Consequently, if HDF5 indicates an error, I've added a check for an existing Python error before trying to create a new exception.
2. Use specific error values on the generated wrapper functions, e.g. `except <hid_t>-1` rather than `except *`. This means Cython checks the return value of the function rather than checking `PyErr_Occurred` in all cases. A comment suggests that this was previously not possible, but it works now.
3. Be prepared to handle an error return value with no HDF5 error information in all wrapper functions, rather than only those with a special `ERROR` flag. I believe this has no performance penalty normally, as the branch is only taken if attempting to retrieve HDF5 error information fails.
4. Avoid converting bytes to unicode and back again while processing error messages.

1 & 2 may reduce the overhead of calling HDF5 functions slightly, but I'm mainly trying to make the lowest level bindings simpler and clearer.

### DIFF
--- a/api_gen.py
+++ b/api_gen.py
@@ -107,6 +107,20 @@ class Line(object):
             raise ValueError("Invalid function signature: {0}".format(self.sig))
         self.args = ", ".join(self.args)
 
+        # Figure out what test and return value to use with error reporting
+        if '*' in self.code or self.code in ('H5T_conv_t',):
+            self.err_condition = "==NULL"
+            self.err_value = f"<{self.code}>NULL"
+        elif self.code in ('int', 'herr_t', 'htri_t', 'hid_t', 'hssize_t', 'ssize_t') \
+                or re.match(r'H5[A-Z]+_[a-zA-Z_]+_t', self.code):
+            self.err_condition = "<0"
+            self.err_value = f"<{self.code}>-1"
+        elif self.code in ('unsigned int', 'haddr_t', 'hsize_t', 'size_t'):
+            self.err_condition = "==0"
+            self.err_value = f"<{self.code}>0"
+        else:
+            raise ValueError("Return code <<%s>> unknown" % self.code)
+
 
 raw_preamble = """\
 # cython: language_level=3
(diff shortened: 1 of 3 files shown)


scopatz:
I like this a lot more than what is there now. Kudos @takluyver!

scopatz:
A news entry is probably needed

takluyver:
I'm happy to add a news entry if you want, but I think release notes or "what's new" pages should focus on changes that could affect users - new features, bugs fixed, changes in how you build it, and so on. If this works as it's meant to, it shouldn't make any difference to what h5py does or how to build it.

scopatz:
It seems like it is an API exposure change

takluyver:
I think all of the affected functions are internal - this is the level
below the documented 'low level' API. Maybe I'm missing something, though.

scopatz:
You are probably right.  This is fine without news

takluyver:
@kif, I'd appreciate your input on this if you have time to have a look.

Otherwise, I'll merge it in a couple of days unless anyone objects or wants more time to review it.

kif:
Sorry, I have been on other subjects those days.
I'll have a look.

kif:
About the generated files by api_gen.py:
I am not that familiar with the declaration of exception in cdef function, especially the 
`except <herr_t>-1:`
The `except -1` is heavily described in the documentation of Cython, but I never saw the `<herr_t>-1` ?
Do you have a reference for this ? is it a coming from the pyrex era ? 

Second point, maybe it would make sense to provide the user with this numerical error value in the RuntimeError generated (that's minor).

Beside this it simplifies the code generated which is nice.

takluyver:
I don't have a reference for the `except <herr_t>-1` - I just found it by trial and error. Using `except -1` without the type in there causes Cython to fail with errors like this:

```
Error compiling Cython file:
------------------------------------------------------------
...
cdef herr_t H5Tset_norm(hid_t type_id, H5T_norm_t norm) except -1
cdef H5T_pad_t H5Tget_inpad(hid_t type_id) except -1
cdef herr_t H5Tset_inpad(hid_t type_id, H5T_pad_t inpad) except -1
cdef H5T_cset_t H5Tget_cset(hid_t type_id) except -1
cdef herr_t H5Tset_cset(hid_t type_id, H5T_cset_t cset) except -1
cdef H5T_str_t H5Tget_strpad(hid_t type_id) except -1
                                                   ^
------------------------------------------------------------

h5py/defs.pxd:419:52: Cannot assign type 'long' to 'H5T_str_t'

Error compiling Cython file:
------------------------------------------------------------
...
cdef herr_t H5Tset_norm(hid_t type_id, H5T_norm_t norm) except -1
cdef H5T_pad_t H5Tget_inpad(hid_t type_id) except -1
cdef herr_t H5Tset_inpad(hid_t type_id, H5T_pad_t inpad) except -1
cdef H5T_cset_t H5Tget_cset(hid_t type_id) except -1
cdef herr_t H5Tset_cset(hid_t type_id, H5T_cset_t cset) except -1
cdef H5T_str_t H5Tget_strpad(hid_t type_id) except -1
                                                   ^
------------------------------------------------------------

h5py/defs.pxd:419:52: Exception value incompatible with function return type
```

Whereas repeating the type appears to make it work. Do you think it's worth asking on the Cython issue tracker to check that this is intentional?

> Second point, maybe it would make sense to provide the user with this numerical error value in the RuntimeError generated (that's minor).

It already includes the error condition (e.g. `==NULL` or `<0`). I'm not aware of any HDF5 functions which use different negative numbers to convey different kinds of error, so I'd leave that as is until we have reason to think more detail would be useful.

kif:
My bad, thanks for opening my eyes: this `<herr_t> -1` is just a type conversion of the value -1.
I thought it was a way to catch any negative value (which would have been great), but it is just a  `-1`.

Then it is OK for me.

takluyver:
Thanks!

PR #1458, opened 2019-12-04
Fix lzf filter build with HDF5 1.10.x

t20100:
This PR updates the C macro handling of the `lzf` filter to make it compatible with hdf5 1.10.x.
I tested with hdf5 1.10.5 and 1.8.21 (I haven't tested 1.6 or 1.8.x <1.8.3.... but they should still be supported).

This is one little step towards eventually removing the `H5_USE_16_API` HDF5 1.6 compatibility flag.


scopatz:
This LGTM

Issue #744, opened 2016-08-03
DimensionManager .items() and .values() methods broken


jjhelmus:
The `items` and `values` methods of the DimensionManager class seem to be broken as can be seen in this snippet:

``` Python
import h5py

f = h5py.File('foo.h5', 'w')
f['x1'] = [1, 2]
f['data'] = [1, 2]
f['data'].dims.create_scale(f['x1'])
f['data'].dims[0].attach_scale(f['x1'])

for i in f['data'].dims.items():
    pass

for i in f['data'].dims.values():
    pass
```

```
$ python bug.py
Traceback (most recent call last):
  File "bug.py", line 9, in <module>
    for i in f['data'].dims.items():
  File "/Users/jhelmus/anaconda/lib/python3.5/site-packages/h5py/_hl/base.py", line 338, in __iter__
    yield (key, self._mapping.get(key))
  File "/Users/jhelmus/anaconda/lib/python3.5/_collections_abc.py", line 595, in get
    return self[key]
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper (-------src-dir--------/h5py/_objects.c:2582)
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper (-------src-dir--------/h5py/_objects.c:2541)
  File "/Users/jhelmus/anaconda/lib/python3.5/site-packages/h5py/_hl/dims.py", line 157, in __getitem__
    if index > len(self) - 1:
TypeError: unorderable types: DimensionProxy() > int()
```

The `__getitem__` method of the class seems to be expecting an integer but is being passed a DimensionProxy instance.

jjhelmus:
Changing the __iter__ method of the DimensionsManager class to `yield i` rather than `yield self[i]`would solve this issue.  Would be happy to put together a PR if desired.

kmuehlbauer:
This is still an issue with h5py `2.10.0`, although the ErrorMessage changed slightly:

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-4-2fe68ce88f85> in <module>
      7 f['data'].dims[0].attach_scale(f['x1'])
      8 
----> 9 for i in f['data'].dims.items():
     10     pass
     11 

/home/kai/miniconda/envs/wradlib_38/lib/python3.8/site-packages/h5py/_hl/base.py in __iter__(self)
    366         with phil:
    367             for key in self._mapping:
[...]

PR #1464, opened 2019-12-11
FIX: remove `base.MappingHDF5` from `dims.DimensionManager`

kmuehlbauer:
- remove `base.MappingHDF5` from `dims.DimensionManager` to not add unneeded Mapping properties
- resolves https://github.com/h5py/h5py/issues/744

Update: Description changed and force-pushed

### DIFF
--- a/h5py/_hl/dims.py
+++ b/h5py/_hl/dims.py
@@ -136,7 +136,7 @@ class DimensionProxy(base.CommonStateObject):
                % (self.label, self._dimension, id(self._id)))
 
 
-class DimensionManager(base.MappingHDF5, base.CommonStateObject):
+class DimensionManager(base.CommonStateObject):
 
     """
         Represents a collection of dimension associated with a dataset.
--- a/news/dims-collection.rst
+++ b/news/dims-collection.rst
@@ -1,0 +1,29 @@
+New features
+------------
+
+* <news item>
+
+Deprecations
+------------
+
+* <news item>
+
+Exposing HDF5 functions
+-----------------------
+
+* <news item>
+
+Bug fixes
+---------
+
+* Remove `base.MappingHDF5` from `dims.DimensionManager` class definition to handle as collection only, resolves https://github.com/h5py/h5py/issues/744.
+
+Building h5py
+-------------
+
+* <news item>
+
+Development
+-----------
+
+* <news item>


scopatz:
This seems like an API break

kmuehlbauer:
@scopatz Why?

kmuehlbauer:
Maybe a bit more context:
`DimensionManager` inherits from `base.MappingHDF5`. So `key`, `value` and `item` are exposed to the public API, but are broken by design. This PR fixes this. If `DimensionManager` is thought to work as sequence, then it should not inherit from Mapping but from Sequence, or do I miss something (which might be the case)?

takluyver:
Presumably this is an API break in the sense that code doing `for dim in ds.dims` would need to change?

My feeling is that the `ds.dims` interface is more like a sequence - the 'keys' are always a consecutive group of integers from 0 to N-1, if I'm reading it right. It sounds like most mapping methods are broken at the moment, so maybe we should just make it a mapping? I guess that would break `__contains__` (`i in ds.dims`), but it seems unlikely much code is using that. Do you see anything else it will break? How do equality checks work? I'm looking at the list of Mapping methods in https://docs.python.org/3/library/collections.abc.html#module-collections.abc

kmuehlbauer:
@scopatz Maybe this inheritance from `base.MappingHDF5` isn't needed at all (by looking at the code).

I tried to follow the code in git commit history. It was copied over from `AttributeManager` but the inheritance wasn't ever changed. The `MappingHDF5` properties (keys, values, items) are never used, weren't even indented to be used (citing the docstring: "...Represents a collection of dimensions..."). Also the documentation has no sign of using keys, values, items, but shows that iteration is sequence like.

kmuehlbauer:
@takluyver Did see your comment just after I sent mine. I removed the `MappingHDF5` inheritance and everything seems to run smoothly (tox-ing). Should I just update this PR?

takluyver:
I'd be inclined to do that. Don't forget to add a [news entry](https://regro.github.io/rever-docs/news.html) as well.

I'll leave it a day or two in case anyone else chimes in with an argument in favour of keeping and fixing the mapping interface.

tacaswell:
That the output of `__iter__` can not got into the input to `__getitem__` is a pretty strong indication that this was not intended to be a maping.

PR #1466, opened 2019-12-17
Use versionadded directive for new chunk query functions

### LINKED ISSUE #1449: h5py.h5d.DatasetID.get_num_chunks missing in 2.10.0 release
The method is documented here http://api.h5py.org/h5d.html#h5py.h5d.DatasetID.

The method can be found in the source code of Git repository.

The method is missing in source code published on Pypi.

takluyver:
Closes #1449

Also moved the listing of the HDF5 functions into the correct section (H5F -> H5D).

This change doesn't need a news entry.


takluyver:
As this is a minor change, I'll merge it in a couple of days if there are no objections.

scopatz:
This looks great! Thanks @takluyver

PR #1468, opened 2019-12-18
Support split files

vbrankov:
Adds support for split HDF5 files.

I revived this old pull request: https://github.com/h5py/h5py/pull/718

### DIFF
--- a/h5py/h5p.pyx
+++ b/h5py/h5p.pyx
@@ -114,7 +114,10 @@ OBJECT_CREATE = lockcls(H5P_OBJECT_CREATE)
 CRT_ORDER_TRACKED = H5P_CRT_ORDER_TRACKED
 CRT_ORDER_INDEXED = H5P_CRT_ORDER_INDEXED
 
-DEFAULT = H5P_DEFAULT
+DEFAULT = None   # In the HDF5 header files this is actually 0, which is an
+                 # invalid identifier.  The new strategy for default options
+                 # is to make them all None, to better match the Python style
+                 # for keyword arguments.
 
 
 # === Property list functional API ============================================
@@ -1078,7 +1081,6 @@ cdef class PropFAID(PropInstanceID):
         - h5fd.MULTI
         - h5fd.SEC2
         - h5fd.STDIO
-        - h5fd.SPLIT
         """
         return H5Pget_driver(self.id)
 
--- a/h5py/h5p.pyx
+++ b/h5py/h5p.pyx
@@ -1038,12 +1038,12 @@ cdef class PropFAID(PropInstanceID):
         H5Pset_fapl_stdio(self.id)
 
     @with_phil
-    def set_fapl_split(self, const char* meta_ext, hid_t meta_plist_id, const char*raw_ext, hid_t raw_plist_id):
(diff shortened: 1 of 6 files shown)


kif:
Interesting ...

Could you please add a small test-case to validate it is working as expected ?

h5py/h5p.pyx
```diff
@@ -114,10 +114,7 @@ OBJECT_CREATE = lockcls(H5P_OBJECT_CREATE)
 CRT_ORDER_TRACKED = H5P_CRT_ORDER_TRACKED
 CRT_ORDER_INDEXED = H5P_CRT_ORDER_INDEXED
 
-DEFAULT = None   # In the HDF5 header files this is actually 0, which is an
-                 # invalid identifier.  The new strategy for default options
-                 # is to make them all None, to better match the Python style
-                 # for keyword arguments.
+DEFAULT = H5P_DEFAULT
```
takluyver: I don't think we need to change this for this PR - look at `set_fapl_family` for an example of how to accept a property list with a default.

If you want to change it anyway, that should be a separate PR.

vbrankov:
@kif [h5py/tests/test_file.py](https://github.com/h5py/h5py/pull/1468/files#diff-be0da2120f746776b9dcfa86d7338f51) contains a test.  Please let me know if I should make it more extensive.

h5py/h5p.pyx
```diff
@@ -1078,7 +1081,6 @@ cdef class PropFAID(PropInstanceID):
         - h5fd.MULTI
         - h5fd.SEC2
         - h5fd.STDIO
-        - h5fd.SPLIT
```
tacaswell: I think we want to keep this line?

tacaswell:
This look good to me.  I left one comment about a documentation update that I think got lost in the shuffle.

You can use `File` as a context manager which saves you from having to explicitly call close in the tests, but I don't think that is worth holding the PR over.

takluyver:
This should also be mentioned in the file drivers section of the [files high level documentation](https://github.com/h5py/h5py/blob/master/docs/high/file.rst).

scopatz:
LGTM too

h5py/h5p.pyx
vbrankov: This mode doesn't exist.  Split is implemented using the MULTI mode.

vbrankov: At one point I introduced the mode, but then I realized it doesn't really exist so I removed the comment.

vbrankov:
I take it back regarding the h5fd.SPLIT.  Looks like I'm wrong.  Let me see for a moment.

takluyver:
I think you might be right, actually. [The docs](https://confluence.hdfgroup.org/display/HDF5/Virtual+File+Drivers) suggest that H5FD_SPLIT exists, but I can't find it in the HDF5 source code - `H5Pset_fapl_split` calls `H5Pset_fapl_multi`.

So I think we just need a mention of the 'split' driver in the high-level file docs, and this is OK.

vbrankov:
@takluyver I just came to the same conclusion.

I added some documentation for the 'split' high level driver.

takluyver:
Thanks @vbrankov !

PR #1472, opened 2020-01-11
Fix doc markup for track_order parameter

takluyver:
I noticed that this was slightly wrong in the docs.


takluyver:
I'll merge this in a day if no-one objects, as it's a fairly trivial change.

PR #1473, opened 2020-01-13
Test intermediate groups creation - closes #1432

### LINKED ISSUE #1432: Check correct behaviour of require_dataset
Following the discussion in #1431, it would be useful to check that [require_dataset](https://github.com/h5py/h5py/blob/7073098eed26904f2077bac98ee450894b99468e/h5py/_hl/group.py#L198) is working as expected.

I will write some unit tests to check that intermediate groups are correctly created.

pierlauro:

### DIFF
--- a/h5py/tests/test_group.py
+++ b/h5py/tests/test_group.py
@@ -1038,3 +1038,30 @@ class TestMutableMapping(BaseGroup):
         Group.__delitem__
         Group.__iter__
         Group.__len__
+
+class TestRequire(TestCase):
+    """
+    Test that intermediate groups are automatically created
+    if present in path of a new dataset/group
+    """
+
+    def setUp(self):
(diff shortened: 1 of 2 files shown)


h5py/tests/test_group.py
```diff
@@ -1038,3 +1038,30 @@ def test_validity(self):
         Group.__delitem__
         Group.__iter__
         Group.__len__
+
+class TestRequire(TestCase):
```
takluyver: There's already a `TestRequire` class in the file above. Reusing the name means that doesn't run, which is why we're seeing a coverage drop. Can you add the test methods to that?

takluyver:
LGTM. I'll leave it another day in case anyone else wants to have a look, otherwise I'll merge it.

PR #1481, opened 2020-01-23
Fix object location for intersphinx

flying-sheep:
Currently, people have to use ``:class:`Dataset` `` to link to `h5py.Dataset`, which isn’t the way it should be; it should read ``:class:`h5py.Dataset` `` (or ``:class:`~h5py.Dataset` `` if you want it to *read* as plain “Dataset”). If everything would link their stuff unnamespaced, there would be clashes all the time!

This fixes that, all attributes, classes and functions are now properly namespaced in `objects.inv`.

### DIFF
--- a/docs/high/attr.rst
+++ b/docs/high/attr.rst
@@ -1,3 +1,4 @@
+.. currentmodule:: h5py
 .. _attributes:
 
 
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -1,3 +1,4 @@
+.. currentmodule:: h5py
 .. _dataset:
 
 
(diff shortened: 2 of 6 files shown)


takluyver:
Thanks @flying-sheep

flying-sheep:
No problem! Also you’re everywhere :smile:

takluyver:
I could say the same about you! :wink: I contributed to virtual dataset support in h5py, and got invited to be a co-maintainer of it.

flying-sheep:
Haha yeah, I think you share my love of just fixing and improving all the parts of the open source ecosystem that you ever touched.

takluyver:
There's definitely a degree of truth to that, although in this case the h5py work was directly enabling something that we wanted to do at work.

flying-sheep:
That counts as improving :wink:

PR #1482, opened 2020-01-23
Fixes for empty region references

takluyver:
- When a region reference points to an empty dataset, or an empty selection on a scalar dataset, return the appropriate `Empty` instance rather than a 1D, length 1 array.
- When a region reference points to an empty selection on a normal (simple) dataset, return an empty array rather than an array with 1 value per dimension (!).
  - Ideally this would also get the right shape, but it always gives (0, 0). I don't know if this is something we can fix or a limitation of HDF5.

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -656,7 +656,8 @@ class Dataset(HLObject):
             sid = h5r.get_region(args[0], self.id)
             mshape = sel.guess_shape(sid)
             if mshape is None:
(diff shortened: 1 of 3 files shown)


scopatz:
This LGTM!

takluyver:
Thanks @scopatz! For this and the other PRs you just LGTMed, I'll leave them a couple of days to give other people a chance to review, and then merge them if there are no more comments.

Issue #135, opened 2012-12-27
reading of numpy structure scalar gives tuple


andrewcollette:
_Original author: jlewi.in...@gmail.com (October 22, 2010 16:14:17)_

OS: Fedora 10
hdf5: 1.8.4
h5py: 1.3.0
numpy: 1.5.1rc1

I can write a numpy structure scalar but when I read it, I get a tuple.
If its an array it works fine. 
below is some code to reproduce the problem

```python
import numpy as np
dt=np.dtype([("i1",np.int),("i2",np.int)])
# make it a scalar

data=np.empty(1,dtype=dt)
data=data[0]
data["i1"]=23
data["i2"]=45

fpath="/tmp/test"
f=h5py.File(fpath,"r+")

f.create_dataset("/data", data=data)
rdata=f["data"].value
# data is  a tuple not a structure array scalar
```

My config is:

```
# Summary of h5py config

HDF5: 1.8.4
1.6 API: True
1.8 API: True
Diagnostic mode: False
Complex names: ('r', 'i')
```

_Original issue: http://code.google.com/p/h5py/issues/detail?id=135_

PR #1484, opened 2020-01-24
Remove special handling for reading a single value

takluyver:
This special case was added in response to issue #135. As far as I can tell, it made the `if arr.shape == ():` branch near the end of the method unreachable, because it forced the destination array to have at least 1 dimension.

`arr.item()` copies a single element from an array into a standard Python type where possible, causing the behaviour seen in #135 (a numpy struct converted to a tuple). In contrast, `arr[()]` extracts a numpy scalar from a 0D array, which is what we want.

There are already several tests that fail if I back out the special-case but leave `arr.item()` as is, e.g. `TestSingleElement.test_compound` in `test_slicing.py`.

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -695,11 +695,7 @@ class Dataset(HLObject):
         if selection.nselect == 0:
             return numpy.ndarray(selection.array_shape, dtype=new_dtype)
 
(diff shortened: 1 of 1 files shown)


scopatz:
I think that you are right that this was not being hit, and so LGTM

takluyver:
Thanks! I'll merge this in a couple of days unless anyone requests changes or more time to review it.

PR #1485, opened 2020-01-24
Add simple tests for inspecting virtual datasets

takluyver:
While looking at the codecov information, I noticed that `ds.is_virtual` and `ds.virtual_sources()` were not exercised by tests. This adds a simple test to address that.


scopatz:
Thanks @takluyver! Since this is a test-only change, I am going to go ahead and merge.

PR #1487, opened 2020-02-01
Enable fancy indexing in Cython reader code

takluyver:
This is another step towards replacing some of the Python selection code. This implements selection with a list or array of indices in the Cython reader pathway.

There shouldn't be much overhead for simple cases: we only check for array-like arguments after checking the options for simple indexing (integers, slices, ellipsis). So when you use those, the only new code that runs is a simple if to check if an array was found.

This also partly addresses #993 (partly because it only affects reads which can use this code).

### DIFF
--- a/h5py/_reader.pyx
+++ b/h5py/_reader.pyx
@@ -134,7 +134,10 @@ cdef class Reader:
                 if array_ix != -1:
                     raise TypeError("Only one indexing vector or array is currently allowed for fancy indexing")
 
+                # Convert negative indices to positive
                 a[a < 0] += l
+
+                # Bounds check
                 if np.any((a < 0) | (a > l)):
                     if l == 0:
                         msg = "Fancy indexing out of range for empty dimension"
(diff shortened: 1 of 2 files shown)


h5py/_reader.pyx
```diff
@@ -113,6 +123,35 @@ cdef class Reader:
...
+                    raise TypeError("Indexing arrays must have integer dtypes")
+                if array_ix != -1:
+                    raise TypeError("Only one indexing vector or array is currently allowed for fancy indexing")
+
+                a[a < 0] += l
+                if np.any((a < 0) | (a > l)):
+                    if l == 0:
+                        msg = "Fancy indexing out of range for empty dimension"
+                    else:
+                        msg = f"Fancy indexing our of range for (0-{l-1})"
```
scopatz: I thought you could fancy index with a negative index.Is this an optimization of some kind?

takluyver: Line 137 should be shifting negative indexes to positive ones - I'll add a comment to make that clear.

scopatz: Ahh thanks!

takluyver:
Close/reopen to rerun tests with master.

takluyver:
The tests have passed on this now; does anyone want to do further review?

scopatz:
LGTM again

scopatz:
Feel free to merge

takluyver:
Thanks Anthony  :-)

Issue #1492, opened 2020-02-07
Test failures with HDF5 1.10.6


takluyver:
Two tests recently started failing in the CI. It looks like this coincides with those builds being switched from HDF5 1.10.5 to 1.10.6. It's entirely possible that this change will affect other platforms; I think the Mac builds are the first to get the new HDF5 version, as the others either build a specified HDF5 version, or get it from Ubuntu repos.

```
_______________________ TestRegionRefs.test_empty_region _______________________

self = <h5py.tests.test_dataset.TestRegionRefs testMethod=test_empty_region>

    def test_empty_region(self):
        ref = self.dset.regionref[:0]
        out = self.dset[ref]
>       assert out.size == 0
E       AssertionError: assert 10000 == 0
E        +  where 10000 = array([[   0,    1,    2, ...,   97,   98,   99],\n       [ 100,  101,  102, ...,  197,  198,  199],\n       [ 200,  201...9797, 9798, 9799],\n       [9800, 9801, 9802, ..., 9897, 9898, 9899],\n       [9900, 9901, 9902, ..., 9997, 9998, 9999]]).size

py37-test-deps/lib/python3.7/site-packages/h5py/tests/test_dataset.py:1174: AssertionError
______________________ TestRegionRefs.test_scalar_dataset ______________________

self = <h5py.tests.test_dataset.TestRegionRefs testMethod=test_scalar_dataset>

    def test_scalar_dataset(self):
        ds = self.f.create_dataset("scalar", data=1.0, dtype='f4')
        sid = h5py.h5s.create(h5py.h5s.SCALAR)
    
        # Deselected
        sid.select_none()
        ref = h5py.h5r.create(ds.id, b'.', h5py.h5r.DATASET_REGION, sid)
>       assert ds[ref] == h5py.Empty(np.dtype('f4'))
E       AssertionError: assert array(1., dtype=float32) == Empty(dtype=dtype('float32'))
E        +  where Empty(dtype=dtype('float32')) = <class 'h5py._hl.base.Empty'>(dtype('float32'))
E        +    where <class 'h5py._hl.base.Empty'> = h5py.Empty
E        +    and   dtype('float32') = <class 'numpy.dtype'>('f4')
E        +      where <class 'numpy.dtype'> = np.dtype

py37-test-deps/lib/python3.7/site-packages/h5py/tests/test_dataset.py:1184: AssertionError
```

takluyver:
[...]

PR #1497, opened 2020-02-21
Mark region ref tests as expected failures with HDF5 1.10.6

takluyver:
This suppresses the test failures described in #1492.

So far, I can't get a response from HDF group about whether this is a bug, but I assume it must be: I can't see any mention of it in the release notes, and empty region references are something that worked previously.

I've assumed for now that it will only occur in HDF5 1.10.6, but that may need to be revised in the future.


scopatz:
This LGTM for the moment!

takluyver:
Thanks @scopatz

Issue #1480, opened 2020-01-23
Rationalise access to fields of structured data


takluyver:
At present, h5py allows field names and indexing to be mixed up arbitrarily in an indexing expression, although this is not legal in numpy:

```
In [43]: ds['y', 0, 'x']                                                        
Out[43]: (81., 9.)

In [44]: arr['y', 0, 'x']                                                       
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-44-fbe26c74f579> in <module>
----> 1 arr['y', 0, 'x']

IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices
```

This seems like there's significant potential for confusion. numpy indexing requires that you select either fields (string or list of strings) or indexes (with all the normal indexing functionality.

If I was designing this from scratch, I would either match numpy's interface, or make field selection explicit through a property or method (e.g. `ds.fields('y', 'x')[0]`). The latter option in particular would simplify the code, because `__getitem__` wouldn't need to handle strings.

However, I don't know if it's realistic to make a change like this to an established API, even in a major release. We could introduce a new interface and keep the old one (perhaps behind a config option), but the 3.0 release seems like the best chance to make backwards incompatible changes like this. Thoughts?

takluyver:
For reference, numpy allows indexing with a single name `arr['x']` or a list of names `arr[['x', 'y']]`. It doesn't allow multiple names outside a list (no `arr['x', 'y']`), nor any mixture of names and indices in (no `arr['x', 0]`) - you have to do field access and indexing/slicing separately. I can't find anything like the `.fields()` method #1508 would add - the only way to select fields seems to be with indexing.

I'd be in favour of restricting field access through h5py to match numpy - only allow a single field name or a list of field names - and making `ds['x']` give a wrapper object, so you need another `[:]` to read all the data. But I suspect this is not a reasonable change to make without a deprecation cycle, even for 3.0.

mikpom:
Greetings!

Currently there is a notice that mixed usege of index and field names is discouraged. However how setting a certain field in some range is possible then?

I mean 

```
ds = file.create_dataset('data', (100,), dtype=[('F1', h5py.string_dtype('ascii')), ('F2', np.int_)])
ds[0:10, 'F2'] = 100

```
How can I set first 10 entries of field F2 to 100 without using mixed indexing? Alternatively I will set on a copy and nothing will change in the dataset.

[...]

PR #1508, opened 2020-03-08
Rethink field access in compound datatype as a dataset.fields() method

takluyver:
This is my idea to address #1480:

```python
# Old way - mix field names and indexing:
x = f['coords']['x', :10]

# New way:
x = f['coords'].fields('x')[:10]
```

The old way continues working with these changes. I'd favour deprecating and/or restricting it, because at present it allows names and indexing to be mixed arbitrarily (`ds['y', 0, 'x']`), which seems confusing.

The new API allows for one thing that's not possible with the current API: extracting a single field, but keeping it as a compound dtype (`.fields(['x'])`). I can't immediately see a case where this is useful, but it's possible in numpy - `arr[['x']]` is not the same as `arr['x']`.

### DIFF
--- a/h5py/_hl/attrs.py
+++ b/h5py/_hl/attrs.py
@@ -20,7 +20,6 @@ import uuid
 from .. import h5, h5s, h5t, h5a, h5p
 from . import base
 from .base import phil, with_phil, Empty, is_empty_dataspace
-from .dataset import readtime_dtype
 from .datatype import Datatype
 
 
@@ -59,7 +58,7 @@ class AttributeManager(base.MutableMappingHDF5, base.CommonStateObject):
         if is_empty_dataspace(attr):
             return Empty(attr.dtype)
 
-        dtype = readtime_dtype(attr.dtype, [])
+        dtype = attr.dtype
         shape = attr.shape
 
         # Do this first, as we'll be fiddling with the dtype for top-level
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -32,22 +32,6 @@ _LEGACY_GZIP_COMPRESSION_VALS = frozenset(range(10))
 MPI = h5.get_config().mpi
 
 
-def readtime_dtype(basetype, names):
-    """ Make a NumPy dtype appropriate for reading """
-
-    if len(names) == 0:  # Not compound, or we want all fields
-        return basetype
-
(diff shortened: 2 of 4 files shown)


h5py/_hl/attrs.py
```diff
@@ -59,7 +58,7 @@ def __getitem__(self, name):
         if is_empty_dataspace(attr):
             return Empty(attr.dtype)
 
-        dtype = readtime_dtype(attr.dtype, [])
```
takluyver: This presumably did something at some time in the past, but in master at present, `readtime_dtype` is a no-op if `len(names) == 0`, so we can skip it here.

h5py/_hl/dataset.py
```diff
@@ -602,6 +621,14 @@ def iter_chunks(self, sel=None):
...
         return ChunkIterator(self, sel)
 
+    @cached_property
+    def _fast_read_ok(self):
+        """Is this dataset suitable for simple reading"""
+        return (
+            self._extent_type == h5s.SIMPLE
+            and isinstance(self.id.get_type(), (h5t.TypeIntegerID, h5t.TypeFloatID))
+        )
+
```
takluyver: This is unchanged, I just moved it to better group related methods together.

h5py/_hl/dataset.py
```diff
@@ -682,8 +711,6 @@ def __getitem__(self, args, new_dtype=None):
             arr = numpy.ndarray(selection.mshape, dtype=new_dtype)
             for mspace, fspace in selection:
                 self.id.read(mspace, fspace, arr, mtype)
-            if len(names) == 1:
-                arr = arr[names[0]]
```
takluyver: This (and the same thing below) are now handled by `FieldsWrapper`.

scopatz:
LGTM

takluyver:
Thanks @scopatz ! As this adds a new method, I'll leave it a few days for other people to have a look.

Any thoughts on deprecating or restricting field names in `ds[...]`? Let's have that discussion on #1480 so it's not interrupted when this is merged.

takluyver:
@tacaswell @aragilar @aparamon @kif - I'll merge this in a couple of days unless anyone objects or wants more time to examine it. It adds a new `.fields()` method for datasets with compound datatypes.

PR #1513, opened 2020-03-15
Misc. test cleanups

takluyver:
A couple of miscellaneous test cleanups I spotted a need for.

### DIFF
--- a/h5py/tests/test_group.py
+++ b/h5py/tests/test_group.py
@@ -51,19 +51,6 @@ class BaseGroup(TestCase):
         if self.f:
             self.f.close()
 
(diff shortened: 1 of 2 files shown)


h5py/tests/test_group.py
```diff
@@ -51,19 +51,6 @@ def tearDown(self):
         if self.f:
             self.f.close()
 
-class TestRepr(BaseGroup):
```
takluyver: This class was replaced by another `TestRepr` below in the same file, so it wasn't getting run.

scopatz:
LGTM! Thanks @takluyver

PR #1515, opened 2020-03-15
ENH: use char* not array.array, clean up some warnings

### LINKED ISSUE #1514: ENH: Don't use non-portable array.array in cython
It turns out there is no clean C interface to the `array.array` object in Python. Cython reproduces the CPython internal `arrayobject` in its [`array.pxd`](https://github.com/cython/cython/blob/master/Cython/Includes/cpython/array.pxd#L2) code, but this does not work on other implementations (like PyPy). Cython's recommendation in this case, if one wishes to be portable, is to [avoid its use](https://github.com/cython/cython/issues/1818). In general, [importing from CPython](https://github.com/h5py/h5py/blob/master/h5py/h5d.pyx#L21) will cause protability problems.

Would you be open to a PR to change the [use of `array.array`](https://github.com/h5py/h5py/blob/master/h5py/h5d.pyx#L481) to a `bytearray`?

mattip:
closes #1514 

Replace `array.array` with a `char*` and cleanup some NumPy warnings

### DIFF
--- a/h5py/tests/test_dataset.py
+++ b/h5py/tests/test_dataset.py
@@ -1247,7 +1247,7 @@ class TestVlen(BaseDataset):
         self.assertArrayEqual(ds[1], np.arange(0))
         self.assertArrayEqual(ds[2], np.array([1, 2, 3]))
         self.assertArrayEqual(ds[1], np.arange(0))
-        ds[0:2] = np.array([np.arange(5), np.arange(4)])
+        ds[0:2] = np.array([np.arange(5), np.arange(4)], dtype=object)
         self.assertArrayEqual(ds[0], np.arange(5))
         self.assertArrayEqual(ds[1], np.arange(4))
         ds[0:2] = np.array([np.arange(3), np.arange(3)])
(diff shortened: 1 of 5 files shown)


h5py/tests/test_dataset.py
```diff
@@ -1247,7 +1247,7 @@ def test_int(self):
         self.assertArrayEqual(ds[1], np.arange(0))
         self.assertArrayEqual(ds[2], np.array([1, 2, 3]))
         self.assertArrayEqual(ds[1], np.arange(0))
-        ds[0:2] = np.array([np.arange(5), np.arange(4)])
+        ds[0:2] = np.array([np.arange(5), np.arange(4)], dtype=object)
```
mattip: These "ragged arrays" will emit a warning for the upcoming NumPy 1.19 release since they cannot be turned into arrays of numerics.

setup_build.py
```diff
@@ -43,7 +43,9 @@ def localpath(*args):
    'libraries'      : ['hdf5', 'hdf5_hl'],
    'include_dirs'   : [localpath('lzf')],
    'library_dirs'   : [],
-   'define_macros'  : [('H5_USE_16_API', None)]
+   'define_macros'  : [('H5_USE_16_API', None),
+                       ('NPY_NO_DEPRECATED_API', 0),
+                      ]
```
mattip: Theoretically this should be `NPY_1_7_API_VERSION` not `0` to take advantage of the "new" (as of 2010) NumPy interfaces, but that is the subject of a different PR. At least this way the annoying warning will not be printed. NumPy has no intention of removing the old API version interfaces, so this is harmless.

mattip:
Not sure how codecov decided this PR reduced coverage in `h5py/_hl/filters.py`

scopatz:
Codecov never really works IMO. :wink:

takluyver:
Thanks, this LGTM.

I've had codecov work beautifully on smaller projects. My best guess is that with multiple CI services running, it doesn't correctly judge when they're all done, so it sometimes cuts off before some results are submitted. Or maybe some jobs fail to upload at all. :-/

Issue #1519, opened 2020-03-21
write_direct_chunk only takes byte data type


weninc:
The write_direct_chunk function only takes bytes as a data type, leading to unnecessary conversions and memory copies from other common types like ndarray.   
To make the function more flexible it should allow any python object that exposes the Py_buffer interface

takluyver:
Fixed by #1520

PR #1520, opened 2020-03-21
use buffer interface in write_direct_chunk

weninc:
#1519

### DIFF
--- a/h5py/h5d.pyx
+++ b/h5py/h5d.pyx
@@ -26,6 +26,10 @@ from .h5p cimport PropID, propwrap
 from ._proxy cimport dset_rw
 
 from ._objects import phil, with_phil
+from cpython cimport PyObject_GetBuffer, \
+                     PyBUF_ANY_CONTIGUOUS, \
+                     PyBuffer_Release
+
 
 # Initialization
 import_array()
@@ -405,7 +409,7 @@ cdef class DatasetID(ObjectID):
 
     IF HDF5_VERSION >= (1, 8, 11):
 
-        def write_direct_chunk(self, offsets, bytes data, filter_mask=0x00000000, PropID dxpl=None):
+        def write_direct_chunk(self, offsets, data, filter_mask=0x00000000, PropID dxpl=None):
             """ (offsets, bytes data, uint32_t filter_mask=0x00000000, PropID dxpl=None)
 
             This function bypasses any filters HDF5 would normally apply to
@@ -428,6 +432,7 @@ cdef class DatasetID(ObjectID):
             cdef hsize_t *offset = NULL
             cdef size_t data_size
             cdef int rank
+            cdef Py_buffer view
 
             dset_id = self.id
             dxpl_id = pdefault(dxpl)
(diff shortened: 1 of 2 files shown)


h5py/h5d.pyx
```diff
@@ -405,7 +409,7 @@ cdef class DatasetID(ObjectID):
 
     IF HDF5_VERSION >= (1, 8, 11):
 
-        def write_direct_chunk(self, offsets, bytes data, filter_mask=0x00000000, PropID dxpl=None):
+        def write_direct_chunk(self, offsets, data, filter_mask=0x00000000, PropID dxpl=None):
```
scopatz: we probably shouldn't change the API  like this

weninc: But this only relaxes the type constraint and it doesn't change how the function is called. So it should be transparent to anyone using the API and calling it with bytes still works without any change. It just allows other types as well

takluyver:
I actually think that allowing numpy arrays directly here is a *bad* idea - if you can pass an ndarray, some people will assume that it takes account of the metadata, e.g. doing automatic conversions based on the dtype, or using the array shape somehow, and then get confused when the 'wrong' data is stored.

However, you can get a memoryview of an array's data with `arr.data`, without copying it. So maybe a suitable middle-ground is to allow bytes or memoryviews?

weninc:
Its a very low level function, so I don't think people would expect the numpy metadata would be accounted for. You have to be very careful with this function anyway especially with compressed data. We are trying to use this to store large amounts of precompressed data from large image detectors and we can't afford copying our data into bytes objects.
Memoryviews of arr.data would be fine for us, but I tried them and can't get it to work:
```
dset.id.write_direct_chunk((i, 0, 0), data.data)
TypeError: Argument 'data' has incorrect type (expected bytes, got memoryview)
```

takluyver:
Fair enough, I guess the buffer interface is the obvious way to do this. But I'd like it to be noted in the docstring that passing an array just saves the raw data from the array, ignoring the shape and dtype completely. The docstring is also used in the low-level API docs at https://api.h5py.org/ .

The CI is also asking you to add yourself to the `.authors.yml` file.

h5py/h5d.pyx
scopatz: But it removes a positional argument, which causes all downstream calls of this function to break :cry:

takluyver: @scopatz I think you're misreading the change - it's not removing an argument, just removing the `bytes` type from the `data` argument (a bit of C like Cython syntax).

weninc:
Great thanks. I updated the docstring to reflect the ndarry metadata issue and added myself to the .authors.yml file. Let me know if you need anything else.

h5py/h5d.pyx
scopatz: Oh! I am! Sorry.  Serves me right for trying to reveiw on my phone

h5py/h5d.pyx
```diff
@@ -405,13 +409,18 @@ cdef class DatasetID(ObjectID):
...
+            """ (offsets, data, uint32_t filter_mask=0x00000000, PropID dxpl=None)
 
             This function bypasses any filters HDF5 would normally apply to
             written data. However, calling code may apply filters (e.g. gzip
             compression) itself before writing the data.
 
+            `data` is a Python object that implements the Py_buffer interface.
+            In case of a ndarray the shape and dtype are ignored. It's the
+            users responsibility to make sure they are compatible with the
```
takluyver:
```suggestion
            user's responsibility to make sure they are compatible with the
```

Sorry to nitpick :wink:

takluyver:
Thanks! Other than that tiny change, LGTM. I'll give it a day or two in case anyone else wants to review it, then I'll merge it if not.

h5py/h5d.pyx
weninc: oh yes of course :sweat_smile:

kif:
Hi Clemens,

I appreciate very much your contribution as this is something we are also struggling with.

The H5DOwrite_chunk is called in a *nogil* environment, hence the  Py_buffer.len and Py_buffer.buf should be Python free. According to the doc, it is the case.

+1 for merging this PR.

PR #1521, opened 2020-03-22
Release the GIL in H5Idec_ref

### LINKED ISSUE #1516: Performance issue in h5py in multi-threaded environment
In the scope of the development of BLISS (https://bliss.gitlab-pages.esrf.fr/bliss/master/) at ESRF
we found a performance issue related to h5py in a multi-threaded server environment (server unresponsive for seconds).

Basically the thread is blocking when closing the file, GIL not released. After a quick profiling it goes down to h5py._hl.files (line 400 - 422) and most of the time is spent in H5I_dec_ref which does not release the GIL.
We tested to release the GIL there and test passes without crash. I wonder if you know any callback which might be called at the file-driver level which may be implemented in Python and cause GIL-related crashes ?
If there ain't any, would you consider a PR which declares the function from H5I as GIL free ?
Cheer,

Jerome

 * Operating System Linux, various flavours
 * Python version 3.7 ananconda, or system
 * h5py version: 2.10, master 
 * HDF5 version: 1.10.4, 1.10.5 (1.10.6?)

kif:
The *dec_ref* function can be taking ages when closing large datasets. This releases the **GIL** to  ensure the system remains responsive. I checked in the base-code, those functions are not used that often and checked that the modification should not trigger python based call-back functions.

close #1516
PS: The  news folder will be upgraded after the CI passes.

### DIFF
--- a/h5py/api_functions.txt
+++ b/h5py/api_functions.txt
@@ -208,8 +208,8 @@ hdf5:
   ssize_t    H5Iget_name( hid_t obj_id, char *name, size_t size)
   hid_t      H5Iget_file_id(hid_t obj_id)
   int        H5Idec_ref(hid_t obj_id) nogil
-  int        H5Iget_ref(hid_t obj_id)
-  int        H5Iinc_ref(hid_t obj_id)
+  int        H5Iget_ref(hid_t obj_id) nogil
+  int        H5Iinc_ref(hid_t obj_id) nogil
   htri_t     H5Iis_valid( hid_t obj_id )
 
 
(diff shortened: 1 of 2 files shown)


takluyver:
Close/reopen to rerun tests with changes in master.

takluyver:
I suspect this only really makes sense for `dec_ref`, because the refcount reaching 0 triggers closing things, and probably flushing data to disk if necessary. But I don't think the other two are used anywhere particularly performance critical, so I'm happy with it as is.

takluyver:
On further reflection, I think I would prefer that we are selective about releasing the GIL where it makes sense, and of these, I think it only makes sense for `dec_ref`.

takluyver:
@kif I've made this specific to `dec_ref` now, as I'm conscious of the advice that releasing the GIL doesn't make sense for trivial functions, and I think `inc_ref` and `get_ref` should be trivial.

kif:
As you want. declaring dec_ref as nogil will fix the issue we have here.

takluyver:
Great, thanks. Merging!

PR #1506, opened 2020-03-06
Try to add HDF5 1.12.0 to Azure pipelines

takluyver:
For #1504.

PR #1522, opened 2020-03-22
Pin HDF5 1.10.6 on Mac CI jobs

takluyver:
h5py is not working with HDF5 1.12.0 yet (see #1506). This will hopefully let the CI keep working until it is.


takluyver:
Merging this myself so that we can have passing tests again.

PR #1524, opened 2020-04-01
MNT: account for changes to cython

tacaswell:
In Cython 3.0 the default binding behavior changes from False to True (see https://github.com/cython/cython/pull/2864)

This explicitly marks several functions to not bind.


scopatz:
LGTM!

takluyver:
Looks good - do we want a news entry for it?

PR #1526, opened 2020-04-02
Reunify selection code

takluyver:
This should largely complete my work on the Cython fast selection code. The Cython selection machinery is used in all cases, and the old pure-Python code to process slicing & indexing is removed. There's a little bit of ugliness because the Cython `_selector` module has to call into the high level `selections` module (to make selection objects), but I think it's worth it.

Other changes:

- Various fixes to the Cython selection code - removing the pure-Python fallback revealed some bugs.
- The previous `Reader` class in Cython has been separated: each Reader now has a `Selector`. Selectors only know about the dataspace and slicing/indexing it; Readers also know about the datatype.
- I have tried to make selection objects more like immutable objects: you create them with all the relevant information, rather than modifying them after creation. They're not strictly immutable (especially PointSelection), but I think this is still good design in general.

I've checked with asv that this doesn't reintroduce significant overhead for simple read patterns. Here's the 'many small reads' benchmark:

Screenshot from 2020-04-02 17-23-38


takluyver:
The coverage drop appears to be because I've removed Python code that was covered. The Cython code that replaces it should all be exercised, but the coverage measurements don't count it yet.

scopatz:
This is awesome!

takluyver:
Can I interest anyone in reviewing this? @scopatz's comment is encouraging, but I'm expecting some more detailed feedback before it lands. It's a moderately substantial change in the core of what h5py does, though it's passing all the tests (including one that was previously an expected failure :-).

h5py/_hl/dataset.py
```diff
@@ -409,13 +409,27 @@ def size(self):
             self._cache_props['size'] = size
         return size
 
+    @property
+    def _selector(self):
```
tacaswell, line 413: There is a race condition here between two threads on with a read-only file, but I think it is mostly harmless (as the failure mode is we generate the selector multiple times and then future calls are cached.

h5py/_hl/selections.py
```diff
@@ -228,35 +224,17 @@ def mshape(self):
...
+        return tuple(x for x, s in zip(self.mshape, scalar) if not s)
 
-        self._sel = (start, count, step, scalar)
-
-        # array shape drops dimensions where a scalar index was selected
-        self._array_shape = tuple(x for x, y in zip(count, scalar) if not y)
-
-        return self
+    def __init__(self, shape, spaceid=None, hyperslab=None):
+        super(SimpleSelection, self).__init__(shape, spaceid)
```
tacaswell, line 231:
```suggestion
        super().__init__(shape, spaceid)
```

h5py/_selector.pyx
```diff
@@ -59,14 +66,16 @@ cdef class Reader:
         cdef:
```
tacaswell, line 66: The docscring on the line above is now wrong (Selector not Reader) but the UI won't let me suggest changes that far from a diff.

h5py/_hl/selections.py
```diff
@@ -17,10 +17,9 @@
 import numpy as np
 
 from .base import product
-from .. import h5s, h5r
+from .. import h5s, h5r, _selector
 
-
-def select(shape, args, dsid):
+def select(shape, args, dataset=None):
```
tacaswell, line 22: Even though this is technically private API we should our users are clever and found it.  I think we should document the (internal) API changes out of an abundance of caution.

h5py/_selector.pyx
```diff
@@ -200,6 +228,58 @@ cdef class Reader:
...
+
+cdef class Reader:
+    cdef hid_t dataset
+    cdef Selector selector
+    cdef TypeID h5_memory_datatype
+    cdef int np_typenum
+
+    def __cinit__(self, DatasetID dsid):
+        self.dataset = dsid.id
+        self.selector = Selector(dsid.get_space())
```
tacaswell, line 272: Maybe rename this `self._selector`?  This is definitely part of the private API (given that we mutate it `read`).

I am normally much more composition over inheritance, but does sub-classing `Reader` from `Selector` make sense here given that they used to be one class?  I am 75% on the side of leaving the change as-is, but thinking out-loud.

h5py/_hl/selections.py
```diff
@@ -349,83 +327,14 @@ def mshape(self):
...
-        for idx in range(len(mshape)):
-            if idx in sequenceargs:
-                mshape[idx] = len(sequenceargs[idx])
-            elif scalar[idx]:
-                mshape[idx] = -1
-
-        self._mshape = tuple(abs(x) for x in mshape)  # Convert -1 back to 1
-        self._array_shape = tuple(x for x in mshape if x >= 0)
+    def __init__(self, shape, spaceid=None, mshape=None, array_shape=None):
+        super(FancySelection, self).__init__(shape, spaceid)
```
tacaswell, line 331:
```suggestion
        super().__init__(shape, spaceid)
```

tacaswell:
I'm mostly :+1: on this, left a couple of questions / comments and trivial code changes.

Moving away from relying on the side-effects of `__getitem__` can only be a good thing.

I think we have to document the API changes as I suspect we have some users who have found and are abusing these functions / classes but I don't think we need a deprecation period (you use private API you take your chances).

h5py/_hl/dataset.py
takluyver: Yup, it should be no problem to have multiple selectors/readers on a single dataset. The cache is not to enforce that there's only one at once, it's just to avoid needlessly re-initialising one.

In fact, two threads should never be using a single selector/reader at the same time, because they're stateful objects. All the cases where a cached one could be used are protected by the `phil` lock.

h5py/_selector.pyx
takluyver: Thanks, good catch.

h5py/_selector.pyx
takluyver: It's not accessible from Python, because it's a cdef-ed attribute not marked as public. I've not been worrying about third party Cython code trying to use it - do you think we should?

I wondered about inheriting as well. I leant towards composition, because a `Reader` knows about an extra dimension of things (datasets and datatypes), whereas a `Selector` operates simply on a dataspace. Also, `__cinit__` doesn't quite behave like a normal Python method:

> If your extension type has a base type, any existing __cinit__() methods in the base type hierarchy are automatically called before your __cinit__() method.

([Cython docs])

So it wouldn't be easy for `Reader` defined as a subclass to accept a DatasetID instead of a SpaceID. There are bound to be ways around that, but it makes things less neat.

GDYendell:
Apologies if this comment is out of place, but this PR was mentioned on #1122 and it caught my attention.

I have recently been investigating the difference in speed between reading large datasets (~GBs) in a single selection
```
data = dataset[...]
```
vs iterating the dataset and reading each chunk one at a time while appending to a numpy array (which seems to be substantially faster, in some cases)
```
array = np.empty(elements, dtype=dataset.dtype)
for chunk_start, chunk_end in pairwise(range(0, elements + 1, chunk_size)):
    array[chunk_start:chunk_end] = dataset[chunk_start:chunk_end]
```
I initially thought it was a VDS issue, but I don't think it is. I also think I have determined the same is not true with the C API. Is this change related at all?

takluyver:
I don't know if I've seen that specific one before, but there have been various odd performance issues. I've deliberately solved at least one such issue in master (by creating a distinction between `selection.mshape` and `selection.array_shape`), and it's possible that my refactoring work over several PRs has unwittingly solved some other things. Though it's also possible that it's introduced new problems...

If you've got time, please do try your test case with this branch or with master once it's merged (I'm just about to rebase it and try to work your changes in). If you still see some odd slowness, we should try to work out what's going on.

h5py/_hl/dataset.py
tacaswell: :+1:

h5py/_selector.pyx
tacaswell: I was thinking of this as public to python, but given that it is not easily accessible, I have no concerns.  Could @kif weigh in on if we need to worry about the cython side extension?

I on fully on board with leaving it as composition.

h5py/_selector.pyx
```diff
@@ -141,10 +145,25 @@ cdef class Selector:
...
+            # MultiBlockSlice exposes h5py's separate count & block parameters
+            # to allow more complex repeating selections.
+            if isinstance(a, MultiBlockSlice):
+                (
+                    self.start[dim_ix],
+                    self.stride[dim_ix],
+                    self.count[dim_ix],
+                    self.block[dim_ix],
+                ) = a.indices(l)
+                print("MBS indices", a.indices(l))
```
tacaswell, line 162:
```suggestion
```

Remove a debugging print.

tacaswell:
Why re-order the MultiBlockSlice indices?

takluyver:
I wanted to have them consistently in the same order as `H5Sselect_hyperslab`. @GDYendell reordered the constructor arguments of MultiBlockSlice at my request, so I'm just bringing the `.indices()` method back into line.

I'm considering, though, replacing the `indices()` method with a `count_for_length()` method returning a single number, as the others are all static anyway.

takluyver:
As it's had some review now, I'll merge this at the end of the week if there are no objections. It's fine to object if you just want more time to look at it, of course.

scopatz:
Windows 3.8 is failing FWIW

takluyver:
Thanks @scopatz  - it was a network error getting something from NuGet. I've re-run it and it's passed now.

PR #1528, opened 2020-04-06
Document ndim in dataset documentation

### LINKED ISSUE #1517: Document Dataset.ndim attribute
I just noticed that the [dataset reference docs](https://docs.h5py.org/en/latest/high/dataset.html#reference) don't mention the `ndim` attribute. Other basic numpy-like attributes like `size` and `shape` are documented, so I guess this was just an oversight.

cydanil:
Hi,
I'm new here, thanks for taking the time to look at this!

This aims to resolve #1517 which says that the documentation for `Dataset.ndim` is lacking.
I added it below references to `size` as it seemed like the right place.

Thanks!
Cyril

### DIFF
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -15,6 +15,7 @@ NumPy operations like slicing, along with a variety of descriptive attributes:
 
   - **shape** attribute
   - **size** attribute
+  - **ndim** attribute
   - **dtype** attribute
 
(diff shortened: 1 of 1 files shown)


kif:
LGTM

takluyver:
Thanks Cyril! I knew we'd get you contributing. :wink:

PR #1529, opened 2020-04-07
Document __bool__ property in file/group/dset

### LINKED ISSUE #1363: Is there an easy way of knowing whether an HDF5 file is opened or closed?
Is there a way of knowing whether an HDF5 file is opened or closed?

Here are my specifications

* Operating System: Mac OS Mojave (10.14.6)
 * Python version: 2.7.15+
 * Where Python was acquired: system Python
 * h5py version: 2.7.1 (`h5py.__version__`)
 * hdf5 version: 1.10.0

cydanil:
This PR aims to resolve #1363: documenting how to easily check if a file descriptor is still valid, as well as groups and datasets.

I also quickly updated the `tempfile.TemporaryFile` example in `file.rst`, where explicitly setting the create flag is required as of version 3.0.

Thanks!
Cyril

### DIFF
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -414,6 +414,21 @@ Reference
 
         NumPy-style slicing to write data.  See :ref:`dataset_slicing`.
 
+    .. method:: __bool__()
+
+        Check that the dataset is accessible.
+        A dataset could be inaccessible for several reasons. For instance, the
+        dataset, or the file it belongs to, may have been closed elsewhere.
+
+        >>> f = h5py.open(filename)
+        >>> dset = f["MyDS"]
+        >>> f.close()
+        >>> if dset:
+        ...     print("datset accessible")
+        ... else:
+        ...     print("dataset is inaccessible")
+        dataset unaccessible
+
     .. method:: read_direct(array, source_sel=None, dest_sel=None)
 
         Read from an HDF5 dataset directly into a NumPy array, which can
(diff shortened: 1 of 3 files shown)


scopatz:
Thanks @cydanil! This looks good to me

docs/high/dataset.rst
```diff
@@ -614,3 +614,16 @@ Reference
     .. attribute:: parent
 
         :class:`Group` instance containing this dataset.
+
+    .. attribute:: __bool__
```
takluyver, line 618: `__bool__()` is a method, so we should use the `.. method::` directive.

docs/high/dataset.rst
```diff
@@ -614,3 +614,16 @@ Reference
...
+
+        Check that the dataset is accessible:
```
takluyver, line 620: I think we should say that you can use this to check if the file it belongs to has been closed. It's clear in the example, but a bit of repetition doesn't hurt to make things clear.

There are other ways you can make an object invalid, like calling `ds.id.close()`, so it's not only telling you if the file is still open, but I think that's the main useful scenario.

docs/high/dataset.rst
cydanil: Ah yes, I also relocated the entries in the files accordingly.

docs/high/dataset.rst
cydanil: That makes sense, I've clarified that datasets can be inaccessible for various reasons.

cydanil:
Thanks for your feedbacks! I addressed them.

docs/high/group.rst
```diff
@@ -195,6 +195,19 @@ Reference
         Create a new link, or automatically create a dataset.
         See :ref:`group_links`.
 
+    .. method:: __bool__()
+
+        Check that the file the group belongs to is open:
```
takluyver, line 200: To be very picky, the same thing you said for datasets also applies to groups - a group also has an HDF5 ID which can be closed. I'm not sure there's any particular reason to do that explicitly when working with h5py, but we may as well be precise.

Other than that, this LGTM. :-)

cydanil: Ah, of course. I hadn't tried that.

takluyver:
Thanks!

cydanil:
Thank you all!

Issue #1475, opened 2020-01-17
Unspecified error in H5Dget_storage_size (return value ==0)


ajelenak:
`H5Dget_storage_size()` correctly reports zero byte size for contiguous HDF5 datasets that were created but with no data written to their files. H5py, however, interprets return value of zero as an error code. The function's description page at the HDF [portal](https://portal.hdfgroup.org/display/HDF5/H5D_GET_STORAGE_SIZE) states:

> Note that `H5D_GET_STORAGE_SIZE` does not differentiate between 0 (zero), the value returned for the storage size of a dataset with no stored values, and 0 (zero), the value returned to indicate an error.

#### Proposed solution

Modify `DatasetID.get_storage_size()` method:
```python
    def get_storage_size(self):
        """ () => LONG storage_size

            Determine the amount of file space required for a dataset.  Note
            this only counts the space which has actually been allocated; it
            may even be zero.
        """
        cdef haddr_t offset
        offset = H5Dget_offset(self.id)
        if offset == HADDR_UNDEF:
            return 0
        else:
            return H5Dget_storage_size(self.id)
```
`H5Dget_offset()` is guaranteed to return `HADDR_UNDEF` for such HDF5 datasets and this can be used to avoid calling `H5Dget_storage_size()`.

#### h5py configuration
```
h5py    2.10.0
HDF5    1.10.6
Python  3.8.1 (default, Jan  5 2020, 21:31:55) [Clang 9.1.0 (clang-902.0.39.2)]
sys.platform    darwin
sys.maxsize     9223372036854775807
numpy   1.18.1
```

takluyver:
Thanks for reporting, that's a frustrating ambiguity.

Unfortunately, checking `H5Dget_offset()` has other issues - that only works for contiguous datasets, and returns `HADDR_UNDEF` for all other layout types. But maybe we can use [H5Dget_space_status](https://portal.hdfgroup.org/display/HDF5/H5D_GET_SPACE_STATUS) instead to check if space has been allocated for the dataset. Do you want to make a PR?

Also, check what happens in the corner case of a zero-size dataset, where it doesn't need any space to store data.

ajelenak:
I sincerely appologize for not following through with this sooner but hope we can do it quickly now.

Unfortunately, there does not seem to be a single HDF5 function that can help here. Because `H5Dget_strorage_size()` is applicable to both contiguous and chunked datasets, below is my second approach:

```python
    def get_storage_size(self):
[...]

PR #1531, opened 2020-04-11
Fixes h5py/h5py#1475

ajelenak:
The most important change here is in `api_gen.py` which for the `H5Dget_storage_size()` function now generates this custom wrapper code:
```python
cdef hsize_t H5Dget_storage_size(hid_t dset_id) except <hsize_t>-1:
    cdef hsize_t r
    set_default_error_handler()
    r = _hdf5.H5Dget_storage_size(dset_id)
    if r==0:
        if set_exception():
            return <hsize_t>-1
    return r
```
`r=0` is now a valid return value unless the HDF5 library reports an error.

Also included are one test and an update to the `DatasetID.get_storage_size()` method's docstring.

### DIFF
--- a/api_gen.py
+++ b/api_gen.py
@@ -113,7 +113,11 @@ class Line(object):
             self.err_value = f"<{self.code}>-1"
         elif self.code in ('unsigned int', 'haddr_t', 'hsize_t', 'size_t'):
             self.err_condition = "==0"
-            self.err_value = f"<{self.code}>0"
+            if self.fname == 'H5Dget_storage_size':
+                # Special case: https://github.com/h5py/h5py/issues/1475
+                self.err_value = f"<{self.code}>-1"
+            else:
+                self.err_value = f"<{self.code}>0"
         else:
             raise ValueError("Return code <<%s>> unknown" % self.code)
 
@@ -253,7 +257,21 @@ cdef {0.code} {0.fname}({0.sig}) except {0.err_value}:
 
 """
         else:
-            imp = """\
+            if self.line.fname == 'H5Dget_storage_size':
+                # Special case: https://github.com/h5py/h5py/issues/1475
+                imp = """\
+cdef {0.code} {0.fname}({0.sig}) except {0.err_value}:
+    cdef {0.code} r
+    set_default_error_handler()
+    r = _hdf5.{0.fname}({0.args})
+    if r{0.err_condition}:
+        if set_exception():
+            return {0.err_value}
+    return r
+
+"""
+            else:
+                imp = """\
 cdef {0.code} {0.fname}({0.sig}) except {0.err_value}:
     cdef {0.code} r
     set_default_error_handler()
--- a/api_gen.py
+++ b/api_gen.py
@@ -113,11 +113,7 @@ class Line(object):
             self.err_value = f"<{self.code}>-1"
         elif self.code in ('unsigned int', 'haddr_t', 'hsize_t', 'size_t'):
             self.err_condition = "==0"
-            if self.fname == 'H5Dget_storage_size':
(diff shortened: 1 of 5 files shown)


api_gen.py
```diff
@@ -113,7 +113,11 @@ def __init__(self, text):
             self.err_value = f"<{self.code}>-1"
         elif self.code in ('unsigned int', 'haddr_t', 'hsize_t', 'size_t'):
             self.err_condition = "==0"
-            self.err_value = f"<{self.code}>0"
+            if self.fname == 'H5Dget_storage_size':
+                # Special case: https://github.com/h5py/h5py/issues/1475
+                self.err_value = f"<{self.code}>-1"
```
takluyver, line 118: I think the best thing to do for this function is to use the `except?` syntax, which the [Cython docs describe](http://docs.cython.org/en/latest/src/userguide/language_basics.html#error-return-values) say is for:

> If all possible return values are legal and you can’t reserve one entirely for signalling errors

(`hsize_t` is unsigned, so -1 can't be represented. I *guess* -1 will be converted to the largest possible `unsigned long long`, which would be implausible as a size. But I'd rather not rely on that guess when Cython has a simple, documented way to handle this case.)

takluyver:
Housekeeping points: the CI is asking you to add yourself in `.authors.yml`, and we should also make a [news entry](https://regro.github.io/rever-docs/news.html) about the fix.

api_gen.py
ajelenak: `hsize_t` is declared in `api_types_hdf5.pxd` as `ctypedef long long hsize_t`, which I think is a signed integer type.

takluyver: That may be a mistake. It's unsigned in the HDF5 code: https://bitbucket.hdfgroup.org/projects/HDFFV/repos/hdf5/browse/src/H5public.h?at=refs%2Ftags%2Fhdf5-1_10_6#186

The two lines below it in `api_types_hdf5.pxd` explicitly use `signed long long`, so I think `hsize_t` was meant to be different from those.

tacaswell:
Does this need to be put behind a hdf5 version check or has this always been the behavior of hdf5?

takluyver:
I'm guessing this has always been the case - [the docs](https://portal.hdfgroup.org/display/HDF5/H5D_GET_STORAGE_SIZE) don't mention it as a new change, and presumably it's always been possible to have 0 bytes stored for a given dataset.

@ajelenak the CI is still complaining about the `.authors.yml` file - I think because the commits are with a different email. Here's the message:

```
'Aleksandar Jelenak' does not have 'ajelenak@users.noreply.github.com' listed! Please add the email to 'Aleksandar Jelenak' in '.authors.yml'. For example:

- name: Aleksandar Jelenak
  email: ajelenak@users.noreply.github.com

Or add an alternative email:

- name: Aleksandar Jelenak
  email: person@example.com
  alternate_emails:
    - ajelenak@users.noreply.github.com
```

@scopatz maybe rever could take `github: ajelenak` to implicitly associate the `@users.noreply.github.com` address?

ajelenak:
> I'm guessing this has always been the case - [the docs](https://portal.hdfgroup.org/display/HDF5/H5D_GET_STORAGE_SIZE) don't mention it as a new change, and presumably it's always been possible to have 0 bytes stored for a given dataset.

Yes, the documentation page would include version change history if there was one.

> @ajelenak the CI is still complaining about the `.authors.yml` file - I think because the commits are with a different email.

I added the suggested alternate email.

takluyver:
Thanks. This looks good to me. I'll leave it a day or two in case anyone else wants to review, otherwise I'll merge it.

Issue #1338, opened 2019-09-21
String conversions for h5py 3.0


takluyver:
@tacaswell, @aragilar and myself sat down yesterday to discuss how various string-like types should be mapped between Python and HDF5. We have decided we do want to make some changes for h5py 3.0. This will inevitably break some code using h5py, but we think it will create a more consistent API, which is also a better fit with Python 3's separation of bytes & str.

We have divided the rules into three kinds of conversion:

```python
# 1. Writing to an existing dataset
f['data'][:10] = obj
# Also applies to creating datasets with data and a specified dtype
f.create_dataset('data', dtype=x, data=obj)

# 2. Creating a new dataset without a specified dtype
f['data'] = obj
f.create_dataset('data', data=obj)

# 3. Reading data
obj = f['data'][:]
```

We aim to ensure that `f['data2'] = f['data1'][:]` will always create the new dataset with the same data type as the copied one. We believe the rules below preserve this - please comment if you notice a case we've missed.

## 1. Writing to existing datasets

Everything rejects numpy U dtypes (UTF-32 fixed width), as in h5py 2.x, because there is no equivalent HDF5 type.

### fixed-width ascii

- accept bytes
- accept str and encode to ascii, error on invalid ascii
  - If you need to write non-ascii data, encode it first and pass bytes
- check that it will fit in width, raise if not
 
### fixed width utf-8
 - accept bytes, just check length
 - accept str, encode to utf-8

### variable width ascii
 - accept any bytes (except NULL)
 - accept str, encode to ascii, error on non-ascii
 
### variable length utf-8
- accept any bytes (except NULL)
- accept str, encode to utf-8

### Opaque
- accept bytes, np.void()
- reject str

## 2. Creating datasets

- numpy object/string array with 'tagged' dtype -> follow the tag
  - tagged means a dtype with string metadata created by h5py to indicate
    string charset & width
- numpy string array not tagged -> fixed width ascii
- numpy object array of bytes not tagged -> void opaque **(see discussion in comments)**
  - List of bytes -> opaque
- numpy object array of str not tagged -> variable length utf-8
  - list of str -> variable length utf-8
- numpy void array -> opaque
- numpy array of U type -> raise

## 3. Reading data

- ASCII Fixed -> numpy 'S' (tagged with ASCII encoding)
- UTF-8 Fixed -> numpy 'S' (tagged with UTF-8 encoding)
[...]

PR #1535, opened 2020-04-19
Make creating a new dataset from strings/bytes more consistent

takluyver:
This aims to make h5py's behaviour more consistent when you create a new dataset/attribute from string objects, without specifying a dtype, e.g. `f['a'] = x`:

1. A list of strings/bytes should result in the same datatype as a single object. Previously, a single string/bytes object was stored as a variable-length HDF5 string, a list of bytes was stored as fixed length, and a list of str was rejected. Now, they all result in variable-length strings in the file.
2. Treat numpy arrays with an `object` dtype but no h5py tag like lists/tuples. I've referred to these in the code as 'object collections' - as distinct from arrays where the dtype tells us what to do.
3. More similarity between datasets and attributes: attributes had a special case to allow lists/tuples of strings. This makes the same thing work for creating datasets.

This is a step towards #1338. It differs somewhat from what we agreed there, but I'm having doubts about that, as I mentioned in a comment there. I also think the consistency is an improvement even if we decide to change the type relations later.

### DIFF
--- a/h5py/_hl/base.py
+++ b/h5py/_hl/base.py
@@ -88,6 +88,38 @@ def guess_dtype(data):
         return None
 
 
+def is_float16_dtype(dt):
+    if dt is None:
+        return False
+
+    dt = np.dtype(dt)  # normalize strings -> np.dtype objects
+    return dt.kind == 'f' and dt.itemsize == 2
+
+
+def array_for_new_object(data, specified_dtype=None):
+    """Prepare an array from data used to create a new dataset or attribute"""
+
+    # We mostly let HDF5 convert data as necessary when it's written.
+    # But if we are going to a float16 datatype, pre-convert in python
+    # to workaround a bug in the conversion.
+    # https://github.com/h5py/h5py/issues/819
+    if is_float16_dtype(specified_dtype):
+        as_dtype = specified_dtype
+    else:
+        as_dtype = guess_dtype(data)
+
+    data = np.asarray(data, order="C", dtype=as_dtype)
+
+    # In most cases, this does nothing. But if data was already an array,
+    # and guess_dtype made a tagged version of the dtype it already had
+    # (e.g. an object array of strings), asarray() doesn't replace its
+    # dtype object. This gives it the tagged dtype:
+    if as_dtype is not None:
+        data.dtype = as_dtype
+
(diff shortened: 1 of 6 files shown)


h5py/_hl/base.py
```diff
@@ -11,6 +11,7 @@
     Implements operations common to all high-level objects (File, etc.).
 """
 
+import numpy as np
```
scopatz, line 14: Shouldn't the numpy import come after the standard library imports?

scopatz:
Nice work here, @takluyver!

h5py/_hl/base.py
takluyver: Looking through the other `_hl` modules, it doesn't look like we've got a consistent way of organising imports. Do we want to impose one?

I tend to combine stdlib and third-party imports in one alphabetical list, mixing `from` imports in with regular imports, but separate local imports. But I'm not very consistent about any of that, and I'm happy to follow another convention if we like.

scopatz: I guess I always follwo pep8, where the 3rd party imports come after stdlib ones. We should probably impose at least that :wink:

takluyver: Ah, I didn't know PEP 8 made recommendations about import order. I've rearranged these ones now.

takluyver:
@scopatz - to check, are you satisfied with the *concept* here as well as the implementation? I think making the conversion more consistent is a benefit in any case, but what I've done here is not exactly what @tacaswell, @aragilar and I agreed a year ago (see [this comment](https://github.com/h5py/h5py/issues/1338#issuecomment-615337758) for why that's the case), and I'd really like to get some consensus on what we're doing with strings. 3.0 is probably our best chance to overhaul this.

scopatz:
@takluyver - yes. I am happy with the concept implemented here too. It seems more consistent which is a big :+1:

takluyver:
Thanks! Then I'll merge this on Friday if there are no objections.

takluyver:
I've added a news entry, but not tried to update the *Strings in HDF5* documentation page. As there are more changes to strings due (see plans in #1338), I'm planning to wait and then rewrite it in one go later on, rather than adjusting it with each PR.

PR #1542, opened 2020-05-02
Update to use HDF5 1.8 API as the base, rather than 1.6

takluyver:
Build with the `H5_USE_18_API` option, and update APIs changed in 1.8 to use the new variants of the functions. We already dropped support for HDF5 1.6 some time ago, but this wasn't cleaned up.

Where HDF5 1.8 introduced version 2 of a function, we have three options:

1. Explicitly use the old version, e.g. `H5Aiterate1`.
2. Use the new function, e.g. `H5Aiterate2`
3. Use the compatibility macro, e.g. `H5Aiterate`. This is equivalent to option 2 when we use the `H5_USE_18_API` option.

These changes at the moment are a mixture of options 2 & 3. We should be systematic about that, but I wasn't sure which one was preferable, so I stuck to making the fewest possible changes, which varies from case to case.

### DIFF
--- a/h5py/api_functions.txt
+++ b/h5py/api_functions.txt
@@ -237,12 +237,9 @@ hdf5:
   hid_t     H5Oopen_by_addr(hid_t loc_id, haddr_t addr)
   hid_t     H5Oopen_by_idx(hid_t loc_id, char *group_name, H5_index_t idx_type, H5_iter_order_t order, hsize_t n, hid_t lapl_id)
 
-  1.6.0-1.11.2 herr_t    H5Oget_info(hid_t loc_id, H5O_info_t *oinfo)
-  1.11.6 herr_t    H5Oget_info1(hid_t loc_id, H5O_info_t *oinfo)
-  1.6.0-1.11.2 herr_t    H5Oget_info_by_name(hid_t loc_id, char *name, H5O_info_t *oinfo, hid_t lapl_id)
-  1.11.6 herr_t    H5Oget_info_by_name1(hid_t loc_id, char *name, H5O_info_t *oinfo, hid_t lapl_id)
-  1.6.0-1.11.2 herr_t    H5Oget_info_by_idx(hid_t loc_id, char *group_name,  H5_index_t idx_type, H5_iter_order_t order, hsize_t n, H5O_info_t *oinfo, hid_t lapl_id)
-  1.11.6 herr_t    H5Oget_info_by_idx1(hid_t loc_id, char *group_name,  H5_index_t idx_type, H5_iter_order_t order, hsize_t n, H5O_info_t *oinfo, hid_t lapl_id)
+  herr_t    H5Oget_info(hid_t loc_id, H5O_info_t *oinfo)
+  herr_t    H5Oget_info_by_name(hid_t loc_id, char *name, H5O_info_t *oinfo, hid_t lapl_id)
+  herr_t    H5Oget_info_by_idx(hid_t loc_id, char *group_name,  H5_index_t idx_type, H5_iter_order_t order, hsize_t n, H5O_info_t *oinfo, hid_t lapl_id)
 
   herr_t    H5Olink(hid_t obj_id, hid_t new_loc_id, char *new_name, hid_t lcpl_id, hid_t lapl_id)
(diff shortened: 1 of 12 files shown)


takluyver:
Following a discussion with @tacaswell, we're going to prefer the compatibility macros (option 3). There are arguments both ways, but the reasons we found decisive were:

1. We can't use versioned functions consistently, because the `H5Xfunc1` name doesn't exist until someone makes `H5Xfunc2`. So we're using the macros at least some of the time anyway. We'd rather "work with the system than against it".
2. There are 'function families' which all change to a new version together, and should probably be kept in line. E.g. several `H5O` functions got a new version in 1.12. It seems like a good idea to keep these in sync, which is easier if we use the macros rather than versioning them ourselves.

The biggest counterargument is that using the macros effectively means changing unrelated functions together when we move to a new API flavour.

This isn't meant to be a strict rule: we'll use the macros by default, but we can use the versioned function if it makes more sense for what we want to do.

h5py/api_functions.txt
```diff
@@ -241,12 +237,9 @@ hdf5:
   hid_t     H5Oopen_by_addr(hid_t loc_id, haddr_t addr)
   hid_t     H5Oopen_by_idx(hid_t loc_id, char *group_name, H5_index_t idx_type, H5_iter_order_t order, hsize_t n, hid_t lapl_id)
 
-  1.6.0-1.11.2 herr_t    H5Oget_info(hid_t loc_id, H5O_info_t *oinfo)
-  1.11.6 herr_t    H5Oget_info1(hid_t loc_id, H5O_info_t *oinfo)
-  1.6.0-1.11.2 herr_t    H5Oget_info_by_name(hid_t loc_id, char *name, H5O_info_t *oinfo, hid_t lapl_id)
-  1.11.6 herr_t    H5Oget_info_by_name1(hid_t loc_id, char *name, H5O_info_t *oinfo, hid_t lapl_id)
-  1.6.0-1.11.2 herr_t    H5Oget_info_by_idx(hid_t loc_id, char *group_name,  H5_index_t idx_type, H5_iter_order_t order, hsize_t n, H5O_info_t *oinfo, hid_t lapl_id)
-  1.11.6 herr_t    H5Oget_info_by_idx1(hid_t loc_id, char *group_name,  H5_index_t idx_type, H5_iter_order_t order, hsize_t n, H5O_info_t *oinfo, hid_t lapl_id)
+  herr_t    H5Oget_info(hid_t loc_id, H5O_info_t *oinfo)
```
tacaswell, line 240: Are we still being saved by the caching on azure here?  This looks like an API that was changed for 1.12 so will escape the 1.8 related pinning.

takluyver: I don't think there's a problem being hidden by caching here. On the second commit of this PR, HDF5 1.12 was [built from source](https://dev.azure.com/h5pyappveyor/h5py/_build/results?buildId=462&view=logs&j=f3407ca5-1e8e-5831-d467-f40b15dcbaf4&t=96643522-c40b-5680-4fd9-7ab105c6ebd9). [Cache isolation](https://docs.microsoft.com/en-us/azure/devops/pipelines/release/caching?view=azure-devops#cache-isolation-and-security) seems to give PRs read-only access to the cache for master, so the cached build that caused problems in PR #1506 shouldn't affect any other PR.

[H5Oget_info](https://confluence.hdfgroup.org/display/HDF5/H5O_GET_INFO) was introduced in 1.8, and versioned in 1.12. So the duplication I'm removing was a necessary workaround when we were requesting the 1.6 API, because in 1.12 it would float to the newest version. It was added in #1335. Now that we're requesting the 1.8 API, 1.12 knows to give us the older version, so this simplification is possible.

There's one function, [H5Pencode](https://confluence.hdfgroup.org/display/HDF5/H5P_ENCODE), which was added in 1.10 and versioned in 1.12. We don't currently wrap that, but if we do, we'll need a similar workaround.

takluyver:
I'll merge this in a couple of days unless anyone objects. I've discussed the idea with @tacaswell, and the specific changes involved are largely mechanical.

h5py/api_functions.txt
tacaswell: Ah, I was inferring from the version spec in this file that the functions were available in 1.6 without doing the legwork to sort out what was going on, sorry about that :sheep: .

Issue #917, opened 2017-08-28
Import h5py fails in serial code after being built with parallel support.


leonfoks:
I have h5py built with parallel support and it is working fine with no issues in parallel.  However, if I try to "import h5py" in a serial run i.e.

```
python -c 'import h5py'
```
instead of
```
srun (or mpirun) --mpi=pmi2 python -c 'import h5py'
```
it fails with the message attached to the bottom of this issue.  The message is related to the mpi library not being able to initialize because it was not run using srun or mpirun.

I believe that the user should still be able to import and run h5py serially despite a parallel install.
Perhaps mpi4py should be initialized when the user opens a file with the mpio driver, and not at import?

OS: Scientific linux
Python: 3.6.0 :: Anaconda 4.3.1
Openmpi: 1.10.2
HDF5: 1.8.18
mpi4py: 2.0.0
h5py: 2.8.0.post0

Thanks!

```
--------------------------------------------------------------------------
It looks like orte_init failed for some reason; your parallel process is
likely to abort.  There are many reasons that a parallel process can
fail during orte_init; some of which are due to configuration or
environment problems.  This failure appears to be an internal failure;
here's some additional information (which may only be relevant to an
Open MPI developer):

  PMI2_Job_GetId failed failed
  --> Returned value (null) (14) instead of ORTE_SUCCESS
--------------------------------------------------------------------------
--------------------------------------------------------------------------
It looks like orte_init failed for some reason; your parallel process is
likely to abort.  There are many reasons that a parallel process can
fail during orte_init; some of which are due to configuration or
environment problems.  This failure appears to be an internal failure;
here's some additional information (which may only be relevant to an
Open MPI developer):

  orte_ess_init failed
  --> Returned value (null) (14) instead of ORTE_SUCCESS
--------------------------------------------------------------------------
--------------------------------------------------------------------------
It looks like MPI_INIT failed for some reason; your parallel process is
likely to abort.  There are many reasons that a parallel process can
fail during MPI_INIT; some of which are due to configuration or environment
problems.  This failure appears to be an internal failure; here's some
additional information (which may only be relevant to an Open MPI
developer):

  ompi_mpi_init: ompi_rte_init failed
  --> Returned "(null)" (14) instead of "Success" (0)
--------------------------------------------------------------------------
*** An error occurred in MPI_Init_thread
*** on a NULL communicator
*** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
***    and potentially your MPI job)
```

[...]

PR #1544, opened 2020-05-03
tell mpi4py not to automatically initialize MPI at import

boegel:
This is an (incomplete) attempt at fixing #917.

The issue is that anything beyond a simple `import mpi4py` will trigger an `MPI_Init` call, which is problematic in a number of contexts. That's annoying, since it means `import h5py` will trigger an `MPI_Init` call, which often results in a hard crash, even when there's no intention to leverage the MPI support at all...

This can be avoided by instructing `mpi4py` to not initialize MPI automatically on import via `mpi4p.rc.initialize` ('rc' is runtime configuration here), but then the `mpi4py.MPI.Init()` needs to be done whenever any MPI calls will actually be used...

I'm clearly still figuring out the last part, but I'm stabbing in the dark a bit since I'm not very familiar with the `h5py` codebase, so I would appreciate some pointers on this.
Is there a central place where doing the `MPI.Init()` would make sense? Or am I overlooking a location in my current changes?

Also, the tests are currently just exiting without producing much helpful information, I can't actually tell where the MPI error is coming from...
Is there a way to make `pytest` not hide that information?

Overall, would this change make sense to get merged (once it's done properly)?

### DIFF
--- a/h5py/__init__.py
+++ b/h5py/__init__.py
@@ -15,6 +15,15 @@
 from warnings import warn as _warn
 import atexit
 
+# Tell mpi4py to not automatically initialize MPI at import;
+# required to avoid problems with 'import h5py' in some environment
+# when h5py was built with MPI support
+try:
+    import mpi4py
+    mpi4py.rc.initialize = False
+except ImportError:
+    pass
+
 
 # --- Library setup -----------------------------------------------------------
 
--- a/h5py/__init__.py
+++ b/h5py/__init__.py
@@ -21,6 +21,7 @@ import atexit
 try:
     import mpi4py
     mpi4py.rc.initialize = False
+    mpi4py.rc.finalize = True
 except ImportError:
     pass
 
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -45,6 +45,8 @@ def _set_fapl_mpio(plist, **kwargs):
         raise ValueError("h5py was built without MPI support, can't use mpio driver")
 
     import mpi4py
+    if not mpi4py.MPI.Is_initialized():
+        mpi4py.MPI.Init()
     kwargs.setdefault('info', mpi4py.MPI.Info())
     plist.set_fapl_mpio(**kwargs)
 
--- a/h5py/tests/common.py
+++ b/h5py/tests/common.py
@@ -58,9 +58,11 @@ class TestCase(ut.TestCase):
         return tempfile.mktemp(suffix, prefix, dir=dir)
 
     def mktemp_mpi(self, comm=None, suffix='.hdf5', prefix='', dir=None):
+        from mpi4py import MPI
         if comm is None:
-            from mpi4py import MPI
             comm = MPI.COMM_WORLD
+        if not MPI.Is_initialized():
+            MPI.Init()
         fname = None
         if comm.Get_rank() == 0:
             fname = self.mktemp(suffix, prefix, dir)
--- a/h5py/tests/common.py
+++ b/h5py/tests/common.py
@@ -57,18 +57,6 @@ class TestCase(ut.TestCase):
             dir = self.tempdir
         return tempfile.mktemp(suffix, prefix, dir=dir)
 
-    def mktemp_mpi(self, comm=None, suffix='.hdf5', prefix='', dir=None):
-        from mpi4py import MPI
-        if comm is None:
-            comm = MPI.COMM_WORLD
-        if not MPI.Is_initialized():
-            MPI.Init()
-        fname = None
-        if comm.Get_rank() == 0:
-            fname = self.mktemp(suffix, prefix, dir)
-        fname = comm.bcast(fname, 0)
-        return fname
-
     def setUp(self):
         self.f = h5py.File(self.mktemp(), 'w')
 
--- a/h5py/tests/test_file.py
+++ b/h5py/tests/test_file.py
@@ -699,6 +699,17 @@ class TestPickle(TestCase):
 # hence no subclassing TestCase
 @pytest.mark.mpi
 class TestMPI(object):
+
+    @pytest.fixture(autouse=True)
+    def init_mpi(self):
+        # make sure MPI is initialized;
+        # this is required because:
+        # i) mpi4py is told not to initialize on import by h5py
+        # ii) the mpi_file_name fixture provides by pytest-mpi calls MPI_Comm_rank
+        from mpi4py import MPI
+        if not MPI.Is_initialized():
+            MPI.Init()
+
     def test_mpio(self, mpi_file_name):
         """ MPIO driver and options """
         from mpi4py import MPI
--- a/tox.ini
+++ b/tox.ini
@@ -35,7 +35,7 @@ commands =
     test: python -c "import sys; print('64 bit?', sys.maxsize > 2**32)"
     test: python {toxinidir}/ci/fix_paths.py
     test-!mpi4py: python -m pytest --pyargs h5py --cov=h5py -rxXs --cov-config={toxinidir}/.coveragerc {posargs}
-    test-mpi4py: mpirun -n {env:MPI_N_PROCS:2} {envpython} -m pytest --pyargs h5py -rxXs --with-mpi {posargs}
+    test-mpi4py: mpirun -n {env:MPI_N_PROCS:2} {envpython} -m pytest --verbose --pyargs h5py -rxXs --with-mpi {posargs}
 changedir =
     test: {toxworkdir}
 passenv =


aragilar:
@boegel The issue is MPI, pytest is being run under MPI, and the MPI process is dying, giving the output you can see. My strategy for debugging this was lots of print functions (as tools like pdb do not work under MPI - if you have one, I'd welcome the suggestion). I was hoping to have time in March to look at improving MPI support, but that didn't end up happening.

With respect to the idea of delaying initialisation, that seems reasonable to me. What I suggest doing is creating a context manager, and then a decorator around the context manager, similar to the lock system we have, which should make it easy to ensure that MPI is only initialised when needed.

boegel:
@aragilar Thanks for the feedback, much appreciated!

And then how/where should that context manager be used? 
With lock system, do you mean the stuff in https://github.com/h5py/h5py/blob/master/h5py/_locks.pxi?

h5py/__init__.py
```diff
@@ -15,6 +15,15 @@
 from warnings import warn as _warn
 import atexit
 
+# Tell mpi4py to not automatically initialize MPI at import;
+# required to avoid problems with 'import h5py' in some environment
+# when h5py was built with MPI support
+try:
+    import mpi4py
+    mpi4py.rc.initialize = False
```
takluyver, line 23: As someone not using MPI, I'd quite like to preserve the current pattern where a non-MPI build of h5py never tries to touch mpi4py at all.

Does it work to set this setting everywhere we already import `mpi4py`, without adding an import?

boegel: The problem is that setting the `rc.initialize` to `False` must be done as already as possible, before anything is imported from `mpi4py`.

So checking whether `h5py` was built with MPI support is difficult, because importing `h5py` submodules will trigger imports from `mpi4py`, and then it's too late (the `MPI_Init` done by `mpi4py` is already done).

I can move this to wherever `mpi4py` is used first, but only once I'm sure where it is imported first, that's not clear to me yet (nor is it clear to me why the tests are failing, but I still need to deep dive into it).

takluyver: I think we only currently import mpi4py in one place (`_set_fapl_mpio`), which you've already found, assuming Cython `cimport` doesn't trigger initialisation.

takluyver: The failure on Azure is also there on master; I'll open a separate issue about that. The Travis failure is in the MPI test run, so that's all yours. :wink:

takluyver:
Btw, pytest-mpi was created by @aragilar for h5py, so if we need changes there, that should be possible.

boegel:
Seems like I've finally figured out the problem...

Using `pytest --capture=no --verbose` no, this error message popped up:

```
=============================== MPI Information ================================
*** The MPI_Comm_rank() function was called before MPI_INIT was invoked.
*** This is disallowed by the MPI standard.
*** Your MPI job will now abort.
```

It's probably not a bad idea to add `--capture=no` in `tox.ini` to avoid that the errors produced when running under `mpirun` don't get swallowed and you're surfing blind (that certainly would've saved me a lot of time).

It took me a while to realize what was calling `MPI_Comm_rank()`, since `h5py` itself isn't doing that at all.
After lots of fiddling around (which included reproducing the Travis test environment in a Docker container...), I finally noticed the `mpi_file_name` fixture provided by `pytest-mpi`, which was indeed the culprit (see https://github.com/aragilar/pytest-mpi/blob/a8e7b56301611621004e8551b5bf785bc2810c57/pytest_mpi/__init__.py#L141).

Making sure that MPI is initialized before running any of the MPI tests by adding an `autorun` fixture in `TestMPI` fixed the failing tests .

I'm not sure whether this should be handled by `pytest-mpi`, maybe it should. It's simple enough: make sure that MPI is initialized before doing any MPI calls, since although mpi4py initializes by default, it can be disabled (like I'm doing here).

On the other hand, it's `h5py` deviating from the default `mpi4py` behavior of initializing at import, so maybe it's better suited to be dealt with here. I'll let others be the judge of that.

Please let me know what (else) needs to change for this PR to be eligible for merging!

h5py/__init__.py
boegel: `mpi4py` is also imported from `h5py/h5p.pyx`, so not only in `__set_fapl_mpio`.

I'm not sure that disabling `mpi4py.rc.initialize` can be moved easily elsewhere without re-introducing the problem that the proposed changes are fixing, it should be done as early as possible...

Setting `mpi4py.rc.initialize` to `False` in `__set_fapl_mpio` doesn't make sense, since you actually *want* it to be initialized at that point (that's when you're sure that MPI is actually required).

takluyver: In `h5p.pyx` it's a Cython `cimport` - does even that trigger initialisation? If so, can we suppress it in the same Cython file?

I'm quite keen that a non-MPI build of h5py shouldn't touch mpi4py at all, even if it's there. Maybe @aragilar has some more ideas about how to achieve that.

boegel: It's not entirely clear to me when exactly the `mpi4py` auto-initialize is triggered.
Certainly not at `import mpi4py`, that's clear (or this would never work).

Maybe @dalcinl can clarify?

dalcinl: A bare `import mpi4py` does not trigger `MPI_Init()`, however `from mpi4py import MPI` or `import mpi4py.MPI` does, unless using the `mpi4py.rc.initialize = False` hackery.

That initialize hackery was not meant to be used by libraries, but end users that really need to workaround issues in their main, entry point scripts to inject pre-MPI_Init stuff if really needed.

My take on this is that `from mpi4py import MPI` should not be issued for non-MPI builds of HDF5, and maybe only issueed if the user somehow requests (via imports or calls) the use of MPI-specific features, of couse this is probably not so easy to implement in h5py and there are potential risks of init-deadlocks (talk is cheap, and I'm doing it quite irreponsibly as usual these days).

dalcinl: > it's a Cython `cimport` - does even that trigger initialisation?

Cython's `cimport` actually implies a Python `import` behind the scenes.

boegel: Thanks for the feedback!

It looks like the `MPI_init` that `mpi4py` does by default is then either triggered by the `from mpi4py.MPI import` statement in https://github.com/h5py/h5py/blob/78e665c1991a2a1fbdffa99e3d5f87597881a082/h5py/api_types_ext.pxd#L15
or by the `from mpi4py.mpi_c cimport` statement in https://github.com/h5py/h5py/blob/2660313fcfdb927303d96848014a639b6230abb0/h5py/h5p.pyx#L39

Could the latter imports also trigger the `MPI_Init`?

dalcinl: It is be the first one. The second one involves a pxd file that do not expose Python types, just all the pure-C extern declarations corresponding to the MPI API from the usual `mpi.h` include file.

In petsc4py, in order to not impose a hard compile-time dependency on mpi4py (PETSc can be built without MPI), I'm using some hackery that may help h5py:

https://bitbucket.org/petsc/petsc4py/src/54650a1436415513083e5856677a51f594860e80/src/PETSc/petscmpi.pxi#lines-36

Of course, you should be aware of the potential issues of doing imports at the function level.

boegel: It seems to me that's the proper way forward, to not always blindly do the `from mpi4py.MPI import ...` when `h5py` was built with MPI support enabled.

Disabling the auto-init done by `mpi4py` is biting us in the ass already when using an `h5py` installation that was patched like this, in scripts where `h5py` and `mpi4py` are combined, since whether or not auto-init will happen now depends on the order in which they are imported... 😱 

I'm willing to put more time into this, but I'm not familiar with Cython stuff at all, so this goes beyond what I'm capable of tackling myself in a reasonable amount of time (at least not without significant help from someone more familiar with this).

Is anyone up for helping out, maybe doing a short pair programming session?
Perhaps @aragilar or @takluyver?

takluyver: Thanks @dalcinl for the clarifications.

Would it work to switch the `from mpi4py.MPI cimport ...` in `api_types_ext.pxd` to `from mpi4py.libmpi cimport ...`, like in `h5p.pyx`? If I've understood correctly, the second form lets us make the C types available for Cython but doesn't trigger initialisation.

dalcinl: > Would it work to switch the `from mpi4py.MPI cimport ...`

Probably not, because you are *cimporting* the `Comm` class from `mpi4py.MPI`, that is a Python type defined and PyType_Init'ed in `mpi4py.MPI` at import time, and it is not available in `mpi4py.libmpi`, which is just pure extern C MPI stuff.

Folks, if you want to get rid of the compile-time mpi4py dependency (and then maybe runtime dependency if you copy over the `libmpi.pxd` file from mpi ), then you have to:

1. Remove the `cimport` using `mpi4py.MPI`
2. Rewrite `set_fapl_mpio()` something like this:

```
cdef extern from "Python.h":
    ctypedef size_t Py_uintptr_t

def set_fapl_mpio(self, comm, info=None):
    from mpi4py import MPI
    if info is None: info = MPI.INFO_NULL  # info is usually optional!
    assert isinstance(comm, MPI.Intracomm)  # XXX do better!
    assert isinstance(info, MPI.Info)  # XXX do better!
    cdef Py_uintptr_t _comm = MPI._handleof(comm)
    cdef Py_uintptr_t _info = MPI._handleof(info)
    H5Pset_fapl_mpio(self.id, <MPI_Comm>_comm, <MPI_Info>_info)
```
Now mpi4py is no longer a runtime dependency of `h5py`, and it will be imported on-demand. If you do not like the import every time, maybe you can use a private cdef global var initialized to `None` to cache the `MPI` module object.

PS: Be aware that the `_handleof()` utility routine may not be available in older versions of mpi4py previous to 2.0. Do you really need to support that? If yes, Maybe you can still use Cython's conditional compilation with `IF`.

takluyver: Yup, that's roughly the conclusion I was coming to as well - we only use `Comm` and `Info` as argument types for a couple of methods, and we can just use runtime type checking there.

It looks like mpi4py 2.0 came out in October 2015, so I think we can reasonably rely on that, unless there's some specific reason people might prefer to keep using the 1.x series for a long time.

@boegel do you want to have a go at implementing @dalcinl's suggestion? It might be easiest to do it in a new PR, since I think it should make the other changes unnecessary.

boegel: I can take a stab at it, but like I mentioned I'm not very familiar with Cython.

I'll try and apply the patch proposed by @dalcinl, see if that fixes the runtime issue I'm interested in fixing, and whether the `h5py` tests approve of the change.

It makes sense to tackle that in a new PR indeed, since it's a different approach.

I'll probably have some questions along the way, but I can put those forward in the new PR as they pop up.

takluyver: @boegel - are you still interested in having a go at that? Or would you rather I picked it up>

boegel: I wouldn't mind looking into it, but I can't make any promises as to when I will get around to it (seems unlikely in the coming days tbh).

If you have a go at it, I'm happy to test what you come up with though, keep me posted!

takluyver: OK, I've had a go in #1552. The MPI tests on CI are passing again, but I haven't tested it beyond that, so having someone who actually uses MPI try it would be great.

boegel:
superseded by #1552, so closing this...

PR #1552, opened 2020-05-19
Avoid importing mpi4py.MPI on import

takluyver:
Follow on from #1544, incorporating @dalcinl's suggestion. Hopefully this solves various strange behaviours in MPI builds.

This moves the minimum requirement for mpi4py to 2.0, released October 2015. We're only testing with the latest version (currently 3.0.3) in any case.


h5py/h5p.pyx
```diff
@@ -1197,16 +1192,20 @@ cdef class PropFAID(PropInstanceID):
 
     IF MPI:
         @with_phil
-        def set_fapl_mpio(self, Comm comm not None, Info info not None):
+        def set_fapl_mpio(self, comm, info):
```
dalcinl, line 1195: IMHO, this should use a default `info=None` which should be translated to `MPI_INFO_NULL`. Anyway, maybe that should be discussed/done in another PR.

takluyver: That makes sense, but it's not high priority - users wouldn't usually use this API directly, and the higher-level API already sets a default:

https://github.com/h5py/h5py/blob/2660313fcfdb927303d96848014a639b6230abb0/h5py/_hl/files.py#L43-L49

takluyver:
BTW @dalcinl this now uses the 'private' functions `_handleof` and `_addressof`. Is it safe to rely on those as part of mpi4py's API? Or is there a better way to get to the underlying MPI objects without cimport-ing `mpi4py.MPI`?

dalcinl:
> Is it safe to rely on those as part of mpi4py's API?

Yes, it is. These routines are indeed "private", but not because they are an internal implementation detail. They are kind of "hidden" utility functions for library implementors who really know what they are doing (e.g. h5py) to do exactly the kind of tricks done in this PR, that is, make third-party extension modules able to use mpi4py as an OPTIONAL runtime dependency.

PS: Maybe I should really use attributes for this (e.g. `comm.handle`). But if I ever do that, I'm not going to remove `_handleof()`, so h5py can safely rely on them.

h5py/h5p.pyx
```diff
@@ -1197,16 +1193,20 @@ cdef class PropFAID(PropInstanceID):
...
             """ (Comm comm, Info info)
 
             Set MPI-I/O Parallel HDF5 driver.
 
             Comm: An mpi4py.MPI.Comm instance
             Info: An mpi4py.MPI.Info instance
             """
-            H5Pset_fapl_mpio(self.id, comm.ob_mpi, info.ob_mpi)
-
+            from mpi4py import MPI
```
dalcinl: Look at the failures on Travis. `MPI` is also a `DEF` (Cython preprocessor) variable.

Maybe you should do `import mpi4py.MPI`, or `from mpi4py.MPI import Comm, Info, _handleof`

takluyver:
Thanks @dalcinl for your help. The tests are passing now, at least.

takluyver:
Ping @boegel - I'm still hoping you'll be able to try with this PR to see if it solves your issue. :-)

boegel:
> Ping @boegel - I'm still hoping you'll be able to try with this PR to see if it solves your issue. :-)

Sorry for the radio silence, I'll try to get around to testing this ASAP...

takluyver:
Thanks! No rush, I just wanted to make sure it didn't drop off the radar completely.

boegel:
@takluyver Just took this for a quick spin: the `import h5py` indeed doesn't trigger an `MPI_Init` anymore (I can easily tell, since before it was triggering a crash because triggering an `MPI_Init` without using `mpirun` results in a hard crash in the latest Intel MPI 2019 update 7, I believe that's called "progress").

I've also tested the [Parallel HDF5 example](https://docs.h5py.org/en/stable/mpi.html#using-parallel-hdf5-from-h5py), which still works like a charm.

So, lgtm!

takluyver:
Thanks @boegel!

I'll give this a couple of days in case anyone else wants to take a look, and then I'll merge it if there are no objections.

scopatz:
I am not an mpi4py expert, but the code here seems fine

PR #1567, opened 2020-06-13
Read HDF5 strings as bytes, add dataset.asstr() wrapper to get strings

takluyver:
This is the next - and possibly more controversial - part of #1338: what to do when reading HDF5 string data.

With this PR, we'd always return (arrays of) bytes when you slice a dataset - either numpy bytes for fixed length strings, or Python bytes objects for variable length strings. If you want `str` objects, you would get them through the `dset.asstr()` wrapper (could also be called `dset.decode()`, as discussed in #1338). Or get the bytes and decode them as needed.

Advantages:

- More consistent: previously, HDF5 strings were converted to different Python types depending on their encoding & size (variable or not). With this, they are always represented as bytes - there's still a difference between fixed size (numpy) bytes and Python bytes objects, but they're conceptually similar and have similar methods (e.g. `b.decode()`).
  - `dset.asstr()` provides a consistent way to get str for all of these possibilities
- Read any data: Strings in the file may not be valid UTF-8, even if the datatype says they are. It's not easy to read such data with h5py at present, because it's always trying to decode them.
- API design: It's easier & more efficient to handle decoding as an extra layer on top of an API that gives you bytes, than it is to work around a core API that tries to be helpful and gives you unicode when you want bytes.

Disadvantages:

- `str` is the idiomatic way to represent text in Python, so it will feel weird that HDF5 strings aren't converted to that by default.
- ~Strings in attributes are read as bytes, and there's no equivalent of `dset.asstr()`, because `.attrs['x']` reads and converts the attribute immediately.~ (Changed, see below)

### DIFF
--- a/h5py/_conv.pyx
+++ b/h5py/_conv.pyx
@@ -27,7 +27,6 @@ from numpy cimport npy_intp, NPY_WRITEABLE, NPY_C_CONTIGUOUS, NPY_OWNDATA
 cnp._import_array()
 
 from cpython.object cimport PyObject
-from cpython.unicode cimport PyUnicode_DecodeUTF8
 from cpython.ref cimport Py_INCREF, Py_XDECREF, Py_XINCREF
 
 cdef PyObject* Py_None = <PyObject*> None
@@ -134,26 +133,16 @@ cdef int conv_vlen2str(void* ipt, void* opt, void* bkg, void* priv) except -1:
         char** buf_cstring = <char**>ipt
         PyObject* tmp_object
         bytes tmp_bytes
-        unicode tmp_unicode
         conv_size_t *sizes = <conv_size_t*>priv
         char* buf_cstring0
 
     buf_cstring0 = buf_cstring[0]
 
-    # When reading we identify H5T_CSET_ASCII as a byte string and
-    # H5T_CSET_UTF8 as a utf8-encoded unicode string
-    if sizes.cset == H5T_CSET_ASCII:
-        if buf_cstring0 == NULL:
-            tmp_bytes =  b""
-        else:
-            tmp_bytes = buf_cstring0 # Let cython converts char* -> bytes for us
-        tmp_object = <PyObject *>tmp_bytes
-    elif sizes.cset == H5T_CSET_UTF8:
-        if buf_cstring0 == NULL:
-            tmp_unicode =  u""
-        else:
-            tmp_unicode = PyUnicode_DecodeUTF8(buf_cstring0, strlen(buf_cstring0), NULL)
-        tmp_object = <PyObject *>tmp_unicode
+    if buf_cstring0 == NULL:
+        tmp_bytes =  b""
+    else:
+        tmp_bytes = buf_cstring0 # Let cython converts char* -> bytes for us
+    tmp_object = <PyObject *>tmp_bytes
 
     # Since all data conversions are by definition in-place, it
     # is our responsibility to free the memory used by the vlens.
--- a/h5py/_hl/attrs.py
+++ b/h5py/_hl/attrs.py
@@ -76,6 +76,13 @@ class AttributeManager(base.MutableMappingHDF5, base.CommonStateObject):
         arr = numpy.ndarray(shape, dtype=dtype, order='C')
         attr.read(arr, mtype=htype)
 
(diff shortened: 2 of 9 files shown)


aragilar:
I'm not entirely happy with the bytes-from-attrs change, given we've always tried to return unicode from attributes (the bytes-from-datasets change I'm fine with, given how datasets interact with numpy), and I suspect strings are much more commonly stored in attributes than in datasets (hence having a larger breaking change). Could we decode attrs as UTF-8 with surrogate escapes maybe (inside the __getitem__ on the high-level API)?

takluyver:
Yeah, returning bytes for string attributes does feel a bit awkward, and I can see the case for decoding them with surrogate-escaping.

My uncertainties are all around where we put the boundaries:

1. Should we do this for all string types? Only vlen strings? Or only vlen UTF-8 strings?
2. Should it be only scalar string attributes (individual strings) or also arrays of strings?
3. Do we try to parallel it in datasets for consistency? Or treat attributes specially?

My current preferences are 1) vlen strings, regardless of encoding (which I don't think means much in HDF5), but not fixed length strings 2) same for arrays, 3) no, only for attributes - use `.asstr()` for datasets.

takluyver:
I've implemented it with my suggested options from the previous comment.

Decoding strings in attributes regardless of the encoding recorded in the file is a change from h5py 2.x (which only decoded those tagged UTF-8), so is that justified? My reasoning is that it's a string datatype, even if labelled ASCII. Decoding only UTF-8 fitted well with Python 2, where the str/bytes type was frequently used for text in any ASCII-compatible encoding. In Python 3, bytes represent binary data, not text, so we don't have *two types of string* to return data as.

(Of course, this suggests that we should always try to decode all strings. See my initial post on why I don't think that's the best option for datasets. It's a finer call for fixed-length strings in attributes, but if the fixed size is an important detail, I think that should be preserved, which means leaving it as bytes)

Why decode as UTF-8 even when it's tagged ASCII? If it's valid ASCII, decoding as UTF-8 gives the same answer. If it's UTF-8 but *not* ASCII, decoding as UTF-8 is probably right. If it's neither, most of the non-ASCII bytes will be surrogate-escaped either way, except when a sequence of bytes happens to be valid as UTF-8.

takluyver:
Ping @aragilar @tacaswell @scopatz - what do you think of this set of options? I'm trying to strike a balance between consistency between cases, choices that seem 'natural', and not gratuitously breaking existing interfaces.

- Datasets:
  - HDF5 fixed-length strings read as numpy byte strings (e.g. `np.dtype('S10')`)
  - HDF5 variable length strings read as Python bytes objects
  - Either kind can be read as Python str objects using `dset.asstr()[:]`
- Attributes:
  - HDF5 fixed length strings read as numpy byte strings
  - HDF5 variable length strings read as Python str objects (decoded as UTF-8 with surrogateescape to avoid losing information)

scopatz:
I think it depends on what the goal is here.  If the goal is to match what HDF5 does, then I think reading a bytes or byte arrays makes the most sense

takluyver:
I guess we can decide what the goal is to some extent. h5py is "a Pythonic interface to the HDF5 binary data format" (from the docs) or "a thin, pythonic wrapper around the HDF5" (from the README). Those broad definitions give us a fair bit of room to decide what the priorities are.

As I see it, we're trying to fit in with three systems - HDF5, Python and Numpy. But they treat strings quite differently: Python (3) strings are made of unicode code points, whereas HDF5 strings are bytes tagged with an encoding (which may be wrong). Numpy has no native equivalent of HDF5's variable length string datatype, except arrays of Python objects.

So, do we want to be more faithful to HDF5 (exposing bytes to the user), more faithful to Python (strings are str), or some compromise position inbetween? The current proposal here is a compromise, but somewhat closer to HDF5 than the design in h5py 2.x. I'm also aiming for more consistency between similar situations.

scopatz:
I am fine with being closer to HDF5 FWIW

takluyver:
Thanks @scopatz !

I'll merge this in another week unless anyone objects.

aragilar:
I played with how strings on attributes behave, and I think @takluyver new changes fix the worries I had about returning bytes.

PR #1573, opened 2020-06-25
Add nbytes attribute to datasets

### LINKED ISSUE #1571: Support `nbytes` attribute from NumPy
Enhancement request: support the `nbytes` attribute from NumPy arrays.

Currently, in order to get the approximate size of a h5py dataset, you have to do something like:
```
dset.size * np.dtype(dset.dtype).itemsize
```
Where as with NumPy arrays, you can directly access this value by:
```
arr.nbytes
```
It'd be much more convenient if this could be wrapped into `h5py`'s datasets as well.
Thanks!

nitrocalcite:
Enhancement: fix #1571

### DIFF
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -574,6 +574,8 @@ Reference
         as datasets may be compressed when written or only partly filled with data.
         This value also does not include the array overhead, as it only describes the size of the data itself.
         Thus the real amount of RAM occupied by this dataset may be slightly greater.
+        
+        .. versionadded:: 3.0
 
     .. attribute:: ndim
 
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -574,7 +574,7 @@ Reference
         as datasets may be compressed when written or only partly filled with data.
         This value also does not include the array overhead, as it only describes the size of the data itself.
         Thus the real amount of RAM occupied by this dataset may be slightly greater.
-        
+
         .. versionadded:: 3.0
 
     .. attribute:: ndim
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -572,7 +572,7 @@ Reference
         Integer giving the total number of bytes required to load the full dataset into RAM (i.e. `dset[()]`).
         This may not be the amount of disk space occupied by the dataset,
         as datasets may be compressed when written or only partly filled with data.
-        Also note that this value does not include the array overhead, as it only describes the size of the data itself.
+        This value also does not include the array overhead, as it only describes the size of the data itself.
         Thus the real amount of RAM occupied by this dataset may be slightly greater.
 
     .. attribute:: ndim
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -569,9 +569,11 @@ Reference
 
     .. attribute:: nbytes
 
-        Integer giving the total number of bytes occupied by the dataset elements.
-        Note that this is approximate, as it does not include the dataset overhead.
-        It only describes the size of the data itself.
+        Integer giving the total number of bytes required to load the full dataset into RAM (i.e. `dset[()]`).
+        This may not be the amount of disk space occupied by the dataset,
+        as datasets may be compressed when written or only partly filled with data.
+        Also note that this value does not include the array overhead, as it only describes the size of the data itself.
+        Thus the real amount of RAM occupied by this dataset may be slightly greater.
 
     .. attribute:: ndim
 
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -17,6 +17,7 @@ NumPy operations like slicing, along with a variety of descriptive attributes:
   - **size** attribute
   - **ndim** attribute
   - **dtype** attribute
+  - **nbytes** attribute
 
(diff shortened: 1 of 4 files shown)


takluyver:
Thanks! There's another potential confusion I think we should try to steer people around: this may be totally different to the size of the data in the file, both because the dataset may not be fully allocated (you can create a large dataset and only fill part of it), and because the stored data may be compressed.

Perhaps it's most accurate to describe this as the amount of memory required to read the whole of the dataset (`dset[()]`).

h5py/_hl/dataset.py
```diff
@@ -396,6 +396,14 @@ def size(self):
             self._cache_props['size'] = size
         return size
 
+    @property
+    def nbytes(self):
+        """Numpy-style attribute giving the dataset size as the number of bytes"""
+        size = self.size
+        if size is None:  # if we are empty, return None
+            return None
+        return numpy.dtype(self.dtype).itemsize * size
```
takluyver, line 405: I think `self.dtype` should always be a numpy dtype object already.

h5py/_hl/dataset.py
```diff
@@ -396,6 +396,14 @@ def size(self):
             self._cache_props['size'] = size
         return size
 
+    @property
+    def nbytes(self):
+        """Numpy-style attribute giving the dataset size as the number of bytes"""
+        size = self.size
+        if size is None:  # if we are empty, return None
+            return None
```
takluyver, line 404: Or 0? It's nice to have a consistent return type where possible.

h5py/_hl/dataset.py
nitrocalcite:
```suggestion
        return self.dtype.itemsize * size
```

h5py/_hl/dataset.py
nitrocalcite: I was also thinking that, but I was concerned that it doesn't mimic `size`.
As in, NumPy's `size` & `nbytes` returns zero for empty datasets:
```
>>> np.empty((0,)).size
0
>>> np.empty((0,)).nbytes
0
```
But `h5py` returns None for empty
```
>>>h5py.Empty(dtype=np.uint16).size

```
Since I think of `nbytes` as "`size`, but scaled" I figured `None` would make sense.  It seems to me either option is logical but `None` is marginally more consistent.  

```suggestion
            return 0
```

takluyver: Numpy doesn't have anything equivalent to `h5py.Empty`, as far as I know: `np.empty((0,))` makes a 1D array of zero length ('empty' there means memory is allocated but not initialised). Both numpy and HDF5 can represent a 0D dataset with 1 entry, but for some reason HDF5 also has a way to store a 0D object with no entries. I don't really know what it's for.

So, no guidance from numpy. h5py used `shape=None` as a convenient way to create an empty dataset (distinguishing it from `shape=()` which makes a scalar). So it made sense for `dset.shape` to give None back. And I guess `dset.size` followed from there.

I'd be inclined not to extend the special case to `nbytes`, but maybe we should get input from other maintainers. If empty dataset support was new, I might also be in favour of making `size` 0. But changing it after the fact is a tougher sell.

docs/high/dataset.rst
```diff
@@ -566,6 +567,14 @@ Reference
 
         Integer giving the total number of elements in the dataset.
 
+    .. attribute:: nbytes
+
+        Integer giving the total number of bytes required to load the full dataset into RAM (i.e. `dset[()]`).
+        This may not be the amount of disk space occupied by the dataset,
+        as datasets may be compressed when written or only partly filled with data.
+        Also note that this value does not include the array overhead, as it only describes the size of the data itself.
```
takluyver, line 575:
```suggestion
        This value also does not include the array overhead, as it only describes the size of the data itself.
```

Something I picked up [from Wikipedia](https://en.wikipedia.org/wiki/Wikipedia:It_should_be_noted) years ago - "note that" can almost always be removed with no change in the meaning.

h5py/_hl/dataset.py
takluyver: OK, as none of the other maintainers have commented on this, I'll call it in favour of 0. We've already got other ways to distinguish an empty dataspace from a regular one where one dimension is 0, so I think consistently returning an int here is better.

nitrocalcite:
I think this covers all the things we've discussed...anything else that should be done before this can be merged?

Thanks for working with me on this!

docs/high/dataset.rst
```diff
@@ -566,6 +567,14 @@ Reference
 
         Integer giving the total number of elements in the dataset.
 
+    .. attribute:: nbytes
+
+        Integer giving the total number of bytes required to load the full dataset into RAM (i.e. `dset[()]`).
+        This may not be the amount of disk space occupied by the dataset,
+        as datasets may be compressed when written or only partly filled with data.
+        This value also does not include the array overhead, as it only describes the size of the data itself.
+        Thus the real amount of RAM occupied by this dataset may be slightly greater.
```
takluyver, line 576:
```suggestion
        Thus the real amount of RAM occupied by this dataset may be slightly greater.
        
        .. versionadded:: 3.0
```

takluyver:
Thanks!

PR #1574, opened 2020-06-26
Fix writing Python integers as uint64

### LINKED ISSUE #1547: uint64 dataset writing incorrect values
I noticed some of my data being changed while attempting to write a uint64 dataset to an H5 file.
Example of trying to write the same dataset in two ways:

```python
import h5py
import numpy as np

first = 9344170403720592284
second = 3671708271
arr = [first, second]

with h5py.File(r"C:\test.h5", 'w') as h5_file:
    name = "test1"
    h5_file.create_dataset(name, data=arr, dtype=np.uint64)
    bad = h5_file[name][0]
    print(bad)

    name = "test2"
    np_arr = np.array(arr, dtype=np.uint64)  # manually create NumPy array
    h5_file.create_dataset(name, data=np_arr)
    good = h5_file[name][0]
    print(good)
```

Output:
```
9344170403720591360
9344170403720592284
```

In "test1", the last byte of that "bad" uint64 value is being set to 0.
For "test2", I am manually constructing the `np.uint64` array and then passing it into the `create_dataset()` method - this works as expected...though I was expecting "test1" to basically be doing the same thing.

Interestingly,  omitting the "second" value from the list results in the correct output.

```python
import h5py
import numpy as np

first = 9344170403720592284
second = 3671708271
arr = [first]    # no second value

with h5py.File(r"C:\test.h5", 'w') as h5_file:
    # ...same as above
```

Output:
```
9344170403720592284
9344170403720592284
```
Is this all expected behavior?

```
h5py    2.10.0
HDF5    1.10.5
Python  3.8.2 (tags/v3.8.2:7b3ab59, Feb 25 2020, 23:03:10) [MSC v.1916 64 bit (AMD64)]
sys.platform    win32
numpy   1.18.4
```

takluyver:
Closes #1547.

I discovered that the same flaw appeared in writing to an existing dataset, and with creating & modifying attributes.

This should fix all four cases, by not relying on numpy's dtype inference when we have a target dtype to aim for. But if a numpy array is passed in, we still leave the conversion to HDF5 rather than asking numpy to do it, which would create a temporary copy in memory.

### DIFF
--- a/h5py/_hl/attrs.py
+++ b/h5py/_hl/attrs.py
@@ -222,12 +222,15 @@ class AttributeManager(base.MutableMappingHDF5, base.CommonStateObject):
             if not name in self:
                 self[name] = value
             else:
-                value = numpy.asarray(value, order='C')
-
                 attr = h5a.open(self._id, self._e(name))
 
                 if is_empty_dataspace(attr):
                     raise IOError("Empty attributes can't be modified")
+
+                # If the input data is already an array, let HDF5 do the conversion.
+                # If it's a list or similar, don't make numpy guess a dtype for it.
+                dt = None if isinstance(value, numpy.ndarray) else attr.dtype
+                value = numpy.asarray(value, order='C', dtype=dt)
 
                 # Allow the case of () <-> (1,)
                 if (value.shape != attr.shape) and not \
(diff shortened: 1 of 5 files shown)


scopatz:
This looks very nice, thanks @takluyver!

takluyver:
Thanks @scopatz! I'll merge this in a couple of days if there are no more comments.

If you have another moment to spare, I'm looking for input on the questions of string types in #1567. :slightly_smiling_face:

Issue #1580, opened 2020-07-09
SWMR querying: AttributeError: 'File' object has no attribute '_swmr_mode'


miketrumpis:
Hi h5py devs. Thank you for the incredible software!

First off:

```
Summary of the h5py configuration
---------------------------------

h5py    2.10.0
HDF5    1.10.4
Python  3.8.2 (default, May 23 2020, 11:49:39) 
[Clang 11.0.0 (clang-1100.0.33.8)]
sys.platform    darwin
sys.maxsize     9223372036854775807
numpy   1.18.4
```

I am trying to find out whether and how to re-open an HDF5 file for read-only access in a subprocess: i.e. in SWMR mode or just in simple read mode. I am trying to get that info by querying a ``h5py._hl.dataset.Dataset`` object about the status of its parent file.

So a brief example

```python
import numpy as np
import h5py as h5py

h5 = h5py.File('foo.h5', 'w', libver='latest')
h5['array'] = np.ones((10, 20))
h5.swmr_mode = True
array = h5['array']
array.file.swmr_mode
```

However, I am getting this attribute error.

```python
AttributeError: 'File' object has no attribute '_swmr_mode'
```

And the same result if the file was opened again in 'r' mode.

I don't understand the difference between ``h5`` and ``array.file`` (they are equal but not identical). But I do see in the code that the attribute in question is only defined conditionally in the constructor (alternatively, in the property setter). Should it be defaulted outside the conditional?

https://github.com/h5py/h5py/blob/2881b7cc1d24989e9d8d54814c4855522f7a9607/h5py/_hl/files.py#L428-L431

miketrumpis:
For clarity, the attribute error is coming from the "swmr_mode" property getter.

miketrumpis:
Related: setting ``h5.swmr_mode = True`` changes ``h5.mode``

```python
>>> h5 = h5py.File('foo.h5', 'w', libver='latest')                                                                                
>>> h5['array'] = np.ones((10, 20))                                                                                               
>>> h5.mode                                                                                                                       
'r+'
>>> h5.swmr_mode = True                                                                                                           
>>> h5.mode is None                                                                                                               
True
```

And the same None value occurs with this read-mode construction: ``h5 = h5py.File('foo.h5', 'r', libver='latest', swmr=True)``

Is that proper behavior? Makes it difficult to know whether a dataset can be written to.

takluyver:
[...]

PR #1582, opened 2020-07-13
Addresses file mode and SWMR mode inconsistencies pointed out in gh-1580

miketrumpis:
* Remove _swmr_mode private attribute
* Both FID and group/dataset "file" attribute have consistent modes
  before and after swmr_mode setter
* Both FID and group/dataset "file" attribute have consistent
  swmr_mode getters

### DIFF
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -254,14 +254,10 @@ class File(Group):
     @with_phil
     def mode(self):
         """ Python mode used to open file """
+        write_intent = h5f.ACC_RDWR
         if swmr_support:
-            intent = self.id.get_intent() & (h5f.ACC_SWMR_WRITE | h5f.ACC_RDWR)
-        else:
-            intent = self.id.get_intent() & h5f.ACC_RDWR
-        if 0 < intent:
-            return 'r+'
-        else:
-            return 'r'
+            write_intent |= h5f.ACC_SWMR_WRITE
+        return 'r+' if self.id.get_intent() & write_intent else 'r'
 
     @property
     @with_phil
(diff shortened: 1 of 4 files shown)


h5py/_hl/files.py
```diff
@@ -254,8 +254,14 @@ def driver(self):
...
-        return {h5f.ACC_RDONLY: 'r',
-                h5f.ACC_RDWR: 'r+'}.get(self.id.get_intent())
+        if swmr_support:
+            intent = self.id.get_intent() & (h5f.ACC_SWMR_WRITE | h5f.ACC_RDWR)
+        else:
+            intent = self.id.get_intent() & h5f.ACC_RDWR
+        if 0 < intent:
+            return 'r+'
+        else:
+            return 'r'
```
takluyver, line 264: I think we can condense this a bit:

```suggestion
        write_intent = h5f.ACC_RDWR
        if swmr_support:
            write_intent |= h5f.ACC_SWMR_WRITE
        return 'r+' if self.id.get_intent() & write_intent else 'r'
```

h5py/_hl/files.py
```diff
@@ -288,19 +294,21 @@ def atomic(self, value):
 
     if swmr_support:
         @property
+        @with_phil
         def swmr_mode(self):
             """ Controls single-writer multiple-reader mode """
-            return self._swmr_mode
+            return 0 < (self.id.get_intent() & (h5f.ACC_SWMR_READ | h5f.ACC_SWMR_WRITE))
```
takluyver, line 300:
```suggestion
            return bool(self.id.get_intent() & (h5f.ACC_SWMR_READ | h5f.ACC_SWMR_WRITE))
```

Seems a bit clearer. The comparison is backwards to my way of thinking. ;-)

miketrumpis:
I like the changes. Now getters / setters are defined even for old hdf5 versions. So in addition to ditching the AttributeError, setting swmr_mode raises (RuntimeError) if the library doesn't support SWMR.

h5py/_hl/files.py
```diff
@@ -286,21 +288,23 @@ def atomic(self, value):
...
+    def swmr_mode(self, value):
+        # pylint: disable=missing-docstring
+        if swmr_support:
             if value:
                 self.id.start_swmr_write()
-                self._swmr_mode = True
             else:
                 raise ValueError("It is not possible to forcibly switch SWMR mode off.")
+        else:
+            raise RuntimeError('SWMR support is not available for version {}.{}.{}.'.format(*hdf5_version))
```
takluyver, line 307:
```suggestion
            raise RuntimeError('SWMR support is not available in HDF5 version {}.{}.{}.'.format(*hdf5_version))
```

Just to make it totally clear that the version number refers to HDF5, not h5py.

h5py/tests/test_dataset_swmr.py
```diff
@@ -36,9 +36,17 @@ def test_flush_raises(self):
...
+        """ The SWMR getter should just be False
+        """
+        assert not self.f.swmr_mode
+
+    def test_set_swmr_mode_raises(self):
+        """ If the SWMR feature is not available, setting swmr_mode = True
+        should raise a RuntimeError
+        """
+        with self.assertRaises(RuntimeError):
+            self.f.swmr_mode = True
```
takluyver, line 49: Let's assert that the attribute remains False after this as well.

miketrumpis:
Ok--pending version checks. Do you want me to squash commits and PR again?

takluyver:
No, don't worry about that, I think the history is OK. Can I get you to add a news snippet about these changes, though? See the information at https://regro.github.io/rever-docs/news.html .

miketrumpis:
Should be done.

takluyver:
Thanks!

Issue #1584, opened 2020-07-15
Test failure with  h5py version 2.10.0


dota17:
To assist reproducing bugs, please include the following:
 * Operating System (e.g. Windows 10, MacOS 10.11, Ubuntu 16.04.2 LTS, CentOS 7)
> Ubuntu 18.04
 * Python version (e.g. 2.7, 3.5)
> 3.6.8
 * Where Python was acquired (e.g. system Python on MacOS or Linux, Anaconda on
   Windows)
>  Linux

```
Summary of the h5py configuration
---------------------------------

h5py    2.10.0
HDF5    1.10.4
Python  3.6.8 (default, Jan 14 2019, 11:02:34)
[GCC 8.0.1 20180414 (experimental) [trunk revision 259383]]
sys.platform    linux
sys.maxsize     9223372036854775807
numpy   1.18.0

```
 * The full traceback/stack trace shown (if it appears)

```
================================================================================================== FAILURES ==================================================================================================
_________________________________________________________________________________________ TestFileOpen.test_default __________________________________________________________________________________________

self = <h5py.tests.test_file.TestFileOpen testMethod=test_default>

    def test_default(self):
        """ Default semantics in the presence or absence of a file """
        fname = self.mktemp()

        # No existing file; error
        with pytest.raises(OSError):
            with File(fname):
                pass


        # Existing readonly file; open read-only
        with File(fname, 'w'):
            pass
        os.chmod(fname, stat.S_IREAD)
        # Running as root (e.g. in a docker container) gives 'r+' as the file
        # mode, even for a read-only file.  See
        # https://github.com/h5py/h5py/issues/696
        exp_mode = 'r+' if os.stat(fname).st_uid == 0 and platform != "win32" else 'r'
        try:
            with File(fname) as f:
                self.assertTrue(f)
>               self.assertEqual(f.mode, exp_mode)
E               AssertionError: 'r' != 'r+'
E               - r
E               + r+
E               ?  +

h5py/tests/test_file.py:56: AssertionError
```

I've noticed there is a discussion in #696, and fixed in #724,  but the same problem still appear now in my computer.
It's really annoying me. Do you have any suggestions?  Thanks in advance!
Cc: @takluyver @tacaswell

takluyver:
It looks like you've got the opposite problem - #724 made it expect read-write mode when opening the file as root, but you're getting read-only mode.

It should be gone in master anyway, because the default mode is changing to read-only anyway.

dota17:
[...]

PR #1585, opened 2020-07-15
Fix default file mode in test as root

dota17:
Close: #1584
@takluyver  please take a look.

### DIFF
--- a/.authors.yml
+++ b/.authors.yml
@@ -124,6 +124,13 @@
   num_commits: 2
   first_commit: 2016-11-08 18:01:54
   github: cgohlke
+- name: Chen
+  email: chenguopingdota@163.com
+  aliases:
+  - chenguoping
+  num_commits: 1
+  first_commit: 2020-07-15 20:34:45
+  github: dota17
 - name: Carlos Pascual
   email: cpascual@cells.es
   num_commits: 2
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks. The CI failures is asking for your details to be in `.authors.yml` - can you check the message and add yourself?

dota17:
> Thanks. The CI failures is asking for your details to be in `.authors.yml` - can you check the message and add yourself?

Done 😃

takluyver:
I think it now wants `chenguoping` in your alias list. Sorry about this, I don't quite understand what rever needs, but we do need to get it passing, otherwise it will fail on master and all new PRs.

dota17:
> I think it now wants `chenguoping` in your alias list. Sorry about this, I don't quite understand what rever needs, but we do need to get it passing, otherwise it will fail on master and all new PRs.

Ok, never mind, I understand. I will have a try again.

takluyver:
It must want both your name on Github ('Chen') and the name you're committing as ('chenguoping').

@scopatz can we simplify our use of rever at all? It's nice to credit people by their preferred name, but I feel like it makes life unnecessarily difficult for people making a small contribution like this.

dota17:
@takluyver well, finally, it passed! 😄 
I think we can add a note in `contributing.rst` to inform people who opened their first PR in h5py how to do this kind of thing.
and I'm willing to do this.

takluyver:
Thanks for sticking with it! Yes, that would be very welcome if you want to improve that.

PR #1590, opened 2020-07-22
Clean some code

dota17:
As the title described, this little patch just clean some code.
Cc: @takluyver

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -100,12 +100,10 @@ def make_new_dset(parent, shape=None, dtype=None, data=None, name=None,
             raise TypeError("Conflict in compression options")
         compression_opts = compression
         compression = 'gzip'
-    if dcpl is None:
-        plist = h5p.create(h5p.DATASET_CREATE)
-    else:
-        plist = dcpl
-    dcpl = filters.fill_dcpl(plist, shape, dtype, chunks, compression, compression_opts,
-                             shuffle, fletcher32, maxshape, scaleoffset, external)
+    dcpl = filters.fill_dcpl(
+        dcpl or h5p.create(h5p.DATASET_CREATE), shape, dtype,
+        chunks, compression, compression_opts, shuffle, fletcher32,
+        maxshape, scaleoffset, external)
 
     if fillvalue is not None:
         fillvalue = numpy.array(fillvalue)
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -100,10 +100,12 @@ def make_new_dset(parent, shape=None, dtype=None, data=None, name=None,
             raise TypeError("Conflict in compression options")
         compression_opts = compression
         compression = 'gzip'
-    dcpl = filters.fill_dcpl(
-        dcpl or h5p.create(h5p.DATASET_CREATE), shape, dtype,
-        chunks, compression, compression_opts, shuffle, fletcher32,
-        maxshape, scaleoffset, external)
+    if dcpl is None:
+        plist = h5p.create(h5p.DATASET_CREATE)
+    else:
+        plist = dcpl
+    dcpl = filters.fill_dcpl(plist, shape, dtype, chunks, compression, compression_opts,
+                             shuffle, fletcher32, maxshape, scaleoffset, external)
 
     if fillvalue is not None:
         fillvalue = numpy.array(fillvalue)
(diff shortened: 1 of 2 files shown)


h5py/_hl/dataset.py
```diff
@@ -100,10 +100,12 @@ def make_new_dset(parent, shape=None, dtype=None, data=None, name=None,
             raise TypeError("Conflict in compression options")
         compression_opts = compression
         compression = 'gzip'
-    dcpl = filters.fill_dcpl(
-        dcpl or h5p.create(h5p.DATASET_CREATE), shape, dtype,
```
takluyver, line 104: I actually quite like the `x or make_default()` pattern for things like this.

h5py/_hl/files.py
```diff
@@ -141,25 +141,24 @@ def make_fapl(driver, libver, rdcc_nslots, rdcc_nbytes, rdcc_w0, **kwds):
...
-                'page': h5f.FSPACE_STRATEGY_PAGE,
-                'aggregate': h5f.FSPACE_STRATEGY_AGGR,
-                'none': h5f.FSPACE_STRATEGY_NONE
-            }
-            fs_strat_num = strategies.get(fs_strategy, -1)
-            if fs_strat_num == -1:
-                raise ValueError("Invalid file space strategy type")
-
-            plist.set_file_space_strategy(fs_strat_num, fs_persist, fs_threshold)
+    plist = h5p.create(h5p.FILE_CREATE)
```
takluyver, line 144: This does we'll always call into HDF5 to create a property list, even when we don't need to. That's probably not a problem for file creation, which you don't do very often, but in other cases we might want to check if it affects performance.

h5py/_hl/dataset.py
dota17: Mm-hmm, then I will revert this part.

h5py/_hl/files.py
dota17: Then I can put it below if and elif statement

h5py/_hl/files.py
```diff
@@ -141,25 +141,25 @@ def make_fapl(driver, libver, rdcc_nslots, rdcc_nbytes, rdcc_w0, **kwds):
...
-            if fs_strat_num == -1:
-                raise ValueError("Invalid file space strategy type")
-
-            plist.set_file_space_strategy(fs_strat_num, fs_persist, fs_threshold)
+        plist.set_link_creation_order(
+            h5p.CRT_ORDER_TRACKED | h5p.CRT_ORDER_INDEXED)
+        plist.set_attr_creation_order(
+            h5p.CRT_ORDER_TRACKED | h5p.CRT_ORDER_INDEXED)
+    elif fs_strategy:
+        plist = h5p.create(h5p.FILE_CREATE)
```
takluyver, line 151: This isn't quite right - it should be possible to set both `track_order` and `fs_strategy` - they're not mutually exclusive.

I think for file creation your previous solution - creating the property list every time - is probably OK. I just wanted to note that it might have a small overhead.

dota17: Oh, I thought you don't like that. 
Anyway, they both have small overhead, but there are always trade-offs. I prefer the previous solution too.

h5py/_hl/files.py
```diff
@@ -141,25 +141,24 @@ def make_fapl(driver, libver, rdcc_nslots, rdcc_nbytes, rdcc_w0, **kwds):
...
-                raise ValueError("Invalid file space strategy type")
-
-            plist.set_file_space_strategy(fs_strat_num, fs_persist, fs_threshold)
+    plist = h5p.create(h5p.FILE_CREATE)
+    if track_order:
+        plist.set_link_creation_order(
+            h5p.CRT_ORDER_TRACKED | h5p.CRT_ORDER_INDEXED)
+        plist.set_attr_creation_order(
+            h5p.CRT_ORDER_TRACKED | h5p.CRT_ORDER_INDEXED)
+    elif fs_strategy:
```
takluyver, line 150:
```suggestion
    if fs_strategy:
```

These two options aren't mutually exclusive.

h5py/_hl/files.py
```diff
@@ -141,25 +141,24 @@ def make_fapl(driver, libver, rdcc_nslots, rdcc_nbytes, rdcc_w0, **kwds):
...
+            'page': h5f.FSPACE_STRATEGY_PAGE,
+            'aggregate': h5f.FSPACE_STRATEGY_AGGR,
+            'none': h5f.FSPACE_STRATEGY_NONE
+        }
+        fs_strat_num = strategies.get(fs_strategy, -1)
+        if fs_strat_num == -1:
+            raise ValueError("Invalid file space strategy type")
+
+        plist.set_file_space_strategy(fs_strat_num, fs_persist, fs_threshold)
     else:
```
takluyver, line 162: No, sorry, this doesn't work either, because now the `else` is only attached to the second `if`, so it does the wrong thing if `track_order` is set but not `fs_strategy`.

At this point I think we should just leave this function as it was. We've proved that it's easy to get it wrong when trying to simplify it.

dota17: Oops, you are right.

takluyver:
Thanks for your patience with this :-)

PR #1595, opened 2020-07-27
Fix error assigning to a fancy-indexed selection

takluyver:
Closes #1593.

I got a condition back to front in #1526.

PR #1004, opened 2018-02-20
ENH,WIP: Add support for including cython in coverage analysis

aragilar:
Adds support for optionally building the h5py cython code with coverage support, WIP until coverage is setup for the rest of h5py.

PR #1598, opened 2020-07-28
Add some testcases

dota17:
Boost the code coverage from 85% to 88%
Cc: @takluyver

### DIFF
--- a/h5py/tests/test_base.py
+++ b/h5py/tests/test_base.py
@@ -14,6 +14,7 @@
 """
 
 from h5py import File
+from h5py._hl.base import is_hdf5, Empty
 from .common import ut, TestCase, UNICODE_FILENAMES
 
 import numpy as np
@@ -41,6 +42,78 @@ class TestName(BaseTest):
         grp = self.f.create_group(None)
         self.assertIs(grp.name, None)
 
+class TestParent(BaseTest):
+
+    """
+        test the parent group of the high-level interface objects
+    """
+
+    def test_object_parent(self):
+        # Anonymous objects
+        try:
+            grp = self.f.create_group(None)
+            bar_parent = grp.parent
+        except ValueError:
+            pass
+
+        # Named objects
+        grp = self.f.create_group("bar")
+        sub_grp = grp.create_group("foo")
+        parent = sub_grp.parent
+        self.assertEqual(r'<HDF5 group "/bar" (1 members)>', repr(parent))
+
+class TestMapping(BaseTest):
+
+    """
+        Test if the registration of Group, AttributeManager as a
+        Mapping behaves as expected
+    """
+
+    def setUp(self):
+        data = ('a', 'b')
+        self.f = File('foo.hdf5', 'w')
+        self.grp = self.f.create_group('bar')
+        self.attr = self.f.attrs.create('x', data)
+
+    def TearDown(self):
+        if self.f:
+            self.close()
+
+    def test_keys(self):
+        key_1 = self.f.keys()
+        self.assertEqual(r"<KeysViewHDF5 ['bar']>", repr(key_1))
+        key_2 = self.grp.keys()
+        self.assertEqual(r"<KeysViewHDF5 []>", repr(key_2))
+
+    def test_values(self):
+        value_1 = self.f.values()
+        self.assertEqual(r'ValuesViewHDF5(<HDF5 file "foo.hdf5" (mode r+)>)', repr(value_1))
+        value_2 = self.grp.values()
+        self.assertEqual(r'ValuesViewHDF5(<HDF5 group "/bar" (0 members)>)', repr(value_2))
+
+    def test_items(self):
+        item_1 = self.f.items()
+        self.assertEqual(r'ItemsViewHDF5(<HDF5 file "foo.hdf5" (mode r+)>)', repr(item_1))
+        item_2 = self.grp.items()
+        self.assertEqual(r'ItemsViewHDF5(<HDF5 group "/bar" (0 members)>)', repr(item_2))
+
+class TestFileType(BaseTest):
+
+    """
+        Test if a file is a HDF5 type
+    """
+
+    def test_is_hdf5(self):
+        filename = File("data.hdf5", "w").filename
+        fname = os.path.basename(filename)
+        fid = is_hdf5(fname)
+        self.assertTrue(fid)
+        # non-existing HDF5 file
+        filename = tempfile.mktemp()
+        fname = os.path.basename(filename)
+        fid = is_hdf5(fname)
+        self.assertFalse(fid)
+
 class TestRepr(BaseTest):
 
     """
@@ -67,6 +140,11 @@ class TestRepr(BaseTest):
         self.f['type'] = np.dtype('f')
         typ = self.f['type']
         self._check_type(typ)
+
+    def test_empty(self):
+        data = Empty(dtype='f')
+        self.assertNotEqual(Empty(dtype='i'), data)
+        self._check_type(data)
 
     @ut.skipIf(not UNICODE_FILENAMES, "Filesystem unicode support required")
     def test_file(self):
--- a/h5py/tests/test_base.py
+++ b/h5py/tests/test_base.py
@@ -64,7 +64,7 @@ class TestParent(BaseTest):
 class TestMapping(BaseTest):
 
     """
-        Test if the registration of Group, AttributeManager as a
+        Test if the registration of Group as a
         Mapping behaves as expected
     """
 
@@ -80,8 +80,6 @@ class TestMapping(BaseTest):
 
     def test_keys(self):
         key_1 = self.f.keys()
-        print(type(key_1))
-        print(dir(key_1))
         self.assertIsInstance(repr(key_1), str)
         key_2 = self.grp.keys()
         self.assertIsInstance(repr(key_2), str)
--- a/h5py/tests/test_base.py
+++ b/h5py/tests/test_base.py
@@ -98,20 +98,6 @@ class TestMapping(BaseTest):
         item_2 = self.grp.items()
         self.assertIsInstance(repr(item_1), str)
 
-class TestFileType(BaseTest):
-
-    """
-        Test if a file is a HDF5 type
-    """
-
-    def test_is_hdf5(self):
-        filename = File(self.mktemp(), "w").filename
-        fid = is_hdf5(filename)
-        self.assertTrue(fid)
-        # non-existing HDF5 file
-        filename = tempfile.mktemp()
-        fid = is_hdf5(filename)
-        self.assertFalse(fid)
 
 class TestRepr(BaseTest):
 
@@ -157,3 +143,12 @@ class TestRepr(BaseTest):
                 os.unlink(fname)
             except Exception:
                 pass
+
+def test_is_hdf5():
+    filename = File(tempfile.mktemp(), "w").filename
+    fid = is_hdf5(filename)
+    assert fid is True
+    # non-existing HDF5 file
+    filename = tempfile.mktemp()
+    fid = is_hdf5(filename)
+    assert fid is False
--- a/h5py/tests/test_base.py
+++ b/h5py/tests/test_base.py
@@ -50,17 +50,16 @@ class TestParent(BaseTest):
 
     def test_object_parent(self):
         # Anonymous objects
-        try:
-            grp = self.f.create_group(None)
-            bar_parent = grp.parent
-        except ValueError:
-            pass
+        grp = self.f.create_group(None)
+        # Parent of an anonymous object is undefined
+        with self.assertRaises(ValueError):
+            grp.parent
 
         # Named objects
         grp = self.f.create_group("bar")
         sub_grp = grp.create_group("foo")
-        parent = sub_grp.parent
-        self.assertEqual(r'<HDF5 group "/bar" (1 members)>', repr(parent))
+        parent = sub_grp.parent.name
+        self.assertEqual(parent, "/bar")
 
 class TestMapping(BaseTest):
 
@@ -81,21 +80,23 @@ class TestMapping(BaseTest):
 
     def test_keys(self):
         key_1 = self.f.keys()
-        self.assertEqual(r"<KeysViewHDF5 ['bar']>", repr(key_1))
+        print(type(key_1))
+        print(dir(key_1))
+        self.assertIsInstance(repr(key_1), str)
         key_2 = self.grp.keys()
-        self.assertEqual(r"<KeysViewHDF5 []>", repr(key_2))
+        self.assertIsInstance(repr(key_2), str)
 
     def test_values(self):
         value_1 = self.f.values()
-        self.assertEqual(r'ValuesViewHDF5(<HDF5 file "foo.hdf5" (mode r+)>)', repr(value_1))
+        self.assertIsInstance(repr(value_1), str)
         value_2 = self.grp.values()
-        self.assertEqual(r'ValuesViewHDF5(<HDF5 group "/bar" (0 members)>)', repr(value_2))
+        self.assertIsInstance(repr(value_2), str)
 
     def test_items(self):
         item_1 = self.f.items()
-        self.assertEqual(r'ItemsViewHDF5(<HDF5 file "foo.hdf5" (mode r+)>)', repr(item_1))
+        self.assertIsInstance(repr(item_1), str)
         item_2 = self.grp.items()
-        self.assertEqual(r'ItemsViewHDF5(<HDF5 group "/bar" (0 members)>)', repr(item_2))
+        self.assertIsInstance(repr(item_1), str)
 
 class TestFileType(BaseTest):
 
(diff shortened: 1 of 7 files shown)


h5py/tests/test_base.py
```diff
@@ -41,6 +42,78 @@ def test_anonymous(self):
...
+        item_2 = self.grp.items()
+        self.assertEqual(r'ItemsViewHDF5(<HDF5 group "/bar" (0 members)>)', repr(item_2))
+
+class TestFileType(BaseTest):
+
+    """
+        Test if a file is a HDF5 type
+    """
+
+    def test_is_hdf5(self):
```
takluyver, line 106: We now use pytest, so if you're familiar with it, you can write new tests as simple functions, rather than needing classes like this. There are some examples at the bottom of `test_dataset.py`.

(I don't particularly want to convert all the existing tests, but it's convenient for writing new tests)

h5py/tests/test_base.py
```diff
@@ -41,6 +42,78 @@ def test_anonymous(self):
...
+        filename = File("data.hdf5", "w").filename
```
takluyver, line 107: This should use a temp file rather than creating something in the cwd.

h5py/tests/test_dataset.py
```diff
@@ -165,6 +166,63 @@ def test_empty_create_via_Empty_class(self):
         self.f.create_dataset('foo', data=h5py.Empty(dtype='f'))
         self.assertTrue(is_empty_dataspace(self.f['foo'].id))
 
+    def test_create_incompatible_data(self):
+        data = self.f.create_dataset('dset', (3,))
+        with self.assertRaises(ValueError):
+            self.f.create_dataset('bar', shape=4, data=data)
```
takluyver, line 172: If you're just testing the difference in shape, can you make `data` a numpy array? It's slightly confusing to have it be another dataset, unless that's the point of the test.

h5py/tests/test_dataset.py
```diff
@@ -165,6 +166,63 @@ def test_empty_create_via_Empty_class(self):
...
+        dset = self.f.create_dataset("dset", (100,), dtype='int64')
+        source_dset = self.f.create_dataset("source_dset", (100,), dtype='int64')
+        dest_dset = self.f.create_dataset("dest_dset", (100,), dtype='int64')
+        empty_dset = self.f.create_dataset("edset", dtype='int64')
+
+        source_sel = sel.select((100,), ..., source_dset)
+        dest_sel = sel.select((100,), ..., dest_dset)
+        arr = np.zeros((100,), dtype='int32')
+
+        if arr.flags.c_contiguous:
```
takluyver, line 191: I don't think we need this check - a newly created array is always contiguous.

h5py/tests/test_dataset.py
```diff
@@ -165,6 +166,63 @@ def test_empty_create_via_Empty_class(self):
...
+    """
+
+    def test_read_direct(self):
+        dset = self.f.create_dataset("dset", (100,), dtype='int64')
+        source_dset = self.f.create_dataset("source_dset", (100,), dtype='int64')
+        dest_dset = self.f.create_dataset("dest_dset", (100,), dtype='int64')
+        empty_dset = self.f.create_dataset("edset", dtype='int64')
+
+        source_sel = sel.select((100,), ..., source_dset)
+        dest_sel = sel.select((100,), ..., dest_dset)
```
takluyver, line 188: You don't need a `dest_dset` here - your destination is an array, so `dataset` can be None.

h5py/tests/test_dataset.py
```diff
@@ -165,6 +166,63 @@ def test_empty_create_via_Empty_class(self):
...
+
+class TestReadDirectly(BaseDataset):
+
+    """
+        Feature: Read data directly from Dataset into a Numpy array
+    """
+
+    def test_read_direct(self):
+        dset = self.f.create_dataset("dset", (100,), dtype='int64')
+        source_dset = self.f.create_dataset("source_dset", (100,), dtype='int64')
```
takluyver, line 183: We shouldn't have separate datasets here. There's only one dataset involved in calling `read_direct` - the one it's reading from.

h5py/tests/test_dataset.py
```diff
@@ -165,6 +166,63 @@ def test_empty_create_via_Empty_class(self):
...
+        if arr.flags.c_contiguous:
+            # read empty dataset
+            with self.assertRaises(TypeError):
+                empty_dset.read_direct(arr, np.s_[0:10], np.s_[50:60])
+
+            dset.read_direct(arr, np.s_[0:10], np.s_[50:60])
+            self.assertEqual(dset.shape, (100,))
+
+            dset.read_direct(arr, source_sel, dest_sel)
+            self.assertEqual(dset.shape, (100,))
```
takluyver, line 200: These assertions don't really make sense. Reading from a dataset will never change the dataset shape.

If we want to check something, it would be better to make the initial contents of `arr` and `dset` different, and then check that `arr` has been updated with the data from `dset`.

h5py/tests/test_dataset.py
```diff
@@ -165,6 +166,63 @@ def test_empty_create_via_Empty_class(self):
...
+        Feature: Write Numpy array directly into Dataset
+    """
+
+    def test_write_direct(self):
+        dset = self.f.create_dataset('dset', (100,), dtype='int32')
+        empty_dset = self.f.create_dataset("edset", dtype='int64')
+
+        arr = np.ones((100,), dtype='int64')
+
+        if arr.flags.c_contiguous:
```
takluyver, line 214: Again, we don't need this check, as far as I know.

h5py/tests/test_dataset.py
```diff
@@ -165,6 +166,63 @@ def test_empty_create_via_Empty_class(self):
...
+            # write into empty dataset
+            # FIXME: write data into empty datase should be allowed
+            with self.assertRaises(TypeError):
+                empty_dset.write_direct(arr, np.s_[0:10], np.s_[50:60])
+
+            dset.write_direct(arr, np.s_[0:10], np.s_[50:60])
+            self.assertEqual(dset.shape, (100,))
+
+            dset.write_direct(arr)
+            self.assertEqual(dset.shape, (100,))
```
takluyver, line 224: These don't really make sense here either. `write_direct` doesn't affect the shape - only explicitly calling `dset.resize()` does that.

We could check for 1s in the relevant bit of the dataset after writing to it.

h5py/tests/test_dataset.py
```diff
@@ -217,6 +275,15 @@ def test_dtype_conflict(self):
...
             self.f.require_dataset('foo', (10, 3), 'S10')
 
+    def test_dtype_exact(self):
+        """ require_dataset with exactly dtype match """
+
+        dset = self.f.create_dataset('foo', (10, 3), 'f')
+        dset2 = self.f.require_dataset('foo', (10, 3), 'f', exact=True)
+        self.assertEqual(dset, dset2)
+        with self.assertRaises(TypeError):
+            self.f.require_dataset('foo', (10, 3), 'i', exact=True)
```
takluyver, line 285: This would also fail with `exact=False`, because `np.can_cast('f', 'i') == False`. To properly check `exact=True`, we should use a pair of dtypes which would be OK without it, e.g. create it with `'i8'` (int64) and require it with `'i4'` (int32).

h5py/tests/test_dataset.py
```diff
@@ -495,6 +567,17 @@ def test_float_fails_without_options(self):
 
         with self.assertRaises(ValueError):
             dset = self.f.create_dataset('foo', (20, 30), dtype=float, scaleoffset=True)
+    def test_non_integer(self):
```
takluyver, line 570:
```suggestion

    def test_non_integer(self):
```

h5py/tests/test_dataset.py
```diff
@@ -713,6 +796,14 @@ def test_track_times(self):
         similar = self.f.create_dataset_like('lenovo', orig)
         self.assertEqual(0, h5py.h5g.get_objinfo(similar._id).mtime)
 
+    def test_special_case(self):
```
takluyver, line 799:
```suggestion
    def test_maxshape(self):
```

'Special case' is (ironically) a very generic term.

h5py/tests/test_dataset.py
```diff
@@ -762,6 +853,9 @@ def test_create_1D(self):
         dset = self.f.create_dataset('bar', 20, maxshape=20)
         self.assertEqual(dset.maxshape, (20,))
 
+        dset = self.f.create_dataset('fun', (1,), maxshape=True)
```
takluyver, line 856: Please don't test this. It's not really intended or documented to work, it just happens because `True == 1` in Python.

h5py/tests/test_file.py
```diff
@@ -250,6 +258,9 @@ def test_backing(self):
         fid = File(fname, 'r')
         assert 'foo' in fid
         fid.close()
+        # When driver is None, any kwds is invalid
```
takluyver, line 261:
```suggestion
        # keywords for other drivers are invalid when using the default driver
```

h5py/tests/test_selections.py
```diff
@@ -26,59 +36,113 @@ class TestTypeGeneration(TestCase):
...
+    def test_selection(self):
+        dset = self.f.create_dataset('dset', (100,100))
+        regref = dset.regionref[0:100, 0:100]
+
+        # args is numpy.ndarray, return a FancySelection
+        st = sel.select((10,), np.array([1,2,3], dtype='b'), dset)
+        self.assertIsInstance(st, sel.FancySelection)
+
+        self.assertEqual(st.mshape, (3,100))
+        self.assertEqual(st.array_shape, ())
```
takluyver, line 109: This test only passes because of a bug (see #1595) and will fail when it is fixed.

This kind of thing is what makes writing tests hard work - we need to understand the code and test for what it should be doing, not just what it is doing now.

h5py/tests/test_selections.py
```diff
@@ -26,59 +36,113 @@ class TestTypeGeneration(TestCase):
...
+
+        # expand shape
+        self.assertEqual(st.expand_shape(()), ())
```
takluyver, line 112: This is also not the right behaviour (part of the same bug)

h5py/tests/test_selections.py
```diff
@@ -26,59 +36,113 @@ class TestTypeGeneration(TestCase):
...
+        with self.assertRaises(TypeError):
+            st.expand_shape((4,100))
+
+        # broadcast in FancySelection
+        st.broadcast((5,100))
```
takluyver, line 117: This doesn't actually test anything - `.broadcast()` is a generator, so you need to iterate over it in order to run it.

h5py/tests/test_selections.py
```diff
@@ -26,59 +36,113 @@ class TestTypeGeneration(TestCase):
...
+        self.assertEqual(st2.mshape, (1,100))
+        self.assertEqual(st2.array_shape, (100,))
+
+        # expand shape
+        self.assertEqual(st2.expand_shape(()), (1,1))
+        with self.assertRaises(TypeError):
+            st2.expand_shape((4,100))
+
+        # broadcast in SimpleSelection
+        st2.broadcast((5,100))
```
takluyver, line 132: Same again - this doesn't test anything.

h5py/tests/test_selections.py
```diff
@@ -26,59 +36,113 @@ class TestTypeGeneration(TestCase):
...
+
+        # args is RegionReference, but dataset is None
+        with self.assertRaises(TypeError):
+            sel.select((100,), regref, None)
+
+        # args is RegionReference, but its shape doesn't match dataset shape
+        with self.assertRaises(TypeError):
+            sel.select((100,), regref, dset)
```
takluyver, line 140: No test of successfully using a RegionReference?

h5py/tests/test_selections.py
```diff
@@ -26,59 +36,113 @@ class TestTypeGeneration(TestCase):
...
+
+        # args is a Selection instance, return a FancySelection
+        st3 = sel.select((100,100), st, dset)
+        self.assertIsInstance(st, sel.FancySelection)
```
takluyver, line 144:
```suggestion
        assert st3 == st
```

h5py/tests/test_dataset.py
dota17: Yes, I know, but when `exact= False`, then it's another error type, I just want to cover these lines: https://github.com/h5py/h5py/blob/14e5542db20f1bbcdadd0a4d6a70fc05b191ceba/h5py/_hl/group.py#L225-L227

h5py/tests/test_dataset.py
dota17: Ok, will do.

h5py/tests/test_dataset.py
dota17: yes, I can.

h5py/tests/test_selections.py
dota17: Ok.

h5py/tests/test_selections.py
dota17: I would to say that some of these tests in this class are just to test the uncovered methods 🤔

h5py/tests/test_selections.py
dota17: Sure, there is. I will add them.

h5py/tests/test_base.py
```diff
@@ -41,6 +42,78 @@ def test_anonymous(self):
...
+            grp = self.f.create_group(None)
+            bar_parent = grp.parent
+        except ValueError:
+            pass
+
+        # Named objects
+        grp = self.f.create_group("bar")
+        sub_grp = grp.create_group("foo")
+        parent = sub_grp.parent
+        self.assertEqual(r'<HDF5 group "/bar" (1 members)>', repr(parent))
```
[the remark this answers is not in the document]
takluyver: Yup, that looks OK.

h5py/tests/test_base.py
```diff
@@ -41,6 +42,63 @@ def test_anonymous(self):
...
+        # Named objects
+        grp = self.f.create_group("bar")
+        sub_grp = grp.create_group("foo")
+        parent = sub_grp.parent.name
+        self.assertEqual(parent, "/bar")
+
+class TestMapping(BaseTest):
+
+    """
+        Test if the registration of Group, AttributeManager as a
```
takluyver, line 67: Note that your tests here don't actually do anything with `AttributeManager`.

h5py/tests/test_base.py
```diff
@@ -80,3 +143,12 @@ def test_file(self):
                 os.unlink(fname)
             except Exception:
                 pass
+
+def test_is_hdf5():
+    filename = File(tempfile.mktemp(), "w").filename
+    fid = is_hdf5(filename)
+    assert fid is True
```
takluyver, line 150:
```suggestion
    assert is_hdf5(filename)
```

The name `fid` is a bit confusing - normally I'd guess that means 'file ID'.

h5py/tests/test_base.py
```diff
@@ -80,3 +143,12 @@ def test_file(self):
...
+    # non-existing HDF5 file
+    filename = tempfile.mktemp()
+    fid = is_hdf5(filename)
+    assert fid is False
```
takluyver, line 154:
```suggestion
    assert not is_hdf5(filename)
```

h5py/tests/test_dataset.py
```diff
@@ -165,6 +166,58 @@ def test_empty_create_via_Empty_class(self):
...
+        empty_dset = self.f.create_dataset("edset", dtype='int64')
+        arr = np.zeros((100,))
+        arr2 = np.zeros((200,))
+
+        # read empty dataset
+        with self.assertRaises(TypeError):
+            empty_dset.read_direct(arr, np.s_[0:10], np.s_[50:60])
+
+        dset.read_direct(arr2, np.s_[0:10], np.s_[50:60])
+        self.assertEqual(dset.shape, (100,))
```
takluyver, line 192: This assertion still doesn't really make sense.

Maybe use `np.ones()` to create `arr2`, and then assert `arr2[50] == 0` after reading zeros into it?

h5py/tests/test_dataset.py
```diff
@@ -165,6 +166,58 @@ def test_empty_create_via_Empty_class(self):
...
+        empty_dset = self.f.create_dataset("edset", dtype='int64')
+        arr = np.ones((100,))
+        arr2 = np.zeros((200,))
+
+        # write into empty dataset
+        with self.assertRaises(TypeError):
+            empty_dset.write_direct(arr, np.s_[0:10], np.s_[50:60])
+
+        dset.write_direct(arr2, np.s_[0:10], np.s_[50:60])
+        self.assertEqual(dset.shape, (100,))
```
takluyver, line 215: Similarly here - it will pass, but it's not really relevant to what you're testing.

h5py/tests/test_dataset.py
takluyver: It's a `TypeError` either way, though. And the point of tests is to check that things work as we expect, not just to cover lines of code. So I think it's a better test if it uses types which would be OK with `exact=False` but not with `exact=True`.

kif:
Increasing the coverage is always a good idea ... 
Unfortunately, there is no coverage for the Cython code.

takluyver:
@aragilar did do some work towards that in #1004 - it should be possible, but we didn't get it working yet.

dota17:
@takluyver @aragilar @kif I have a look at the project [Cython](https://github.com/cython/cython), it seems that they made it.

FYI: 
+ https://github.com/cython/cython/blob/084a25f55d0b4cf8b4c3cd496ec57bb3e2f57f71/tests/run/coverage_api.srctree

+ https://github.com/cython/cython/blob/5ba68b7cfff0b00dddf2d7ba828ee93e80c11b51/Cython/Coverage.py

takluyver:
Thanks!

PR #1600, opened 2020-07-29
refactory code in h5py/h5p.pyx

dota17:
Class `PropGCID` and class `PropFCID` inherit from class `PropOCID`, they have some common methods that can be moved into common class.
CC: @takluyver

### DIFF
--- a/h5py/h5p.pyx
+++ b/h5py/h5p.pyx
@@ -353,27 +353,6 @@ cdef class PropFCID(PropOCID):
         return (addr, size)
 
 
-    @with_phil
-    def set_link_creation_order(self, unsigned int flags):
-        """ (UINT flags)
-
-        Set tracking and indexing of creation order for links added to this group
-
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks - but `PropOCID` (object creation property list) is also the parent class for `PropDCID` (dataset creation) and `PropTCID` (datatype creation), which shouldn't have those methods. So I think a bit of duplication is preferable to over-generalising.

takluyver:
And I don't think we should make `PropFCID` inherit from `PropGCID` either. While [H5Pset_link_creation_order](https://confluence.hdfgroup.org/display/HDF5/H5P_SET_LINK_CREATION_ORDER) works on a group or file creation property list, there are others which are only documented as being [for group creation](https://confluence.hdfgroup.org/display/HDF5/H5P_SET_LINK_PHASE_CHANGE), and others which are [for dataset or group creation](https://confluence.hdfgroup.org/display/HDF5/H5P_SET_DEFLATE).

So I really would just leave them duplicated. They're very simple wrappers around HDF5 functions, anyway.

PR #1603, opened 2020-07-30
Stricter rules for writing into vlen string datasets

takluyver:
This is part of the string changes described in #1338. Specifically, it affects writing to existing variable length string datasets. Changes:

- Writing `None` to these datasets is no longer allowed. Previously it was equivalent to storing an empty string.
- Only str and bytes objects are now accepted. Previously, any Python object would be automatically converted as `str(obj)`, in an uncanny echo of Javascript.
- ASCII vlen string datasets now only accept str made of ASCII characters, rather than using UTF-8 encoding.
  - bytes objects are still stored with no check of the text encoding, if you need to store something in the 'wrong' encoding.

### DIFF
--- a/h5py/_conv.pyx
+++ b/h5py/_conv.pyx
@@ -179,14 +179,7 @@ cdef int conv_str2vlen(void* ipt, void* opt, void* bkg, void* priv) except -1:
             temp_object = temp_object.encode('utf-8')
 
         elif not isinstance(temp_object, bytes):
-            # There is not test on this !
-            if sizes.cset == H5T_CSET_ASCII:
-                encoding = 'ascii'
-            elif sizes.cset == H5T_CSET_UTF8:
-                encoding = 'utf-8'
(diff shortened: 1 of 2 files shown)


aragilar:
Looks like there was some network issue between appveyor and github, I've told appveyor to rebuild.

aragilar:
For the actual changes, these seem sensible (not sure why we were trying to autoconvert to str...).

takluyver:
Thanks! The re-run appears to have succeeded.

I'm not sure why auto conversion was desirable either. My best guess is that it's due to the conversion infrastructure - we register a converter to go from Python objects to vlen strings, not specifically Python strings. Maybe in that context it seemed to make sense to convert any Python objects.

aragilar:
I think given I'm making things less strict (in terms of matching types) in #1608, I'm going to merge this so there isn't a conflict between the two PRs.

Issue #565, opened 2015-04-16
Conversion error reading vlen string dataset


jreadey:
This file contains one dataset with a vlen string type: https://github.com/HDFGroup/h5serv/blob/master/testfiles/vlen_string_dset_utc.h5. 

h5dump output looks normal, but if I open with h5py I get a conversion error:

```
f = h5py.File("vlen_dset_utc.h5", "r")
ds1 = f["ds1"]
print ds1[0]
```

Outputs:

```
KeyError: 'Conversion function not found (No appropriate function for conversion path)'
```

Any ideas what is wrong?

tacaswell:
I do not remember the details off the top of my head, but I think this is related to numpy not having a good data type to map to vlen strings (as the 'native' string types are fixed length).  I do not recall the reason not to use 'object' as the dtype in this case though.

andrewcollette:
We do support vlen strings using the 'O' type... the function h5py.special_dtype can be used to construct the appropriate type.

I get the same error, but weirdly I can create vlen arrays from h5py with no problem.  I can't find any difference between the data types in the two files.

```
import h5py
f = h5py.File("vlen.hdf5")
ds = f.create_dataset("x", (100,), dtype=h5py.special_dtype(vlen=str))
ds[0] = "Hello"
f.close()
```

jreadey:
Right, compare with this file: https://github.com/HDFGroup/h5serv/blob/master/testfiles/vlen_string_nullterm_dset.h5.   Looks like exactly the same type and dataset attributes but it loads fine with h5py while the original one does not.

jreadey:
This is the current path for the file illustrating the problem: https://github.com/HDFGroup/hdf5-json/blob/master/data/hdf5/vlen_string_dset_utc.h5.

aragilar:
So the distinction seems to be spacepad vs nullterm between the files, I wonder if there some additional conversion happening along the line which lets spacepad work?

takluyver:
We believe that this was fixed by #1608. If people can still replicate it after that, please either reopen this issue or create a new one.

PR #1608, opened 2020-08-01
Be less strict about convert to/from vlen strings

### LINKED ISSUE #428: Possible byte order glitch related to opaque types
From the list:

I have a problem to open a hdf5 file with h5py (using h5py-2.2.1) containing variable length strings when the file is created on a x86 cluster (little endian) and read on a power cluster (big endian).

I do not have any issue with fixed length strings.

I can read and write a file with h5py containing variable length strings if I read it using the same system I used to create the file. I have this error when I try to read the power file on the x86 cluster and the x86 file on the power cluster :

Failed to find converter for 8 -> PYTHON:OBJECT

I am able to dump the info from the files using h5dum without any problem (any cluster can dump the info from both files without any issue).

Comparing the byte information (the files have exactly the same size), only a few bytes are different between the 2 files. I was wondering if one of this byte was referring to the endianness of the file which could impact how the length of the strings are stored.

Here is the code I used to read and then write the file on both systems :

To write:

```
import h5py
file = h5py.File('vlstring_p7.h5','w')
str_type = h5py.new_vlen(str)
dataset = file.create_dataset("DSvariable",(4000,), dtype=str_type)
data=()
for i in range(1000):
    data += ("Parting", " is such", " sweet", " sorrow...")
dataset[...] = data
file.close()
```

To read:

```
import h5py
file = h5py.File('vlstring_p7.h5', 'r')
dataset = file['DSvariable']
data_out = dataset[...]
for i in range(4):  
    print "DSvariable[",i,"]", "'"+data_out[i]+"'", "has length", len(data_out[i])

print data_out
file.close()
```

aragilar:
This may fix #428, and possibly other issues with slightly mismatching types.

I am slightly wary of this possibly allowing more things through, so I'd appreciate if people could check whether this actually fixes the issue (as suggested by https://github.com/h5py/h5py/issues/428#issuecomment-202388952).

@tillea Are you able to test this on the additional Debian arches which you disabled, and see if this fixes things?


aragilar:
This also seems to fix #565 (for the given HDF5 files).

takluyver:
Is it possible to add a test for the thing this fixes? Either forcing the creation of a big-endian file somehow, or committing a small sample file from a big-endian machine? I think @kif works with some POWER systems and may be able to help.

I also see this comment from Andrew on #428:

> In this case, the converter function would probably have to be modified to check H5Tis_variable_str() so that it doesn't try to convert non-string VLEN types.

Maybe that's outdated - I see we are registering the converter specifically against a variable string datatype. But it would be good to check.

I also still want to understand this 'hard' vs 'soft' distinction better... I'll go and look for more info.

takluyver:
Aha, this might be it:

> A conversion path contains a source and destination datatype and each path contains a hard conversion function and/or a soft conversion function. The only difference between hard and soft functions is the way in which the library chooses which function applies: A hard function applies to a specific conversion path while a soft function may apply to multiple paths. When both hard and soft functions apply to a conversion path, then the hard function is favored and when multiple soft functions apply, the one defined last is favored. 

http://davis.lbl.gov/Manuals/HDF5-1.6.1/Datatypes.html#Datatypes-DataConversion

aragilar:
I wonder if this happened to just work because the `vlen2ndarray` is defined afterwards (in which case we need should warn about ordering inside the function).

For tests, I'll start with the file from #565, and see if I can create more broken files which this fixes.

aragilar:
The ordering does make a difference: if I move the string converter down to the end, I get a segfault when running the test. I suspect I'll need to add the `INIT` logic to exclude non-string types.

aragilar:
And because I'm trying to convert to a python object (so cannot rely on HDF5's type checking), I still can't be restrictive enough to prevent incorrect types being passed through... I'll see if moving this up to the top with the type checks just on the HDF5 type will mean that we're still fixing #565.

takluyver:
I think I'm getting there with the help of [this document](https://support.hdfgroup.org/HDF5/doc/Supplements/dtype_conversion/Conversion.html) and reading some code.

Hard conversion functions are registered for data types, with exactly one function for each source+destination data type pair. Soft conversion functions are registered for data type *classes*, as in [H5Tget_class](https://confluence.hdfgroup.org/display/HDF5/H5T_GET_CLASS), with a list of functions for each source+destination class pair. So in this case, that's between STRING and OPAQUE classes. Then when it finds the conversion function, it steps backwards through this list, trying to initialise each one, and uses the first one where initialisation succeeded (returned >= 0).

So we should add an init function that checks both `H5Tis_variable_str` for the string side and for the `"PYTHON:OBJECT"` tag on the opaque side.

aragilar:
Oh, that's how you check for a python class, I was wondering how that's being done. I'll try that then, and see if I avoid segfaults.

aragilar:
I've pushed tests using the files from #565, I'm still experimenting with the converter which is still segfaulting (which I haven't pushed yet), but this should make it easier for others to test the change on their systems.

tacaswell:
> The soft conversions tend to be more rigid although this method is slower than the hard conversions because of all the bit operations during the conversions.

The mixed metaphors here....

tacaswell:
Would it be worth putting debugging side-effects in to the conversion functions if we are not 100% sure which of them are being hit?

aragilar:
That's probably worth doing anyway, given there may be subtle bugs hiding due to ordering which we haven't seen yet.

aragilar:
Quick update: I've stopped the segfaulting, but now I'm being too restrictive on the types allowed. I've also posted on the HDF forum asking if there was some way of getting the name of the converter that will be used for two types, as it looks like `H5Tfind` only gives back the function.

aragilar:
Ok, I've now got the type checking working. I'd still like people to check on different architectures, and see if this fixes #428 (@rposts I think you ran into a similar issue?).

aragilar:
Looks like an internal HDF5 assert is failing for 1.12, I'll install 1.12 and try to work out why.

aragilar:
Interestingly I don't get an error if I use the Debian version of 1.12, but if I build it from source (using the same args as the CI setup), then I get an error.

aragilar:
This should now work on 1.12 (I think HDF5 got smarter about tracking unfreed memory) - someone who understands cython better than I do should improve my comment about why we need the try-finally wrapper (as opposed to doing checks via if).

h5py/_conv.pyx
```diff
@@ -121,12 +132,72 @@ cdef herr_t init_generic(hid_t src, hid_t dst, void** priv) except -1:
...
+    # This complexity is needed to sure:
+    #   1) That ctag is freed
+    #   2) We don't segfault (for some reason a try-finally statement is needed,
+    #   even if we do (what I think are) the right steps in copying and freeing.
+    cdef char* ctag = NULL
+    try:
+        if H5Tget_class(obj) == H5T_OPAQUE:
+            ctag = H5Tget_tag(obj)
+            if ctag != NULL:
+                tag = str(ctag)
```
takluyver, line 153: Is it practical to compare the `char*` strings in C (`strcmp`?), rather than as Python objects?

h5py/_conv.pyx
```diff
@@ -121,12 +132,72 @@ cdef herr_t init_generic(hid_t src, hid_t dst, void** priv) except -1:
...
+                if tag == _H5PY_PY_TAG:
+                    return True
+        return False
+    finally:
+        IF HDF5_VERSION >= (1, 8, 13):
```
takluyver, line 158: Do we have a minimum 1.8.x version we support? It [looks like](https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-1.8/) 1.8.13 was released in 2014, so I'd say we could reasonably ignore older versions by now.

takluyver:
I don't think I'm any better at Cython, but do you have the variant you tried with `if` handy to paste in a comment?

h5py/_conv.pyx
aragilar: I copied this from https://github.com/h5py/h5py/blob/master/h5py/h5t.pyx#L635, after trying all the if related experiments. We could do a overall cleanup for HDF5 versions at some point (like we did for py3).

aragilar:
I don't, but the general gist was something like:
```
cdef char* ctag = NULL
if H5Tget_class(obj) == H5T_OPAQUE:
    ctag = H5Tget_tag(obj)
    tag = str(ctag)
    if ctag != NULL:
        free(ctag)
    if tag == _H5PY_PY_TAG:
        return True
return False
```
with lots of different methods tried to ensure that tag pointed to a completely different buffer to ctag. I'm not sure it's that important overall, but it would be good to have a better explanation than "it segfaults otherwise" (as even reading the C code that was produced, I couldn't see a problem).

h5py/_conv.pyx
aragilar: That may work better, I'll try that.

takluyver:
The changes here look OK to me. I'd obviously like to get someone to test with moving files between big-endian and little-endian systems - as in #428 - but I'm not sure how to do that.

rposts:
> 
> 
> Ok, I've now got the type checking working. I'd still like people to check on different architectures, and see if this fixes #428 (@rposts I think you ran into a similar issue?).

I'll give it a try in next few days - Thanks. @aragilar

rposts:
> 
> 
> > Ok, I've now got the type checking working. I'd still like people to check on different architectures, and see if this fixes #428 (@rposts I think you ran into a similar issue?).
> 
> I'll give it a try in next few days - Thanks. @aragilar

@aragilar - tried x86-s390x test using your branch - seems to work now. Thanks!

aragilar:
@rposts Would you be able to generate a small test file that we can use?

takluyver:
I've opened #1636 about the missing test, so I can merge this.

PR #1609, opened 2020-08-01
Fix creating datasets from nested lists of str/bytes objects & writing to fixed UTF-8 datasets

takluyver:
Fix creating datasets from nested lists of str/bytes objects.

These now behave the same as using a single object or a simple list. See https://github.com/h5py/h5py/issues/1338#issuecomment-667538812 .

EDIT: Now also allows storing str objects with any unicode characters into a fixed-length UTF-8 string dataset. Previously, str objects with non-ASCII characters would error.

### DIFF
--- a/h5py/_hl/base.py
+++ b/h5py/_hl/base.py
@@ -38,26 +38,32 @@ def is_hdf5(fname):
         return False
 
 
-def object_collection_of(data):
-    """Check if data is a list/tuple/numpy object array (without h5py tag)
+def find_item_type(data):
+    """Find the item type of a simple object or collection of objects.
 
-    If so, and its contents are homogeneous, return their type.
-    Otherwise, return None.
+    E.g. [[['a']]] -> str
+
+    The focus is on collections where all items have the same type; we'll return
+    None if that's not the case.
 
     The aim is to treat numpy arrays of Python objects like normal Python
     collections, while treating arrays with specific dtypes differently.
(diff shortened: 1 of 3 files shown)


h5py/_hl/base.py
```diff
@@ -75,10 +81,7 @@ def guess_dtype(data):
         if isinstance(data, h5r.Reference):
             return h5t.ref_dtype
 
-        item_type = object_collection_of(data)
-        if item_type is None:
-            # Potentially scalar
-            item_type = type(data)
+        item_type = find_item_type(data)
```
tacaswell, line 84: There is a path through `find_item_type` that returns `None`, but the code that used to be here ensured that `item_type` was never `None` before going into the rest of the code.

takluyver: It did, but then the only things we do with item_type are to check `item_type is bytes` and `item_type is str` immediately below. The case which the extra `if item_type is None` previously handled - a single str/bytes object - is now handled inside `find_item_type()`.

`item_type` will be None if there is a non-homogeneous collection (different types in a list) or a numpy array with a more specific dtype already.

aragilar:
This looks good to me.

takluyver:
Thanks! I'll wait until the US maintainers can take a look later today, but I'll merge it if there are no objections. I think this completes the string work for 3.0. :slightly_smiling_face:

PR #1611, opened 2020-08-01
BLD: require at least cython 0.29

### LINKED ISSUE #1538: Certain versions of numpy & cython cause problems
RedHat
3.6
   Windows)
system python
 * h5py version (e.g. 2.6)
latest from branch
 * HDF5 version (e.g. 1.8.17)
1.10.6
File "/nas/longleaf/home/XXX/.local/lib/python3.6/site-packages/h5py/__init__.py",
line 45, in <module>
    from ._conv import register_converters as _register_converters, \
  File "h5py/h5t.pxd", line 15, in init h5py._conv
  File "h5py/h5t.pyx", line 262, in init h5py.h5t
  File "h5py/h5t.pyx", line 257, in h5py.h5t._get_available_ftypes
AttributeError: 'numpy.dtype' object has no attribute 'typeobj'

`h5py.version.info` 
Not available. Traceback given above.
However, this error occurs only when compiling from sources pulled from the master branch on github. pip install --no-binary=h5py h5py works fine.  

This error occurs when the available version of  numpy<=1.17. It works with numpy>=1.18.3 
I suggest to update the requirements. Right now, it checks for numpy >=1.7.

tacaswell:
closes #1538

For reasons that are not clear to me cython 0.25 to 0.29 compile but do not work (but cython 0.23 and 0.24 do at least import).  I think this is fallout from https://github.com/h5py/h5py/pull/1390.    Rather than trying to sort out what changed in Cython, I propose we require at least cython 0.29 for all versions of Python.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -10,14 +10,11 @@ deps =
     test: pytest-cov
     test: pytest-mpi>=0.2
 
-    py36-deps: cython>=0.23
-    py37-deps: cython>=0.29
-
     py36-deps: numpy>=1.12
     py37-deps: numpy>=1.14.3
     py38-deps: numpy>=1.17.4
 
-    py36-mindeps: cython==0.23
+    py36-mindeps: cython==0.29
     py37-mindeps: cython==0.29
     py38-mindeps: cython==0.29.14
 
--- a/docs/build.rst
+++ b/docs/build.rst
@@ -78,6 +78,7 @@ Source installation
 To install h5py from source, you need three things installed:
 
 * A supported Python version with development headers
+* Cython >=0.29
 * HDF5 1.8.4 or newer with development headers
 * A C compiler
 
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ RUN_REQUIRES = [NUMPY_DEP, "cached-property"]
 # these are required to build h5py
 # RUN_REQUIRES is included as setup.py test needs RUN_REQUIRES for testing
 # RUN_REQUIRES can be removed when setup.py test is removed
-SETUP_REQUIRES = RUN_REQUIRES + [NUMPY_DEP, 'Cython>=0.25', 'pkgconfig']
+SETUP_REQUIRES = RUN_REQUIRES + [NUMPY_DEP, 'Cython>=0.29', 'pkgconfig']
 
 # Needed to avoid trying to install numpy/cython on pythons which the latest
 # versions don't support


tox.ini
```diff
@@ -10,14 +10,11 @@ deps =
     test: pytest-cov
     test: pytest-mpi>=0.2
 
-    py36-deps: cython>=0.23
-    py37-deps: cython>=0.29
```
takluyver, line 14: Should Cython still be encoded anywhere in the tox config for the `deps` builds? Or is it sufficient to have it as a dependency in the packaging info?

aragilar: We used to have a `nodeps` tox env for testing without having numpy/cython preinstalled, not sure why that was dropped (possibly it was making the CI more flakey?)

takluyver: Ah, so that might wind up using the dicey setuptools method for installing build dependencies as eggs. @tacaswell and I were talking yesterday about using PEP-518, which should avoid setuptools complicating things.

aragilar: I think it's the optional build deps (i.e. mpi4py) that cause the problem, I don't think we can do some trick with extras or something to get it to work (based on my knowledge of PEP-518).

aragilar:
I'm not exactly sure how https://github.com/MacPython/h5py-wheels/issues/11 affects this, we may need to change how the wheels are built for 3.0.

tox.ini
takluyver: Aha, right, I'd forgotten about that. No, I can't see any way of expressing that at present.

takluyver: I went to post a question about this on Discourse, but found a recent thread about it, and actually it should be possible, as [Paul Ganssle explains](https://discuss.python.org/t/options-to-build-the-same-package-in-different-ways-with-different-build-dependencies/4458/10?u=takluyver).

PEP 518 specifies how to express fixed dependencies to run the build system at all (like setuptools). Then PEP 517 defines how to call the build system, including how to ask it for extra dependencies, which can depend on environment variables or config files. So we'd conditionally add mpi4py to `setup_requires`. Apparently, when setuptools is used as a PEP 517 backend, it doesn't try to install `setup_requires` as eggs.

aragilar: Oooh, that seems smart (and that could open up options for being able to generate correct dependency metadata based on what was used to build, rather than trying to lock numpy/cython down statically via pyproject.toml - have a release_deps option which uses minimum values, but also allow for building against the latest version to check how newer versions interact).

takluyver:
I'm going to merge this as is, and try to experiment with using `pyproject.toml` in the build.

PR #1612, opened 2020-08-01
BLD: use json to pickle to stash build configuration

tacaswell:
While sorting out #1611 I found we are using a pickle to stash configuration information between steps of the build process.  The information we are stashing is a `Dict[str, str]` so switch to json.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ h5py/*.dll
 h5py/*.so
 *.hdf5
 *.pkl
+h5config.json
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -11,6 +11,7 @@ h5py/*.dll
 h5py/*.so
 *.hdf5
 *.pkl
```
takluyver, line 13:
```suggestion
```

I'm assuming the only reason this was there was for the pickle config file?

PR #1613, opened 2020-08-01
Update documentation for strings

takluyver:
I've been working on making the string interfaces more consistent for h5py 3.0. This updates the documentation, trying to present the rules concisely.

To do: examples.

### DIFF
--- a/docs/strings.rst
+++ b/docs/strings.rst
@@ -3,17 +3,50 @@
 Strings in HDF5
 ===============
 
-The Most Important Thing
-------------------------
+.. note::
 
-If you remember nothing else, remember this:
+   The rules around reading & writing string data were redesigned for h5py
+   3.0. Refer to `the h5py 2.10 docs <https://docs.h5py.org/en/2.10.0/strings.html>`__
+   for how to store strings in older versions.
 
-    **All strings in HDF5 hold encoded text.**
+Reading strings
+---------------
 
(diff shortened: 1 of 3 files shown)


docs/strings.rst
```diff
@@ -3,164 +3,68 @@
...
-    >>> dt = h5py.string_dtype()
-    >>> dset = f.create_dataset("name", (100,), dtype=dt)
+Storing strings
+---------------
 
-They can store any character a Python unicode string can store, with the
-exception of NULLs.  In the file these are created as variable-length strings
-with character set H5T_CSET_UTF8.
+When creating a new dataset or attribute, Python ``str`` or ``bytes`` objects
+will be treated as variable-length strings, marked as UTF-8 and ASCII respectively.
```
tacaswell, line 26: I do not recall why we can not use this flag to also mark the bytes as `UTF-32`.

tacaswell:
This looks good to me.

docs/strings.rst
aragilar: HDF5 has no UTF-32 option (just ASCII and UTF-8, it's effectively an enum), or do you mean we convert from UTF-32 to UTF-8 and store that?

takluyver: Yup, it's a fixed list of 2 options, rather than an arbitrary tag - see [H5Tset_cset](https://confluence.hdfgroup.org/display/HDF5/H5T_SET_CSET).

PR #1614, opened 2020-08-02
Add commentary to release instructions to drive discussion

### LINKED ISSUE #1300: Document release process

aragilar:
This should close #1300 (once it's done) - people should feel free to push changes (I figured it was easiest to discuss this via a PR).

This currently has a lot of comments in the rst file, hopefully we can replace this comments with actual text so everyone knows what the release process is.

In terms of things that have been done in #1300, we have the API docs updating automatically now via github pages.

I think the main thing is working out how to handle the wheels (can rever have multiple stages?).

### DIFF
--- a/docs/release_guide.rst
+++ b/docs/release_guide.rst
@@ -1,5 +1,10 @@
 .. _release_guide:
 
+..
+    This is derived from the matplotlib release guide:
+    https://matplotlib.org/devdocs/devel/release_guide.html#all-releases
+    but there will be differences.
+
 Release Guide
 =============
 h5py uses `rever <https://regro.github.io/rever-docs/>`_ for release management.
@@ -21,8 +26,17 @@ that everything you need to perform the release is correctly installed
 and that you have the correct permissions. All rever commands should be
 run in the root level of the repository.
 
+..
+    rever currently is set to upload to PyPI, which will likely cause issues as
+    we need to upload the sdist and wheels at the same time
 
-**Step 1 (repeat until successful)**
+..
+    We don't have any of the github stats set up, so we can ignore that I think
+    For docs, running tox should be sufficient.
+
+
+Step 1: Pre-release checks (repeat until successful)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 .. code-block:: sh
 
@@ -32,7 +46,11 @@ Resolve any issues that may have come up, and keep running ``rever check``
 until it passes. After it is successful, simply pass the version number
 you want to release (e.g. ``X.Y.Z``) into the rever command.
 
-**Step 2**
+..
+    This doesn't run the tests, should we have it run tox?
+
+Step 2: Create the release
+^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 .. code-block:: sh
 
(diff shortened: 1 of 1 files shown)


docs/release_guide.rst
```diff
@@ -21,8 +26,17 @@ that everything you need to perform the release is correctly installed
 and that you have the correct permissions. All rever commands should be
 run in the root level of the repository.
 
+..
+    rever currently is set to upload to PyPI, which will likely cause issues as
+    we need to upload the sdist and wheels at the same time
```
takluyver, line 31: Do we? I thought it was possible to upload them separately, so long as each file is only uploaded once.

aragilar: https://github.com/h5py/h5py/issues/1300#issuecomment-531407658 was what I was going off, I think the issue is if there's no wheel for the latest version, pip tries to build from source, which errors, meaning the install fails. I guess it depends on how long it takes to get and upload the wheels (the longer it is, the more people get a build error - I imagine a hour or so it would be fine, longer I'm not so sure).

takluyver: Ah, gotcha. I was thinking of limitations on uploading to PyPI, but it makes sense that we don't want a release uploaded without wheels for any length of time.

aragilar:
I'm not exactly sure how https://github.com/MacPython/h5py-wheels/issues/11 affects this, we may need to change how the wheels are built for 3.0.

aragilar:
I wonder if something like this could work: we do everything up to the PyPI upload via rever, which instead uploads the sdist somewhere (TBD, ideally the place we upload the wheels) and pushes the new version to the h5py-wheels repo. The wheels for mac/linux get uploaded via CI (based on the sdist we uploaded), and then we upload the wheels from Christoph Gohlke. Then there's a final script which does the copying over of the sdist/wheels to PyPI.

docs/release_guide.rst
```diff
@@ -42,3 +60,57 @@ You probably want to make sure (with ``git tag``) that the new version
 number is available. If any release activities fail while running this
 command, you may safely re-run this command. You can also safely undo
 previously run activities. Please see the rever docs for more details.
+
+..
+    I'm guessing we want this to just push the tag to GitHub.
+    Things that could differ from matplotlib in this step
+    * Empty commit
```
tacaswell, line 67: The empty commit is because Matplotlib uses versioneer to inject the version which injects the names of all of the references to the tagged commit into the tarball so if you have both the tag and a branch and then later just a tag the hash of the tarball will change.

docs/release_guide.rst
```diff
@@ -42,3 +60,57 @@ You probably want to make sure (with ``git tag``) that the new version
...
+    * Tag signing (rever appears not to have that) - probably not needed
```
tacaswell, line 68: I'm +.5 on doing the signing.

docs/release_guide.rst
```diff
@@ -42,3 +60,57 @@ You probably want to make sure (with ``git tag``) that the new version
...
+    * The format for the github release (we may not care as long as it's
```
tacaswell, line 69: For Matplotlib we just use the body of the tag message.

docs/release_guide.rst
```diff
@@ -42,3 +60,57 @@ You probably want to make sure (with ``git tag``) that the new version
...
+      consistent)
+
+    We should also work out what need to happen for zenodo
```
tacaswell, line 72: This should be triggered by the creating the release on github.

takluyver:
My experiences so far for 3.0.0rc1 (#1695):

- `rever check` worked OK, but I got emails from Github warning me about a deprecated authentication API.
- I ran subsequent steps one by one (`rever -a authors`) instead of just letting it go, to keep control over the process.
- `authors` complained about one Github username (from PR #1576). I don't really know why, nor why the check we run in CI didn't already complain - it's otherwise quite picky about the `.authors.yml` file. I added the username by hand and it was happy.
- `version_bump` silently missed the `rc1` from the tuple that ultimately defines `h5py.__version__`. I think the regex is not quite right. Fixed manually.
- `changelog`
  - Called the file `3.0.0.rc1.rst` - we'd ideally like it to be `3.0.rst`.
  - Didn't put a title on the page ("What's new in h5py 3.0")
  - Didn't add the new file to `whatsnew/index.rst`
  - Got some entries in the wrong section - e.g. my recent addition in #1691 went under 'Building h5py', although I had put it under 'Bug fixes' in the PR.
  - Left some `<news item>` entries in the changelog.
  - Split 'building h5py' into two sections

Even if all that is fixed, I think the canonical release process should stop and let the maintainer check things manually at that point. No automated process can turn a bunch of little fragments into decent release notes by itself - you want someone to rearrange them, straighten out the wording, fix links, etc.

takluyver:
I continued stepping through the actions one at a time. `tag`, `push_tag` and `ghrelease` all seemed to work with no issues. The `pypi` action gave up with `RuntimeError: password not in pypi section of /home/takluyver/.pypirc`. That's quite true - my password is not in `.pypirc`, because twine & flit can both retrieve the password with the `keyring` library, allowing it to be stored more securely.

PR #1615, opened 2020-08-02
Add current version and build date to API docs

aragilar:
Currently the changes to the API docs are subtle: there may be a function added or modified, but this isn't noticeable when browsing. This adds the version to the title, and the build date to the page (so it's more noticeable that things are changing).

### DIFF
--- a/docs_api/conf.py
+++ b/docs_api/conf.py
@@ -112,7 +112,7 @@ html_theme = 'nature'
 
 # The name for this set of Sphinx documents.  If None, it defaults to
 # "<project> v<release> documentation".
-html_title = "Low-level API for h5py"
+html_title = "Low-level API for h5py (v{})".format(release)
 
 # A shorter title for the navigation bar.  Default is the same as html_title.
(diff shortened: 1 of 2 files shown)


docs_api/conf.py
```diff
@@ -112,7 +112,7 @@
 
 # The name for this set of Sphinx documents.  If None, it defaults to
 # "<project> v<release> documentation".
-html_title = "Low-level API for h5py"
+html_title = "Low-level API for h5py (v{})".format(release)
```
takluyver, line 115: N.B. the title is what shows up in the tab. The end of this is often hidden - I use an extension to have a list of tabs in the sidebar, but I still wouldn't see the ending of this:

Screenshot from 2020-08-02 13-28-28

It's also possible to put the version in the content as `|version|` - [Sphinx docs](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#default-substitutions) - which might be useful, at least on the front page.

PR #1619, opened 2020-08-03
Try to use pyproject.toml & PEP-517 builds

takluyver:


takluyver:
This appears to be working now - at least, the CI passes. cc @aragilar @tacaswell 

I've kind of fudged the check if MPI is enabled - it only looks at the environment variable. I'm wondering about getting rid of the stateful config system (where you can do `setup.py configure <options>` before building), and require all non-default options to be specified as environment variables for the build. That would simplify the build scripts, and reduce the chance of confusion where it picks up some previously specified option the user wasn't aware of. I already made one step towards this by not using the HDF5 version from the json config file. Thoughts?

tacaswell:
I am definitely in favor of moving to ENV configuration of `pip install ...` so that the install process looks the same coming from the repo and coming from pypi and the same with the autodetection / specification

takluyver:
Great - are you OK with me adding that into this PR, or would you rather I started a separate one?

tacaswell:
I am having trouble getting this to install under my py310 environment. 

```
pip install -v --no-build-isolation --no-use-pep517 .
```

fails with 

```
ERROR: Disabling PEP 517 processing is invalid: project specifies a build backend of setuptools.build_meta in pyproject.toml
```


and 

```
pip install -v --no-build-isolation  .
```

fails trying to install numpy an a virtual env but it fails with a main branch setuptools as well.

```
...
    subprocess.CalledProcessError: Command '['/home/tcaswell/.virtualenvs/bleeding/bin/python3', '-m', 'pip', '--disable-pip-version-check', 'wheel', '--no-deps', '-w', '/tmp/tmpsdkrszl7', '--quiet', 'numpy==1.17.4']' returned non-zero exit status 1.
...
```

The current env is:

```
✔ 11:47:12 $ pip list                                                                                                                                                
Package     Version
----------- ------------------------
Cython      3.0a6
decorator   4.4.2
flake8      3.8.3
mccabe      0.6.1
networkx    2.5rc1.dev20200804150820
numpy       1.20.0.dev0+593ef5f
pip         20.2.1
pkgconfig   1.5.1
pybind11    2.5.dev1
pycodestyle 2.6.0
pyflakes    2.2.0
scipy       1.6.0.dev0+034b1d3
setuptools  49.2.1
six         1.15.0
wheel       0.34.2
```

takluyver:
I'm guessing numpy 1.17.4 doesn't build on Python 3.10? Is there any version on PyPI that does, or does this only work if you can install numpy from git?

tacaswell:
numpy 1.19.1 will install on py310 so we can fix the current problem, but that still leaves a window between cpython / numpy sorting out their differences and numpy making a release when h5py can not be installed.  This seems like a problem if we want to test that the dev versions of numpy and h5py work together.

takluyver:
I'm puzzled why it tries to install anything when using `--no-build-isolation`. As far as I can see, that should assume that everything relevant is already installed in your target environment.

tacaswell:
Maybe because we are setting `numpy==VER` so it is deciding my version of numpy is too-new?

If you are making wheels that you plan to give to others, you need to pin numpy back, but if you are installing for your own uses it is probably ok to only set a minimum.

setup.py
```diff
@@ -29,20 +29,39 @@
...
 VERSION = '2.10.0'
 
-NUMPY_DEP = 'numpy>=1.7'
+# Minimum supported versions of Numpy & Cython depend on the Python version
+NUMPY_MIN_VERSIONS = [
+    # Numpy    Python
+    ('1.12',   "=='3.6'"),
+    ('1.14.3', "=='3.7'"),
+    ('1.17.4', "=='3.8'"),
+    ('1.17.4', "=='3.9'"),
```
tacaswell, line 38: Need to check that this actually works for py39.

takluyver:
Gah, it's setuptools getting in the way again. :angry: `pip install --no-build-isolation` does the right thing, but then setuptools gets in on the act and tries to install the `setup_requires` itself.

I'll see if I can find a way for `setup_requires` to only be passed at the right times...

takluyver:
OK, I've hopefully got it working so it doesn't try to install anything when you use `--no-build-isolation`.

takluyver:
I believe what you ran into is a bug in setuptools - I don't think it should install anything itself when used as a PEP 517 backend. I've opened an issue about it: https://github.com/pypa/setuptools/issues/2303

tacaswell:
:+1: This works for me!  Thank you for remotely debugging setuptools for me :)

takluyver:
Great! In that case, I'm going to take out the numpy version specification for Python 3.10, because I think that's so far from release that we can't make any claims about what will be compatible with it.

Have you verified the numpy requirement for Python 3.9?

aragilar:
I notice we're specifying a specific version for numpy, but allowing a range for cython. Did we want to lock down cython also?

takluyver:
The rationale for numpy is that if you build with a newer numpy, you can't then use it with an older numpy. So building with the oldest supported numpy means the built package should work with the full range of supported versions. That doesn't apply to Cython, as far as I know - it only touches Cython at build time, so the version conflict at runtime can't come up.

We could arbitrarily pin a specific Cython version to make builds more consistent - but that also makes it slightly harder to benefit if the next version of Cython fixes a bug (e.g. a problem on some platform we don't usually test). And we can't - at this level - pin the compiler version or the HDF5 version, so it doesn't give us much consistency.

I think I'm inclined to leave Cython with a range, but maybe it should be a capped range, like `Cython>=0.29, <1.0`.

aragilar:
I definitely agree with dropping the stateful build system (I can't see that it would work with pip, especially with build isolation).

Keeping the cython range opens seems fine to me (I was more checking if that was intended, and not a typo).

Looking at mpi4py changes (https://bitbucket.org/mpi4py/mpi4py/src/master/CHANGES.rst), we may want a open range, or track minimum supported python versions there also (if versions before 3.0.2 cannot build on python 3.8).

takluyver:
Good point; I had just made a quick guess for the mpi4py version. I don't suppose you could figure out what version is needed for each supported Python version (3.6 upwards)? I know you use MPI, so this might be easier for you to experiment with than me.

aragilar:
I think based on my experiments at https://travis-ci.com/github/aragilar/mpi4py_test/builds/178982226, 3.0 (which came out in 2017...) seems to be a reasonable target for 3.6-3.7, with 3.8 or higher requiring 3.0.3 (I'm guessing this is due to cython rebuild in 3.0.3 sdist). 3.0.0 does appear to run on py38, but given the comment about cython and py38 in the mpi4py changelog, I don't see too much of an issue requiring a newer version (and I suspect MPI users would use --no-build-isolation as they probably want a specific build of mpi4py).

takluyver:
:+1:

takluyver:
I've specified those MPI versions.

I'll work removing the stateful config in another PR.

aragilar:
This looks good to me, I'll hold off merging in case anyone else has comments.

takluyver:
I'd still like to work out what numpy version (if any) we should specify for Python 3.9 before merging. :-)

takluyver:
The word from setuptools is that we shouldn't rely on `setup_requires` long term - building with optional features is not really something setuptools aims to support, and `setup_requires` may be deprecated at some point in the future.

For 3.0, I think this approach is reasonable - it brings us up to speed with PEP 517 & 518 without requiring invasive changes. Beyond that, we might want to look at either switching to a different build system (enscons?) or completely rethinking how h5py and MPI fit together (e.g. a separate package which adds MPI support).

takluyver:
From setuptools [version 49.5](https://setuptools.readthedocs.io/en/latest/history.html#v49-5-0), setuptools won't try to install `setup_requires` itself when it's used as a PEP-517 backend. So I'll remove the 'egg_info' check again. We should still probably look at either a different build system or a different way of organising the MPI support in the future, but this will work for now.

I'm also going to remove the specified numpy version for Python 3.9 for now. We haven't worked out what we want it to be, and we're currently specifying 1.17.4, so we'll get the same result with a Python `>='3.8'` condition.

aragilar:
Half-baked idea: is there anything in PEP 517 preventing us from writing a backend that calls other backends as needed? E.g. defer most of the hooks to setuptools, with a different backend doing the "configuration"?

takluyver:
That's roughly possible, although there's no separate configuration step in PEP 517. We could write a little backend in which the build & metadata hooks just call the setuptools hooks, but defines the `get_requires_for_build_*` hooks itself. That would be a useful way forwards if setuptools does get rid of `setup_requires`.

I'm going to add an environment variable switch to turn off `setup_requires` if necessary - I find it useful to do `setup.py build` during development, for instance. Then I'll merge this in a couple of days if no-one objects.

PR #1622, opened 2020-08-04
Relative path to virtual datasets

### LINKED ISSUE #1546: Use '.' filename in virtual datasets referring to their own file
```py
>>> import h5py
>>> f = h5py.File('test.hdf5', 'w')
>>> f.create_dataset('base', data=[1, 2, 3])
<HDF5 dataset "base": shape (3,), type "<i8">
>>> layout = h5py.VirtualLayout(f['base'].shape)
>>> vs = h5py.VirtualSource(f['base'])
>>> layout[:] = vs[:]
>>> f.create_virtual_dataset('virtual', layout)
<HDF5 dataset "virtual": shape (3,), type "<f4">
>>> f['virtual'][:]
array([1., 2., 3.], dtype=float32)
```

```
$ mv test.hdf5 test2.hdf5
```

```py
>>> import h5py
>>> f = h5py.File('test2.hdf5', 'r')
>>> f['virtual'][:]
array([0., 0., 0.], dtype=float32)
```

It gives all 0s because the virtual dataset hard-codes the filename `test.hdf5`, which no longer exists. 

```py
>>> f['virtual'].virtual_sources()
[VDSmap(vspace=<h5py.h5s.SpaceID object at 0x11d8332f0>, file_name='test.hdf5', dset_name='/base', src_space=<h5py.h5s.SpaceID object at 0x11d833410>)]
```

So it just returns the empty buffer unchanged. I would have expected an error here instead. I looked into the h5py and hdf5 sources and I from what I could tell an error should be raised when a file isn't found. I can't really tell why it isn't.

I also would have expected it to use a relative filename by default when `VirtualSource` is passed a dataset (`'.'` can be used to reference the same file), but maybe there's a good reason not to do this.

```py
>>> print(h5py.version.info)

h5py    2.10.0
HDF5    1.10.5
Python  3.7.6 | packaged by conda-forge | (default, Mar 23 2020, 22:45:16)
[Clang 9.0.1 ]
sys.platform    darwin
numpy   1.18.1
```

(this is actually the git h5py)

tmichela:
Before opening a pull request, please:

- [x] Run simple static checks with `tox -e pre-commit`
- [x] Run the tests with e.g. `tox -e py37-test-deps`
- [x] Add a release note in the news/ folder (copy TEMPLATE.rst)

---
Hi!

this should hopefully fix #1546 

when creating a virtual dataset, check if each source filename matches the file it's being created in, and if so, use '.' as the filename.

### DIFF
--- a/docs/vds.rst
+++ b/docs/vds.rst
@@ -113,7 +113,7 @@ Reference
    When `creating a virtual dataset <creating_vds_>`_, paths to sources present
    in the same file are changed to a ".", refering to the current file (see
    `set virtual <https://portal.hdfgroup.org/display/HDF5/H5P_SET_VIRTUAL>`_).
-   This will keep such sources valid is the file is renamed.
+   This will keep such sources valid in case the file is renamed.
 
    :param path_or_dataset:
        The path to a file, or a :class:`Dataset` object. If a dataset is given,
--- a/docs/vds.rst
+++ b/docs/vds.rst
@@ -112,7 +112,7 @@ Reference
 
    When `creating a virtual dataset <creating_vds_>`_, paths to sources present
    in the same file are changed to a ".", refering to the current file (see
-   `set virtual <https://portal.hdfgroup.org/display/HDF5/H5P_SET_VIRTUAL>`_).
+   `H5Pset_virtual <https://portal.hdfgroup.org/display/HDF5/H5P_SET_VIRTUAL>`_).
    This will keep such sources valid in case the file is renamed.
 
    :param path_or_dataset:
--- a/docs/vds.rst
+++ b/docs/vds.rst
@@ -31,6 +31,7 @@ HDF5 dataset.
 The HDF Group has documented the VDS features in detail on the website:
 `Virtual Datasets (VDS) Documentation <https://support.hdfgroup.org/HDF5/docNewFeatures/NewFeaturesVirtualDatasetDocs.html>`_.
 
+.. _creating_vds:
 
 Creating virtual datasets in h5py
 ---------------------------------
@@ -108,6 +109,11 @@ Reference
 
    Instantiate this class to represent an entire source dataset, and then
    slice it to indicate which regions should be used in the virtual dataset.
+
+   When `creating a virtual dataset <creating_vds_>`_, paths to sources present
+   in the same file are changed to a ".", refering to the current file (see
+   `set virtual <https://portal.hdfgroup.org/display/HDF5/H5P_SET_VIRTUAL>`_).
+   This will keep such sources valid is the file is renamed.
 
    :param path_or_dataset:
        The path to a file, or a :class:`Dataset` object. If a dataset is given,
--- a/h5py/_hl/group.py
+++ b/h5py/_hl/group.py
@@ -168,10 +168,15 @@ class Group(HLObject, MutableMappingHDF5):
             """
             from .vds import VDSmap
             # Encode filenames and dataset names appropriately.
-            sources = [VDSmap(vspace, filename_encode(file_name),
-                              self._e(dset_name), src_space)
-                       for (vspace, file_name, dset_name, src_space)
-                       in layout.sources]
+            sources = []
+            for vspace, file_name, dset_name, src_space in layout.sources:
+                if file_name == self.file.filename:
+                    # use relative path if the source dataset is in the same
+                    # file, in order to keed the virtual dataset valid in case
+                    # the file is renamed.
+                    file_name = '.'
+                sources.append(VDSmap(vspace, filename_encode(file_name),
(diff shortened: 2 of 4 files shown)


scopatz:
CC @asmeurer

asmeurer:
It looks good to me. The h5py VDS docs should probably be updated to mention `'.'` and this behavior.

h5py/_hl/group.py
```diff
@@ -168,10 +168,15 @@ def create_virtual_dataset(self, name, layout, fillvalue=None):
...
             # Encode filenames and dataset names appropriately.
-            sources = [VDSmap(vspace, filename_encode(file_name),
-                              self._e(dset_name), src_space)
-                       for (vspace, file_name, dset_name, src_space)
-                       in layout.sources]
+            sources = []
+            for vspace, file_name, dset_name, src_space in layout.sources:
+                if file_name == self.file.filename:
+                    # use relative path if the source dataset is in the same
+                    # file, in order to keed the virtual dataset valid in case
```
takluyver, line 175:
```suggestion
                    # file, in order to keep the virtual dataset valid in case
```

h5py/tests/test_vds/test_highlevel_vds.py
```diff
@@ -338,5 +339,64 @@ def test_index_layout(self):
     def tearDown(self):
         shutil.rmtree(self.tmpdir)
 
+@ut.skipUnless(vds_support,
+               'VDS requires HDF5 >= 1.9.233')
+class RelativeLinkTestCase(ut.TestCase):
+
+    def setUp(self):
+        self.tmpdir = tempfile.mkdtemp()
```
takluyver, line 347: h5py is using pytest now - we haven't converted all the old tests, but you're welcome to write new tests as functions with fixtures if you like.

dota17: should we convert all unittest to pytest?  if so, I'd like to have a try.

docs/vds.rst
```diff
@@ -109,6 +110,11 @@ Reference
    Instantiate this class to represent an entire source dataset, and then
    slice it to indicate which regions should be used in the virtual dataset.
 
+   When `creating a virtual dataset <creating_vds_>`_, paths to sources present
+   in the same file are changed to a ".", refering to the current file (see
+   `set virtual <https://portal.hdfgroup.org/display/HDF5/H5P_SET_VIRTUAL>`_).
+   This will keep such sources valid is the file is renamed.
```
dota17, line 116: This will keep such sources valid `in case` the file is renamed.

h5py/tests/test_vds/test_highlevel_vds.py
takluyver: I think we should leave them - converting them all is a lot of work, and even if you're volunteering to do that, someone else has to review a lot of changes. As we've seen, even seemingly simple changes can break things. And h5py wouldn't really be any better for it - the existing unittest-style tests do what they're supposed to.

Other maintainers may disagree, but unless one of them makes that case, please don't start converting them. If you're looking for something else to do, we tagged a few issues [good first issue](https://github.com/h5py/h5py/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22) for the sprint last weekend.

docs/vds.rst
takluyver: Well spotted, thanks. I suspect 'is' was meant to be 'if'.

docs/vds.rst
```diff
@@ -109,6 +110,11 @@ Reference
    Instantiate this class to represent an entire source dataset, and then
    slice it to indicate which regions should be used in the virtual dataset.
 
+   When `creating a virtual dataset <creating_vds_>`_, paths to sources present
```
takluyver, line 113: Huh, I didn't know that worked! I always use `:ref:` to do that.

docs/vds.rst
```diff
@@ -109,6 +110,11 @@ Reference
...
+   in the same file are changed to a ".", refering to the current file (see
+   `set virtual <https://portal.hdfgroup.org/display/HDF5/H5P_SET_VIRTUAL>`_).
```
takluyver, line 115:
```suggestion
   `H5Pset_virtual <https://portal.hdfgroup.org/display/HDF5/H5P_SET_VIRTUAL>`_).
```

docs/vds.rst
tmichela: > I suspect 'is' was meant to be 'if'.

Indeed :)

thanks

h5py/tests/test_vds/test_highlevel_vds.py
dota17: Ok, got it 😃

takluyver:
Thanks!

PR #1624, opened 2020-08-06
Fix cross-reference of constants in API docs

takluyver:
I noticed cross-referencing wasn't fully working in the API docs, because functions wrapped in `@with_phil` appeared to be part of `h5py._objects`, rather than the module they should be associated with. This fixes it.

There's even a special extension which recognises `OBJ*` references and links to the constant group. :-)

### DIFF
--- a/h5py/_objects.pyx
+++ b/h5py/_objects.pyx
@@ -54,7 +54,7 @@ def with_phil(func):
         with _phil:
             return func(*args, **kwds)
 
-    functools.update_wrapper(wrapper, func, ('__name__', '__doc__'))
+    functools.update_wrapper(wrapper, func)
     return wrapper
 
 # --- End locking code --------------------------------------------------------
--- a/h5py/h5f.pyx
+++ b/h5py/h5f.pyx
@@ -215,7 +215,7 @@ def get_obj_count(object where=OBJ_ALL, int types=H5F_OBJ_ALL):
         special constant OBJ_ALL, to count objects in all files.
 
     type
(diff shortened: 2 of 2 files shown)


aragilar:
Nice. I did notice we do sometimes sometimes refer to the HDF5 constant (with it's own namespace), which then isn't picked up. Do we want to go through the docstrings and update those (which would then link between objects, but lose some clarity between it and the HDF5 constant it represents)?

takluyver:
Sorry, I don't quite follow what you mean - can you point to an example?

aragilar:
https://api.h5py.org/h5f.html#h5py.h5f.FileID.get_intent would be an example: `H5F_ACC_RDONLY ` vs `ACC_RDONLY` (or `h5py.h5f.ACC_RDONLY `). `H5F_ACC_RDONLY` is likely more searchable via google (and will likely lead to the HDF5 docs), whereas `ACC_RDONLY` will create a hyperlink within our docs (but not be as google friendly, at least for me the former has the HDF5 docs as the first link, whereas the latter find the h5py api page and then a bunch of mailing list posts).

takluyver:
Aha, gotcha. I think the h5py docs should probably refer to the constants with the names you need to use in the h5py API. Maybe each constants section in the low-level API docs should have a reminder like "these correspond to HDF5 C constants with the prefix H5F_".

aragilar:
This I think is ready to merge, if the relabelling of things will be in a different PR?

takluyver:
Yup, I'll do that in another PR.

PR #1633, opened 2020-08-09
Update Windows CI to newer HDF5

### LINKED ISSUE #839: Support the entire set of filename on Windows

aragilar:
Heads up, this will take a while to run, looks like 15-20 minutes per environment, I suspect this may lock up appveyor for a while (once the cache is filled, things will be much faster).

Appears to close #839 (as 1.10.6 uses UTF-8 now).

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -23,19 +23,39 @@ environment:
       TOXENV: "py36-test-deps"
       HDF5_VSVERSION: "14"
       HDF5_DIR: "C:\\hdf5\\%HDF5_VERSION%\\%HDF5_VSVERSION%"
-      HDF5_VERSION: "1.10.1"
+      HDF5_VERSION: "1.10.6"
 
     - PYTHON: "C:\\Python37"
       TOXENV: "py37-test-deps"
       HDF5_VSVERSION: "14"
       HDF5_DIR: "C:\\hdf5\\%HDF5_VERSION%\\%HDF5_VSVERSION%"
-      HDF5_VERSION: "1.10.1"
+      HDF5_VERSION: "1.10.6"
 
     - PYTHON: "C:\\Python38"
       TOXENV: "py38-test-deps"
       HDF5_VSVERSION: "14"
       HDF5_DIR: "C:\\hdf5\\%HDF5_VERSION%\\%HDF5_VSVERSION%"
-      HDF5_VERSION: "1.10.1"
+      HDF5_VERSION: "1.10.6"
+
+    ### USING HDF5 1.12 ###
+
+    - PYTHON: "C:\\Python36"
+      TOXENV: "py36-test-deps"
+      HDF5_VSVERSION: "14"
+      HDF5_DIR: "C:\\hdf5\\%HDF5_VERSION%\\%HDF5_VSVERSION%"
+      HDF5_VERSION: "1.12.0"
+
+    - PYTHON: "C:\\Python37"
+      TOXENV: "py37-test-deps"
+      HDF5_VSVERSION: "14"
+      HDF5_DIR: "C:\\hdf5\\%HDF5_VERSION%\\%HDF5_VSVERSION%"
+      HDF5_VERSION: "1.12.0"
+
+    - PYTHON: "C:\\Python38"
+      TOXENV: "py38-test-deps"
+      HDF5_VSVERSION: "14"
+      HDF5_DIR: "C:\\hdf5\\%HDF5_VERSION%\\%HDF5_VSVERSION%"
+      HDF5_VERSION: "1.12.0"
 
 install:
   # We need wheel installed to build wheels
--- a/h5py/_hl/compat.py
+++ b/h5py/_hl/compat.py
@@ -3,8 +3,9 @@ Compatibility module for high-level h5py
 """
 import sys
 from os import fspath, fsencode, fsdecode
+from ..version import hdf5_built_version_tuple
 
-WINDOWS_ENCODING = "mbcs"
+WINDOWS_ENCODING = "utf-8" if hdf5_built_version_tuple >= (1, 10, 6) else "mbcs"
 
 
 def filename_encode(filename):
--- a/h5py/tests/test_h5d_direct_chunk.py
+++ b/h5py/tests/test_h5d_direct_chunk.py
@@ -10,55 +10,55 @@ class TestWriteDirectChunk(TestCase):
     def test_write_direct_chunk(self):
 
         filename = self.mktemp().encode()
-        filehandle = h5py.File(filename, "w")
+        with h5py.File(filename, "w") as filehandle:
 
-        dataset = filehandle.create_dataset("data", (100, 100, 100),
-                                            maxshape=(None, 100, 100),
-                                            chunks=(1, 100, 100),
-                                            dtype='float32')
+            dataset = filehandle.create_dataset("data", (100, 100, 100),
+                                                maxshape=(None, 100, 100),
+                                                chunks=(1, 100, 100),
+                                                dtype='float32')
 
-        # writing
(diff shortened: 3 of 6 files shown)


appveyor.yml
```diff
@@ -23,19 +23,39 @@ environment:
...
+      TOXENV: "py37-test-deps"
+      HDF5_VSVERSION: "14"
+      HDF5_DIR: "C:\\hdf5\\%HDF5_VERSION%\\%HDF5_VSVERSION%"
+      HDF5_VERSION: "1.12.0"
+
+    - PYTHON: "C:\\Python38"
+      TOXENV: "py38-test-deps"
+      HDF5_VSVERSION: "14"
+      HDF5_DIR: "C:\\hdf5\\%HDF5_VERSION%\\%HDF5_VSVERSION%"
+      HDF5_VERSION: "1.12.0"
```
takluyver, line 58: I think Appveyor was often the last one to finish already; we may decide it's not worth running the full product of Python versions and HDF5 versions, especially as we have Windows tests on Azure too. But I'm happy to see how it goes for a while before that.

h5py/tests/test_h5d_direct_chunk.py
```diff
@@ -10,56 +10,56 @@ class TestWriteDirectChunk(TestCase):
...
-            read_data = filehandle["data"][i]
-            numpy.testing.assert_array_equal(array[i], read_data)
+        with h5py.File(filename, "r") as filehandle:
+            for i in range(10):
+                read_data = filehandle["data"][i]
+                numpy.testing.assert_array_equal(array[i], read_data)
 
 
 @ut.skipUnless(h5py.version.hdf5_version_tuple >= (1, 10, 2), 'Direct Chunk Reading requires HDF5 >= 1.10.2')
+@ut.skipIf('gzip' not in h5py.filters.encode, "DEFLATE is not installed")
```
takluyver, line 36: We do still have gzip on at least some of the Windows builds, though, right?

h5py/_hl/compat.py
```diff
@@ -3,8 +3,9 @@
 """
 import sys
 from os import fspath, fsencode, fsdecode
+from ..version import hdf5_built_version_tuple
 
-WINDOWS_ENCODING = "mbcs"
+WINDOWS_ENCODING = "utf-8" if hdf5_built_version_tuple >= (1, 10, 6) else "mbcs"
```
takluyver, line 8: Where did you find that this is the case? I've found [the blog post announcing 1.10.6](https://www.hdfgroup.org/2019/12/release-of-hdf5-1-10-6-newsletter-170/) and the [release notes](https://confluence.hdfgroup.org/display/support/HDF5%201.10.6#releasenotes), neither of which seem to mention it. I also found [forum](https://forum.hdfgroup.org/t/non-english-characters-in-hdf5-file-name/4627) [threads](https://forum.hdfgroup.org/t/wide-string-equivalent-of-h5fcreate-or-h5fopen/5482) about the issue, but no-one saying it's been fixed. :confused:

aragilar: #839 has comments about the fix arriving. Interestingly https://jira.hdfgroup.org/browse/HDFFV-11026?jql=labels%20%3D%20Windows suggests it wasn't announced due to problems with testing (it's not clear which encoding the reporter used).

takluyver: @epourmal do you have any newer information about unicode filenames on Windows? There are some hints that it was done for 1.10.6, while [HDFFV-10691](https://jira.hdfgroup.org/browse/HDFFV-10691) is 'in progress' and seems to say it might be done for 1.12.1.

If this is now supported, do we pass UTF-8 filenames to `H5Fopen`, or do we need to do something special for it?

takluyver:
@aragilar do you want to separate the CI change (which I think is ready to merge) from the filesystem encoding change, which I'd still like to get confirmation of?

aragilar:
The reason I included the filename fix in the PR is the Windows CI would fail otherwise (and I think that particular test was introduced in response to earlier Windows issues with unicode filenames...). I suspect even if utf-8 is not supported properly yet, we're going to need confirmation of what the encoding we're supposed to be using for 1.10.6 and later (as mbcs is failing).

takluyver:
Looking at the code, I can see it's opening files with a UTF-8 function [in 1.10.6](https://bitbucket.hdfgroup.org/projects/HDFFV/repos/hdf5/browse/src/H5win32defs.h?at=refs/tags/hdf5-1_10_6#56), and [not in 1.10.5](https://bitbucket.hdfgroup.org/projects/HDFFV/repos/hdf5/browse/src/H5win32defs.h?at=refs/tags/hdf5-1_10_5#57).

[This is the commit](https://bitbucket.hdfgroup.org/projects/HDFFV/repos/hdf5/commits/750b5c293076b6a446088fa3020e4e0787d489d7#) that changed it, and it claims to solve a number of issues. Yet the ones I can see - [HDFFV-10691](https://jira.hdfgroup.org/browse/HDFFV-10691) and [HDFFV-10413](https://jira.hdfgroup.org/browse/HDFFV-10413) are still showing as unresolved and with 'fix version' as a future version.

I guess if this makes the tests pass, we should do it. It would be good if HDF group would document it, though!

Issue #412, opened 2014-03-07
MPI issue with zero-length slices in collective mode


andrewcollette:
From the list (https://groups.google.com/forum/#!topic/h5py/eIRkkpXruoA):

First of all, thanks for the fantastic work on h5py and the parallel implementation!  I'm using the parallel framework to write distributed portions (numpy arrays) of a global data array to a single dataset.  To improve performance, I wanted to use the high-level context manager, from Andrew Collette's branch, for collective writes as described in this thread:

https://groups.google.com/forum/#!topic/h5py/wDgTjGho0dY

This works well and seems to be improving the write-time scaling for our small test cases.  The issue is that occasionally I only want to write a subspace of the global array (one particular plane, pencil, or entry), to which some processes generally contribute no data, and this seems to cause my current implementation to hang when I add the collective context manager.

I've attached a simple example of what I'm doing, which considers a 2D array distributed along the first dimension.  

To write the first global column to a dataset, I just write the first column of the local data to the proper part of the global column dataset, which works fine with and without the collective context manager.

To write the first global row, however, only process 0 contributes data and the rest write empty slices to empty slices of the global row dataset.  This works without the collective context manager, but hangs with it during the process 0 write.

Now for this simple example we could clearly just do a single non-collective write from process 0, but the real program is generalized to higher dimensions and slices along any axes of the global array, so generally I will still need/want collective writes while some processes won't have any data to contribute.

It seems like two ways forward might be to either:
1) Somehow specify just a subset of processes to take part in a collective write (maybe through a flag to the context manager?).  No idea how feasible this is.
or 
2) Somehow include length-0 contributions from some processes in the collective context.

PR #1634, opened 2020-08-09
Fix for zero length collective IO - rebased

### LINKED ISSUE #965: MPI hang during collective writes due to erroneous elision of ranks with nothing to write

aragilar:
Closes #965, rebased version of #1206. As there were a bunch of changes which were adding something, then removing it, I've squashed eveything as it made it much easier to rebase.


aragilar:
It seems the tests are broadcasting, not sure why?

takluyver:
`test_collective_write` and `test_collective_write_empty_rank` both write a scalar (0D) value into a 2D selection, i.e. broadcasting it.

We implement [some special casing](https://github.com/h5py/h5py/blob/0076bc4880bfa8b311338b21ddcd0e28cb27c443/h5py/_hl/dataset.py#L912-L937) which in this case pre-broadcasts the 0D data to 1D. I'd suggest that the broadcasting check should happen before that - it's an implementation detail, and it's still logically broadcasting if we do it with numpy before passing data to HDF5.

kburns:
I just want to add a "+1" for this since it's something we've been working around for many years (#412 !!).  I haven't poked around in the h5py internals before, but I might be able to help out with a little direction.

PR #1637, opened 2020-08-09
Add missing objects to H5P API docs

takluyver:
Also rearrange the H5D page to match other similar page (H5F, H5G), with a 'Functional API' section at the top.

### DIFF
--- a/docs_api/h5d.rst
+++ b/docs_api/h5d.rst
@@ -2,7 +2,18 @@ Module H5D
 ==========
 
 .. automodule:: h5py.h5d
-    :members:
+
+Functional API
+--------------
+
+.. autofunction:: open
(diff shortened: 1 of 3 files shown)


aragilar:
This looks good, were there more changes, or can I merge this?

Minor aside: our docstring style is not exactly consistent (I usually copy the same style as surrounding code), did we want to standardise on something?

takluyver:
This was all I was thinking of doing for this PR.

We should probably try to be roughly consistent within the high-level and the low-level APIs - it doesn't necessarily matter if the convention's a bit different between those. I tend to go for fairly loose guidelines for docstrings, but I know the arguments for having stricter standards, so if you want to have a common format I'll try to stick to it.

Issue #1053, opened 2018-06-07
Unable to create attribute (Object header message is too large)


kitchoi:
* Operating System: Windows 7, Mac OSX 10.11
 * Python version: 2.7, 3.6
 * Where Python was acquired: EDM
 * h5py version: 2.7.0
 * HDF5 version: 1.8.16

Attempting to write a long array of strings as an HDF5 attribute results in an error.
To reproduce:

```
import os
import shutil
import tempfile

import numpy
import h5py

print(h5py.version.info)

attrs = numpy.arange(3117, dtype=numpy.int64).astype("S")
tempdir = tempfile.mkdtemp()
filepath = os.path.join(tempdir, "tmp.h5")

try:
    with h5py.File(filepath, "w") as fhandle:
        fhandle.attrs["Name"] = attrs
finally:
    os.remove(filepath)
    shutil.rmtree(tempdir)
```

Traceback:
```
---------------------------------

h5py    2.7.0
HDF5    1.8.16
Python  3.6.0 |Enthought, Inc. (x86_64)| (default, Mar  3 2017, 03:26:01) 
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.57)]
sys.platform    darwin
sys.maxsize     9223372036854775807
numpy   1.13.3

h5py version: 2.7.0
hdf5 version: 1.8.16
Traceback (most recent call last):
  File "VM_shared/create_attr.py", line 18, in <module>
    fhandle.attrs["Name"] = attrs
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper (/Users/vagrant/workspace/lay-eggs/c25aeab8/pisi/tmp/h5py-2.7.0-2/work/h5py-2.7.0/h5py/_objects.c:2846)
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper (/Users/vagrant/workspace/lay-eggs/c25aeab8/pisi/tmp/h5py-2.7.0-2/work/h5py-2.7.0/h5py/_objects.c:2804)
  File "/Users/kchoi/.edm/envs/h5py36/lib/python3.6/site-packages/h5py/_hl/attrs.py", line 93, in __setitem__
    self.create(name, data=value, dtype=base.guess_dtype(value))
  File "/Users/kchoi/.edm/envs/h5py36/lib/python3.6/site-packages/h5py/_hl/attrs.py", line 188, in create
[...]

PR #1638, opened 2020-08-11
Exposing H5Pset_attr_phase_change function when creating Attributes

dota17:
Close issue #1053 

the new added test case in h5py/tests/test_attrs.py still failed, but work well in h5py/tests/test_h5p.py, I haven't figure it out. Maybe I have missed something important, so please allow me to mark this PR as WIP.
@takluyver @aragilar @tacaswell

### DIFF
--- a/h5py/_hl/attrs.py
+++ b/h5py/_hl/attrs.py
@@ -126,6 +126,9 @@ class AttributeManager(base.MutableMappingHDF5, base.CommonStateObject):
             # First, make sure we have a NumPy array.  We leave the data type
             # conversion for HDF5 to perform.
             if not isinstance(data, Empty):
+                if numpy.size(data) > 3116:
+                    plist = self._id.get_create_plist()
+                    plist.set_attr_phase_change(0, 0)
                 data = base.array_for_new_object(data, specified_dtype=dtype)
 
             if shape is None:
--- a/h5py/_hl/attrs.py
+++ b/h5py/_hl/attrs.py
@@ -126,9 +126,6 @@ class AttributeManager(base.MutableMappingHDF5, base.CommonStateObject):
             # First, make sure we have a NumPy array.  We leave the data type
             # conversion for HDF5 to perform.
             if not isinstance(data, Empty):
-                if numpy.size(data) > 3116:
-                    plist = self._id.get_create_plist()
-                    plist.set_attr_phase_change(0, 0)
                 data = base.array_for_new_object(data, specified_dtype=dtype)
 
             if shape is None:
--- a/h5py/h5a.pyx
+++ b/h5py/h5a.pyx
@@ -47,6 +47,9 @@ def create(ObjectID loc not None, char* name, TypeID tid not None,
         Link access property list for obj_name
     """
 
+    if space.get_simple_extent_npoints() > 3116:
+        # reset the dataspace extent
+        space.set_extent_simple(())
     return AttrID(H5Acreate_by_name(loc.id, obj_name, name, tid.id,
             space.id, H5P_DEFAULT, H5P_DEFAULT, pdefault(lapl)))
 
--- a/h5py/h5a.pyx
+++ b/h5py/h5a.pyx
@@ -47,9 +47,6 @@ def create(ObjectID loc not None, char* name, TypeID tid not None,
         Link access property list for obj_name
     """
 
-    if space.get_simple_extent_npoints() > 3116:
-        # reset the dataspace extent
-        space.set_extent_simple(())
     return AttrID(H5Acreate_by_name(loc.id, obj_name, name, tid.id,
             space.id, H5P_DEFAULT, H5P_DEFAULT, pdefault(lapl)))
 
--- a/news/attr_phase_change.rst
+++ b/news/attr_phase_change.rst
@@ -12,7 +12,7 @@ Deprecations
 Exposing HDF5 functions
 -----------------------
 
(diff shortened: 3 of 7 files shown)


h5py/_hl/attrs.py
```diff
@@ -126,6 +126,9 @@ def create(self, name, data, shape=None, dtype=None):
             # First, make sure we have a NumPy array.  We leave the data type
             # conversion for HDF5 to perform.
             if not isinstance(data, Empty):
+                if numpy.size(data) > 3116:
+                    plist = self._id.get_create_plist()
+                    plist.set_attr_phase_change(0, 0)
```
takluyver, line 131: From the docs for [H5Pset_attr_phase_change](https://portal.hdfgroup.org/display/HDF5/H5P_SET_ATTR_PHASE_CHANGE), this is something that has to be done when creating the object/group to which the attributes belong, not when adding an attribute. `get_create_plist()` returns a *copy* of the creation property list - it can't be used to modify the object.

I'd recommend exposing the get/set functions in the low-level API first, without trying to use them in the high-level API.

dota17: Ok, got it. thanks a lot~

dota17: the problem is we create the group/dataset before creating attributes, and in many cases, we don't even create attributes, we can't know the value info about the attributes before we create it, so we can't decide how to set its value.

h5py/h5a.pyx
```diff
@@ -47,6 +47,9 @@ def create(ObjectID loc not None, char* name, TypeID tid not None,
         Link access property list for obj_name
     """
 
+    if space.get_simple_extent_npoints() > 3116:
+        # reset the dataspace extent
+        space.set_extent_simple(())
```
takluyver, line 52: I don't fully understand this, and it now seems completely separate from the get/set_attr_phase_change methods. The changes to expose those two HDF5 methods in the low-level API look fine - would you like to do those changes by themselves first, and we can merge them pretty quickly, and then discuss this in a separate PR?

If you want to keep them together, can you explain what this is meant to do? I don't think modifying the extent (shape) of the data space makes sense. And what is the significance of 3116?

dota17: I just want to try to modify the data space of an attribute to solve the problem (at least test cases pass). This is really different from the previous changes. I will still try to use set/get_attr_phase_change, but I still have some questions, as I comment before: 
> the problem is we create the group/dataset before creating attributes, and in many cases, we don't even create attributes, we can't know the value info about the attributes before we create it.

In my understanding, we should set `max_compact&min_dense` parameter in set_attr_phase_change() according the size of an attribute when creating a group/dataset, that is to say, before we set the group/dataset, we need to know the size of the attribute exactly. But the reality is that we don't know.

so I don't know how to made it. the problem in https://github.com/h5py/h5py/issues/1053#issue-330325080 is that object header message space is limited, so I realize we can change attribute's data space to break its limitation. but I know this is away from our original plan.
The critial value is from #1053, I don't know why it's 3117, but not 65535:

> We generally consider the maximum size of an attribute to be 64K bytes. The library has two ways of storing attributes larger than 64K bytes: in dense attribute storage or in a separate dataset. Using dense attribute storage is described in this section, and storing in a separate dataset is described in the next section.

so there are two ways:
a) set `set_attr_phase_change(0,0)` in the right place to store an attribute in dense (this will be very similar to store large number of attributes).
b) create a new dataset to store that large attribute.

@takluyver what do you think?

takluyver: As far as I know, there isn't anything we can do to *automatically* solve #1053. What you're doing here makes the error disappear, but it doesn't store all the data. I've just tried adding a line to your test to check the value of the attribute:

```python
np.testing.assert_array_equal(grp.attrs['data'], data)
```

And only the first value from the array is stored. This is worse than the error!

Exposing the attr_phase_change methods means someone who knows they need to store larger attributes can do so, even if requires a few lines of extra code. That's a useful step by itself.

dota17: Do you mean, at the moment, we just exposing these two methods? if so, then I will remove the related tests in test_attrs.py

P.S. in my local, if I turn on the track_order when I creating file/group/dataset, then these test cases will pass all. 
but if I test with the code, but outsite of the h5py project, there will occur `record is already in B-tree` error.

takluyver:
Please, leave the defaults when creating objects alone for now, and just expose these two methods so users can decide what to do with them. If setting these limits affects the order tracking, or anything else, that's useful information for those issues, but it doesn't necessarily mean we want to change the defaults.

dota17:
> Please, leave the defaults when creating objects alone for now, and just expose these two methods so users can decide what to do with them. If setting these limits affects the order tracking, or anything else, that's useful information for those issues, but it doesn't necessarily mean we want to change the defaults.

OK, will do tomorrow~

news/attr_phase_change.rst
```diff
@@ -0,0 +1,30 @@
...
+
+Deprecations
+------------
+
+* <news item>
+
+Exposing HDF5 functions
+-----------------------
+
+* H5Pset_attr_phase_change - https://support.hdfgroup.org/HDF5/doc/RM/RM_H5T.html#Property-SetAttrPhaseChange
```
takluyver, line 15:
```suggestion
* H5Pset_attr_phase_change - https://portal.hdfgroup.org/display/HDF5/H5P_SET_ATTR_PHASE_CHANGE
```

dota17:
@takluyver 👍 Thanks for your help.

takluyver:
Thanks! I'll merge this once the CI has finished.

PR #1642, opened 2020-08-14
Fix reporting errors from fileobj driver

### LINKED ISSUE #1639: BUG: issue with writing string datasets to a file buffer
Writing an `h5py.string_dtype()` to a `h5py.File` creating using a writable buffer (i.e. `h5py.File(open(..., "wb"), "w")`) raises an error. Writing to a file created internally by `h5py.File` (i.e. `h5py.File("...", "w")`) does not have this issue. Writing a numerical dataset to a writable buffer does not have this issue.

SSCCE:

```py
import h5py
import numpy

def test(file):
    group = file.create_group("group")
    data = numpy.array('6b035227-da6b-4d71-88fc-f3594748cee9', dtype=object)
    dataset = group.create_dataset("data", data=data, dtype=h5py.string_dtype())

file = h5py.File("/tmp/test1", "w")
test(file)  # works
file_like = open("/tmp/test2", "wb")
file = h5py.File(file_like, "w")
test(file)  # raises
```

Traceback is:

```
  File "h5py/h5fd.pyx", line 158, in h5py.h5fd.H5FD_fileobj_read
io.UnsupportedOperation: read

The above exception was the direct cause of the following exception:

  File "help.py", line 14, in <module>
    test(file)  # raises
  File "help.py", line 7, in test
    dataset = group.create_dataset("data", data=data, dtype=h5py.string_dtype())
  File "/home/dallan/.conda/envs/py37/lib/python3.7/site-packages/h5py/_hl/group.py", line 136, in create_dataset
    dsid = dataset.make_new_dset(self, shape, dtype, data, **kwds)
  File "/home/dallan/.conda/envs/py37/lib/python3.7/site-packages/h5py/_hl/dataset.py", line 170, in make_new_dset
    dset_id.write(h5s.ALL, h5s.ALL, data)
  File "h5py/h5d.pyx", line 222, in h5py.h5d.DatasetID.write
  File "h5py/_proxy.pyx", line 164, in h5py._proxy.dset_rw
  File "h5py/_proxy.pyx", line 93, in h5py._proxy.H5PY_H5Dwrite
  File "h5py/defs.pyx", line 495, in h5py.defs.H5Dwrite
  File "h5py/h5fd.pyx", line 169, in h5py.h5fd.H5FD_fileobj_write
SystemError: <built-in method seekable of _io.FileIO object at 0x7f0a11eeffa0> returned a result with an error set
```

h5py    2.10.0
HDF5    1.10.4
Python  3.7.7 (default, May  7 2020, 21:25:33) 
[GCC 7.3.0]
sys.platform    linux
numpy   1.19.0

Platform is specifically Ubuntu 18.04.

takluyver:
The error value for `herr_t` is -1; returning a different value on errors means HDF5 thinks operations have succeeded when they haven't, resulting in confusing error messages, as in #1639.

Closes #1639.

I'm expecting this to fail CI at first, based on my local testing. I want to see if it fails for all HDF5 versions or just some.

### DIFF
--- a/docs/high/file.rst
+++ b/docs/high/file.rst
@@ -125,6 +125,12 @@ a better option may be to store temporary data on disk using the functions in
 
 .. literalinclude:: ../../examples/bytesio.py
 
+.. warning::
+
+   When using a Python file-like object for an HDF5 file, make sure to close
+   the HDF5 file before closing the file object it's wrapping. If there is an
+   error while trying to close the HDF5 file, segfaults may occur.
+
 .. _file_version:
 
 Version bounding
--- a/h5py/h5fd.pyx
+++ b/h5py/h5fd.pyx
@@ -122,7 +122,7 @@ cdef void *H5FD_fileobj_fapl_copy(PyObject *old_fa) with gil:
     Py_INCREF(<object>new_fa)
     return new_fa
 
-cdef herr_t H5FD_fileobj_fapl_free(PyObject *fa) except 1 with gil:
+cdef herr_t H5FD_fileobj_fapl_free(PyObject *fa) except -1 with gil:
     Py_DECREF(<object>fa)
     return 0
 
@@ -134,7 +134,7 @@ cdef H5FD_fileobj_t *H5FD_fileobj_open(const char *name, unsigned flags, hid_t f
     f.eoa = 0
     return f
 
-cdef herr_t H5FD_fileobj_close(H5FD_fileobj_t *f) except 1 with gil:
+cdef herr_t H5FD_fileobj_close(H5FD_fileobj_t *f) except -1 with gil:
     Py_DECREF(<object>f.fileobj)
     stdlib_free(f)
     return 0
@@ -150,7 +150,7 @@ cdef haddr_t H5FD_fileobj_get_eof(const H5FD_fileobj_t *f, H5FD_mem_t type) exce
     (<object>f.fileobj).seek(0, libc.stdio.SEEK_END)
     return (<object>f.fileobj).tell()
 
-cdef herr_t H5FD_fileobj_read(H5FD_fileobj_t *f, H5FD_mem_t type, hid_t dxpl, haddr_t addr, size_t size, void *buf) except 1 with gil:
+cdef herr_t H5FD_fileobj_read(H5FD_fileobj_t *f, H5FD_mem_t type, hid_t dxpl, haddr_t addr, size_t size, void *buf) except -1 with gil:
     cdef unsigned char[:] mview
     (<object>f.fileobj).seek(addr)
     if hasattr(<object>f.fileobj, 'readinto'):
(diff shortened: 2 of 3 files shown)


takluyver:
Good consistency there: docs and static checks are OK, everything that tries to run the tests fails - all platforms, all HDF5 versions.

The failure is triggered by a couple of tests like this, where a file object is created but then deliberately invalidated:

https://github.com/h5py/h5py/blob/75a964bfe3e7ed0ff215435c60a02ebca915f5c3/h5py/tests/test_file2.py#L219-L223

This is what I believe happens:

1. Inside `File.close()`, h5py gets another reference to the file ID, and calls `h5i.dec_ref` on it until the ref count hits 0 and HDF5 tries to close it.
2. HDF5 frees some of the internal state relating to the file.
3. Closing (which may include calling methods like write, flush, truncate as well as close itself) fails.
4. HDF5 bails out with an error before cleaning up the ID
5. When we call `FileID.close()`, it's still wrapping what HDF5 claims is a valid ID, so it tries to close it again
6. HDF5 tries to access memory it freed in step 2, and segfaults.

I can avoid this specific sequence of events, but I can't currently see how to avoid an ID that HDF5 believes is valid but really isn't. And HDF5 seems to keep an internal list of IDs to clean up when the program exits, which triggers a segfault again.

takluyver:
I've resorted to skipping one test to avoid the tests segfaulting.

I thought about making the driver lie, and report success even when closing failed, so that HDF5 would clean up its file reference. This is how it was working before this PR. It might be palatable if only the driver's close function needed to do that, but even the write function would need to lie, which doesn't seem acceptable.

I also tried to expose and use [H5Iremove_verify](https://portal.hdfgroup.org/display/HDF5/H5I_REMOVE_VERIFY) to remove the ID from HDF5's tracking after we failed to close it in the test. But this doesn't work with types defined by HDF5 itself, only with extension types you register.

@epourmal: Depending on how far custom file drivers are meant to be supported, I think there's a flaw in HDF5 here. An error while closing a file seems to leave it in an invalid state, where `H5Iis_valid` says its ID is valid, but `H5Idec_ref` can trigger a segfault.

takluyver:
I'll merge this in a couple more days unless anyone objects.

ajelenak:
This issue was reported to The HDF Group and I investigated it using this code:
```python
from io import BytesIO
import h5py
fileobj = BytesIO()
f = h5py.File(fileobj, 'w')
fileobj.close()
f.close()
```
with Python 3.6.12, HDF5-1.12.0, and h5py code from commit a1605fa.

The Python process on my Mac is killed on `f.close()` by a `SIGABORT` signal raised by a failed HDF5 assertion:

    Assertion failed: (H5AC_cache_is_clean(f, 3)), function H5F__dest, file H5Fint.c, line 1215.
    Abort trap: 6

during the process of closing the HDF5 file.

I did not notice anything wrong in the HDF5 library's behavior. Once the reference count of the file HDF5 identifier reached one and `H5Idec_ref` was called again, the library started the process of closing that file and a failed flush operation eventually led to the failed assertion. Before starting the file close process, the reference count of the file identifier was greater than one so `H5Iis_valid` correctly  reported the identifier as valid every time it was called.

takluyver:
That sounds somewhat different to what I saw. In either case, we generally try to preserve the guarantee that mistakes in Python code can't cause a hard crash (segfault, SIGABRT, or similar). But in this case, I can't see any way of ensuring that in h5py, short of lying to HDF5 that the operations are succeeding when they're not.

From my perspective, HDF5 should be able to survive bugs in custom file drivers (or code using them). It can return errors - even refuse to do any further HDF5 operations at all - but it shouldn't bring down the process.

ajelenak:
As far as the HDF5 library, I did not observe any invalid state of its internal information, which is good. At the time the assertion I reported failed, the library had already started reporting errors so that information could perhaps be used to avoid aborting execution. But this is moving into the feature request territory and such things take time to review, design, and eventually implement.

Issue #1548, opened 2020-05-16
Bug when nesting vlen_dtype and assigning a list with a single nested array e.g. [array([1,2,3])]


malpunek:
### What happens
I wanted to store nested vlen arrays of ints. It looks it's working, but breaks when I'm trying to assign a list with a single element (that element being an array).
### Code

```python
import h5py
import numpy as np


with h5py.File("test.hdf5", "w") as f:
    dt = h5py.vlen_dtype(h5py.vlen_dtype(np.dtype("int32")))
    caps = f.create_dataset("show_bug", (10,), dtype=dt)

    caps[0] = []  # OK
    caps[1] = [np.array([1, 2, 3]), np.array([4, 5]), np.array([67, 89, 120, 140])]  # OK
    caps[2] = [np.arange(20), np.arange(10)]  # OK
    caps[3] = [np.array([0, 1, 2, 3]), np.array([])]  # OK
    caps[4] = [np.array([0, 1, 2, 3])]  # Fails
```

### Expected result
One of the two:
  * h5py to work properly (preferred)
  * h5py to spit in my face with an error "You can't nest vlen dtypes" upon creating the dtype, with an additional warning in the docs

### Actual result
Ugly stacktrace:

```
Traceback (most recent call last):
  File "bug.py", line 12, in <module>
    caps[3] = [np.array([0, 1, 2, 3])]  # Fails
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "/home/malpunek/.local/share/virtualenvs/msc-fMydB2_4/lib/python3.8/site-packages/h5py/_hl/dataset.py", line 708, in __setitem__
    self.id.write(mspace, fspace, val, mtype, dxpl=self._dxpl)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py/h5d.pyx", line 222, in h5py.h5d.DatasetID.write
  File "h5py/_proxy.pyx", line 163, in h5py._proxy.dset_rw
  File "h5py/defs.pyx", line 3536, in h5py.defs.H5Tconvert
  File "h5py/_conv.pyx", line 771, in h5py._conv.ndarray2vlen
  File "h5py/_conv.pyx", line 808, in h5py._conv.conv_ndarray2vlen
[...]

Issue #1447, opened 2019-11-27
Exceptions are not documented


gregreen:
Exceptions are not documented at Readthedocs. For example, the documentation page for "Datasets" says the following:

> As empty datasets cannot be sliced, some methods of datasets such as read_direct will raise an exception if used on a empty dataset.

What sort of exception will be raised? This is important information to know, because catch-all try/except blocks are not good form.

A search through the documentation doesn't reveal any information on what exception classes exist. For example, searching for "Exception" leads to only a few generic references to "exception" (no class provided) and change logs noting the removal of the "H5Error" class.

aragilar:
At a guess, without reading the code, I would expect a `ValueError`, but we could definitely provide more guidance.

If someone wants to do this, I'd suggest the starting point would be to look through the high-level code (https://github.com/h5py/h5py/tree/master/h5py/_hl - attrs.py and dataset.py would be the first to look at), and see what exceptions we raise.

dota17:
Actually, it will raise a `TypeError` exception according to the test case I added before:
https://github.com/h5py/h5py/blob/8cc83410995996330c85867d0e56cbe922b709f8/h5py/tests/test_dataset.py#L221-L223

Issue #1119, opened 2018-10-22
Exception type on invalid index


woutdenolf:
Invalid indexing of a dataset raises ValueError as opposed to the usual IndexError. This has unexpected consequences. Here is a MWE:
```python
import h5py

class Minimal(object):
    
    def __init__(self,dataset):
        self.dataset = dataset

    def __getitem__(self,index):
        return self.dataset[index]
            
class MinimalFix(object):
    
    def __init__(self,dataset):
        self.dataset = dataset

    def __getitem__(self,index):
        try:
            return self.dataset[index]
        except ValueError as e:
            raise IndexError(e)
        
with h5py.File('test.h5',mode='w') as f:
    f['array'] = [range(5)]*4
    for slc in MinimalFix(f['array']):
        pass
    for slc in Minimal(f['array']):
        pass
```
It's an easy fix (change ValueError with IndexError in h5py._hl_selection._translate_int). I was going to submit a PR but it breaks a lot of tests. They expect specifically ValueError exceptions. Any chance this can be changed?

aparamon:
Your proposal does seem reasonable to me, and I'm all for consistency with Python and NumPy. But there must be consensus also with @tacaswell, @aragilar, @takluyver.

In [my recent PR](https://github.com/h5py/h5py/pull/1098), I found [h5py.get_config()](http://docs.h5py.org/en/latest/config.html#library-configuration) object really helpful. Now I think that for this and similar issues where improvement is only achievable by incompatible changes (read: string/bytes handling), `h5py.get_config()` could be (ab)used to both provide the enhancement *and* have perfect backwards compatibility by default. Do you agree that a one-liner
````py
h5py.get_config().IndexError_on_invalid_index = True
````
right after `import h5py` would be good enough compromise, in case your change might not be enabled by default?

P.S: While writing this, I realized that `h5py.get_config()` additionally providing a context manager could make it more widely usable. Not sure if it was intended this way though :-) (and threading...)

**Comment by tacaswell** [comment]
Does the exception type depend on the version of hdf5?  There is an internal mapping between the hdf5 error state and what python exception we raise.

If this changed, we should just fix it, if it has always raised `ValueError`, then we should probably instead raise

```python
class Both(ValueError, IndexError):
   ...
```

for now and think about deprecating the raising of `ValueError` in the future (it is hard to warn on though...I don't see how to warn only if the user catches the `ValueError`....).

[...]

PR #1643, opened 2020-08-15
Document supplement

dota17:
### major change:
~~rename exception type `ValueError` to `IndexError`, related issue #1119~~
+ document exception type in docs/high/dataset.rst, related issue #1447 

~~add some note when creating ragged arrays, related issue #1548~~

CC: @takluyver

### DIFF
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -420,7 +420,7 @@ dataset::
 
 The dtype of the dataset can be accessed via ``<dset>.dtype`` as per normal.
 As empty datasets cannot be sliced, some methods of datasets such as
-``read_direct`` will raise an exception if used on a empty dataset.
+``read_direct`` will raise an **TypeError** exception if used on a empty dataset.
 
 Reference
 ---------
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -420,7 +420,7 @@ dataset::
 
 The dtype of the dataset can be accessed via ``<dset>.dtype`` as per normal.
 As empty datasets cannot be sliced, some methods of datasets such as
-``read_direct`` will raise an **TypeError** exception if used on a empty dataset.
+``read_direct`` will raise a ``TypeError`` exception if used on a empty dataset.
 
 Reference
 ---------
--- a/docs/special.rst
+++ b/docs/special.rst
@@ -91,11 +91,6 @@ For example, you can create a "ragged" array of integers::
     >>> dset[0] = [1,2,3]
     >>> dset[1] = [1,2,3,4,5]
 
-But unfortunately, NumPy in particular has no real support for "ragged" array, so
-everything is workarounds. The `awkward-array <https://github.com/scikit-hep/awkward-array>`_
-package has some functionality built on top of h5py for saving ragged array as HDF5
-datasets.
-
 Single elements are read as NumPy arrays::
 
     >>> dset[0]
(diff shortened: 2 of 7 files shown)


docs/special.rst
```diff
@@ -91,6 +91,11 @@ For example, you can create a "ragged" array of integers::
     >>> dset[0] = [1,2,3]
     >>> dset[1] = [1,2,3,4,5]
 
+But unfortunately, NumPy in particular has no real support for "ragged" array, so
+everything is workarounds. The `awkward-array <https://github.com/scikit-hep/awkward-array>`_
```
takluyver, line 95: I think this is some of my words from an issue? I was writing a bit casually; I might need to have another go at it.

I'm also not sure whether to point people to awkward-array. I thought it allowed reading and writing ragged arrays as HDF5 vlen data, but it actually has its own way of saving its arrays in an HDF5 file, and it's only useful if you use the same tool to read it.

dota17: Yes, it's from https://github.com/h5py/h5py/issues/1548.
Then, I will  wait for your further suggestion.

h5py/h5g.pyx
```diff
@@ -406,7 +406,7 @@ cdef class GroupID(ObjectID):
         Retrieve a copy of the group creation property list used to
         create this group.
         """
-        return PropGCID(H5Gget_create_plist(self.id))
```
aragilar, line 409: Why the change from PropGCID to propwrap?

dota17: because I found that the `get_create_plist` in h5d, h5f, h5t do the same thing. 😃

docs/high/dataset.rst
```diff
@@ -420,7 +420,7 @@ dataset::
 
 The dtype of the dataset can be accessed via ``<dset>.dtype`` as per normal.
 As empty datasets cannot be sliced, some methods of datasets such as
-``read_direct`` will raise an exception if used on a empty dataset.
+``read_direct`` will raise an **TypeError** exception if used on a empty dataset.
```
tacaswell, line 423: Maybe use `TypeError` instead which if intersphinx is set up correctly will link to the cpython docs.

dota17: do you mean link `TypeError` to the cpthon docs?

aragilar: @dota17 
```
``TypeError``
```
or
```
`TypeError`
```
are what you want to replace `**TypeError**` with (I think the first, but maybe the second, I can never remember which one).

takluyver: The first one (double backticks) is an inline code snippet in rst. I'd use that here.

@tacaswell was suggesting that we use a role to link to the definition of `TypeError` in the Python docs. But we don't have intersphinx configured to link to the Python docs at the moment, and I don't think the [description of TypeError](https://docs.python.org/3/library/exceptions.html#TypeError) is that useful to us anyway.

h5py/_hl/group.py
```diff
@@ -475,10 +475,11 @@ def copy(self, source, dest, name=None,
        Example:
 
         >>> f = File('myfile.hdf5')
```
takluyver, line 477:
```suggestion
        >>> f = File('myfile.hdf5', 'w')
```

Like in the example you modified.

docs/high/dataset.rst
takluyver, line 423:
```suggestion
``read_direct`` will raise a ``TypeError`` exception if used on a empty dataset.
```

h5py/h5g.pyx
takluyver: That's reasonable enough. For future reference, it's a good idea to keep pull requests focused, so a documentation PR doesn't change code at all (except code in docs & examples). But this instance is OK.

Issue #1474, opened 2020-01-17
virtual dataset gets created with dtype=float32


asmeurer:
```
>>> print(h5py.version.info)
Summary of the h5py configuration
---------------------------------

h5py    2.10.0
HDF5    1.10.5
Python  3.7.3 | packaged by conda-forge | (default, Mar 27 2019, 15:43:19)
[Clang 4.0.1 (tags/RELEASE_401/final)]
sys.platform    darwin
sys.maxsize     9223372036854775807
numpy   1.16.5

```

If you create a VirtualLayout with dtype=None and then create a virtual dataset out of it, the dtype of the dataset becomes float32. 
 
```py
>>> import h5py
>>> f = h5py.File('test.hdf5', 'w')
>>> import numpy as np
>>> data = np.ones(10)
>>> data.dtype
dtype('float64')
>>> f['data'] = data
>>> vs = h5py.VirtualSource(f['data'])
>>> vl = h5py.VirtualLayout((10,))
>>> vl[0:10] = vs
>>> f.create_virtual_dataset('virtual_data', vl)
<HDF5 dataset "virtual_data": shape (10,), type "<f4">
>>> f['virtual_data'].dtype
dtype('<f4')
>>> vs.dtype
dtype('<f8')
>>> print(vl.dtype)
None
```

I would expect the dtype to either default to be the same as the source, or, if that is not possible (it seems like it might not be from the way the API works), make the dtype a required argument. At the very least, I would expect the default dtype to be float64, since that is the default dtype used by h5py when creating a new dataset. 

If you specify the dtype it works as expected:

```py
>>> vl = h5py.VirtualLayout((10,), dtype=np.float64)
>>> vl[0:10] = vs
>>> f.create_virtual_dataset('virtual_data2', vl)
<HDF5 dataset "virtual_data2": shape (10,), type "<f8">
>>> f['virtual_data2'].dtype
dtype('<f8')
```

CC @scopatz

takluyver:
> I would expect the default dtype to be float64, since that is the default dtype used by h5py when creating a new dataset.

This depends how you do it. If I create an empty dataset with no data, the default is float32; I think this is where I copied from for virtual datasets.

```
In [6]: ds = f.create_dataset('a', (10,))                                       

In [7]: ds                                                                      
Out[7]: <HDF5 dataset "a": shape (10,), type "<f4">
```

Whereas I guess you tried creating a dataset from a list of floats, which gets float64 when we convert it to a numpy array:

```
[...]

PR #1644, opened 2020-08-15
Fix virtual dataset

dota17:
### major change:
+ specify dtype in VirtualLayout, fix issue #1474 
+ change the default dtype to float64 for dataset (in numpy, the default for float is float64 regardless of the system).

CC: @takluyver

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -77,7 +77,7 @@ def make_new_dset(parent, shape=None, dtype=None, data=None, name=None,
     else:
         # Validate dtype
         if dtype is None and data is None:
-            dtype = numpy.dtype("=f4")
+            dtype = numpy.dtype("=f8")
         elif dtype is None and data is not None:
             dtype = data.dtype
         else:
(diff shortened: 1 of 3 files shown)


takluyver:
The idea from #1474 was to require a dtype in `VirtualLayout`, i.e. not provide any default. The obvious default in most cases is whatever the dtype of the source data is, but that's a bit tricky for various reasons (see [this comment](https://github.com/h5py/h5py/issues/1474#issuecomment-667207562)), so it's better to make it explicit than to surprise the user with a guess.

I'm not sure about changing the default dtype if you create a dataset with no dtype and no data. Matching numpy is nice, but so is consistency from one version to the next. Even when we're heading for 3.0, I don't think we should change everything just because we can.

takluyver:
@tacaswell @aragilar & anyone else who's interested: thoughts on changing the default dtype? I'm -0.5 on it.

aragilar:
If anything, I would think not having a default type would be better (either you provide the data, or you provide the type), -1 from me.

dota17:
Ok @aragilar  @takluyver I will leave it as-is.

takluyver:
@dota17 to be clear, we're suggesting a change from the current code. Currently it defaults to float32, and this PR makes the default float64. We're suggesting removing the default (in `VirtualLayout`), so the user has to pass a dtype in explicitly.

I'm going to close this because I think it would be easier to do this change in a new pull request, but we can reopen this one if you prefer.

PR #1646, opened 2020-08-21
Fix some todos

dota17:
Fix some legacy TODOs.

CC: @takluyver @aragilar

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -520,20 +520,6 @@ class Dataset(HLObject):
 
     @property
     @with_phil
-    def track_order(self):
-        """ Whether dataset creation order are tracked (T/F)"""
-        dcpl = self._dcpl
-        return dcpl.get_attr_creation_order() > 0
-
-    @property
-    @with_phil
-    def track_times(self):
-        """ Whether times associated with an object are tracked (T/F)"""
-        dcpl = self._dcpl
-        return dcpl.get_obj_track_times()
-
-    @property
-    @with_phil
     def scaleoffset(self):
         """Scale/offset filter settings. For integer data types, this is
         the number of bits stored, or 0 for auto-detected. For floating
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -520,6 +520,20 @@ class Dataset(HLObject):
 
     @property
     @with_phil
+    def track_order(self):
+        """ Whether dataset creation order are tracked (T/F)"""
+        dcpl = self._dcpl
+        return dcpl.get_attr_creation_order() > 0
+
+    @property
+    @with_phil
+    def track_times(self):
+        """ Whether times associated with an object are tracked (T/F)"""
+        dcpl = self._dcpl
+        return dcpl.get_obj_track_times()
+
+    @property
+    @with_phil
     def scaleoffset(self):
         """Scale/offset filter settings. For integer data types, this is
         the number of bits stored, or 0 for auto-detected. For floating
--- a/h5py/_hl/group.py
+++ b/h5py/_hl/group.py
@@ -252,12 +252,8 @@ class Group(HLObject, MutableMappingHDF5):
         """
         for k in ('shape', 'dtype', 'chunks', 'compression',
                   'compression_opts', 'scaleoffset', 'shuffle', 'fletcher32',
-                  'fillvalue'):
+                  'fillvalue', 'track_times', 'track_order'):
             kwupdate.setdefault(k, getattr(other, k))
-        # TODO: more elegant way to pass these (dcpl to create_dataset?)
-        dcpl = other.id.get_create_plist()
-        kwupdate.setdefault('track_times', dcpl.get_obj_track_times())
-        kwupdate.setdefault('track_order', dcpl.get_attr_creation_order() > 0)
 
         # Special case: the maxshape property always exists, but if we pass it
         # to create_dataset, the new dataset will automatically get chunked
@@ -267,8 +263,7 @@ class Group(HLObject, MutableMappingHDF5):
 
         return self.create_dataset(name, **kwupdate)
 
-    def require_group(self, name):
-        # TODO: support kwargs like require_dataset
+    def require_group(self, name, track_order=None):
         """Return a group, creating it if it doesn't exist.
 
         TypeError is raised if something with that name already exists that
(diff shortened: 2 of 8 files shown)


h5py/_hl/group.py
```diff
@@ -267,16 +263,15 @@ def create_dataset_like(self, name, other, **kwupdate):
 
         return self.create_dataset(name, **kwupdate)
 
-    def require_group(self, name):
```
aragilar, line 270: Should we be checking what `track_order` is, and raising if the group doesn't match the properties?

h5py/h5p.pyx
```diff
@@ -873,9 +873,9 @@ cdef class PropDCID(PropOCID):
             size = H5Pget_virtual_dsetname(self.id, index, NULL, 0)
             name = <char*>emalloc(size+1)
             try:
-                # TODO check return size
-                H5Pget_virtual_dsetname(self.id, index, name, <size_t>size+1)
-                src_dset_name = bytes(name).decode('utf-8')
+                len = H5Pget_virtual_dsetname(self.id, index, name, <size_t>size+1)
+                if len > 0:
+                    src_dset_name = bytes(name).decode('utf-8')
```
aragilar, line 878: What happens when `len == 0`? What's returned?

h5py/tests/test_dataset.py
```diff
@@ -816,14 +816,18 @@ def test_no_chunks(self):
     def test_track_times(self):
         orig = self.f.create_dataset('honda', data=np.arange(12),
                                      track_times=True)
+        self.assertEqual(orig.track_times, True)
```
aragilar, line 819: It would be nicer to use `assertTrue`/`assertFalse` or `assert` and `assert not`

h5py/h5p.pyx
dota17: oh, the docs said: `Returns a non-negative value if successful; otherwise returns a negative value.`,  
so, should we do it lile this: `if len >= 0`?

h5py/tests/test_dataset.py
dota17: ok, will do.

aragilar:
The order tracking looks good to me (though the `require_group` bit I'm unsure about, given the issues around `require_dataset`).

I'm less certain around the cython string checking, given src_dset_name is defined within an if scope (I'm not sure that case is well tested).

h5py/_hl/group.py
dota17: Sorry, I don't get it,  in line #274, it will call create_group, then it will check `track_order` in it.

aragilar: This is similar to https://github.com/h5py/h5py/pull/897, where we don't know if the arguments are trying to check whether the group is being tracked or not (e.g. `g.require_group("test", True)` could mean "create this group if it doesn't exist, and start tracking changes" or "create this group and track changes, if the group doesn't exist"). We need to know how the user wants tracking to be treated (and if we should error).

h5py/h5p.pyx
aragilar: The issue is around `src_dset_name` possibly not being defined, and should we have a fallback value or an error or do something else?

h5py/_hl/group.py
dota17: I think I understand your word now.
but sadly, I can't do it like in reuqire_dataset:
https://github.com/h5py/h5py/blob/fba98fd5b1b0ed39074b292323980096a5637448/h5py/_hl/group.py#L224-L234
cause `group` have no property like dataset. so I can't check it like: `if not track_order == grp.track_order`.

Actually, I want to check `original group` and `require group` in testcase before, but I realized it didn't work.

h5py/h5p.pyx
dota17: Yes, we should do something when an error is detected.

dota17: @aragilar should we tackle it like this: 

https://github.com/HDFGroup/hdf5/blob/e54f5a14d28cf9eb1226ce7feab2e2600d3984b4/java/src/jni/h5pDCPLImp.c#L1286-#L1300

h5py/_hl/group.py
```diff
@@ -267,17 +263,24 @@ def create_dataset_like(self, name, other, **kwupdate):
...
+                return self.create_group(name, track_order)
             grp = self[name]
+            if track_order is None:
+                track_order = h5.get_config().track_order
+            if grp._id.get_create_plist().get_attr_creation_order() == 0:
+                order = False
+            else:
+                order = True
+            if not order == track_order:
+                raise TypeError("tack_order do not match (existing %s vs new %s)" % (order, track_order))
```
takluyver, line 283: It's not clear to me if this kind of check makes sense. `require_dataset()` checks shape & dtype - the fundamental components of what a dataset is - but all of the extra parameters (including `track_order`) are only used if it creates a new dataset.

I think it's pragmatic for details like this to be unchecked - you can tell `require_*` how to create an object if necessary, but still use an existing object even if it doesn't match all your preferences. Nobody is rushing to 'fix' the checks in `require_dataset()`, and it's nearly a year since I suggested on #897 that we leave it as-is.

h5py/h5p.pyx
takluyver: Negative (error) return values are already checked and converted to Python exceptions in the autogenerated function wrappers (look in `defs.pyx`). So there's no need to handle them here.

Do we need a special case when there are 0 bytes? It may not make sense to have an empty string, but if that's what HDF5 gives us with no error, we can convert it to an empty Python string without introducing any ambiguity or inaccuracy.

So it's not clear to me what these three TODOs are for. Maybe we can just remove the TODOs and leave the code alone.

takluyver (commented):
TODOs in code can be surprisingly tricky - they often mean it's not clear how something should be done, in which case it's not easy to fix them without a fair bit of discussion.

h5py/_hl/dataset.py
```diff
@@ -518,6 +518,20 @@ def fletcher32(self):
...
+        """ Whether dataset creation order are tracked (T/F)"""
+        dcpl = self._dcpl
+        return dcpl.get_attr_creation_order() > 0
+
+    @property
+    @with_phil
+    def track_times(self):
+        """ Whether times associated with an object are tracked (T/F)"""
+        dcpl = self._dcpl
+        return dcpl.get_obj_track_times()
```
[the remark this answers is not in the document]
dota17: This wasn't the first time exposing `get_obj_track_times` in the high level API, it comes from here https://github.com/h5py/h5py/blob/15ed70a9561340a552dc7873d03db70c7489393e/h5py/_hl/group.py#L263 so I just move it into a method, and let these properties stay at the same place.

h5py/_hl/group.py
dota17: well, I will leave it as-is, just remove this kind of TODO.

dota17:
rerun to pass static-check.

PR #1652, opened 2020-08-28
FIX: init `self.mpi` as None in `setup_configure.py`

kmuehlbauer:
This fixes building `h5py` when invoked with `--mpi` (without set env var).

XRef: https://github.com/conda-forge/h5py-feedstock/pull/79 (break) and https://github.com/conda-forge/h5py-feedstock/pull/80 (fix)

### DIFF
--- a/setup_configure.py
+++ b/setup_configure.py
@@ -59,7 +59,7 @@ def get_env_options():
         'hdf5_libdir': os.environ.get('HDF5_LIBDIR'),
         'hdf5_pkgconfig_name': os.environ.get('HDF5_PKGCONFIG_NAME'),
         'hdf5_version': os.environ.get('HDF5_VERSION'),
-        'mpi': os.environ.get('HDF5_MPI'),
+        'mpi': True if os.environ.get('HDF5_MPI') == "ON" else None,
     }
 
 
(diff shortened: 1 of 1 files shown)


aragilar:
`"ON"` is explicitly documented in https://docs.h5py.org/en/stable/build.html#building-against-parallel-hdf5, so I'm not sure this is the right fix. Given the plan seems to be to move away from using flags (PEP 517/pip don't work well with flags), and use environment variables, wouldn't it make more sense to have the conda-forge feedstock use environment variables rather than flags.

kmuehlbauer:
@aragilar You're completely correct. I've changed the expression to explicitly check for `HDF5_MPI="ON"` otherwise set `self.mpi=None`. This prevents the current code from breaking if the command line switch (`--mpi`) is used. 

If using environment variables is the recommended way, it absolutely makes sense to switch to it in conda-forge. Thanks for pointing this out.

takluyver:
I'll merge this - in the context of the current code, it's a correct fix. But as @aragilar mentioned, we're thinking of getting rid of the `setup.py configure` mechanism, where you modify stored config before building, in favour of setting build options using the environment variables.

Issue #1654, opened 2020-08-31
DeprecationWarning: tostring() is deprecated. Use tobytes() instead.


xiao623:
I just got these warnings when building the libray:
```bash
build/lib.linux-x86_64-3.6/h5py/tests/test_dataset.py::TestExternal::test_contents
  /data/h5py/build/lib.linux-x86_64-3.6/h5py/tests/test_dataset.py:723: DeprecationWarning: tostring() is deprecated. Use tobytes() instead.
    assert contents == testdata.tostring()

build/lib.linux-x86_64-3.6/h5py/tests/test_h5d_direct_chunk.py: 10 warnings
  /data/h5py/build/lib.linux-x86_64-3.6/h5py/tests/test_h5d_direct_chunk.py:24: DeprecationWarning: tostring() is deprecated. Use tobytes() instead.
    dataset.id.write_direct_chunk((index, 0, 0), a.tostring(), filter_mask=1)
```
Do we need to fix this？

takluyver:
Yup - it's not urgent, but it should be easy enough.

PR #1656, opened 2020-09-03
Update ndarray comparison in tests

cydanil:
This a simple PR aiming to silence the `DeprecationWarning`s documented in #1654 , in which numpy arrays were being compared using their `tostring` method.

Thanks :)


takluyver:
Thanks!

PR #1657, opened 2020-09-05
Make arrays contiguous prior to H5T conversion

### LINKED ISSUE #1649: Writing sliced NumPy array stores wrong values
Environment
 * Operating System: Ubuntu 18.04.4 LTS
 * Python version: 3.8.3
 * Where Python was acquired: Anaconda on Linux
 * h5py version: 2.10.0
 * HDF5 version: 1.10.4

When writing a sliced NumPy array to a dataset, the value stored in the dataset does not match the value being written unless `.copy()` is called on the sliced array first.

```python
import h5py
import numpy as np

f = h5py.File('/tmp/data.hdf5', 'w')
f.create_dataset('x', (10,), dtype=h5py.vlen_dtype('bool'))
x = np.array([True, False, True, True, False, False, False])
print(x[::2]) # [ True  True False False]
f['x'][0] = x[::2]
print(f['x'][0]) # [ True False  True  True]
f['x'][0] = x[::2].copy()
print(f['x'][0]) # [ True  True False False]
```

cydanil:
This PR aims to close #1649, by forcing all arrays fed to `conv_ndarray2vlen` to be contiguous, by using `numpy.ascontiguousarray`.

I unsuccessfully tried to make use of `PyBuffer_ToContiguous` so (with appropriate imports from `cpython.buffer`):

```cython
cdef int conv_ndarray2vlen(void* ipt,
                           void* opt,
                           TypeID intype,
                           TypeID outtype) except -1:
    cdef:
        PyObject** buf_obj = <PyObject**>ipt
        vlen_t* in_vlen = <vlen_t*>opt
        void* data
        cnp.ndarray ndarray
        size_t len, nbytes
        PyObject* buf_obj0
        Py_buffer view
        Py_UCS4 contiguity = u'C'

    buf_obj0 = buf_obj[0]
    ndarray = np.ascontiguousarray(<cnp.ndarray> buf_obj0)
    len = ndarray.shape[0]
    nbytes = len * max(outtype.get_size(), intype.get_size())

    data = emalloc(nbytes)

    PyObject_GetBuffer(ndarray.data, &view, PyBUF_SIMPLE)
    PyBuffer_ToContiguous(data, &view, view.len, contiguity)
    PyBuffer_Release(&view)

    H5Tconvert(intype.id, outtype.id, len, data, NULL, H5P_DEFAULT)

    in_vlen[0].len = len
    in_vlen[0].ptr = data

    return 0
 ```

However, it does not do the job correctly:

```
AssertionError: [ True False False False] != [ True  True False False]
E       assert False
E        +  where False = all(array([ True, False, False, False]) == array([ True,  True, False, False])
E           +array([ True, False, False, False])
E           -array([ True,  True, False, False])
E           Full diff:
E           - array([ True,  True, False, False])
E           ?              -------
E           + array([ True, False, False, False])
E           ?              +++++++)
```

So I used `numpy.ascontiguousarray`, as shown in the code.
I'm not too familiar with the trade-offs of using Python imports in Cython.
Could I get an opinion?

Thanks!
Cyril

### CODE REVIEW

**Comment by takluyver** [comment]
I don't think it's a problem to call Python functions in this code - it should have the GIL, because of this bit:

https://github.com/h5py/h5py/blob/15ed70a9561340a552dc7873d03db70c7489393e/h5py/_conv.pyx#L726-L734

I think using `ascontiguousarray` does a redundant copy of the data when it's not already contiguous - from the original array to the contiguous one, and then from the contiguous array to the malloc-ed buffer. That's not necessarily a blocker, but I think it's worth investigating if there's an easy way to do it in one step - I was hoping `PyBuffer_ToContiguous` would do that, but maybe there's another way.

File: h5py/tests/test_sliced_arrays.py [hunk 1]
```diff
@@ -0,0 +1,28 @@
+import numpy as np
+import h5py
+from h5py import File
+from .common import TestCase
+
+
+class TestSlicedArrays(TestCase):
+    """Sliced Arrays are stored correctly"""
```
takluyver, line 8: Minor naming nitpick: we specifically want to test a non-contiguous array. Slicing with a step > 1 is the easiest way to get that, but a slice like `arr[5:10]` wouldn't show it.

Also, I think this can be added in an existing file - maybe in the class `h5py.tests.test_dataset.TestVlen`.

takluyver:
@cydanil I think I've got it working with the buffer API, and just a single copy. I've made a PR against your branch: https://github.com/cydanil/h5py/pull/1.

takluyver:
Thanks! I'd still like the new test to be integrated with the existing `TestVlen` class.

Other than that, I think this is good. As I've contributed to it, I'll give other maintainers a couple of days to have a look, but if no-one objects I'll merge it.

takluyver:
:+1:

takluyver:
Oh, I almost forgot - can you add a [news fragment](https://regro.github.io/rever-docs/news.html) for this change?

cydanil:
Ok, is that the correct way to do it?

news/non-contiguous-arrays.rst
```diff
@@ -0,0 +1,30 @@
...
+
+Exposing HDF5 functions
+-----------------------
+
+* <news item>
+
+Bug fixes
+---------
+
+* Correctly store non-contiguous arrays, such as numpy slices.
```
takluyver, line 19: Specifically, it's storing them as HDF5 variable length data that was failing before.

Otherwise, this looks good, thanks.

cydanil: Thanks, I've amended the comment.

takluyver: Did something go wrong with committing it? The text I can see on Github still looks the same as before.

cydanil: I forgot to do got add, yes

takluyver:
Thanks @cydanil !

PR #1658, opened 2020-09-08
Add some benchmarks

dota17:
### major changes:
+ fix timeout failure in benchmark_slicing.py
+ remove unnecessary code in benchmark_slicing.py
+ add benchmarks for group/dataset creation etc.

#### TODO: 

- [x] add readme.rst in benchmarks

CC: @takluyver


benchmarks/README.rst
```diff
@@ -0,0 +1,70 @@
...
+
+.. _ASV documentation: https://asv.readthedocs.io/
+
+
+Writing benchmarks
+------------------
+
+See `ASV documentation`_ for basics on how to write benchmarks.
+
+Some things to consider:
```
takluyver, line 47: How much of this is specific to h5py? I think it's a good idea to keep the README quite short - briefly describe how we're using asv in h5py, and point to the ASV documentation for everything else.

benchmarks/asv.conf.json
```diff
@@ -127,12 +126,12 @@
     "html_dir": ".asv/html",
 
     // The number of characters to retain in the commit hashes.
-    // "hash_length": 8,
+    "hash_length": 8,
```
takluyver, line 129: Why are things changing in this config file?

benchmarks/benchmarks/benchmark_dataset.py
```diff
@@ -0,0 +1,164 @@
...
+        self.f.close()
+        self._td.cleanup()
+
+    def time_create_dataset(self):
+        self.f.create_dataset('dataset', (100,), dtype='f')
+
+    def time_create_empty_dataset(self):
+        self.f.create_dataset('empty', dtyp='f')
+
+class TimeSuite:
```
takluyver, line 81: Reusing a name will mean the second class overwrites the first.

benchmarks/benchmarks/benchmark_dataset.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def teardown(self):
+        self.f.close()
+        self._td.cleanup()
+
+    def time_create_dataset(self):
+        self.f.create_dataset('dataset', (100,), dtype='f')
+
+    def time_create_empty_dataset(self):
+        self.f.create_dataset('empty', dtyp='f')
```
takluyver, line 79:
```suggestion
        self.f.create_dataset('empty', dtype='f')
```

Did you try running the benchmarks? I don't think this can have worked.

benchmarks/benchmarks/benchmark_dataset.py
```diff
@@ -0,0 +1,164 @@
...
+            vsource = h5py.VirtualSource(filename, 'data', shape=(100,))
+            self.layout[n] = vsource
+
+        self.f = h5py.File('VDS.h5', 'w', libver='latest')
+
+    def teardown(self):
+        self.f.close()
+
+    def time_virtual_dataset(self):
+        self.f.create_virtual_dataset('vdata', self.layout, fillvalue=-1)
```
takluyver, line 118: I haven't looked at all of these in detail, but does it make sense to time only this bit, leaving the layout/source bits in setup? For that matter, does it make sense to benchmark creating virtual datasets at all? I can probably make a good case that it's not particularly important for performance.

The more general point is that we need to design benchmarks that can tell us something interesting, which isn't always easy.

benchmarks/benchmarks/benchmark_dataset.py
```diff
@@ -0,0 +1,164 @@
...
+        self.f = h5py.File(self.path, 'w')
+        self.dset = self.f.create_dataset('data', data= np.arange(512).reshape(8,8,8))
+        self.indices = slice(2,8).indices(4)
+
+    def teardown(self):
+        self.f.close()
+        self._td.cleanup()
+
+    def time_write_with_indices(self):
+        self.dset[self.indices] = 0
```
takluyver, line 164: I don't think this is doing what you think it's doing.

aragilar:
We don't seem to run the benchmarks as part of the CI, I think we should think about how/where/when we want to run these.

dota17:
> We don't seem to run the benchmarks as part of the CI, I think we should think about how/where/when we want to run these.

I probably know how to do it. I'll try it in next days, and maybe we should also add an item in Sphinx documentation.

takluyver:
I think benchmarks on CI are kind of tricky - you really want to run them on a consistent system with no other significant work competing for the CPU, memory or disk access, so that one run can be compared to the next. I don't think that's what any of the free CI services will offer, for obvious reasons.

I've got a Raspberry Pi sitting around not doing much. Maybe I can get that set up to run benchmarks regularly.

dota17:
> I think benchmarks on CI are kind of tricky - you really want to run them on a consistent system with no other significant work competing for the CPU, memory or disk access, so that one run can be compared to the next. I don't think that's what any of the free CI services will offer, for obvious reasons.

You are right. I think I should stop moving forward on this, because I realize that `asv` can't automatically run on CI, when running on a new machine, it has an interactive process to collect various configuration of the machine.

> I've got a Raspberry Pi sitting around not doing much. Maybe I can get that set up to run benchmarks regularly.

Raspberry Pi YES 🚀 
I have a Raspberry Pi too, it's one of the most fun digital products in the world :)

aragilar:
I wonder if we want to move the benchmarks to their own repository, as that will make it easier to run the latest set of benchmarks against different commits without having to copy files around?

takluyver:
I *think* asv already takes care of running the same benchmark code against library code from different commits. It seems I can write a new benchmark and then run it against older commits, in any case. But I don't particularly have a good argument for having them in the same repo, if you think a separate repo is better.

aragilar:
Cool, I didn't realise asv could do that.

Issue #1653, opened 2020-08-30
'setup.py check' is deprecated


takluyver:
Our `checkreadme` tox environment uses the `setup.py check` command:

https://github.com/h5py/h5py/blob/f525d1c180f6b4bc59726faf833fd50a26d57d83/tox.ini#L80-L81

This prints a warning that it is deprecated, and we should build a distribution and then use `twine check` instead.

PR #1659, opened 2020-09-09
Use 'twine check' to check readme

dota17:
Fix https://github.com/h5py/h5py/issues/1653

cc: @takluyver

### DIFF
--- a/.authors.yml
+++ b/.authors.yml
@@ -128,6 +128,7 @@
   email: chenguopingdota@163.com
   aliases:
     - chenguoping
+    - Chenguoping
   num_commits: 1
   first_commit: 2020-07-15 20:34:45
   github: dota17
--- a/tox.ini
+++ b/tox.ini
@@ -77,9 +77,10 @@ commands=
 
 [testenv:checkreadme]
 skip_install=True
-deps=readme_renderer
+deps=twine
 commands=
-    python setup.py check -s -r
+    python setup.py sdist
+    twine check dist/*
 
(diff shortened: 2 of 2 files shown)


takluyver:
Thanks! It's just the ever-mystifying authors check complaining now:

```
'chenguopingdota@163.com' does not have 'Chenguoping' listed! Please add the name to 'chenguopingdota@163.com' in '.authors.yml'. For example:

- name: Chenguoping
  email: chenguopingdota@163.com

Or add an alias:

- name: Some Body Else
  email: chenguopingdota@163.com
  aliases:
    - Chenguoping
```

dota17:
Yes, I see it, but I have no idea why it complains again 😅

takluyver:
@scopatz is this case sensitive? It's asking for 'Chenguoping', althought 'chenguoping' is already listed as an alias.

takluyver:
For that matter, what does it use aliases for? It seems like the email address in the commit should be enough to identify the author, and we can use the `name:` field in attribution messages.

dota17:
I tried to change email/name in `.author.yml`, or change the local git configuration by `git config --global user.name/user.email`, none of them successed.

takluyver:
I don't exactly know, but we need to make it pass somehow before we can merge the PR, otherwise it will fail for everyone.

dota17:
OK, thanks! The static check turns out green now.

takluyver:
Super. This is ready to go.

PR #1662, opened 2020-09-10
Configure build with environment variables only

takluyver:
This removes the `setup.py configure` command, and the ability to set build configuration in a separate step before building. All the build config is now set by environment variables when you build - as we're already doing on CI.

The motivation is to simplify the build scripts. There's no longer a need to combine the current command line, environment variables and config saved from a previous `setup.py configure` command. This will hopefully make it easier to follow where things come from.

There's still a little bit of state - it saves the normalised config in the JSON file after a successful build, and passes `force=True` to Cython if any of the config has changed since then. I'm not sure if this is necessary - I suspect it isn't - but I left it for now, trying to imitate what the previous code was doing with `config.rebuild_required`.


setup_configure.py
```diff
@@ -43,131 +44,81 @@ def stash_config(dct):
...
-            bool(self.hdf5),
-            bool(self.hdf5_pkgconfig_name)
+            bool(hdf5_includedir or hdf5_libdir),
+            bool(hdf5),
+            bool(hdf5_pkgconfig_name)
         ]) > 1:
             raise ValueError(
                 "Specify only one of: HDF5 lib/include dirs, HDF5 prefix dir, "
                 "or HDF5 pkgconfig name"
             )
```
takluyver, line 97: Should we get rid of this check? It's meant to avoid confusion where a person thinks *this* option applies but the code would actually use *that* option, by rejecting any ambiguity. That might be less likely with fewer ways to specify the same things.

OTOH, "In the face of ambiguity, refuse the temptation to guess"  is still a good principle.

aragilar: I think we'd want to warn at a minimum, even if we don't error.

takluyver: I'll leave it as an error for now. I think it's easy to miss build warnings, especially when there's lots of other output.

kif:
While I don't like the idea of a "setup.py configure", and the risk of having inconsistent path in the saved files,
having to pass every options thing via environment variable makes things more cumbersome, especially on windows.
Why not passing those option directly to build (without pickling them), in addition to the environment variables ?
Of course this feature could be appended later on. 

I asked some packager their point of view as it impact them much.

t20100:
I had a quick look at conda and debian packaging: for the `configure`, both are currently using arguments rather than environment variables (though they could be using env var).

- https://github.com/conda-forge/h5py-feedstock/blob/master/recipe/build.sh
- http://deb.debian.org/debian/pool/main/h/h5py/h5py_2.10.0-9.debian.tar.xz

takluyver:
I'm not surprised that downstream packaging recipes are currently relying on `setup.py configure`, but I think it's reasonable to change how building works in a major release. It's a relatively small group of people who need to handle it, and we're still providing a clear way to configure build options.

> having to pass every options thing via environment variable makes things more cumbersome, especially on windows.
Why not passing those option directly to build (without pickling them), in addition to the environment variables ?

The environment variables are more convenient as soon as you're using `pip install .` rather than `python setup.py install`, as is now recommended. And having just one way to pass the information is simpler, even if it's a bit more verbose.

t20100:
Yes it is definetely better to remove the `setup.py configure` step and it is the right time to do so!

It was more about env. var. versus arguments and what was used for packaging.
BTW, with pip you can still pass args to the build step through `--global-options`.

takluyver:
> BTW, with pip you can still pass args to the build step through --global-options.

I don't think `--global-option` is currently used if it's building with PEP 517 (as it will by default now that we have a `pyproject.toml` file). But even if it's possible, it's fairly ugly: `--global-option="--hdf5-dir /path/to/hdf5"` - and that's even before thinking about quoting the path. Environment variables seem neater.

aragilar:
Looking at Debian's build setup (https://salsa.debian.org/science-team/h5py/-/blob/master/debian/rules - note that most of the code there appears to be there to allow installing a serial and mpi version at the same time), I think switching to environment variables might actually simplify it.

aragilar:
@scopatz @RizzerOnGitHub @terjeros @ArchangeGabriel As the distro maintainers who seem to have been the most recently active, do you have any objections to the dropping of the configure command or requiring environmental variables? If not, I'll merge this on Monday (as it's a long weekend here).

RizzerOnGitHub:
I don't have any objection in principle, so long as it's well documented.  And you've got that covered in the patch to build.rst.

The Debian build tool is pybuild, which runs
```
python3 setup.py configure
python3 setup.py build
python3 setup.py install
```
Operationally your change is to redefine `class configure` as `class BuildConfig`, so the configuration handling is done in `setup.py build` instead of `setup.py configure`.

I can't see any particular reason why Debian can't work with that.

takluyver:
We just merged some substantial changes to the CI in #1674, so I'm going to close & reopen this to re-run the tests with those changes.

ArchangeGabriel:
If I got things right, I just need to replace:
```
python setup.py configure --mpi
python setup.py build
```
with:
```
HDF5_MPI=ON python setup.py build
```
Right? Then definitively not an issue for me/us (Arch).

takluyver:
Yup, that's pretty much it. You may need to take the step after that into consideration as well - if you then run `pip install .` or `python setup.py install`, it's effectively calling `setup.py build` internally, so you may need to make sure that sees `HDF5_MPI=ON` as well.

ArchangeGabriel:
Right, good point, but not an issue either. ;)

takluyver:
I'll merge this in a couple more days if there are no objections. If anyone wants more time to look at it, just let me know.

PR #1665, opened 2020-09-11
Require a dtype in VirtualLayout

### LINKED ISSUE #1474: virtual dataset gets created with dtype=float32

takluyver:
The previous default would convert integer data to floats, which is surprising. I think most people would expect that if you don't specify a dtype, it reflects the dtype of the source data. But that's trickier than it sounds - there might not be any source data, or you might be describing source data that's not available yet. So let's keep it simple, and just require the caller to pass a dtype.

This is an API break, but aimed for a major release. If people were happy with the previous behaviour, all they need to do is pass `VirtualLayout(..., dtype='=f4')` to preserve it.

Closes #1474

### DIFF
--- a/docs/vds.rst
+++ b/docs/vds.rst
@@ -86,7 +86,7 @@ found in the examples folder:
 Reference
 ---------
 
-.. class:: VirtualLayout(shape, dtype=None, maxshape=None)
+.. class:: VirtualLayout(shape, dtype, maxshape=None)
 
    Object for building a virtual dataset.
 
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -139,7 +139,7 @@ def make_new_dset(parent, shape=None, dtype=None, data=None, name=None,
     return dset_id
 
 
-def make_new_virtual_dset(parent, shape, sources, dtype=None, name=None,
+def make_new_virtual_dset(parent, shape, sources, dtype, name=None,
                           maxshape=None, fillvalue=None):
     """ Return a new low-level dataset identifier for a virtual dataset """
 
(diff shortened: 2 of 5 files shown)


takluyver:
(I'm tackling easy issues myself because I want to get a 3.0 release ready)

asmeurer:
+1. This would have saved me some pain.

takluyver:
I'll merge this in a couple of days if there are no objections.

PR #1666, opened 2020-09-11
Bump minimum numpy on Python 3.7 to 1.14.5

takluyver:
Numpy 1.14.3 doesn't have wheels on PyPI for Python 3.7, so installing h5py from source tries to build numpy from source as well, for the PEP 517 build environment. This only appears to affect Python 3.7 - the specific versions for Python 3.6 and 3.8 have numpy wheels available. 1.14.5 is the first release with Python 3.7 wheels available.

I'm also thinking of pointing to the `--no-build-isolation` flag in the docs about installing h5py from source; I suspect it's often what people will want. But that won't be part of this PR.


takluyver:
I'll merge this in a couple more days unless there are objections.

takluyver:
This also makes Python 3.7 CI builds quicker on Linux & Windows, because the `mindeps` tox environment doesn't need to build numpy from source. It doesn't appear to make much difference to the Mac build, for some reason.

kif:
As there are little changes in micro version of numpy, it looks perfectly reasonable to me. I tend also to select the latest version of a minor branch as those are only bugfix version (i.e. upgrading for the user is advised _and_ guaranties there are no API change)

takluyver:
Thanks @kif

PR #1669, opened 2020-09-16
Remove pavement.py & other outdated files related to building on Windows

### LINKED ISSUE #1664: Remove pavement.py files?
It looks like this file was once used to do releases, but I don't think it's been used recently. It hasn't been updated to create wheels, for instance. Even the tool it's written for - [Paver](https://pypi.org/project/Paver/#history) - doesn't seem to have been updated for a couple of years.

There's also `windows/pavement.py`, which is meant to help with building on Windows. But that's only had a few cleanup PRs since 2014, so I suspect it's more of a distraction than a help by now.

Is anyone still using either of those files, or can we clean them up?

takluyver:
Closes #1664.


takluyver:
This one is simple cleanup, so I'll merge it in a day if no-one objects.

Issue #1489, opened 2020-02-04
h5py does not import 'mpi4py.MPI'


1313e:
To assist reproducing bugs, please include the following:
h5py    2.9.0
HDF5    1.10.4
Python  3.6.5 | packaged by conda-forge | (default, Apr  6 2018, 13:39:56) 
[GCC 4.8.2 20140120 (Red Hat 4.8.2-15)]
sys.platform    linux
sys.maxsize     9223372036854775807
numpy   1.17.3

When opening an HDF5-file using `h5py.File('test.hdf5', 'w', driver='mpio')`, it seems that `h5py` assumes that `mpi4py.MPI` has already been imported (as importing just `mpi4py` will not import it because of the way the package works).
Therefore, executing the above, even with `mpi4py` installed, will crash, as [this line](https://github.com/h5py/h5py/blob/84d7b7e58177e2c1b63ae37115bb4a88f6a90049/h5py/_hl/files.py#L48) attempts to access `mpi4py.MPI`.

I know that you can theoretically speaking assume that `mpi4py.MPI` is available when someone uses the `'mpio'` driver, as they have to import `mpi4py.MPI` in order to initialize the MPI processes, but this is still a bug in my opinion.

takluyver:
I think you're right; do you want to fix the import on the line above?

We're now trying to avoid loading `mpi4py.MPI` at import time (see #1552) so we don't unexpectedly trigger MPI initialisation. But when someone opens a file with the mpio driver, we can import it.

takluyver:
@dota17 this should be a one-line fix, and I'd like to have it solved for version 3.0. Would you like to do a PR for it?

dota17:
Of course. I will do it tomorrow.

PR #1670, opened 2020-09-17
Import mpi4py.MPI instead of mpi4py when using 'mpio' driver mode

dota17:
Close https://github.com/h5py/h5py/issues/1489


takluyver:
Thanks, this looks good.

Issue #1604, opened 2020-07-31
Set up CI with HDF5 development version


takluyver:
This is something we've discussed elsewhere - I'll add links if I find the issues. We want to test h5py against latest development versions of HDF5.

This may well break quite often because of changes in HDF5, so I don't want it to be part of the normal CI setup in this repository, which is comparing changes in h5py against a (normally) stable version of HDF5. Rather, I'd like to set up a separate repository with a scheduled (weekly? daily?) CI job which will clone both HDF5 and h5py, build them both, and run h5py's test suite. This would give us an easy way to see if h5py is working with the development branch of HDF5, without getting in the way when it doesn't.

I've tagged this 'good first issue'. It's not an *easy* issue, but it is quite standalone. You'll need to figure out: git, building HDF5 and h5py from source, running h5py's tests with `pytest`, and configuration of some CI system. So far, we're using Azure, Travis & Appveyor - it would be good to use one of those so we're not adding yet another service to deal with.

The HDF5 git repository is at https://bitbucket.hdfgroup.org/scm/hdffv/hdf5.git

takluyver:
In #1311, @pierlauro wrote a Docker file that can do this. That could be a good starting point, either with a CI service that runs a Docker build, or as a reference for the steps needed.

takluyver:
Thanks @epourmal !

PR #1674, opened 2020-09-17
Build & test wheels on Azure pipelines

takluyver:
Using [cibuildwheel](https://cibuildwheel.readthedocs.io/en/stable/setup/#azure-pipelines)

I can't believe it's really this simple, but... let's see what happens.

(Narrator: It wasn't that simple)


takluyver:
I have the Linux builds creating & publishing wheels as Azure pipeline artifacts. The wheels currently lack gzip support. In principle, Mac & Windows should be easier, because you don't need to build in a Docker container. In practice, there are probably some obstacles.

Building Linux wheels is currently quite slow, because it has to build HDF5 in the Docker container each time before building h5py. I think that the way to get round this is to create our own Docker image (manylinux + HDF5), publish it on Dockerhub, and then tell cibuildwheel to use that.

takluyver:
- Linux: working and including gzip (actually zlib) compression now ([files here](https://dev.azure.com/h5pyappveyor/h5py/_build/results?buildId=725&view=artifacts&type=publishedArtifacts))
  - [ ] Automate the creation of the Docker image.
  - [ ] Test the wheels produced by this machinery
- Mac: Failing to find libhdf5 for the build.
- Windows: Building successfully, but wheels are suspiciously small. I suspect it's not bundling HDF5 into the wheel. I don't know of any Windows equivalent of auditwheel/delocate to bundle DLLs.

takluyver:
Manylinux + HDF5 Docker images are now automatically built on DockerHub: https://hub.docker.com/repository/docker/h5py/manylinux2010_x86_64-hdf5/general

I've created an h5py org there, but apparently free organisations are only allowed up to 3 members. We could use another service instead - e.g. quay.io, where the standard manylinux images are hosted.

takluyver:
@h5py/team-h5py I think this is ready for review.

The central idea is to create and test wheels as part of the CI. The wheels are published on each build (e.g. [on build 776](https://dev.azure.com/h5pyappveyor/h5py/_build/results?buildId=776&view=artifacts&type=publishedArtifacts)), so we could manually download them and upload to PyPI after a release. By default, it retains the last 3 builds of a branch for up to 30 days; if you log in to Azure you can manually retain builds. The wheels it has built are then downloaded in a separate job and tested.

The obvious drawback is that the CI takes about twice as long - from ~10 minutes to ~20. This is partly due to the total number of jobs - it will only run 10 in parallel, so more jobs - and partly because building and then testing the wheels is slow - especially on Mac, for some reason. I've trimmed some combinations from the testing matrix to avoid it being even slower, but I'm still interested if there's any other way to speed it up.

Other points I'm still thinking about:

- The Windows builds seem to be missing the deflate filter
- Linux wheels seem to be suspiciously big compared to both the Windows & Mac wheels, and wheels on PyPI (14 MB vs ~3 MB). It looks like the HDF5 library we're bundling is bigger.
- So far I'm just building 64-bit wheels. We could add 32-bit Linux/Windows wheels, but it's yet more CI jobs, and 64-bit CPUs have been common for years now, especially in the environments where HDF5 is mostly used.

Linux wheels are built in an image derived from the manylinux2010 images, defined here: https://github.com/h5py/hdf5-manylinux

takluyver:
Oh, at least part of what's making Mac builds slower is that Numpy 1.17.4 doesn't have a Python 3.8 wheel for Mac. Should we bump that to 1.17.5?

takluyver:
To cut down the size of the Linux wheels, I have reduced the level of debug information included in the compiled h5py modules (following [this comment](https://github.com/cython/cython/issues/2102#issuecomment-413763954)), and disabled debug symbols in libhdf5 entirely.

This gets us to 4.2 MB. Removing the debug symbols entirely (`CFLAGS=-g0`) would make it 3.3 MB, which is close to the last release (2.8 MB, with an older hdf5) - but those symbols might be useful in debugging a crash. So I'm inclined to say 4.2 MB is acceptable.

aragilar:
Nice work @takluyver.

I'm wondering if it makes sense to do the wheel building/uploading in a different repo, to avoid slowing down the CI (like https://github.com/MacPython/h5py-wheels), especially if we end up producing more wheels? Looking at azure pipelines, it doesn't appear there's any major things we can speed up (e.g. using containers would presumably only speed up linux, and that would gain only seconds at best if the timing is accurate), and I think we're caching wheels from other projects.

Otherwise this looks good to me.

takluyver:
We could do that, but I think there's a fair bit of value in having it closely tied to commits in this repository - it's easy to see if something breaks the wheel builds, including pull requests, and having the wheels available can make it more convenient to manually test a change. It's definitely not ideal to make the builds slower, but I'm inclined to say it's worth it. Matplotlib builds appear to take ~30 minutes, so our CI is still not the slowest.

We could - following [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) - drop Python 3.6 support now, which would speed things up, but we'll soon want to add Python 3.9 to the CI, so we'll be back roughly where we are.

[Why do I say that this should be in the repo, but CI with a development version of HDF5 (#1604) should be separate? The development version is a moving target, so when something breaks it's not obvious if it's the latest h5py commit or a change in HDF5, reducing the value of doing it on every commit. That check will also need to build HDF5 from source every time, which is particularly time-consuming.]

takluyver:
I spent a while trying to split up the Mac wheel builds into two jobs to save a bit of time. But ugly glob patterns were spreading everywhere to try to get the right wheels to the right test jobs, and it still wasn't working, so I've abandoned that for now, and reverted to building the Mac wheels one after another in a single job.

aragilar:
Do we want to merge this (as we seems to have all the wheels building now), and leave further improvements to a new PR?

takluyver:
That sounds good to me, if you're happy with the changes here. IIRC, you've done most work on the CI previously, so I was particularly hoping you'd look at this. :slightly_smiling_face:

aragilar:
Done, #1662 I think is the last blocker (we can sort out #1300/#1614 as we do the release process).

takluyver:
Thanks @aragilar !

PR #1675, opened 2020-09-18
Replace a couple more arr.tostring with tobytes

takluyver:
These must have been overlooked before. I noticed the deprecation warnings while running the tests.


dota17:
Just found there is one left in this doc: https://github.com/h5py/h5py/blob/master/docs/strings.rst. I'm not sure whether it's worthy to change too.

takluyver:
Thanks, done!

takluyver:
Merging this one straight away, as it's a trivial fix and it's passed CI.

Issue #1672, opened 2020-09-17
h5py.h5ds functions missing docstrings


takluyver:
Several of the low-level functions in `h5ds` are missing docstrings: https://api.h5py.org/h5ds.html

For low-level functions, the docstring also describes the function signature (parameters and return values). If tackling this, please look at other low-level functions (`h5py.h5*` modules) to see the expected format.

Once this is done, we can remove the `:undoc-members:` option for Sphinx autodoc (which includes functions without a docstring).

dota17:
Great, I think I can tackle this.

PR #1676, opened 2020-09-19
Add missing docstrings in h5py.h5ds

dota17:
Fix https://github.com/h5py/h5py/issues/1672

### DIFF
--- a/h5py/h5ds.pyx
+++ b/h5py/h5ds.pyx
@@ -28,7 +28,7 @@ def set_scale(DatasetID dset not None, char* dimname=''):
 
 @with_phil
 def is_scale(DatasetID dset not None):
-    """(DatasetID dset)
+    """(DatasetID dset) => BOOL
 
     Determines whether dset is a dimension scale.
     """
--- a/h5py/h5ds.pyx
+++ b/h5py/h5ds.pyx
@@ -37,28 +37,54 @@ def is_scale(DatasetID dset not None):
 @with_phil
 def attach_scale(DatasetID dset not None, DatasetID dscale not None, unsigned
                  int idx):
+    """(DatasetID dset, DatasetID dscale, UINT idx)
+
+    Attach Dimension Scale dscale to Dimension idx of Dataset dset.
+    """
     H5DSattach_scale(dset.id, dscale.id, idx)
 
 @with_phil
 def is_attached(DatasetID dset not None, DatasetID dscale not None,
                 unsigned int idx):
+    """(DatasetID dset, DatasetID dscale, UINT idx)
+
+    Report if Dimension Scale dscale is currently attached to Dimension
+    idx of Dataset dset.
+    """
     return <bint>(H5DSis_attached(dset.id, dscale.id, idx))
 
 @with_phil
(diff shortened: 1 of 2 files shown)


dota17:
CI failed in Windows py38-hdf5110:
> ERROR:   py38-test-deps: could not install deps [pytest, pytest-cov, pytest-mpi>=0.2, numpy>=1.17.4]; v = InvocationError("'D:\\a\\1\\s\\.tox\\py38-test-deps\\Scripts\\python.EXE' -m pip install pytest pytest-cov 'pytest-mpi>=0.2' 'numpy>=1.17.4'", 2)

h5py/h5ds.pyx
```diff
@@ -37,28 +37,54 @@ def is_scale(DatasetID dset not None):
...
+    """(DatasetID dset, DatasetID dscale, UINT idx)
+
+    Attach Dimension Scale dscale to Dimension idx of Dataset dset.
+    """
     H5DSattach_scale(dset.id, dscale.id, idx)
 
 @with_phil
 def is_attached(DatasetID dset not None, DatasetID dscale not None,
                 unsigned int idx):
+    """(DatasetID dset, DatasetID dscale, UINT idx)
```
takluyver, line 49:
```suggestion
    """(DatasetID dset, DatasetID dscale, UINT idx) => BOOL
```

h5py/h5ds.pyx
```diff
@@ -76,6 +102,10 @@ def get_label(DatasetID dset not None, unsigned int idx):
 
 @with_phil
 def get_scale_name(DatasetID dscale not None):
+    """(DatasetID dscale) => STRING name_of_scale
+
+    Retrieves name of Dimension Scale dscale into buffer name.
```
takluyver, line 107:
```suggestion
    Retrieves name of Dimension Scale dscale.
```

We wouldn't generally talk about reading into a buffer, unless the function takes the buffer as an input.

takluyver:
Other than those points, this LGTM. I think the test failure was down to a random network error; I've re-run it.

h5py/h5ds.pyx
dota17: > When successful, returns a positive value for TRUE or 0 (zero) for FALSE. Otherwise returns a negative value.

I realized that the returned value are not just Boolean value(True/False), so I removed `=> BOOL` after I added it in the first.

takluyver: The HDF5 C function returns integers, but the h5py wrapper converts them into boolean values. The `<bint>` in the Cython code is casting to a *boolean integer*, a special Cython type to bridge between Python booleans and C. When it's returned as a Python object, it's automatically converted to a Python bool.

We can do this partly because Python has a separate mechanism for exceptions, whereas the C API needs to use negative numbers to indicate errors.

takluyver:
I spotted an existing docstring that was missing `=> BOOL` as well. Once the CI passes, I'll merge this.

dota17:
It seems that the CI should re-run again.

takluyver:
Thanks!

Issue #1005, opened 2018-02-20
Copy fails with complex dtype


CompPhysChris:
I'm trying to copy a complex dataset from one file to another.  I'm running on Linux Mint 18.2 and with Python 3.5.2 from the package manager.  I use h5py 2.7.1 and HDF5 1.8.18.  Below is a test script that reproduces the problem and the output I receive.

```
import numpy as np
import h5py

print('H5PY info:', h5py.version.info)

data = np.random.rand(10)+np.random.rand(10)*1j

h5_file = h5py.File('test.h5', mode='w')

h5_source = h5_file.create_dataset('Data', data=data)

h5_group = h5_file.create_group('Group')

h5_file.copy(h5_source, h5_group)
h5_dest = h5_group['Data']

print(np.all(np.isclose(h5_source[()], h5_dest[()])))
```

> H5PY info: Summary of the h5py configuration
> ---------------------------------
> 
> h5py    2.7.1
> HDF5    1.8.18
> Python  3.5.2 (default, Nov 23 2017, 16:37:01) 
> [GCC 5.4.0 20160609]
> sys.platform    linux
> sys.maxsize     9223372036854775807
> numpy   1.14.0

```
Traceback (most recent call last):
  File "/home/challtdow/workspace/pytest/bin/h5_bugtest.py", line 14, in <module>
    h5_file.copy(h5_source, h5_group)
  File "/usr/local/lib/python3.5/dist-packages/h5py/_hl/group.py", line 372, in copy
    dest_path = pp.basename(h5i.get_name(source[source_path].id))
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "/usr/local/lib/python3.5/dist-packages/h5py/_hl/dataset.py", line 429, in __getitem__
    new_dtype = readtime_dtype(self.id.dtype, names)
  File "/usr/local/lib/python3.5/dist-packages/h5py/_hl/dataset.py", line 44, in readtime_dtype
    raise ValueError("Field names only allowed for compound types")
ValueError: Field names only allowed for compound types
```

DTHaydon:
This is still a problem with: h5py 2.9
A work around is to provide the path to where the object should be create instead of the group.

`h5_file.copy(h5_source, "/".join(h5_group.name, "Data")`

PR #1677, opened 2020-09-22
Fix copy() method's bug when source is Dataset, dest is Group

dota17:
This PR try to fix https://github.com/h5py/h5py/issues/1005. 
In copy(self, source, dest, ...) method, when source is Dataset, dest is Group, the `'.'` will be passed as a slice to the dataset, so there is a crash at `dest_path = pp.basename(h5i.get_name(source[source_path].id))`.

### DIFF
--- a/h5py/_hl/group.py
+++ b/h5py/_hl/group.py
@@ -499,7 +499,10 @@ class Group(HLObject, MutableMappingHDF5):
                     dest_path = name
                 else:
                     # copy source into dest group: dest_name/source_name
-                    dest_path = pp.basename(h5i.get_name(source[source_path].id))
+                    if isinstance(source, dataset.Dataset):
+                        dest_path = pp.basename(h5i.get_name(dest.id))
+                    else:
+                        dest_path = pp.basename(h5i.get_name(source[source_path].id))
 
             elif isinstance(dest, HLObject):
                 raise TypeError("Destination must be path or Group object")
--- a/h5py/_hl/group.py
+++ b/h5py/_hl/group.py
@@ -497,12 +497,11 @@ class Group(HLObject, MutableMappingHDF5):
             if isinstance(dest, Group):
                 if name is not None:
                     dest_path = name
+                elif source_path == '.':
+                    dest_path = pp.basename(h5i.get_name(source.id))
                 else:
                     # copy source into dest group: dest_name/source_name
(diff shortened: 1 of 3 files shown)


h5py/tests/test_group.py
```diff
@@ -877,13 +877,17 @@ def test_copy_group_to_group(self):
...
+        grp = self.f1.create_group("grp")
 
         self.f1.copy(foo, 'bar')
         self.assertArrayEqual(self.f1['bar'], np.array([1,2,3]))
 
         self.f1.copy('foo', 'baz')
         self.assertArrayEqual(self.f1['baz'], np.array([1,2,3]))
 
+        self.f1.copy(foo, grp)
+        self.assertArrayEqual(self.f1['/grp/grp'], np.array([1,2,3]))
```
dota17, line 889: The results are not perfect, after copy, the dataset's name remains the same with the group.

h5py/_hl/group.py
```diff
@@ -499,7 +499,10 @@ def copy(self, source, dest, name=None,
                     dest_path = name
                 else:
                     # copy source into dest group: dest_name/source_name
-                    dest_path = pp.basename(h5i.get_name(source[source_path].id))
+                    if isinstance(source, dataset.Dataset):
```
takluyver, line 502: I think we should actually check `source == '.'` here, rather than its type. The same problem would also affected named datatypes stored in the file.

takluyver: Also, we can use an `elif` to have one less layer of nesting.

h5py/_hl/group.py
```diff
@@ -499,7 +499,10 @@ def copy(self, source, dest, name=None,
...
+                        dest_path = pp.basename(h5i.get_name(dest.id))
```
takluyver, line 503: I think this should use `source.id` rather than `dest.id`. That should fix the issue you noticed when writing the tests.

dota17:
Done. Thanks a lot.

takluyver:
Thanks! Can you add a file in `news/` for it? Here are the details if you need them: https://regro.github.io/rever-docs/news.html

dota17:
> Thanks! Can you add a file in `news/` for it? Here are the details if you need them: https://regro.github.io/rever-docs/news.html

Oh, that should be added in, will do now.

PR #1680, opened 2020-09-28
Ignore .swp file

dota17:
Ignore .swp file in working directory in case adding it into git commit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ MANIFEST
 .eggs/
 *.so
 *~
+*.swp
 .pytest_cache
 .coverage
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -33,3 +33,4 @@ coverage.xml
 
 # Rever
 rever/
+*.swp
```
takluyver, line 36: Putting it here implies that it's related to Rever, which isn't really true. I'd suggest putting it under `*~`, which I think is used in a similar way.

dota17: Done.

takluyver:
Thanks

PR #1681, opened 2020-09-28
Raise IndexError when index elemetns aren't in increasing order

dota17:
### major change:
+ raise IndexError when index elemetns aren't in increasing order
+ add test for bidirectional indices

### DIFF
--- a/h5py/_hl/group.py
+++ b/h5py/_hl/group.py
@@ -497,6 +497,8 @@ class Group(HLObject, MutableMappingHDF5):
             if isinstance(dest, Group):
                 if name is not None:
                     dest_path = name
+                elif source_path == '.':
+                    dest_path = pp.basename(h5i.get_name(source.id))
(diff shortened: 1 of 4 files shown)


takluyver:
Is `IndexError` clearly better than `TypeError` here? The docs for [IndexError](https://docs.python.org/3/library/exceptions.html#IndexError) specifically talk about out of range problems. We don't have to follow that strictly, but unless there's a solid reason to prefer IndexError, I think we should avoid changing what h5py already does.

dota17:
> Is `IndexError` clearly better than `TypeError` here? The docs for [IndexError](https://docs.python.org/3/library/exceptions.html#IndexError) specifically talk about out of range problems. We don't have to follow that strictly, but unless there's a solid reason to prefer IndexError, I think we should avoid changing what h5py already does.

Ahh, thanks for pointing it out.

Issue #1685, opened 2020-10-02
Cannot resize virtual dataset


bomber8013:
It seems that virtual datasets cannot be resized, even though they should be able to as mentioned in the documentation. Virtual layouts can specify unlimited maximum dimensions but cannot be resized.

Version info:
```
Summary of the h5py configuration
---------------------------------

h5py    2.10.0
HDF5    1.10.5
Python  3.7.7 (tags/v3.7.7:d7c567b08f, Mar 10 2020, 10:41:24) [MSC v.1900 64 bit (AMD64)]
sys.platform    win32
sys.maxsize     9223372036854775807
numpy   1.19.2
```
Source code:
```
import h5py
import io
import numpy as np
import numpy.testing as test


def virtual_test():
    shape = (10, 2)
    with h5py.File("virtual.h5", "w") as hfile:
        source_dset: h5py.Dataset = hfile.create_dataset(
            "source",
            data=np.ones(shape, dtype="i8"),
            maxshape=(None, 2),
            chunks=True
        )

        layout0 = h5py.VirtualLayout((10, 1), dtype="i8", maxshape=(None, 1))
        layout0_source0 = h5py.VirtualSource(source_dset)
        layout0[:, 0] = layout0_source0[:, 0]

        virtual_dset0: h5py.Dataset = hfile.create_virtual_dataset("virtual_dset0", layout0)
        test.assert_array_equal(virtual_dset0, np.ones((10, 1), dtype="i8"))

        source_dset[:, 0] = range(10)
        test.assert_array_equal(virtual_dset0, np.arange(10).reshape(10, 1))

        # layout1 = h5py.VirtualLayout((10, 1), dtype="i8")
        # layout1_source0 = h5py.VirtualSource(source_dset)
        # layout1[:, 0] = layout1_source0[:, 1]

        # virtual_dset1 = hfile.create_virtual_dataset("virtual_dset1", layout1)
        # test.assert_array_equal(virtual_dset1, np.ones((10, 1), dtype="i8"))

        # source_dset[:, 1] = range(10, 20)
        # test.assert_array_equal(virtual_dset1, np.arange(10, 20).reshape(10, 1))
        
        source_dset.resize(20, axis=0)
        virtual_dset0.resize(20, axis=0)
        test.assert_array_equal(virtual_dset0[:, 0], source_dset[:, 0])


virtual_test()
```
Error:
```
Traceback (most recent call last):
  File "h5_virtual.py", line 42, in <module>
    virtual_test()
[...]

Issue #1202, opened 2019-04-11
Unlimited dimension for one VDS dimension


mrosanes:
To assist reproducing bugs, please include the following:
 * Operating System: Debian9 
 * Python version: 3.7.3
 * Where Python was acquired: conda
 * h5py version: 2.9.0
 * HDF5 version: 1.10.4 

Hello,

According to VDS specifications, an unlimited dimension for VDS is accepted; but according to https://github.com/h5py/h5py/issues/1167, this feature is not exposed in the high-level API of h5py. At Alba synchrotron we would be really interested in such a feature.

Our use case is performing scans where each point (a source h5 file) is a single image coming from a 2D detector; all the source images would have to be mapped into a single VDS. For cases where we do not know the number of scan points in advance, an extensible VDS with an unlimited dimension would be ideal. (It could even be interesting for scans for which we know the number of points in advance, given that the scans could be stopped/aborted before finishing it).

The use case in which we are interested is described in point **8.5** of the HDF5 VDS requirements:
[Fixed dimension Full Source and Unlimited dimension Virtual Dataset Mapping](https://support.hdfgroup.org/HDF5/docNewFeatures/VDS/HDF5-VDS-requirements-use-cases-2014-12-10.pdf)

(eventually we could also be interested in point **8.3**:  Unlimited dimension Source and Unlimited dimension Virtual Dataset Mapping)

Even if recreating a VDS could be a cheap operation at the beginning, for performing long scans (with a high number of points) it could become more expensive. Having an unlimited dimension in the VDS, would avoid having to recreate the layout, mapping all the sources and recreate the VDS, each time that a new image (scan point) arrives.

bomber8013:
I discovered a possible fix for this issue at https://github.com/h5py/h5py/pull/1686#issuecomment-703189449 . Reposting the snippet below. Example outputs can be seen in referenced comment. 
```
with h5py.File("resize", "w") as f:
    source_dset = f.create_dataset(
        "source",
        data=np.arange(20),
        shape=(10, 2),
        maxshape=(None, 2),
        chunks=(10, 1),
        fillvalue=-1
    )
    layout = h5py.VirtualLayout((10, 1), np.int, (None, 1))
    layout_source = h5py.VirtualSource(source_dset)
    layout[:, 0] = layout_source[:, 1]
    for vspace, fpath, dset, src_space in layout.sources:
        vspace.select_hyperslab((0, 0), (h5py.h5s.UNLIMITED, 1))
[...]

PR #1686, opened 2020-10-02
Fix for issue #1685

bomber8013:
Tox tests pass. Fix seems to be working.

### DIFF
--- a/.authors.yml
+++ b/.authors.yml
@@ -667,3 +667,10 @@
   alternate_emails:
     - alemigo@github.com
     - 32462069+alemigo@users.noreply.github.com
+- name: Alex Porter
+  email: bomber8013@gmail.com
+  github: bomer8013
+  aliases:
+    - bomber8013
+  num_commits: 1
+  first_commit: Fri Oct 2 21:14:36 2020 +1000
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -605,7 +605,7 @@ class Dataset(HLObject):
         fixed.
         """
         with phil:
-            if self.chunks is None and not self.is_virtual:
+            if self.chunks is None and vds_support and not self.is_virtual:
                 raise TypeError("Only chunked datasets can be resized")
 
             if axis is not None:
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -605,7 +605,7 @@ class Dataset(HLObject):
         fixed.
         """
         with phil:
-            if self.chunks is None:
+            if self.chunks is None and not self.is_virtual:
                 raise TypeError("Only chunked datasets can be resized")
 
             if axis is not None:
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -605,7 +605,7 @@ class Dataset(HLObject):
         fixed.
         """
         with phil:
-            if self.chunks is None and not getattr(self, "is_virtual", True):
+            if self.chunks is None and not getattr(self, "is_virtual", False):
                 raise TypeError("Only chunked datasets can be resized")
 
             if axis is not None:
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -605,7 +605,7 @@ class Dataset(HLObject):
         fixed.
         """
         with phil:
-            if self.chunks is None and vds_support and not self.is_virtual:
+            if self.chunks is None and not getattr(self, "is_virtual", True):
                 raise TypeError("Only chunked datasets can be resized")
 
             if axis is not None:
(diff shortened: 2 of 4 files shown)


aragilar:
@bomber8013 Can you add a test for this?

.authors.yml
```diff
@@ -667,3 +667,10 @@
   alternate_emails:
     - alemigo@github.com
     - 32462069+alemigo@users.noreply.github.com
+- name: Alex Porter
+  email: bomber8013@gmail.com
+  github: bomer8013
```
takluyver, line 672:
```suggestion
  github: bomber8013
```

:wink:

takluyver:
Thanks for this. Have you checked if an expanded virtual dataset actually exposes new data? I think I recall that using a selection `source[:]` is equivalent to `source[0:len(source_dataset)]`, so you can grow the virtual dataset but the new space will be empty.

bomber8013:
> Thanks for this. Have you checked if an expanded virtual dataset actually exposes new data? I think I recall that using a selection `source[:]` is equivalent to `source[0:len(source_dataset)]`, so you can grow the virtual dataset but the new space will be empty.

In writing the test for resizing virtual datasets, suggested by @aragilar , I find that what you describe does not happen. I am going to limit the scope of the test to just allowing virtual datasets to resize. I'll make another issue for the behavior or the resizing.

bomber8013:
> > Thanks for this. Have you checked if an expanded virtual dataset actually exposes new data? I think I recall that using a selection `source[:]` is equivalent to `source[0:len(source_dataset)]`, so you can grow the virtual dataset but the new space will be empty.
> 
> In writing the test for resizing virtual datasets, suggested by @aragilar , I find that what you describe does not happen. I am going to limit the scope of the test to just allowing virtual datasets to resize. I'll make another issue for the behavior or the resizing.

In doing more investigation. Applying H5Dset_extent is actually something like a no-op. @takluyver the behavior you describe is taken care of in a snippet like this (adapted from [RFC:	HDF5 Virtual Dataset](https://support.hdfgroup.org/HDF5/docNewFeatures/VDS/HDF5-VDS-requirements-use-cases-2014-12-10.pdf)) examples:
```
with h5py.File("resize", "w") as f:
        "source",
        data=np.arange(20),
        shape=(10, 2),
        maxshape=(None, 2),
        chunks=(10, 1),
        fillvalue=-1
    )
    layout = h5py.VirtualLayout((10, 1), np.int, (None, 1))
    layout[:, 0] = layout_source[:, 1]
    for vspace, fpath, dset, src_space in layout.sources:
        vspace.select_hyperslab((0, 0), (h5py.h5s.UNLIMITED, 1))
        src_space.select_hyperslab((0, 1), (h5py.h5s.UNLIMITED, 1))
    
    virtual_dset: h5py.Dataset = f.create_virtual_dataset("virtual", layout)

print(source_dset[:, :])
print(virtual_dset[:, :])
```
Output:
```
[[ 0  1]
 [ 2  3]
 [ 4  5]
 [ 6  7]
 [ 8  9]
 [10 11]
 [12 13]
 [14 15]
 [16 17]
 [18 19]]
[[ 1]
 [ 3]
 [ 5]
 [ 7]
 [ 9]
 [11]
 [13]
 [15]
 [17]
 [19]]
```
Followed by:
```
source_dset.resize(20, 0)
print(source_dset[:, :])
print(virtual_dset[:, :])
```
Output:
```
[[ 0  1]
 [ 2  3]
 [ 4  5]
 [ 6  7]
 [ 8  9]
 [10 11]
 [12 13]
 [14 15]
 [16 17]
 [18 19]
 [-1 -1]
 [-1 -1]
 [-1 -1]
 [-1 -1]
 [-1 -1]
 [-1 -1]
 [-1 -1]
 [-1 -1]
 [-1 -1]
 [-1 -1]]
[[ 1]
 [ 3]
 [ 5]
 [ 7]
 [ 9]
 [11]
 [13]
 [15]
 [17]
 [19]
 [-1]
 [-1]
 [-1]
 [-1]
 [-1]
 [-1]
 [-1]
 [-1]
 [-1]
 [-1]]
```
I would conclude that the changes I have made so far are pretty much useless, and even the original issue probably be closed and replaced by a new issue regarding the behavior of the virtual datasets. Let me know your thoughts.

takluyver:
So IIUC, it's only possible to set up the 'UNLIMITED' selection you need for this to make sense using the low-level API (`.select_hyperslab()`), not with the high level API. That's not entirely useless, but it may cause some confusion.

My take is that virtual datasets are cheap to create, so if you add more source data, you may as well create a new virtual dataset and discard the old one. So support for unlimited selections & resizing isn't particularly interesting for me, but I'm not against adding that support if we can find a clean way to do it. Ideally that would mean both that it's clear for users and that it doesn't need much extra complexity internally.

See also #1202.

bomber8013:
Thanks for showing me #1202 . After reviewing that, it would seem what I'm trying to get done falls under that ticket. With that in mind, I'm going to close this pull request and the associated issue and work on fixing that one. Thanks!

PR #1687, opened 2020-10-03
Check the asv benchmarks

aragilar:
This adds `asv check` to (quoting command) "Import and check benchmark suite, but do not run benchmarks." I think this should mean we're checking for syntax errors and such, but not actually running the benchmarks (which might be a bit slow...). I've put this on travis just in case it does take some time, but there's nothing special about the python or HDF5 version used.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,13 @@ matrix:
         - pip install doctr
         - doctr deploy . --built-docs docs_api/_build/html
 
+    # Checks the asv benchmarks
+    - python: 3.8
+      env:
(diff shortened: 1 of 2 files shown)


takluyver:
Failing because it can't find Cython when trying to build. I think Cython gets installed in the tox environment where asv is run, but not in the environment asv creates to install h5py. However, it looks like asv disables pip's build isolation, so it's not getting *another* level of environment isolation there.

Possibly relevant: https://github.com/airspeed-velocity/asv/issues/903

PR #1688, opened 2020-10-03
Test that gzip & lzf filters are available in wheels

takluyver:
The new test is disabled by default, and enabled by an environment variable set in the CI for the jobs testing installation from a wheel.

Should we also check for szip? The docs don't guarantee its availability, but I think it's been there previously.


kif:
There are patent issues on szip ... so we cannot have any warranty on its availability.

PR #1689, opened 2020-10-05
Added function to allow unlimited resizing for VDS

bomber8013:
Added `set_unlimited` function to VDSMap to allow setting an axis to have a mapping with an unlimited dimension.

### DIFF
--- a/h5py/_hl/vds.py
+++ b/h5py/_hl/vds.py
@@ -103,10 +103,7 @@ class VirtualSource(object):
     def __getitem__(self, key):
         tmp = copy(self)
         tmp.sel = select(self.shape, key, dataset=None)
-        try:
-            key = tuple(key)
-        except TypeError:
-            key = (key,)
+        key = key if isinstance(key, tuple) else (key,)
         for i, sl in enumerate(key):
             if isinstance(sl, slice):
                 if sl.stop == h5s.UNLIMITED:
--- a/h5py/_hl/vds.py
+++ b/h5py/_hl/vds.py
@@ -22,6 +22,25 @@ class VDSmap(namedtuple('VDSmap', ('vspace', 'file_name',
                                    'dset_name', 'src_space'))):
     '''Defines a region in a virtual dataset mapping to part of a source dataset
     '''
+    def set_unlimited(self, axis: int):
+        """
+        Sets the specified axis for this source to have an unlimited selection
+        """
+        if not isinstance(axis, int):
+            raise TypeError("Axis must be an integer")
+        if axis < 0:
+            raise ValueError("Axis must be greater than 0")
+
+        for space in (self.vspace, self.src_space):
+            rank = space.get_simple_extent_ndims()
+            if not axis < rank:
+                raise ValueError("Axis {} value must be less than rank {}".format(axis, rank))
+
+            start, stride, count, block = space.get_regular_hyperslab()
+            counts = list(count)
+            counts[axis] = h5s.UNLIMITED
+            count = tuple(counts)
+            space.select_hyperslab(start, count, stride, block)
 
 
 vds_support = False
--- a/h5py/_hl/vds.py
+++ b/h5py/_hl/vds.py
@@ -22,25 +22,7 @@ class VDSmap(namedtuple('VDSmap', ('vspace', 'file_name',
                                    'dset_name', 'src_space'))):
     '''Defines a region in a virtual dataset mapping to part of a source dataset
     '''
-    def set_unlimited(self, axis: int):
-        """
-        Sets the specified axis for this source to have an unlimited selection
-        """
-        if not isinstance(axis, int):
-            raise TypeError("Axis must be an integer")
-        if axis < 0:
-            raise ValueError("Axis must be greater than 0")
 
-        for space in (self.vspace, self.src_space):
-            rank = space.get_simple_extent_ndims()
-            if not axis < rank:
-                raise ValueError("Axis {} value must be less than rank {}".format(axis, rank))
-
-            start, stride, count, block = space.get_regular_hyperslab()
-            counts = list(count)
-            counts[axis] = h5s.UNLIMITED
-            count = tuple(counts)
-            space.select_hyperslab(start, count, stride, block)
 
 
 vds_support = False
@@ -121,6 +103,20 @@ class VirtualSource(object):
     def __getitem__(self, key):
         tmp = copy(self)
         tmp.sel = select(self.shape, key, dataset=None)
+        try:
+            key = tuple(key)
+        except TypeError:
+            key = (key,)
+        for i, sl in enumerate(key):
+            if isinstance(sl, slice):
+                if sl.stop == h5s.UNLIMITED:
+                    space = tmp.sel.id
+                    rank = space.get_simple_extent_ndims()
+                    start, stride, count, block = space.get_regular_hyperslab()
+                    counts = list(count)
+                    counts[i] = h5s.UNLIMITED
+                    count = tuple(counts)
+                    space.select_hyperslab(start, count, stride, block)
         return tmp
 
 class VirtualLayout(object):
@@ -149,6 +145,20 @@ class VirtualLayout(object):
 
     def __setitem__(self, key, source):
         sel = select(self.shape, key, dataset=None)
+        try:
+            key = tuple(key)
+        except TypeError:
+            key = (key,)
+        for i, sl in enumerate(key):
+            if isinstance(sl, slice):
+                if sl.stop == h5s.UNLIMITED:
+                    space = sel.id
+                    rank = space.get_simple_extent_ndims()
+                    start, stride, count, block = space.get_regular_hyperslab()
+                    counts = list(count)
+                    counts[i] = h5s.UNLIMITED
+                    count = tuple(counts)
+                    space.select_hyperslab(start, count, stride, block)
         self.sources.append(VDSmap(sel.id,
                                    source.path,
                                    source.name,
--- a/h5py/_hl/vds.py
+++ b/h5py/_hl/vds.py
@@ -32,6 +32,39 @@ if hdf5_version >= h5.get_config().vds_min_hdf5_version:
     vds_support = True
 
 
+def _convert_space_for_key(space, key):
+    """
+    Converts the space with the given key. Mainly used to allow unlimited
+    dimensions in virtual space selection.
+    """
+    key = key if isinstance(key, tuple) else (key,)
+    type_code = space.get_select_type()
+
+    # check for unlimited selections in case where selection is regular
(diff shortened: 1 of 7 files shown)


takluyver:
Oh, I responded to the idea on #1202 before I saw this. Copying my comment from there:

-----

VDSmap is kind of an internal detail (not documented anywhere). I think that ideally, the interface for doing this would be with VirtualSource & VirtualLayout objects, something like:

layout[:h5py.h5s.UNLIMITED, 0] = layout_source[:h5py.h5s.UNLIMITED, 0]

bomber8013:
Oh, duh. I had it like that initially, by switching `None`s in the slice to `h5s.UNLIMITED` but couldn't figure out how to disambiguate that from the case where thry just meant the current size. Do you think it would be better to just expose the h5s constant or introduce something "higher level"?

takluyver:
There are two main possibilities I can think of: either we change a slice with no specified end value (`source[:]`) to mean UNLIMITED, instead of *up to the end of the current data shape*, or we allow some special symbol in slicing to mean an unlimited selection.

I'm leaning towards the second option out of caution. E.g. a dataset can have more than one unlimited dimension in `maxshape`, but a selection for a virtual dataset can only have one unlimited dimension, so `VirtualSource(dset)[:, :]` would be ambiguous. `VirtualSource(dset)[:UNLIMITED, :]` makes it explicit which dimension can grow.

I don't have a strong preference on whether we introduce a new token or use the existing `h5py.h5s.UNLIMITED`. Using the existing constant seems pretty clear to me, but maybe mixing the high-level & low-level APIs like that isn't ideal. :thinking:

scopatz:
CC @asmeurer

h5py/_hl/vds.py
```diff
@@ -102,6 +103,20 @@ def shape(self):
     def __getitem__(self, key):
         tmp = copy(self)
         tmp.sel = select(self.shape, key, dataset=None)
+        try:
+            key = tuple(key)
+        except TypeError:
+            key = (key,)
```
takluyver, line 109:
```suggestion
        key = key if isinstance(key, tuple) else (key,)
```

The `isinstance()` check matches what we do in e.g. `Dataset.__getitem__`.

h5py/_hl/vds.py
```diff
@@ -102,6 +103,20 @@ def shape(self):
...
+        for i, sl in enumerate(key):
+            if isinstance(sl, slice):
+                if sl.stop == h5s.UNLIMITED:
+                    space = tmp.sel.id
+                    rank = space.get_simple_extent_ndims()
+                    start, stride, count, block = space.get_regular_hyperslab()
+                    counts = list(count)
+                    counts[i] = h5s.UNLIMITED
```
takluyver, line 117: Unfortunately, just iterating through and modifying the `i-th` index won't quite work, because you can use ellipsis, like `obj[..., :UNLIMITED]` to slice the last dimension.

`...` is equivalent to filling any missing dimensions (`rank - (len(args) - 1)`) with `slice(None, None)`, which is equivalent to `[:]`. We should be able to rely on `select()` throwing an error if it's used incorrectly (e.g. multiple ellipsis).

h5py/tests/test_vds/test_highlevel_vds.py
```diff
@@ -398,5 +398,49 @@ def test_relative_vds(self):
...
+        ))
+        comp3 = np.vstack((
+            comp1,
+            np.full(shape=(10, 1), fill_value=0)
+        ))
+        with h5.File(self.path, "a") as f:
+            source_dset = f['source']
+            virtual_dset = f['virtual']
+            assert (comp1 == virtual_dset).all()
+            source_dset.resize(20, 0)
```
takluyver, line 437:
```suggestion
            source_dset.resize(20, axis=0)
```

Just a bit clearer. :)

h5py/tests/test_vds/test_highlevel_vds.py
```diff
@@ -398,5 +398,49 @@ def test_relative_vds(self):
...
+            np.full(shape=(10, 1), fill_value=-1)
+        ))
+        comp3 = np.vstack((
+            comp1,
+            np.full(shape=(10, 1), fill_value=0)
+        ))
+        with h5.File(self.path, "a") as f:
+            source_dset = f['source']
+            virtual_dset = f['virtual']
+            assert (comp1 == virtual_dset).all()
```
takluyver, line 436:
```suggestion
            np.testing.assert_array_equal(virtual_dset, comp1)
```

Numpy's specialised testing functions give better error messages if the comparison fails, and can pick up discrepancies that `==` hides by broadcasting. Same for the other assertions.

h5py/tests/test_vds/test_highlevel_vds.py
```diff
@@ -398,5 +398,49 @@ def test_relative_vds(self):
...
+            source_dset.resize(20, 0)
+            assert (comp2 == virtual_dset).all()
```
takluyver, line 438: I'm confused - you're now checking against a `(20, 1)` array, but the virtual dataset should still have shape `(10, 1)`. Surely it doesn't automatically resize when the source dataset resizes?

h5py/tests/test_vds/test_highlevel_vds.py
```diff
@@ -398,5 +398,49 @@ def test_relative_vds(self):
...
+        with h5.File(self.path, "w") as f:
+            source_dset = f.create_dataset(
+                "source",
+                data=np.arange(20),
+                shape=(10, 2),
+                maxshape=(None, 2),
+                chunks=(10, 1),
+                fillvalue=-1
+            )
+            self.layout = h5.VirtualLayout((10, 1), np.int, (None, 1))
```
takluyver, line 417:
```suggestion
            self.layout = h5.VirtualLayout((10, 1), np.int, maxshape=(None, 1))
```

A bit of explicit naming never hurts. :-)

h5py/tests/test_vds/test_highlevel_vds.py
bomber8013: It surely does, no manual resizing required. If this is a bug, its an awfully convenient one.

h5py/tests/test_vds/test_highlevel_vds.py
bomber8013: In my own testing I used the numpy functions, but saw the == in other people's unit tests. A bit of my inner lemming. I'll go ahead and change them.

h5py/tests/test_vds/test_highlevel_vds.py
bomber8013: Thanks!

h5py/_hl/vds.py
bomber8013: Yeah, I was doing the selection changes after using `select()` to catch those kinds of things. I don't think `select()` should be changed at all since the `UNLIMITED` hyperslab selection was added specifically for VDS, and I'm not sure if it actually works with other kinds of selections.

h5py/_hl/vds.py
bomber8013: Good call, thanks!

h5py/_hl/vds.py
takluyver: Thanks, that makes sense. The main use of selection is for reading & writing data, so I think you're right that UNLIMITED only makes sense for virtual dataset mappings, and thus should be handled here. But hopefully we can skip some of the sanity checks around ellipsis, because `select()` already does them.

h5py/tests/test_vds/test_highlevel_vds.py
takluyver: OK! I probably wouldn't design it like that, but I can't see how it could be a bug in h5py - we get the shape from HDF5.

Does that mean that a virtual dataset in a read-only file can resize itself with no warning? We currently assume that it's safe to cache a dataset's shape if it's in a read-only file.

bomber8013: A read-only opened file would never have anything change unless opened in SWMR. So VDS shouldnt affect anything in that way by itself, but your caching decision should be determined off of `writable or swmr` not just `writable`. If that case is true (haven't looked myself), then the cached value shouldn't ever get invalidated.

takluyver: But if a virtual dataset can change shape based on its source data changing, and its source data is in a different file, you could have the virtual dataset open in a read-only file and the source dataset in a read-write file. Or have I missed something?

With SWMR, as far as I know you have to call `.refresh()` to update the dataset metadata, so we're using that to invalidate cached information at the h5py level as well.

bomber8013: If the virtual source is in another file, that file would have to be opened with the same permissions/mode as the original file the virtual dataset is in. So if the original is in write mode, linked dataset is opened in write, etc. Therefore even if the dataset was in a different file, it would have the same constraints.

takluyver: I tweaked your test to illustrate what I mean. It is a problem, but this PR doesn't introduce it, it just makes it easier to see. Luckily, there's an easy workaround: call `.refresh()` to clear the cached shape.

```python
import h5py
import numpy as np
import os.path as osp
import tempfile

tmpdir = tempfile.mkdtemp()
path = osp.join(tmpdir, "resize.h5")
vds_path = osp.join(tmpdir, "resize_vds.h5")
with h5py.File(path, "w") as f:
        "source",
        data=np.arange(20),
        shape=(10, 2),
        maxshape=(None, 2),
        chunks=(10, 1),
        fillvalue=-1
    )
    layout = h5py.VirtualLayout((10, 1), np.int, maxshape=(None, 1))
    layout[:h5py.UNLIMITED, 0] = layout_source[:h5py.UNLIMITED, 1]

with h5py.File(vds_path, "w") as f:
    f.create_virtual_dataset("virtual", layout)


with h5py.File(path, "a") as f, h5py.File(vds_path, 'r') as fv:
    source_dset = f['source']
    virtual_dset = fv['virtual']

    print("Before", virtual_dset.shape)  # (10, 1)
    source_dset.resize(20, axis=0)
    print("After resize", virtual_dset.shape)  # (10, 1)
    virtual_dset.refresh()
    print("After refresh", virtual_dset.shape)  # (20, 1)
```

takluyver: I've opened #1693 about this.

takluyver:
The implementation looks good. I think the one remaining thing to do is to documented the option of using `UNLIMITED` in the [virtual datasets page](https://docs.h5py.org/en/latest/vds.html). Docs are generated from the `docs/` folder in this repo. Can you have a look at adding that? If you're not sure where to put it in, I can have a go.

takluyver:
Thanks @bomber8013 !

PR #1690, opened 2020-10-06
Set up CI for Python 3.9

takluyver:
& start dropping Python 3.6 (following [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html)).

We'll probably want to wait for numpy wheels for Python 3.9 to merge this, because building numpy tends to make the CI slow.


takluyver:
I inadvertantly started this on top of the branch for #1688. I'll fix that if necessary, but I expect that PR will be merged first anyway.

takluyver:
It appears that the `UsePythonVersion` task on Azure pipelines doesn't know about Python 3.9 yet. This is on hold until it does. :-)

aragilar:
I did some searching to get an idea of the timeline when our wheel dependencies would become available:
 * https://github.com/actions/virtual-environments/issues/1740 is tracking when 3.9 reaches azure pipelines (next week apparently).
 * The latest emails on numpy-discussion suggests numpy is waiting on azure also (and they may only provide manylinux2014 wheels, not earlier ones). The release of the wheels is tracked by https://github.com/numpy/numpy/issues/17482.
 * https://github.com/cython/cython/issues/3868 is tracking cython's 3.9 wheels, there doesn't appear to be any messages about 3.9 on either the user or dev mailing lists (though as cython is only a build dependency, it's not vital that it release wheels).

takluyver:
Thanks @aragilar !

takluyver:
I believe this is still missing from the Windows images, and certainly NumPy hasn't made wheels yet. But I'm going to close & reopen to rerun the tests and see how it goes now.

takluyver:
I've just re-run it, and Azure now has Python 3.9 on all platforms. It's still slower because the 3.9 jobs have to build numpy from source (see numpy/numpy#17482).

Building Numpy from source on Mac OS also produces a failure I haven't seen before:

> RuntimeError: Polyfit sanity test emitted a warning, most likely due to using a buggy Accelerate backend. If you compiled yourself, see site.cfg.example for information. Otherwise report this to the vendor that provided NumPy.
> RankWarning: Polyfit may be poorly conditioned

aragilar:
Accelerate is MacOS's BLAS/LAPACK (which is generally buggy for some reason, the numpy/scipy devs would have clearer idea why - I thought they'd dropped support for it). I guess the quickest fix would be to install openblas via homebrew, but I'm not sure exactly what that implies around numpy ABIs/what effect it has around a wheel we produce.

takluyver:
No, I'm not sure about that either. I also don't know on other platforms if the NumPy we build from source will be ABI compatible with wheels released by NumPy. I guess it probably is in most cases - there must be people who have built NumPy differently and then installed h5py from a wheel, and I don't think they've come to complain. But that's probably not super common.

I'm vaguely hoping that NumPy will make wheels for 3.9 pretty quickly and then we can use those. Or do you think we should get h5py 3.0 out sooner, either without wheels for 3.9, or with wheels that we're not quite so confident in?

takluyver:
NumPy did get on the wheels quickly - there is a new 1.19.3 release with wheels for Python 3.9. I have updated the dependency in `setup.py` and `tox.ini`.

takluyver:
OK, after a bit of prodding CI to restart (some of the Mac builds didn't even start the first time round, for some reason), all tests are now passing. This was the last thing I was waiting for before h5py 3.0.

takluyver:
I've decided to restore the wheels for Python 3.6 for this release. NumPy still made Python 3.6 wheels in its new release, and we've made a lot of substantial improvements in h5py 3.0, so I don't want users to have trouble getting it when it's fairly easy for us to build those wheels for a few days more.

PR #1691, opened 2020-10-07
Get processor architecture at compile time

### LINKED ISSUE #1079: Importing h5py before mpi4py causes OpenMPI warnings

takluyver:
Avoids the platform module running a subprocess during import.

Hopefully this might fix #1079.

### DIFF
--- a/h5py/h5t.pyx
+++ b/h5py/h5t.pyx
@@ -33,12 +33,10 @@ import numpy as np
 from .h5 import get_config
 
 from ._objects import phil, with_phil
-import platform
-
 
 cfg = get_config()
 
-MACHINE = platform.machine()
+DEF MACHINE = UNAME_MACHINE  # processor architecture, provided by Cython
 cdef char* H5PY_PYTHON_OPAQUE_TAG = "PYTHON:OBJECT"
 
 # === Custom C API ============================================================
--- a/news/machine-compile-time.rst
+++ b/news/machine-compile-time.rst
@@ -1,0 +1,30 @@
+New features
+------------
+
+* <news item>
+
+Deprecations
+------------
+
+* <news item>
+
+Exposing HDF5 functions
+-----------------------
+
+* <news item>
+
+Bug fixes
+---------
+
+* Avoid launching a subprocess by using ``platform.machine()`` at import time.
+  This could trigger a warning in MPI.
+
+Building h5py
+-------------
+
+* <news item>
+
+Development
+-----------
+
+* <news item>


takluyver:
Tests are passing. I think this is a good idea in general (avoiding work at import time), but it would be good if someone who's been seeing the MPI warnings can see if this fixes them.

jonaslb:
Fixes it for me! :)

takluyver:
Thanks! I'll merge this tomorrow if no-one spots any problems or wants more time to look at it.

kif:
Is cross-compilation an issue ?
Beside this it passes on ppc64le

takluyver:
I wouldn't even know how to go about cross-compiling it. Is there any likely scenario where you'd cross-compile to or from PowerPC? If we don't know of anyone who needs that, I'd say we don't need to worry about it.

aragilar:
Cross-compiling python extensions isn't a thing as far as I know (cross-compiling cpython itself seems quite difficult - http://www.trevorbowen.com/2013/10/07/cross-compiling-python-2-7-5-for-embedded-linux/). I think if this works on x86 and ppc64le, as this is only used to check for the PowerPC float detection issue, then this should be fine to merge, so I'm going to go ahead and do that.

aragilar:
Thanks @takluyver for fixing this.

takluyver:
The underlying issue - that `platform.machine()` launches an unnecessary subprocess - appears to have also been fixed for Python 3.9 (https://github.com/python/cpython/pull/12239 ).

PR #1697, opened 2020-10-10
Allow building with system lzf library via setting H5PY_SYSTEM_LZF=1

### LINKED ISSUE #430: Support building with extern lzf library
The Fedora h5py package must patch h5py to build with our system lzf library.  It would be nice is this was supported upstream.

opoplawski:
Resolves #430

### DIFF
--- a/lzf/lzf_filter.c
+++ b/lzf/lzf_filter.c
@@ -26,7 +26,7 @@
 #include <stdio.h>
 #include <errno.h>
 #include "hdf5.h"
-#include "lzf/lzf.h"
+#include "lzf.h"
 #include "lzf_filter.h"
 
 /* Our own versions of H5Epush_sim, as it changed in 1.8 */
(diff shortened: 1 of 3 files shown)


aragilar:
This looks fine to me, though I'm slightly hesitant to merge this for 3.0 (as it's quite close to the release). Can `H5PY_SYSTEM_LZF` be added to the allowed variables in tox?

opoplawski:
Like this?

takluyver:
Thanks, yup, that's right. :+1: 

I agree with @aragilar that I'd rather not merge this right before releasing 3.0. We'll do so right afterwards. But I'm not familiar enough with compilation & linking to be 100% confident reviewing a change like this, and I don't want to mess up the 3.0 release because I've overlooked something.

Kudos on fixing an issue you filed yourself 6 years ago. :wink:

Issue #1698, opened 2020-10-10
Can't work with h5py files from open file handles


Helveg:
To assist reproducing bugs, please include the following:
 * Operating System Windows 10
 * Python version 3.6
 * Pyenv build
 * h5py version (2.10 & 3.0.0rc1)
 * HDF5 version (e.g. 1.8.17)
 * The full traceback/stack trace shown (if it appears)

```
Summary of the h5py configuration
---------------------------------

h5py    3.0.0rc1
HDF5    1.12.0
Python  3.6.10 (default, Mar 20 2020, 00:41:24) 
[GCC 7.5.0]
sys.platform    linux
sys.maxsize     9223372036854775807
numpy   1.17.4
cython (built with) 0.29.21
numpy (built against) 1.12.0
HDF5 (built against) 1.12.0
```

I thought you could create HDF5 files from file objects but I get "I/O operation on closed file" errors. What's going wrong here?

```
>>> with open("test.hdf5", "w") as f:
...  h = h5py.File(f, "w")
...  h.create_group("ehe")
... 
Traceback (most recent call last):
  File "h5py/_objects.pyx", line 199, in h5py._objects.ObjectID.__dealloc__
  File "h5py/h5fd.pyx", line 169, in h5py.h5fd.H5FD_fileobj_write
ValueError: I/O operation on closed file.
Exception ignored in: 'h5py._objects.ObjectID.__dealloc__'
Traceback (most recent call last):
  File "h5py/_objects.pyx", line 199, in h5py._objects.ObjectID.__dealloc__
  File "h5py/h5fd.pyx", line 169, in h5py.h5fd.H5FD_fileobj_write
ValueError: I/O operation on closed file.
<HDF5 group "/ehe" (0 members)>
```

jangarrevoet:
with h5py.File("test.hdf5", "w") as f:
    f.create_group("ehe")

Helveg:
That's not possible in this situation, I need to write to an already existing file handle. I just reduced my example for the issue report here. It's also featured in the [docs](https://docs.h5py.org/en/stable/high/file.html#python-file-like-objects) that this should be possible starting from v2.9:

> The first argument to File may be a Python file-like object, such as an io.BytesIO or tempfile.TemporaryFile instance. This is a convenient way to create temporary HDF5 files, e.g. for testing or to send over the network.

Helveg:
The error changes slightly with different file modes, for sure I should've opened it in binary mode:

```python
[...]

PR #1704, opened 2020-10-12
Point to alternatives to using Python file-like objects

takluyver:
The `fileobj` driver is probably always going to be more error prone than using HDF5 'normally', e.g. #1698. But it's such an attractive possibility, and h5py makes it so simple, that users are naturally going to be drawn to it. This is a small attempt to point people back to using standard HDF5 file drivers where possible.


takluyver:
I'll merge this tomorrow if there are no objections.

PR #1710, opened 2020-10-15
Highlight incompatibility of changes to string conversion

takluyver:
Following discussion on the mailing list, I wanted to make these changes clearer in the release notes.


takluyver:
I also noticed we hadn't actually highlighted the end of Python 2 support.

Issue #1711, opened 2020-10-19
lzf compression filter crashes on macos with h5py3.0/HDF5 1.12.0 


t20100:
Tested on:
- macos 10.12
- python3.7
- HDF5 1.12.0
- h5py installed from azure pipeline wheel and built from current master with HDF51.12.0 built following `ci/travis/get_hdf5_if_needed.sh`

The following code crashes when calling `File.close`:
```python
In [1]: import h5py, numpy                                                      
In [2]: f = h5py.File("test.h5", "w")                                           
In [3]: f.create_dataset("data", data=numpy.arange(100), compression='lzf')     
Out[3]: <HDF5 dataset "data": shape (100,), type "<i8">
In [4]: f.close()                                                               
Python(72206,0x7fffe44623c0) malloc: *** error for object 0x7f9f808d77f0: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug
Abort trap: 6
```

We have similar issue with `hdf5plugin` compression filters (see https://github.com/silx-kit/hdf5plugin/issues/92).

Crash information:
```
Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	0x00007fffdb660d42 __pthread_kill + 10
1   libsystem_pthread.dylib       	0x00007fffdb74e457 pthread_kill + 90
2   libsystem_c.dylib             	0x00007fffdb5c6420 abort + 129
3   libsystem_malloc.dylib        	0x00007fffdb6b607f free + 530
4   h5z.cpython-37m-darwin.so     	0x000000011977620f lzf_filter + 303 (lzf_filter.c:253)
5   libhdf5.200.dylib             	0x000000011084f6ea H5Z_pipeline + 2938
6   libhdf5.200.dylib             	0x00000001101e017a H5D__chunk_flush_entry + 2042
7   libhdf5.200.dylib             	0x00000001101d8360 H5D__chunk_flush + 544
8   libhdf5.200.dylib             	0x0000000110229641 H5D__flush_real + 625
9   libhdf5.200.dylib             	0x00000001102282d8 H5D_close + 520
10  libhdf5.200.dylib             	0x000000011082e29e H5VL__native_dataset_close + 398
11  libhdf5.200.dylib             	0x00000001107f4f92 H5VL__dataset_close + 722
12  libhdf5.200.dylib             	0x00000001107f4b73 H5VL_dataset_close + 867
[...]

PR #1713, opened 2020-10-20
Add testing lzf compression with actual data

t20100:
This PR adds testing of lzf compression with actual data.

On my macos laptop, this test crashes with HDF5 1.12.0.

### DIFF
--- a/ci/azure-pipelines-steps.yml
+++ b/ci/azure-pipelines-steps.yml
@@ -10,7 +10,7 @@ steps:
     # HDF5_VSVERSION is only for Windows - it's just a fixed string elsewhere.
     # Increment the first number to clear the cache if you change e.g.
     # dependencies or build options.
-    key: 1 | HDF5 | "$(Agent.OS)" | "${{ parameters.vmImage }}" | "$(HDF5_VERSION)" | "$(HDF5_VSVERSION)"
+    key: 2 | HDF5 | "$(Agent.OS)" | "${{ parameters.vmImage }}" | "$(HDF5_VERSION)" | "$(HDF5_VSVERSION)"
     path: $(HDF5_CACHE_DIR)
   condition: and(succeeded(), ne(variables['HDF5_VERSION'], ''))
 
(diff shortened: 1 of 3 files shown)


t20100:
Trying this with compilation option to build libhdf5 1.12.0 in "production" mode (rather than debug mode by default).
This should fix the issue and close #1711.

news/add-lzf-test.rst
```diff
@@ -0,0 +1,9 @@
+Building h5py
+-------------
+
+* Build libhdf5 1.12.x in production mode
```
takluyver, line 4: I'm going to get a bit pickier about news entries after the release, having assembled them into release notes for 3.0. This kind of thing makes sense as a commit message, but release notes should be written addressing someone upgrading from a previous release: what has changed that they might care about? This change is important, but I think we're basically just getting back to the state we were in for previous releases (when wheels were built separately), so it probably doesn't need a news entry.

The 'Building h5py' section is primarily for notes that people might need if they build h5py from source, rather than details of how we build it for the wheels or the tests.

takluyver:
It's using a cached build of HDF5 on Ubuntu. You can increment the number here to invalidate the cache:

https://github.com/h5py/h5py/blob/7a413101ccaf2a8ac853c36aa887ada4e8955d98/ci/azure-pipelines-steps.yml#L11-L13

For some reason it seems to be ignoring the cache for the Mac OS wheel builds. I'm not sure why at the moment, but I'd like to work that out - it saves some build time.

news/add-lzf-test.rst
takluyver: Yup, I think we can drop the news entry for this - thanks. I'll make a PR to expand the documentation on news entries.

t20100: news entry dropped

t20100:
Cache key incremented, let's see how it goes with rebuilt libhdf5 1.12.0.

takluyver:
Thanks! And I've realised why it's not using the cache on Mac OS: there's a hardcoded check for `libhdf5.so` to use the cache, and of course it's `libhdf5.dylib` on Mac. No need to fix that in this PR, though - it works without caching.

takluyver:
Thanks @t20100 for finding & debugging this.

PR #1716, opened 2020-10-20
Only depend on 3rd party cached-property if Py<3.8.

anntzer:
On Py3.8+ we can just use the stdlib version instead.

(given that the new dependency on cached-property is not even in the current whatsnew, I assumed removing it doesn't warrant a whatsnew entry...)


takluyver:
Thanks, this makes sense. We're just about to release 3.0, so I'm holding all changes that don't fix bugs for now, but this can be merged after that.

PR #1718, opened 2020-10-22
Use cached build of HDF5 on Mac CI

takluyver:
The CI script was only checking for `libhdf5.so`, whereas on Mac this needs to be `libhdf5.dylib`.


takluyver:
cc @t20100, this follows something I spotted on your PR #1713.

I've also added a hash of the build script into the cache key, which should mean we don't need to manually increment the number in the key so often.

PR #1727, opened 2020-10-30
Remove Python 3.6 support

takluyver:
This ditches wheel building and tests on Python 3.6, and marks it as supporting 3.7 and above.

I don't think there's much in 3.7 we're eager to start using, so we could keep supporting 3.6 for a while if there's a need - perhaps without building wheels for it. But testing on lots of Python versions takes CI time, and if we don't test it then sooner or later it will break. Dropping it now is OK by [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html). So I propose to leave this for a couple of weeks to see if anyone makes a case for extending 3.6 support.

(I made a bit of a mess by accidentally pushing the first two commits to master. Sorry about that. :-( I've reverted them in master, and re-made them for this PR)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,9 +47,9 @@ matrix:
 
     # MPI tests
     # TODO: We should test with newer versions of HDF5
-    - python: 3.6
+    - python: 3.8
       env:
-      - TOXENV=py36-test-mindeps-mpi4py
+      - TOXENV=py38-test-mindeps-mpi4py
       - CC="mpicc"
       - HDF5_MPI="ON"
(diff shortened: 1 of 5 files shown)


takluyver:
That sounds reasonable enough, especially as I'm hoping to speed up the release cycle a bit (more than 1 release per year!). I've removed the milestone from this PR for now.

takluyver:
I'd like to bring this back into consideration now that 3.1 is out.

aragilar:
I think dropping 3.6 is fine (I think if people are really desperate, they can patch the setup.py file).

PR #1730, opened 2020-10-31
Properly handle reading non-native endian data. Fixes #1792

### LINKED ISSUE #1792: read_direct bug when dest and source have different shape
h5py    3.1.0
HDF5    1.12.0
Python  3.9.1 (default, Dec 29 2020, 09:45:39) 
[Clang 12.0.0 (clang-1200.0.32.28)]
sys.platform    darwin
numpy   1.19.5
cython (built with) 0.29.21
numpy (built against) 1.19.3
HDF5 (built against) 1.12.0

read_object gives an error if a dest_sel is provided but source and dest have a different shape. Example script:
```
f = h5py.File('tmp.h5', 'w')
dset = f.create_dataset("dset", (10,), dtype='int64')
dset[:] = range(10)
arr = np.zeros((12,), dtype='int32')
dset.read_direct(arr, None, np.s_[2:])
print(arr)
```
Running this results in:
```
  File "/Users/jasondet/Downloads/h5py_bug.py", line 9, in <module>
    dset.read_direct(arr, None, np.s_[2:])
  File "/usr/local/lib/python3.9/site-packages/h5py/_hl/dataset.py", line 970, in read_direct
    for mspace in dest_sel.broadcast(source_sel.mshape):
  File "/usr/local/lib/python3.9/site-packages/h5py/_hl/selections.py", line 291, in broadcast
    tshape = self.expand_shape(source_shape)
  File "/usr/local/lib/python3.9/site-packages/h5py/_hl/selections.py", line 264, in expand_shape
    raise TypeError("Can't broadcast %s -> %s" % (source_shape, self.array_shape))  # array shape
TypeError: Can't broadcast (10,) -> (8,)
```

It appears that the problem is in dataset.py line 967:
```
                dest_sel = sel.select(dest.shape, dest_sel, self)
```
In the 3rd argument, self (the source dataset) is provided for select() to make a selection on dest_sel, which seems wrong. Indeed, if I simply change "self" to "None", my example code runs without error producing the expected output:
```
[0 0 0 1 2 3 4 5 6 7 8 9]
```

Xarthisius:
This PR ensures that `H5Dread` stores data in a buffer that preserves endianess. 

Please note that I modified the original behavior of `h5py/tests/test_big_endian_file.py:test_vlen_big_endian`, which was comparing numbers stored with BE with their LE interpration.

### DIFF
--- a/h5py/tests/test_big_endian_file.py
+++ b/h5py/tests/test_big_endian_file.py
@@ -1,34 +1,50 @@
 import pytest
 
+import numpy as np
 from h5py import File
+from .common import TestCase
 from .data_files import get_data_file_path
 
 
 def test_vlen_big_endian():
     with File(get_data_file_path("vlen_string_s390x.h5")) as f:
-        assert f.attrs['created_on_s390x'] == 1
+        assert f.attrs["created_on_s390x"] == 1
 
-        dset = f['DSvariable']
-        assert dset[0] == b'Parting'
-        assert dset[1] == b'is such'
-        assert dset[2] == b'sweet'
-        assert dset[3] == b'sorrow...'
+        dset = f["DSvariable"]
+        assert dset[0] == b"Parting"
+        assert dset[1] == b"is such"
+        assert dset[2] == b"sweet"
+        assert dset[3] == b"sorrow..."
 
-        dset = f['DSLEfloat']
+        dset = f["DSLEfloat"]
         assert dset[0] == 3.14
         assert dset[1] == 1.61
         assert dset[2] == 2.71
         assert dset[3] == 2.41
         assert dset[4] == 1.2
-        assert dset.dtype == '<f8'
+        assert dset.dtype == "<f8"
 
         # Same float values with big endianess
-        assert f['DSBEfloat'][0] == pytest.approx(7.9824696849641e-157)
-        assert f['DSBEfloat'].dtype == '>f8'
+        assert f["DSBEfloat"][0] == 3.14
+        assert f["DSBEfloat"].dtype == ">f8"
 
-        assert f['DSLEint'][0] == 1
-        assert f['DSLEint'].dtype == 'uint64'
+        assert f["DSLEint"][0] == 1
+        assert f["DSLEint"].dtype == "uint64"
 
         # Same int values with big endianess
-        assert f['DSBEint'][0] == 72057594037927936
-        assert f['DSBEint'].dtype == '>i8'
+        assert f["DSBEint"][0] == 1
+        assert f["DSBEint"].dtype == ">i8"
+
+
+class TestEndianess(TestCase):
+    def test_simple_int_be(self):
+        fname = self.mktemp()
+
+        arr = np.ndarray(shape=(1,), dtype=">i4", buffer=bytearray([0, 1, 3, 2]))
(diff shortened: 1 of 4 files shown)


h5py/tests/test_big_endian_file.py
```diff
@@ -1,34 +1,50 @@
...
         assert dset[2] == 2.71
         assert dset[3] == 2.41
         assert dset[4] == 1.2
-        assert dset.dtype == '<f8'
+        assert dset.dtype == "<f8"
 
         # Same float values with big endianess
-        assert f['DSBEfloat'][0] == pytest.approx(7.9824696849641e-157)
-        assert f['DSBEfloat'].dtype == '>f8'
+        assert f["DSBEfloat"][0] == 3.14
```
aragilar, line 28: Why this change (I'm guessing this is now the correct endianess)?

h5py/tests/test_big_endian_file.py
```diff
@@ -1,34 +1,50 @@
...
 
-        assert f['DSLEint'][0] == 1
-        assert f['DSLEint'].dtype == 'uint64'
+        assert f["DSLEint"][0] == 1
+        assert f["DSLEint"].dtype == "uint64"
 
         # Same int values with big endianess
-        assert f['DSBEint'][0] == 72057594037927936
-        assert f['DSBEint'].dtype == '>i8'
+        assert f["DSBEint"][0] == 1
```
aragilar, line 35: As above, why the change?

aragilar:
This looks good to me, thanks @Xarthisius.

takluyver:
Thanks, but I'm a bit concerned that this implementation will come with non-trivial overhead. The point of `Reader` implemented in Cython is to be a fast path for small, simple reads. Using `cython -a` to get annotated output, the new lines light up bright yellow, indicating that the generated code has significant complexity:

Screenshot from 2020-11-01 11-45-04

The generated code goes through things like `__Pyx_PyObject_GetAttrStr` (for `sys.byteorder`), `PyObject_RichCompare` and `__Pyx_PyObject_IsTrue`. We can do better using something like `PyArray_IsNativeByteOrder` from the NumPy C-API (although this doesn't seem to be documented in NumPy, but Cython exposes it). In fact, since the byte order of the dataset won't change, we can precompute a boolean for whether it matches the system byte order when the `Reader` is created, and further simplify each read.

Also, if I'm reading the docs correctly, calling `.byteswap()` on a newly created empty array is redundant, just shuffling around zeros in memory. The bit that matter, AIUI, is `.newbyteorder()`, which changes the dtype for how that memory will be interpreted.

takluyver:
In principle, it would be even better to use `PyArray_SimpleNewFromDescr` and pass the dtype in (including the byte order). But all the array-creation functions which take a full dtype are commented out in the Cython file which exposes Numpy C-API functions. I guess this is because they steal references, which creates complications (https://github.com/cython/cython/issues/2722).

takluyver:
I've opened PR #1731 with the improvements I was talking about.

I'd like to get that in - assuming it works - but then I guess we should do a quick follow-up release to fix this rather serious issue.

Xarthisius:
> Thanks, but I'm a bit concerned that this implementation will come with non-trivial overhead.

Sorry about that. I was looking at Numpy API docs yesterday, but I was unable to come up with a clean solution like you did in #1731, hence I went the "pure python" route.

takluyver:
No worries. Working and slow is definitely better than efficient and broken. :wink:

PR #1731, opened 2020-11-01
Polish Cython fix for reading data in opposite byte order.

takluyver:
Follow up to #1730. This simplifies the code in the fast read path for handling data in the non-native byte order. This should have lower overhead for reading both data that already matches the native byte order, and data that doesn't.


takluyver:
I'll merge this in a couple of days if no-one objects.

PR #1734, opened 2020-11-02
Test and fix creating datasets with names in bytes

### LINKED ISSUE #1732: h5py 3.0 incompatibility with TensorFlow model serialization (multiple versions)
* Operating System: Linux (Ubuntu 18.04), macOS
 * Python version: 3.6, 3.7
 * Where Python was acquired: System (Linux), Pyenv (macOS)
 * h5py version: 3.0.0
 * HDF5 version: 1.12.0
 * The full traceback/stack trace shown:

```
  File "test.py", line 5, in <module>
    model.save('test.h5')
  File "/Users/tgaddair/.venv/horovod/stable/lib/python3.7/site-packages/tensorflow_core/python/keras/engine/network.py", line 1171, in save
    signatures)
  File "/Users/tgaddair/.venv/horovod/stable/lib/python3.7/site-packages/tensorflow_core/python/keras/saving/save.py", line 109, in save_model
    model, filepath, overwrite, include_optimizer)
  File "/Users/tgaddair/.venv/horovod/stable/lib/python3.7/site-packages/tensorflow_core/python/keras/saving/hdf5_format.py", line 103, in save_model_to_hdf5
    save_weights_to_hdf5_group(model_weights_group, model_layers)
  File "/Users/tgaddair/.venv/horovod/stable/lib/python3.7/site-packages/tensorflow_core/python/keras/saving/hdf5_format.py", line 625, in save_weights_to_hdf5_group
    param_dset = g.create_dataset(name, val.shape, dtype=val.dtype)
  File "/Users/tgaddair/.venv/horovod/stable/lib/python3.7/site-packages/h5py/_hl/group.py", line 143, in create_dataset
    if '/' in name:
TypeError: a bytes-like object is required, not 'str'
```

Here are the parameter values being passed from TensorFlow into `create_dataset`:

```
self = <Closed HDF5 group>, name = b'dense/kernel:0', shape = (3, 2)
dtype = dtype('float32'), data = None, kwds = {'track_order': False}
group = <Closed HDF5 group>
```

As you can see, TensorFlow is providing the name as bytes type, not a str, which seems to be the crux of the issue here.

Repro script (tested with TensorFlow 1.15.0 and 2.3.1):

```
import tensorflow as tf

model = tf.keras.models.Sequential()
model.add(tf.keras.layers.Dense(2, input_shape=(3,)))
model.save('test.h5')
```

Rolling back to `h5py==2.10.0` works fine.

takluyver:
Closes #1732

The docs say object names can be passed as bytes, and this works for most operations. I've added some tests of operations which take object names - it seems only creating datasets was broken.

### DIFF
--- a/h5py/_hl/group.py
+++ b/h5py/_hl/group.py
@@ -140,15 +140,10 @@ class Group(HLObject, MutableMappingHDF5):
         with phil:
             group = self
             if name:
-                if '/' in name:
-                    h5objects = [obj for obj in name.split('/') if len(obj)]
-                    name = h5objects[-1]
-                    h5objects = h5objects[:-1]
-
-                    for new_group in h5objects:
-                        group = group.get(new_group) or group.create_group(new_group)
-
                 name = self._e(name)
+                if b'/' in name.lstrip(b'/'):
+                    parent_path, name = name.rsplit(b'/', 1)
+                    group = self.require_group(parent_path)
 
             dsid = dataset.make_new_dset(group, shape, dtype, data, name, **kwds)
             dset = dataset.Dataset(dsid)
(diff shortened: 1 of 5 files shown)


harupy:
@takluyver Will a new version be released once this patch has been merged?

takluyver:
Yes, at some point in the next few days. There are several bugs which people only discovered once 3.0 was released, so I'll try to get fixes in for all of the ones I know about (so long as they're reasonably easy to solve) before doing a release.

harupy:
@takluyver Thanks for the quick reply & fix 👍

takluyver:
I'll merge this in a couple of days if no-one objects.

Issue #1735, opened 2020-11-02
[regression] writing list of integers fails


nschloe:
MWE:
```python
import h5py


f = h5py.File("out.h5", "w")

group = f.create_group("GRO")
dataset = group.create_dataset("NOM", (1,), dtype="3int8")

dataset[0] = [1, 2, 3]
```
It does work with
```python
import numpy
dataset[0] = numpy.array([1, 2, 3])
```

takluyver:
Here's the error:

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-18-dcdd83f83695> in <module>
----> 1 ds[0]  = [1, 2, 3]

h5py/_objects.pyx in h5py._objects.with_phil.wrapper()

h5py/_objects.pyx in h5py._objects.with_phil.wrapper()

~/miniconda3/envs/h5py-dev/lib/python3.7/site-packages/h5py/_hl/dataset.py in __setitem__(self, args, val)
    863             # If it's a list or similar, don't make numpy guess a dtype for it.
    864             dt = None if isinstance(val, numpy.ndarray) else self.dtype
--> 865             val = numpy.asarray(val, order='C', dtype=dt)
    866 
    867         # Check for array dtype compatibility and convert

~/miniconda3/envs/h5py-dev/lib/python3.7/site-packages/numpy/core/_asarray.py in asarray(a, dtype, order)
     83 
     84     """
---> 85     return array(a, dtype, copy=False, order=order)
     86 
     87 

TypeError: object of type 'int' has no len()
```

takluyver:
Thanks, can you try #1737? Once the build on Azure pipelines has completed, there will be wheels there which you can manually download and install if you prefer that to building from source.

nschloe:
Fix confirmed.

PR #1737, opened 2020-11-02
Fix writing a list to a dataset with a subarray dtype

takluyver:
Closes gh-1735


takluyver:
@nschloe confirmed this fixes the issue he saw. I'll merge in a day or so if no-one objects.

Issue #1739, opened 2020-11-03
test_vlen_big_endian failing on s390x


opoplawski:
Fedora Rawhide s390x
Python 3.9.0, pytest-6.0.2, py-1.9.0, pluggy-0.13.1
h5py 3.0

This is from the Fedora package build of h5py:
```
_____________________________ test_vlen_big_endian _____________________________
    def test_vlen_big_endian():
        with File(get_data_file_path("vlen_string_s390x.h5")) as f:
            assert f.attrs['created_on_s390x'] == 1
    
            dset = f['DSvariable']
            assert dset[0] == b'Parting'
            assert dset[1] == b'is such'
            assert dset[2] == b'sweet'
            assert dset[3] == b'sorrow...'
    
            dset = f['DSLEfloat']
>           assert dset[0] == 3.14
E           assert 7.9824696849641e-157 == 3.14
serial/build/lib.linux-s390x-3.9/h5py/tests/test_big_endian_file.py:18: AssertionError
_________________ TestTypeFloatID.test_custom_float_promotion __________________
self = <h5py.tests.test_h5t.TestTypeFloatID testMethod=test_custom_float_promotion>
    def test_custom_float_promotion(self):
        """Custom floats are correctly promoted to standard floats on read."""
    
        # This test uses the low-level API, so we need names as byte strings
        test_filename = self.mktemp().encode()
        dataset = b'DS1'
        dataset2 = b'DS2'
        dataset3 = b'DS3'
        dataset4 = b'DS4'
        dataset5 = b'DS5'
    
        dims = (4, 7)
    
        wdata = np.array([[-1.50066626e-09,   1.40062184e-09,   1.81216819e-10,
                           4.01087163e-10,   4.27917257e-10,  -7.04858394e-11,
                           5.74800652e-10],
                          [-1.50066626e-09,   4.86579665e-10,   3.42879503e-10,
                           5.12045517e-10,   5.10226528e-10,   2.24190444e-10,
                           3.93356459e-10],
                          [-1.50066626e-09,   5.24778443e-10,   8.19454726e-10,
                           1.28966349e-09,   1.68483894e-10,   5.71276360e-11,
                           -1.08684617e-10],
                          [-1.50066626e-09,  -1.08343556e-10,  -1.58934199e-10,
                           8.52196536e-10,   6.18456397e-10,   6.16637408e-10,
                           1.31694833e-09]], dtype=np.float32)
    
[...]

PR #1741, opened 2020-11-04
Fix test of little-endian dtype for big-endian systems

takluyver:
See discussion on #1739.

Educated guesswork: `== 'uint64'` checks for a uint64 in system-native byteorder. Hopefully `== '<u8'` checks for explicitly little-endian uint64.


takluyver:
@opoplawski confirmed (in #1739) that with this, the tests pass on a big-endian s390x machine. So merging. I'll aim for a new release soon, but I'll investigate #1742 first.

Issue #1712, opened 2020-10-20
support for the ros3 driver


satra:
hdf5 created the ros3 driver with compile support for it in both 1.12.0 and 1.10.6. 

is there any timeline for h5py support for this?

takluyver:
Does it work if you build HDF5 with that option and then build h5py yourself?

The HDF5 we bundle into prebuilt wheels for h5py is largely built with the default options. I'd hesitate to enable any non-default features without a proper understanding of what they are and why they're not enabled by default. But that shouldn't stop you using extra features if you build h5py on top of your own HDF5.

satra:
> Does it work if you build HDF5 with that option and then build h5py yourself?

we are checking that out at the moment. will post it here.

unfortunately for our users, mostly neuroscientists, building is not a viable option. at least one path is via conda-forge which had it enabled but appears to not work. the hdf5group is helping debug that. 

https://github.com/conda-forge/hdf5-feedstock/issues/125

hopefully if this clears up at least that route may deliver a working h5py with this support built in.

however, we can definitely build it for any jupyterhub instance via a docker build.

sirloon:
@satra I'm trying to achieve the same thing, I compiled hdf5 lib using `./configure --enable-ros3-vfd` then built h5py (branch 2.10.x), pointing to the HDF5 libs freshly compiled. I couldn't have this ros3 driver active though:

```
> h5py.registered_drivers()
frozenset({'core', 'family', 'fileobj', 'mpio', 'sec2', 'stdio'})
```

Explicit file open with drive='ros3' gives `ValueError: Unknown driver type "ros3"`...

Let me know how it goes and your end. Thanks!

satra:
@takluyver - is there any guidance you can provide as to how we could add a new driver to h5py. is there an easy route through `register_driver`? i don't know what an `fapl` is. do we need to write some kind of shim, or can h5py introspect the built library directly?

takluyver:
A `fapl` is a File Access Property List. HDF5 uses property lists extensively where we'd generally use default arguments in Python. To provide a specific file driver, we need to implement a function which takes a fapl and modifies it to specify the file driver to use. Like this for the mpio driver:

https://github.com/h5py/h5py/blob/27829cb3221dfe3e921eb745d69c47304327c0ae/h5py/_hl/files.py#L43-L50

[...]

PR #1755, opened 2020-11-12
add ros3 functions

satra:
related to #1712 

@takluyver - it looks like the pieces are all in this header file. in fact each of the drivers have their own header file. would this make it simple to generate the bindings?

<details> <summary> hdf5/include/H5FDros3.h </summary>

```
/* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * *
 * Copyright by The HDF Group.                                               *
 * All rights reserved.                                                      *
 *                                                                           *
 * This file is part of HDF5.  The full HDF5 copyright notice, including     *
 * terms governing use, modification, and redistribution, is contained in    *
 * the COPYING file, which can be found at the root of the source code       *
 * distribution tree, or in https://support.hdfgroup.org/ftp/HDF5/releases.  *
 * If you do not have access to either file, you may request a copy from     *
 * help@hdfgroup.org.                                                        *
 * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */

/*
 * Read-Only S3 Virtual File Driver (VFD)
 *
 * Programmer:  John Mainzer
 *              2017-10-10
 *
 * Purpose:    The public header file for the ros3 driver.
 */
#ifndef H5FDros3_H
#define H5FDros3_H

#ifdef H5_HAVE_ROS3_VFD
#define H5FD_ROS3               (H5FD_ros3_init())
#else
#define H5FD_ROS3               (H5I_INVALID_HID)
#endif /* H5_HAVE_ROS3_VFD */

#ifdef H5_HAVE_ROS3_VFD

/****************************************************************************
 *
 * Structure: H5FD_ros3_fapl_t
 *
 * Purpose:
 *
 *     H5FD_ros3_fapl_t is a public structure that is used to pass S3
 *     authentication data to the appropriate S3 VFD via the FAPL.  A pointer
 *     to an instance of this structure is a parameter to H5Pset_fapl_ros3()
 *     and H5Pget_fapl_ros3().
 *
 *
 *
 * `version` (int32_t)
 *
 *     Version number of the H5FD_ros3_fapl_t structure.  Any instance passed
 *     to the above calls must have a recognized version number, or an error
 *     will be flagged.
 *
 *     This field should be set to H5FD_CURR_ROS3_FAPL_T_VERSION.
 *
 * `authenticate` (hbool_t)
 *
 *     Flag TRUE or FALSE whether or not requests are to be authenticated
 *     with the AWS4 algorithm.
 *     If TRUE, `aws_region`, `secret_id`, and `secret_key` must be populated.
 *     If FALSE, those three components are unused.
 *
 * `aws_region` (char[])
 *
 *     String: name of the AWS "region" of the host, e.g. "us-east-1".
 *
 * `secret_id` (char[])
 *
 *     String: "Access ID" for the resource.
 *
 * `secret_key` (char[])
 *
 *     String: "Secret Access Key" associated with the ID and resource.
 *
 ****************************************************************************/

#define H5FD_CURR_ROS3_FAPL_T_VERSION     1

#define H5FD_ROS3_MAX_REGION_LEN         32
#define H5FD_ROS3_MAX_SECRET_ID_LEN     128
#define H5FD_ROS3_MAX_SECRET_KEY_LEN    128

typedef struct H5FD_ros3_fapl_t {
    int32_t version;
    hbool_t authenticate;
    char    aws_region[H5FD_ROS3_MAX_REGION_LEN + 1];
    char    secret_id[H5FD_ROS3_MAX_SECRET_ID_LEN + 1];
    char    secret_key[H5FD_ROS3_MAX_SECRET_KEY_LEN + 1];
} H5FD_ros3_fapl_t;


#ifdef __cplusplus
extern "C" {
#endif

H5_DLL hid_t H5FD_ros3_init(void);
H5_DLL herr_t H5Pget_fapl_ros3(hid_t fapl_id, H5FD_ros3_fapl_t *fa_out);
H5_DLL herr_t H5Pset_fapl_ros3(hid_t fapl_id, H5FD_ros3_fapl_t *fa);

#ifdef __cplusplus
}
#endif

#endif /* H5_HAVE_ROS3_VFD */

#endif /* ifndef H5FDros3_H */
```

</details>


takluyver:
I don't think it makes any difference whether it's in a separate file or not - all of the public interfaces are exposed through `hdf5.h` in any case.

satra:
@takluyver - i am running into a cython issue. i have defined the struct in api_types_hdf5.pxd, but it doesn't seem to get picked up during compilation.

takluyver:
I'm not very familiar with preprocessing & compiling C code, but I think we're referring to things defined in HDF5 header files (`extern from "hdf5.h"`), rather than actually defining them in Cython.

satra:
in that case, i expect it to be there since it is pulling in the hdf5.h file, which internally links to the file. i will dig more. perhaps some variable is not being set.

takluyver:
I think `H5_HAVE_ROS3_VFD` is not set. What I don't know is whether building HDF5 with that enabled records something in the installed header files to automatically enable it, or whether we need to detect it from the shared library and define the flag again to use that feature.

satra:
@takluyver - i have gotten this driver to work and have updated this PR. however, i would very much appreciate if someone from the h5py team could let me know what else i need to fix.

```
In [1]: import h5py as h5

In [2]: f = h5.File("https://dandiarchive.s3.amazonaws.com/girder-assetstore/99/31/99311ec02ab2402f87
   ...: b8ceda2cb8348d", driver='ros3')

In [3]: f.keys()
Out[3]: <KeysViewHDF5 ['acquisition', 'analysis', 'file_create_date', 'general', 'identifier', 'processing', 'session_description', 'session_start_time', 'specifications', 'stimulus', 'timestamps_reference_time', 'units']>

In [4]: f.get('units')
Out[4]: <HDF5 group "/units" (7 members)>

In [5]: foo = f.get('units')

In [6]: foo.items()
Out[6]: ItemsViewHDF5(<HDF5 group "/units" (7 members)>)

In [7]: [val for val in foo.items()]
Out[7]: 
[('cell_type', <HDF5 dataset "cell_type": shape (22,), type "|O">),
 ('electrode_group', <HDF5 dataset "electrode_group": shape (22,), type "|O">),
 ('global_id', <HDF5 dataset "global_id": shape (22,), type "<u2">),
 ('id', <HDF5 dataset "id": shape (22,), type "<i8">),
 ('shank_id', <HDF5 dataset "shank_id": shape (22,), type "<i8">),
 ('spike_times', <HDF5 dataset "spike_times": shape (2759455,), type "<f8">),
 ('spike_times_index',
  <HDF5 dataset "spike_times_index": shape (22,), type "<u4">)]
```

aragilar:
@satra It looks like some of the tests are failing as your email is not in the authors.yml, can you add yourself there?

I think the ros3 needs to be configurable via setup_configure (or some kind of auto-detection set up), similar to MPI.

I'd also suggest making the interface more user-friendly: authenticate should take a python bool, and I would not use default arguments (`version` from what I can see is versioning the struct contents, so should not be exposed to users, and the others should be managed by the high level interface). Also, `get_fapl_ros3` should decode the struct into a tuple (I'd use a dict, but the other drivers use tuples, so for consistency it makes sense to use a tuple).

takluyver:
> I think the ros3 needs to be configurable via setup_configure (or some kind of auto-detection set up), similar to MPI.

Yup. I'd ideally like it to detect whether HDF5 was built with the ros3 driver and automatically wrap it in h5py, because it doesn't seem great to add lots of build options where you just want to enable the same features as for the underlying hdf5 build.

> authenticate should take a python bool, and I would not use default arguments (version from what I can see is versioning the struct contents, so should not be exposed to users, and the others should be managed by the high level interface

I agree on not exposing `version` - that's an implementation detail, not a parameter.

For the others, I would expose the three authentication parameters `aws_region`, `secret_id` and `secret_key` with default arguments. Then we can automatically set `authenticate=True` if any of those parameters are set. That avoids having to explain the relationship between `authenticate` and the other parameters.

satra:
@takluyver - let me know if there is anything else you might want to see changed. i think i've handled all the little testing  and compilation issues. it would be nice i think to add a real test with that driver, but that would also involve building hdf5lib with the driver enabled option. 

also i haven't tested the authentication part. i need to put a file on s3 in a private space and do that.

h5py/h5p.pyx
```diff
@@ -1033,6 +1033,41 @@ cdef class PropFAID(PropInstanceID):
         return (msize, plist)
 
 
+    if ROS3 and HDF5_VERSION >= (1, 10, 6):
```
takluyver, line 1036: Shouldn't need the version check here, right? `ROS3` should only be true if HDF5 is new enough to support it.

setup_configure.py
```diff
@@ -53,18 +53,30 @@ def validate_version(s):
...
+def ros3_enabled():
+    hdf5 = os.environ.get('HDF5_DIR')
+
+    # Specified a prefix dir (e.g. '/usr/local')
+    if hdf5:
+        with open(op.join(hdf5, 'lib', 'libhdf5.settings')) as fp:
+            for line in fp.readlines():
+                if '(Read-Only) S3 VFD: yes' in line:
+                    return True
+    return False
```
takluyver, line 65: This seems a bit brittle. @aragilar has taken a different approach in #1759 - I haven't reviewed that in detail yet, but can I ask you to look at each other's PRs and figure out together what's the best way to detect optional build-time features? :+1:

satra: will coordinate with @aragilar . i'm happy to adopt whatever is the best way to detect build settings.

aragilar: I'm unclear whether `libhdf5.settings` is an explicit part of of the HDF5 install or just a helpful logfile, is it documented somewhere?

satra: if you do `h5cc -showconfig` it shows the exact contents of `libhdf5.settings` and if i remove it i get:

```
# h5cc -showconfig
cat: /usr/local/hdf5/lib/libhdf5.settings: No such file or directory
```

so i suspect this is an essential part of the library.

satra: @aragilar - any thoughts on the settings file?

satra:
@takluyver - any thoughts on what the next steps are? we just merged cross-platform ros3 support for hdf5 in conda-forge, and i would love to get this enabled in h5py.

regarding the collection of settings `libhdf5.settings` is created by the make process and used by `hfcc -showconfig` as noted in the comment above. if it doesn't exist something is wrong with the HDF5 installation.

yarikoptic:
Please let me know if I could be of any help to assist with progressing this PR forward over the finish click on Merge  ;)

aragilar:
I've posted https://forum.hdfgroup.org/t/guarantees-around-libhdf5-settings/7845 to get clarity from the HDF Group around `libhdf5.settings` - I'm still wary on relying on the file to be there (I can't find any documentation on it), or that the format remains static (ideally there would be a function for checking this in HDF5, but I suspect the difference between checking the ABI and that function would be minimal).

aragilar:
The HDF Group views the `libhdf5.settings` as not part of the public interface, so it shouldn't be relied on (https://forum.hdfgroup.org/t/guarantees-around-libhdf5-settings/7845/2). I think #1759 is the was forward then.

@satra @yarikoptic If one of you wants to make the switch to the the design in #1759 that would be helpful (feel free to cherry pick the changes in #1759 to avoid needing to manually copy the code).

satra:
@aragilar - thanks. updated to use ABI and drop settings.

satra:
i don't have a windows machine to debug, but the `HDF5LibWrapper(h5_lib)` function does not seem to work on a windows machine in the test.

yarikoptic:
@satra would you be kind to push a rebase or rewrite of the last commit just to trigger CI rerun -- seems can't access those older appveyor build logs any longer

satra:
@yarikoptic - you should see the failures here.

yarikoptic:
> i don't have a windows machine to debug, but the `HDF5LibWrapper(h5_lib)` function does not seem to work on a windows machine in the test.

what I see is a complete bail out by pip (ran within tox) such as 
```
  Loading library to get build settings and version: D:\a\1\cache\hdf5\1.12.0\lib\hdf5.dll
  error: Could not find module 'D:\a\1\cache\hdf5\1.12.0\lib\hdf5.dll' (or one of its dependencies). Try using the full path with constructor syntax.
  ----------------------------------------
  ERROR: Failed building wheel for h5py
Failed to build h5py
```
sent you a blind attempt at a fix in https://github.com/satra/h5py/pull/1

yarikoptic:
oppa -- a miraculous success (note: I force-pushed tune up to my "fix", abandoning the merge).  Now we have only py39-hdf5110 failing with

```
ERROR:   py39-test-mindeps: could not install deps [pytest, pytest-cov, pytest-mpi>=0.2, cython==0.29.14, numpy==1.19.3]; v = InvocationError("'D:\\a\\1\\s\\.tox\\py39-test-mindeps\\Scripts\\python.EXE' -m pip install pytest pytest-cov 'pytest-mpi>=0.2' cython==0.29.14 numpy==1.19.3", 1)
```
which is unrelated to these changes since happening similarly in https://github.com/h5py/h5py/pull/1785 for a typo fix.

satra:
@yarikoptic - please add yourself to authors.yml file

yarikoptic:
> @yarikoptic - please add yourself to authors.yml file

done

<details>
<summary>although never used such one, so just followed your addition, and I know no `rever`, which seems to fail for me for "authors" action</summary> 

```shell
$> rever check --activities authors
  File "/home/yoh/deb/gits/h5py-upstream/venvs/dev3/bin/rever", line 2, in <module>
    from rever.main import main
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 786, in exec_module
  File "/home/yoh/deb/gits/h5py-upstream/venvs/dev3/lib/python3.9/site-packages/xonsh/imphooks.py", line 123, in get_code
    code = execer.compile(src, glbs=ctx, locs=ctx)
  File "/home/yoh/deb/gits/h5py-upstream/venvs/dev3/lib/python3.9/site-packages/xonsh/execer.py", line 135, in compile
    tree = self.parse(input, ctx, mode=mode, filename=filename, transform=transform)
  File "/home/yoh/deb/gits/h5py-upstream/venvs/dev3/lib/python3.9/site-packages/xonsh/execer.py", line 95, in parse
    tree, input = self._parse_ctx_free(input, mode=mode, filename=filename)
  File "/home/yoh/deb/gits/h5py-upstream/venvs/dev3/lib/python3.9/site-packages/xonsh/execer.py", line 219, in _parse_ctx_free
    tree = self.parser.parse(
  File "/home/yoh/deb/gits/h5py-upstream/venvs/dev3/lib/python3.9/site-packages/xonsh/parsers/base.py", line 497, in parse
    tree = self.parser.parse(input=s, lexer=self.lexer, debug=debug_level)
  File "/home/yoh/deb/gits/h5py-upstream/venvs/dev3/lib/python3.9/site-packages/xonsh/ply/ply/yacc.py", line 335, in parse
    return self.parseopt_notrack(input, lexer, debug, tracking, tokenfunc)
  File "/home/yoh/deb/gits/h5py-upstream/venvs/dev3/lib/python3.9/site-packages/xonsh/ply/ply/yacc.py", line 1122, in parseopt_notrack
    p.callable(pslice)
  File "/home/yoh/deb/gits/h5py-upstream/venvs/dev3/lib/python3.9/site-packages/xonsh/parsers/base.py", line 2221, in p_atom_expr
    p[0] = self.apply_trailers(p[1], p[2])
  File "/home/yoh/deb/gits/h5py-upstream/venvs/dev3/lib/python3.9/site-packages/xonsh/parsers/base.py", line 2215, in apply_trailers
    assert False
AssertionError

```
</details>

aragilar:
Closing and opening to rerun CI.

aragilar:
The only thing missing would be some tests - I'm not sure how easy it would be to set up, but at least having something that could be run locally (if getting it to work on the CI fails), would make me much happier about merging this. Otherwise I think this is ready to merge.

aragilar:
I suspect given the windows tests failed again, this may need a rebase.

satra:
@aragilar - rebased on master. 

regarding tests - i couldn't figure out how hdf5 is being installed. if from conda-forge we can directly add tests to check. if it's being built, we would ideally build a version with and without the ros3 option, but the very least with ros3.

satra:
@aragilar - i added a test, but it's not a good test. it uses functionality that should be tested. neither brew nor apt at this point compile with the driver, so hard to evaluate with or without the compiled driver. 

but looks like everything is passing except coverage.

satra:
let us know if there is something that we should still be doing for this. would love to see a minor release with this patch.

aragilar:
@satra Sorry, I thought I'd hit the submit button on those comments... My main concern was what are the consequences of repeatedly requesting https://s3.us-east-2.amazonaws.com/hdf5ros3/GMODO-SVM01.h5? If nothing major, then I think this can be merged.

h5py/tests/test_file.py
```diff
@@ -780,3 +780,16 @@ def test_swmr_mode_consistency(self):
...
+# unittest doesn't work with pytest fixtures (and possibly other features),
+# hence no subclassing TestCase
+class TestROS3(object):
+    @pytest.mark.skipif(h5py.version.hdf5_version_tuple < (1, 10, 6)
+                        or not h5py.h5.H5PYConfig().ros3,
+                        reason="ros3 file operations were added in HDF5 1.10.6+")
+    def test_ros3(self):
+        """ ROS3 driver and options """
+
+        with File("https://s3.us-east-2.amazonaws.com/hdf5ros3/GMODO-SVM01.h5", 'r', driver='ros3') as f:
```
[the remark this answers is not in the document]
yarikoptic: AFAIK it was provided by hdf5 group as an example url and is public. We could of cause check with them or just ask for forgiveness later: I don't know if they are paying for data egress.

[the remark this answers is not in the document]
aragilar: Oh, cool, I'll post the question on the forum then (I was more concerned that it was one of @satra's files, and either it would produce a surprise Amazon bill or be taken down at some point).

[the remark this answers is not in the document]
aragilar: See https://forum.hdfgroup.org/t/test-url-for-ros3-driver/7986

danieldjewell:
> @satra Sorry, I thought I'd hit the submit button on those comments... My main concern was what are the consequences of repeatedly requesting https://s3.us-east-2.amazonaws.com/hdf5ros3/GMODO-SVM01.h5? If nothing major, then I think this can be merged.

Well, the consequence is that whoever owns that bucket (HDF Group?) is going to be paying for the data transfer... 🤑  That said, like *Mega* kudos to you @satra for asking this question - I don't think this gets asked enough. (Remember that deal where - I think it was D-Link - consumer routers were hard coded in firmware to get time *way* too often over NTP and it ended up kinda DDoSing those NTP servers? Wow, looks like there's an article about it: https://en.wikipedia.org/wiki/NTP_server_misuse_and_abuse) 

I don't think that Amazon's infrastructure will be overloaded ... the only other question is how much data does it transfer? Just headers? Few hundred KiB?

satra:
This is a 30MB file (which may not be good for repeated testing for several reasons). but i can easily create a smaller HDF5 file for testing purposes and serve it from one of our public buckets. i'll do that and update this PR.

satra:
@aragilar  and @danieldjewell  - i replaced the pointer with a 3K file in our bucket. since we plan to use h5py for the foreseeable future, using this pointer is fine.

h5py/tests/test_file.py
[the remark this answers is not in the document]
satra: this has been fixed by shifting to one of our files.

satra:
@aragilar - the tests that are failing appear to be unrelated to this PR. if they are fixed in master, i would be happy to merge and update. otherwise i think this has addressed all the issues you had.

aragilar:
This looks good to me, merging.

PR #1759, opened 2020-11-13
Add configure-time checks for detecting MPI builds

aragilar:
This adds helpers for detecting which functions are available in HDF5 (which should correspond with which features it was built with).

I've only done this with MPI so far, as a test run, but we can expand this to other options (such as ros3 as needed for #1755).


takluyver:
Once I resolved the merge conflicts, it turns out all the changes here have already landed along with the ROS3 support, so I'll close this PR.

PR #1766, opened 2020-11-22
Move MPI to azure

aragilar:
This removes more travis runs, replacing them with azure pipelines.

This will take some time to run, as the cache will be reset from adding MPI.


ci/azure-pipelines-steps.yml
```diff
@@ -48,8 +48,13 @@ steps:
 - ${{ if eq(parameters.installer, 'apt') }}:
   - script: |
       sudo apt-get update
-      sudo apt-get install libhdf5-serial-dev ccache openmpi-bin libopenmpi-dev
+      sudo apt-get install ccache openmpi-bin libopenmpi-dev
```
takluyver, line 51: Is it worth doing this only if HDF5_MPI is set? It's only about 10 seconds, but 10 seconds on each job adds up.

I'm not sure installing ccache actually does anything, given that we're not carrying over its cache files from run to run. Maybe it's worth trying again to get that working, but that's definitely for a separate PR.

PR #1772, opened 2020-12-02
add an example to create a virtual dataset with FancySelection

payno:
A simple example on how we can create FancySelection for VirtualSource (and as a consequence virtual dataset).
It would have been useful to me so maybe it could help someone else in the future.


payno:
In fact I am not sure this is the best way to go. In this example I need to acces `FancySelection` by importing `h5py._hl.selections as selection`and so importing protected '_hl'. But I don't see an other way to accomplish this. Would you have idea ? Is this feature foreseen ?

takluyver:
Does `v_source[[1, 2, 5]]` work for doing fancy selection?

What's the performance like with this? I recall that making many small separate mappings for a virtual dataset caused poor performance when reading from it, and I'd be somewhat concerned that fancy indexing might be similar, as it's lots of small hyperslabs in the source selection.

payno:
> Does `v_source[[1, 2, 5]]` work for doing fancy selection?

In fact it does. It call also 'select'. The same operation could be done by (for the first selection)

```python
v_source_1 = h5py.VirtualSource(file_path, "original_data/d1", shape=(10, 10, 10))
v_source_1 = v_source_1[[2, 3]]
```
which is in fact more convenient thank you. I will update the example.

> What's the performance like with this? I recall that making many small separate mappings for a virtual dataset caused poor performance when reading from it, and I'd be somewhat concerned that fancy indexing might be similar, as it's lots of small hyperslabs in the source selection.

Make sense. I didn't had a look at performances yet.

aragilar:
I think the remaining bit before this can be merged is the removal of `import h5py._hl.selections as selection` (which isn't needed any more), but I'm not up to date with the state of VDS.

payno:
Thanks. In fact I was no more using the 'h5py._hl.selections'  directly. I removed the import and improved file formatting (black - flake). Let me know if there is more improvements to be done.

PR #1780, opened 2020-12-11
Address numpy 1.20 deprecation warnings

vasole:
- Suppress annoying warning concerning the use of None as shape. The warning was triggered when retrieving a dataset as [()]
- Replaces np.float by np.float64
- Replaces np.int by int (it was an alias)
- Replaces np.bool by bool (it was an alias)

### DIFF
--- a/h5py/tests/test_dtype.py
+++ b/h5py/tests/test_dtype.py
@@ -47,7 +47,7 @@ class TestVlen(TestCase):
 
         dt_vb = np.dtype([
             ('foo', vidt),
-            ('logical', np.bool)])
+            ('logical', bool)])
         vb = f.create_dataset('dt_vb', shape=(4,), dtype=dt_vb)
         data = np.array([(a([1, 2, 3]), True),
                          (a([1    ]), False),
@@ -67,11 +67,11 @@ class TestVlen(TestCase):
         dt_vvb = np.dtype([
             ('foo', vidt),
             ('bar', vidt),
-            ('logical', np.bool)])
+            ('logical', bool)])
         vvb = f.create_dataset('dt_vvb', shape=(2,), dtype=dt_vvb)
 
         dt_bvv = np.dtype([
-            ('logical', np.bool),
+            ('logical', bool),
             ('foo', vidt),
             ('bar', vidt)])
         bvv = f.create_dataset('dt_bvv', shape=(2,), dtype=dt_bvv)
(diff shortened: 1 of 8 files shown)


vasole:
One can also use np.bool_ and np.int_ instead of bool and int.

h5py/tests/test_dtype.py
```diff
@@ -47,7 +47,7 @@ def a(items):
 
         dt_vb = np.dtype([
             ('foo', vidt),
-            ('logical', np.bool)])
```
aragilar, line 50: I *think* we want to use the numpy boolean dtype (rather than the python boolean type) here, not sure what it's named though.

aragilar: (It would be the same for the rest of the file)

h5py/tests/test_slicing.py
```diff
@@ -238,7 +238,7 @@ def test_slice_zero_length_dimension(self):
         """ Slice a dataset with a zero in its shape vector
             along the zero-length dimension """
         for i, shape in enumerate([(0,), (0, 3), (0, 2, 1)]):
-            dset = self.f.create_dataset('x%d'%i, shape, dtype=np.int, maxshape=(None,)*len(shape))
+            dset = self.f.create_dataset('x%d'%i, shape, dtype=int, maxshape=(None,)*len(shape))
```
aragilar, line 241: I think we want `np.int64` here, but I don't its going to make too much difference either way.

aragilar:
I told the CI to retry, so we'll see what happens. I don't think it's going to make too much difference whether we use the python or numpy type, but we may want to use the numpy dtype to be consistent. It's probably worth adding some tests to see that the python and numpy types behave the same way.

aragilar:
Otherwise, I think assuming the CI works this time, this can be merged.

h5py/tests/test_dtype.py
vasole: One can use np.bool_ and np.int_ as stated above.

news/warnings.rst
```diff
@@ -0,0 +1,29 @@
+New features
+------------
+
+* Suppress numpy 1.20 deprecation warnings concerning the use of None as shape, np.float, np.int and np.bool
+
+Deprecations
+------------
+
+* Replace np.float by np.float64, np.int by int and np.bool by bool
```
takluyver, line 9:
```suggestion
* <news item>
```

This section is for deprecating things in h5py's API, so no need to mention this.

h5py/tests/test_dtype.py
takluyver: From what I can see [in the numpy docs](https://numpy.org/doc/stable/reference/arrays.dtypes.html#specifying-and-constructing-data-types), specifying `bool` for a dtype is equivalent to `np.bool_` anyway. So I think this is fine as is.

PR #1786, opened 2021-01-06
Use the next patch version of Cython on Python 3.9 to avoid building it

takluyver:
Hopefully this will fix the recent CI failures. Cython now has a fallback pure Python (`py2.py3-none-any`) wheel, and the README seems to indicate that the benefits of the compiled parts are outweighed for CI by the time taken to compile them.


takluyver:
Merging this myself so we have the CI working again.

Issue #1733, opened 2020-11-02
Do we need to move away from Travis CI?


takluyver:
Travis have [made a blog post](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing) which appears to say that, effective immediately, free users will be 'upgraded' to a trial plan with a one-time allocation of 10k credits (equivalent to 1000 minutes). After that, we'd have to request an allocation of OSS credits. It's not very clear, but it sounds like that's a manual process, you might get a one-time allocation rather than a recurring one, and possibly there's even a limited pool of credits for all OSS projects.

If that's the case, Travis looks like a pretty unappealing option compared to our other CI services. It's not clear what's actually going on, though - one of my colleagues got an email this morning saying he would get 1k credits per month. If that was the case, we might be OK if we're careful about what runs on Travis.

We currently use Travis for:

- ppc64le builds (cc @kif) - I'm not aware where else we can do this if Travis is unavailable. But [this IBM blog post](https://developer.ibm.com/technologies/linux/tutorials/travis-ci-on-power/) suggests IBM provided the servers for Travis to help open-source projects, so perhaps they'll ensure there's still some option.
- Low-level API docs - [doctr](https://drdoctr.github.io/) has specific integration with Travis which makes this very convenient, but we could use any CI service with a bit more work.
- MPI tests - @aragilar, was there a specific reason these were on Travis?
- Allowed failure builds on Python nightly and Pypy. We don't pay any attention to these anyway, so no great loss, and they could go to other services if we needed them.

aragilar:
Yeah, it really unclear whether that's a monthly thing, or a one off. For MPI, I did try to move it to azure pipelines, but there were odd issues around it timing out, so I left it on Travis (as it seemed to work there).

I haven't received any emails from Travis about the change (maybe they're doing some kind of rollout about who gets the emails in relation to when the service switches over).

takluyver:
There's also the added complexity that is still - for now - on travis-ci.org, which has no billing. I think everything is meant to migrate to travis-ci.com within a few months.

yousef92-tech:
Why not use GitHub actions instead of Travis CI? You get unlimited minutes/month. I can help with creating and maintaing the pipeline if needed.

takluyver:
We already use other CI services, especially Azure pipelines (which is similar to Github actions). But there are some specific things that are easier (or perhaps even only possible) on Travis, hence this discussion.

kif:
Hi Thomas,

[...]

PR #1788, opened 2021-01-06
Move low-level API docs creation to Github actions

takluyver:
Part of moving CI away from Travis, issue #1733. Github actions is the obvious choice for this, since it is published to Github pages.

Like the previous set-up, it should build the docs for pull requests, so we can check they're not broken, but only publish them on changes in master.

### DIFF
--- a/.github/workflows/lowlevel-api-docs.yml
+++ b/.github/workflows/lowlevel-api-docs.yml
@@ -1,0 +1,46 @@
+name: Build low-level API docs
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  deploy:
+    runs-on: ubuntu-18.04
+    env:
+      HDF5_VERSION: 1.12.0
+      HDF5_DIR: ${{ github.workspace }}/cache/hdf5
+      TOXENV: apidocs
+    steps:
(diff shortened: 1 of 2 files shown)


takluyver:
The build and the artifact publication [are working](https://github.com/takluyver/h5py/actions/runs/466604667) on my fork. I haven't tested publishing to Github pages, but that's using a [popular action](https://github.com/marketplace/actions/github-pages-action) in a moderately simple way.

takluyver:
I'll merge this myself in another day or so, unless anyone objects. It doesn't change the package at all, only part of the CI, and we can always change that further.

aragilar:
Given this is just the docs we're hosting on github pages, I think using github actions (vs azure pipelines) makes sense. I'm slightly wary of using what looks to me like a random person's action, but given this is limited to the API docs, I don't think there's too much to worry about there.

takluyver:
Thanks! There are a number of actions for publishing to Github pages, but no 'official' one that I could see. The one I used seemed to be the most popular.

Unfortunately, the deployment appears to have broken https://api.h5py.org/ - I think because it didn't preserve the `CNAME` file in the gh-pages branch. I'll get on that shortly.

PR #1789, opened 2021-01-07
Deprecate setting default_file_mode to values other than 'r'

takluyver:
This config option was added to help people transition to h5py 3.0, where the default file mode changed to read-only.

In h5py 2.x, opening a file without specifying a mode (`h5py.File('foo.h5')`) would try to guess based on whether the file existed and what permissions you had. From 2.10 (#1143), this issued a warning that the default in h5py 3.0 would be read-only. We added this option as a way for people to silence the warning when they couldn't change the code opening the file, by setting an appropriate file mode externally.

This was intended as a temporary feature to help with the transition, not a permanent part of the API. It is clearly better to pass the mode where the file is opened, rather than modifying global state. This is the first step towards getting rid of it:

1. Setting the default to anything other than `'r'` (which is the default since 3.0 anyway) warns
2. Setting the default to `'r'` warns, and to anything else is an error
3. Remove the `default_file_mode` option

### DIFF
--- a/h5py/h5.pyx
+++ b/h5py/h5.pyx
@@ -10,8 +10,10 @@
 
 include "config.pxi"
 
+from warnings import warn
 from .defs cimport *
 from ._objects import phil, with_phil
+from .h5py_warnings import H5pyDeprecationWarning
 
 ITER_INC    = H5_ITER_INC     # Increasing order
 ITER_DEC    = H5_ITER_DEC     # Decreasing order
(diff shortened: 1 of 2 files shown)


takluyver:
While 3.0 has only been out a couple of months, the change has been signposted (in docs and warnings) since 2.10, over a year ago, so I think we can do this deprecation in 3.2.

aragilar:
I think this makes sense, and definitely should go in 3.1 (dropping in 3.2 I also agree with). I'm merging this for 3.1, any objections about the dropping in 3.2 can be discussed in a separate issue/PR.

takluyver:
Thanks James! To be clear, this is going in for 3.2, as 3.1 already happened (soon after 3.0, to fix some regressions)

PR #1790, opened 2021-01-08
Add CNAME file after building docs on Github Actions

takluyver:
This should fix https://api.h5py.org

### DIFF
--- a/.github/workflows/lowlevel-api-docs.yml
+++ b/.github/workflows/lowlevel-api-docs.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - master
-    pull_request:
+  pull_request:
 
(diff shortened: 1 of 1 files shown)


takluyver:
The CI failure is on building the docs, because api.h5py.org is broken:

```
Warning, treated as error:
failed to reach any of the inventories with the following issues:
intersphinx inventory 'https://api.h5py.org/objects.inv' not fetchable due to <class 'requests.exceptions.HTTPError'>: 404 Client Error: Not Found for url: https://api.h5py.org/objects.inv
```

This PR should fix that once it is in master.

takluyver:
And indeed https://api.h5py.org/ is back now :-)

PR #1791, opened 2021-01-09
Fix chunk_iter mistake

patiencedaur:
I only fixed a typo in an .rst docs file and believe tests are not necessary in this case.


takluyver:
Thanks! You're right, a change like this doesn't need tests.

PR #1796, opened 2021-01-18
Fix read_direct when source and destination have different shapes

### LINKED ISSUE #1773: [BUG] read_direct() does not take shape of array into account
### Short description
The `read_direct()` method on datasets does not always take the shape of the preexisting array into account, but only the shape of the dataset itself. Indeed, looking at the `select()` function in `h5py/_hl/selections.py`, the `shape` argument is never used in the case of both `len(args) != 1` and `dataset is not None`. Thus code where we read into a sliced array is broken, e.g.
```python
with h5py.File(myfile, mode='r') as f:
    f['data'].read_direct(arr, dest_sel=np.s_[:, :size])
```
where `arr` is a preexisting 2D array, with `arr.shape[1] > size`.

### Cause
This problem is not there in version 2.10.0, but appears in 3.0.0 and remains in 3.1.0. I believe the bug is introduced in #1526, specifically 474ef53d2698f6fa8988df2a2ec99525863bc8c4, where (some of) the changes to `h5py/_hl/selections.py` are
```python
-    sel = SimpleSelection(shape)	
-    sel[args]
-    return sel
---
+    if dataset is not None:
+        selector = dataset._selector
+    else:
+        space = h5s.create_simple(shape)
+        selector = _selector.Selector(space)
+    return selector.make_selection(args)
```
Note that in the new code, `shape` is completely neglected if `dataset is not None`.

### Complete example
The code below does the following:

 A) Generate an array of shape `(2, 2)` and write it to disk.
 B) Read the array back in using `[...]`.
 C) Read the array back in using `read_direct()` into an array of matching shape.
 D) Read the array back in using `read_direct()` into a slightly too large array, with properly specified `dest_sel` argument to `read_direct()`.

```python
import numpy as np
import h5py

# Generate data
size = 2
shape = (size, size)
data = np.zeros(shape, dtype=np.double)
count = 0
for i in range(size):
    for j in range(size):
        data[i, j] = count
        count += 1
print('A:', data)

# Write data
with h5py.File('data.hdf5', mode='w') as f:
    dset = f.create_dataset('data', shape, dtype=np.double)
    dset[...] = data

# Read data
with h5py.File('data.hdf5', mode='r') as f:
    data_read = f['data'][...]
print('B:', data_read)

# Read data using read_direct
data_read_direct = -np.ones(shape, dtype=np.double)
with h5py.File('data.hdf5', mode='r') as f:
    f['data'].read_direct(
        data_read_direct,
        dest_sel=np.s_[:, :size],  # does not matter
    )
print('C:', data_read_direct)

# Read data using read_direct into larger array
shape_large = (size, size + 1)
data_read_direct_large = -np.ones(shape_large, dtype=np.double)
with h5py.File('data.hdf5', mode='r') as f:
    f['data'].read_direct(
        data_read_direct_large,
        dest_sel=np.s_[:, :size],  # should matter!
    )
data_read_direct_large_sliced = data_read_direct_large[:, :size]
print('D:', data_read_direct_large_sliced)
```

At each stage the result is printed. We get

> A: [[0. 1.]
 [2. 3.]]
B: [[0. 1.]
 [2. 3.]]
C: [[0. 1.]
 [2. 3.]]
D: [[ 0.  1.]
 [ 3. -1.]]

We see that D fails, with the read effectively being carried out into `data_read_direct_large[...]` rather than `data_read_direct_large[:, :size]`, despite supplying `dest_sel=np.s_[:, :size]`. I believe the problem only occurs when the slice of the array in which to read in the data (well *write the data to*, but from disk to memory) is non-contiguous, which is the case for most slices of multi-dimensional arrays. I stress that this non-contiguousity should be perfectly OK (and used to be), as long as it's encoded by `dest_sel` and the array itself *is* contiguous.

It's not true that the argument `dest_sel` is completely ignored, as leaving it out on D results in

> TypeError: Can't broadcast (2, 2) -> (2, 3)

(which is fair), though setting it correctly does not result in the expected behavior.

### Fix
I'm not well-versed on the h5py internals, so I won't suggest a proper fix, though obviously the `shape` argument to `_hl.selections.select()` must not go unused. For the simple example above, removing
```python
selector = dataset._selector
```
and always going with
```python
space = h5s.create_simple(shape)
selector = _selector.Selector(space)
```
does in fact work, but I don't know if this causes other use cases to break.

Interestingly, `read_direct()` is being tested in `tests/test_dataset.py`, but it appears that only 1D arrays are tried, whereas this problem is inherent to multi-dimensional arrays.

I guess it's worth also checking up on `write_direct()`.

### System specifications
 * OS: CentOS Linux 7, Linux Mint 19.3
 * Python: 3.9 (build from source)
 * h5py: 3.0.0rc1, 3.0.0, 3.1.0
 * HDF5: 1.10.6, 1.12.0
 * NumPy: 1.19.4
 * Cython: 0.29.21

bmerry:
Also strengthened the unit tests to pick up the bug and check a few
other conditions.

Closes #1773 and #1792.

### DIFF
--- a/news/read-direct-fix.rst
+++ b/news/read-direct-fix.rst
@@ -1,0 +1,29 @@
+New features
+------------
+
+* <news item>
+
+Deprecations
+------------
+
+* <news item>
+
+Exposing HDF5 functions
+-----------------------
+
+* <news item>
+
+Bug fixes
+---------
+
+* Fix ``read_direct`` when the source and destination have different shapes.
+
+Building h5py
+-------------
+
+* <news item>
+
+Development
+-----------
+
+* <news item>
--- a/.authors.yml
+++ b/.authors.yml
@@ -850,7 +850,9 @@
   first_commit: 2020-11-06 23:32:13
   github: uellue
 - name: Bruce Merry
-  email: 1963944+bmerry@users.noreply.github.com
+  email: bmerry@ska.ac.za
+  alternate_emails:
+    - 1963944+bmerry@users.noreply.github.com
   num_commit: 1
   first_commit: 2021-01-18 09:30:28
   github: bmerry
--- a/.authors.yml
+++ b/.authors.yml
@@ -850,7 +850,7 @@
   first_commit: 2020-11-06 23:32:13
   github: uellue
 - name: Bruce Merry
-  email: bmerry@ska.ac.za
+  email: 1963944+bmerry@users.noreply.github.com
   num_commit: 1
   first_commit: 2021-01-18 09:30:28
   github: bmerry
--- a/.authors.yml
+++ b/.authors.yml
@@ -849,3 +849,8 @@
   num_commit: 2
   first_commit: 2020-11-06 23:32:13
   github: uellue
+- name: Bruce Merry
+  email: bmerry@ska.ac.za
+  num_commit: 1
+  first_commit: 2021-01-18 09:30:28
+  github: bmerry
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -986,7 +986,7 @@ class Dataset(HLObject):
             if source_sel is None:
                 source_sel = sel.SimpleSelection(source.shape)
             else:
-                source_sel = sel.select(source.shape, source_sel, self)  # for numpy.s_
+                source_sel = sel.select(source.shape, source_sel)  # for numpy.s_
             mspace = source_sel.id
 
             if dest_sel is None:
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -967,7 +967,7 @@ class Dataset(HLObject):
             if dest_sel is None:
                 dest_sel = sel.SimpleSelection(dest.shape)
             else:
-                dest_sel = sel.select(dest.shape, dest_sel, self)
+                dest_sel = sel.select(dest.shape, dest_sel)
 
             for mspace in dest_sel.broadcast(source_sel.mshape):
                 self.id.read(mspace, fspace, dest, dxpl=self._dxpl)
--- a/h5py/tests/test_dataset.py
+++ b/h5py/tests/test_dataset.py
@@ -259,29 +259,43 @@ class TestReadDirectly:
         with pytest.raises(TypeError):
             dset.read_direct(arr)
 
-class TestWriteDirectly(BaseDataset):
+class TestWriteDirectly:
 
     """
         Feature: Write Numpy array directly into Dataset
     """
 
-    def test_write_direct(self):
-        dset = self.f.create_dataset('dset', (100,), dtype='int32')
-        empty_dset = self.f.create_dataset("edset", dtype='int64')
-        arr = np.ones((100,))
-        arr2 = np.ones((200,))
+    @pytest.mark.parametrize(
+        'source_shape,dest_shape,source_sel,dest_sel',
+        [
+            ((100,), (100,), np.s_[0:10], np.s_[50:60]),
+            ((70,), (100,), np.s_[50:60], np.s_[90:]),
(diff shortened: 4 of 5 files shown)


bmerry:
@takluyver I've marked this as a draft because I think we should also fix write_direct (I haven't tested it, but by eyeball it has the same issue). But I think it's worth getting some feedback in case there is anything you'd like changed before I duplicate the work for write_direct.

I wanted to use `pytest.mark.parameterize` for the test, but it doesn't work with unittest.TestCase. Let me know if you're happy to change the base test to not inherit from TestCase so that the test can use pytest's parametrization instead of my home-brewed version.

news/read-direct-fix.rst
```diff
@@ -0,0 +1,29 @@
...
+
+Exposing HDF5 functions
+-----------------------
+
+* <news item>
+
+Bug fixes
+---------
+
+* Fix ``read_direct`` when the source and destination have different shapes.
```
takluyver, line 19:
```suggestion
* Fix :meth:`Dataset.read_direct` when the source and destination have different shapes.
```

This should ultimately create a nice link when it's combined into the release notes.

takluyver (commented):
Thanks! Yes, it's fine to convert the test class you're working with to a pytest-style without inheritance, or add tests as functions. There's a `writable_file` fixture you can use to get a new `File` object for each test. I just don't want to review a mass conversion of all the tests, which is why they are mostly still unittest-style classes at the moment.

I agree with you that `write_direct` does appear to have the same issue. If you're happy to fix both in this PR, that's great.

news/read-direct-fix.rst
bmerry: Should it have a dot before `Dataset` so that Sphinx will search for it? Or is `h5py` already considered the current module for the release notes?

bmerry:
Any idea why the Azure check wants to add my personal email address to `.authors.yml`? I don't even know how it found it; possibly by scraping my Github profile, but I didn't think that was public from Github. I've checked that the Git comments are authored by my work address, and that's the one that's appropriate to associate with my work on h5py.

bmerry:
> Any idea why the Azure check wants to add my personal email address to `.authors.yml`? I don't even know how it found it; possibly by scraping my Github profile, but I didn't think that was public from Github. I've checked that the Git comments are authored by my work address, and that's the one that's appropriate to associate with my work on h5py.

I see it Github did have my personal email set as a public email address. I've told it to keep my email address private, so I'll see what the bot says when I push further changes.

news/read-direct-fix.rst
takluyver: I'm not 100% sure, to be honest. There always has to be some manual editing of the release notes when preparing a release, so I'd catch it then, but I don't think there's any harm in putting the dot anyway if you prefer.

takluyver:
Sorry about the overzealous checks on `.authors.yml`. I'm periodically confused by why it needs some of the information it asks for. Hopefully making your email private will have fixed it.

bmerry:
I've updated to fix write_direct as well.

It seems that there are also problems with using an integer instead of a range in one of the selections (the resulting shape has a length-one dimension instead of the dimension being removed, later leading to shape mismatches), but I'm guessing that is a separate bug. Once this is merged I can file that as a diff against the current tests.

takluyver:
Thanks, this looks good.

> It seems that there are also problems with using an integer instead of a range in one of the selections (the resulting shape has a length-one dimension instead of the dimension being removed, later leading to shape mismatches),

Feel free to tackle that separately, or to open an issue about it. It might be as simple as switching between `mshape` and `array_shape`, but I'm not sure. HDF5 itself has no concept of selecting a lower-dimensional space, so that bit of semantics is something that h5py adds on top.

PR #1799, opened 2021-01-22
FIX: make sure we explicitly close attrs rather than relying on gc

### LINKED ISSUE #1591: PyPy test failures: cannot overwrite an attribute
Testing with PyPy 3.7 v7.3.1 on lnux 64, I have a number of failures. xref gh-1012 for a complete report of an older version on windows. Here I would like to focus on the following failure (from test_attrs.py test_overwrite):
```
from h5py import File
f = File('tempname', 'w')
f.attrs['a'] = 4.0
f.attrs['a'] = 5.0
assert f.attrs['a'] == 5.0
```
The old value is preserved. It seems this is connected to `h5a.delete` somehow? Any hints what could be the problem? Is there a flush or other operation that is dependent on refcount semantics when an object goes out of scope?

tacaswell:
See #1591 (doesn't apparently fix all issues)

### DIFF
--- a/h5py/_hl/attrs.py
+++ b/h5py/_hl/attrs.py
@@ -215,6 +215,8 @@ class AttributeManager(base.MutableMappingHDF5, base.CommonStateObject):
                         attr.close()
                         h5a.delete(self._id, self._e(tempname))
                         raise
+                finally:
+                    attr.close()
 
     def modify(self, name, value):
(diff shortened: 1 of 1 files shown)


takluyver:
This looks fine to me. Looking at the code, I also think we could get rid of the outermost try/except/else, where all the except clause does is re-raise the exception.

Issue #1802, opened 2021-01-28
absolute path  ending up in the egg-info SOURCES.txt


Mingli-Yu:
$ cat SOURCES.txt
ANN.rst
AUTHORS
LICENSE
MANIFEST.in
README.rst
api_gen.py
pylintrc
pytest.ini
setup.cfg
setup.py
setup_build.py
setup_configure.py
tox.ini
/work/h5py/_conv.c
/work/h5py/_errors.c
[snip]

aragilar:
@Mingli-Yu Which OS/python version/h5py version did you run into this, I've tried using latest master on linux (python 3.9) and do not find this to be the case.

takluyver:
I also don't see absolute paths in there, and it's not clear that it's a big problem if they do. I'm going to close this issue and PR #1803 as something we don't need to worry about. We can always reopen if someone offers more details.

PR #1803, opened 2021-01-29
setup_build.py: avoid absolute path

Mingli-Yu:
Prevent absolute path from ending up in the egg-info SOURCES.txt.

Signed-off-by: Mingli Yu <mingli.yu@windriver.com>

### DIFF
--- a/setup_build.py
+++ b/setup_build.py
@@ -21,7 +21,7 @@ from setup_configure import BuildConfig
 
 
 def localpath(*args):
-    return op.abspath(op.join(op.dirname(__file__), *args))
+    return op.join(*args)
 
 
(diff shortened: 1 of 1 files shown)


Mingli-Yu:
This PULL request used to fix https://github.com/h5py/h5py/issues/1802

takluyver:
I think the idea of using `__file__` like that is that it can work regardless of what the cwd is when you run it. But I'm not sure if that's actually something we need to worry about, or if we should assume that it has to be built from the directory containing `setup.py` anyway.

I also don't know if `SOURCES.txt` actually matters for anything. The `.egg-info`directory is something non-standard that setuptools produces, and I don't think SOURCES.txt is part of any official standard.

takluyver:
Closed as we haven't really established that there's a problem to solve - see #1802.

Issue #1807, opened 2021-02-05
Handling corrupt data after crash


HonakerM:
Hello! So I have a system where my program is using an h5py File as a buffer. I want it to be able to recover the buffer after a crash. However sometimes when reopening the file I get a `OSError: Unable to open file (truncated file: eof = 2288632, sblock->base_addr = 0, stored_eof = 2501624) `  what is the cause of this error? How can I avoid it?

This is on h5py version 3.1.0

[Edit] I tried opening the file in swmr mode but while the data is good I get a `file is already open for write/SWMR write (may use <h5clear file> to clear file consistency flag)`

HonakerM:
I also started running into an interesting issue. I created a dataset using the following commands
```
import h5py
file = h5py.File('test.h5df','w',swmr=True, libver='latest')
dset = file.create_dataset("data",shape=(10,4))
file.flush()
```
I then terminated the session to simulate a crash. When I went to reopen the file I got the following errors
```
>>> old_file= h5py.File('test.h5df','r',swmr=True, libver='latest')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/mchonaker/.local/lib/python3.6/site-packages/h5py/_hl/files.py", line 427, in __init__
    swmr=swmr)
  File "/home/mchonaker/.local/lib/python3.6/site-packages/h5py/_hl/files.py", line 190, in make_fid
    fid = h5f.open(name, flags, fapl=fapl)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py/h5f.pyx", line 96, in h5py.h5f.open
OSError: Unable to open file (file is not already open for SWMR writing)
```
I then tried to open the file in write mode and got the following error:
```
>>> old_file= h5py.File('test.h5df','r+',swmr=True, libver='latest')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/mchonaker/.local/lib/python3.6/site-packages/h5py/_hl/files.py", line 427, in __init__
    swmr=swmr)
  File "/home/mchonaker/.local/lib/python3.6/site-packages/h5py/_hl/files.py", line 192, in make_fid
    fid = h5f.open(name, h5f.ACC_RDWR, fapl=fapl)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
[...]

PR #1812, opened 2021-02-12
Warn when swmr=True is passed with mode!='r'

takluyver:
The `h5py.File(..., swmr=True)` option is only used in 'r' (read-only) mode, and currently silently ignored with any other mode. There's a separate API for starting SWMR write mode (`f.swmr_mode=True`). This is liable to cause confusion if people think they're using SWMR when they're not (as in #1807).

This adds a warning when you pass `swmr=True` and mode other than `'r'`. In the future, it may become an error, but this gives people a chance to fix their code without breaking things that currently work.

### DIFF
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -13,6 +13,7 @@
 
 import sys
 import os
+from warnings import warn
 
 from .compat import filename_decode, filename_encode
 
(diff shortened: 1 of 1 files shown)


takluyver:
I'm not sure why codecov reports a substantial coverage drop. That makes no sense for this simple change.

aragilar:
Merging this also. Thanks @takluyver for all the recent fixes!

PR #1813, opened 2021-02-12
Allow H5PY_TEST_CHECK_FILTERS environment variable into tox

takluyver:
I just noticed that the `test_filters_available` wasn't running where it was meant to on CI, and I think this was why. It's meant to verify that wheels have been built with gzip & lzf filters on all platforms.


takluyver:
I checked a few CI logs, and the relevant test appears to be un-skipped now.

PR #1815, opened 2021-02-16
Create appropriate exception types when the exception message has errno

### LINKED ISSUE #1133: Error number (errno) of OSError is None
I am wondering if it's expected that `error.errno` in raised exception is `None` when a hdf5 file is opened in mode `w-` but already exist?

My scenario looks like this:

```python
import h5py;

try:
    with h5py.File("file-that-already-exist.hdf5", "w-") as f:
        # Do something with the file here
except OSError as error:
    print(error.errno)
    # --> None
    print(str(error))
    # --> OSError: Unable to create file (unable to open file: name = 'file-that-already-exist.hdf5', errno = 17, error message = 'File exists', flags = 15, o_flags = a02)
```

Ideally I would like to be able to do:

```python
expecting_error_no_17 = maybe_yes_maybe_no(my_magic_conditions)
try:
    with h5py.File("file-that-already-exist.hdf5", "w-") as f:
        # Do something with the file here
except OSError as error:
    if error.errno == 17 and expecting_error_no_17:
        pass
    else:
        raise
```

The above code doesn't work. I can achieve it by parsing the error message for `errno` but I am wondering why `error.errno` is `None` instead of the error number reported in the error message? Is it expected that `error.errno` is `None` or is this a bug? If it's expected is there a preferred way of getting the error number or is that up to the user?

`h5py.version.info`:
```bash
h5py    2.8.0
HDF5    1.10.3
Python  3.6.6 | packaged by conda-forge | (default, Jul 26 2018, 09:55:02)
[GCC 4.2.1 Compatible Apple LLVM 6.1.0 (clang-602.0.53)]
sys.platform    darwin
numpy   1.15.3
```

I am running macOS 10.12.6.

takluyver:
We agreed with HDF group that the error message format should be stable enough for us to extract the system errno from it (https://github.com/HDFGroup/hdf5/pull/37). This means we can give nicer exceptions, e.g. `FileNotFoundError` when you try to read a file that doesn't exist:

```
In [2]: f = h5py.File('woijkjer')
FileNotFoundError                         Traceback (most recent call last)
<ipython-input-2-7afd330b19e3> in <module>
----> 1 f = h5py.File('woijkjer')

~/miniconda3/envs/h5py-dev/lib/python3.7/site-packages/h5py/_hl/files.py in __init__(self, name, mode, driver, libver, userblock_size, swmr, rdcc_nslots, rdcc_nbytes, rdcc_w0, track_order, fs_strategy, fs_persist, fs_threshold, **kwds)
    435                                fapl, fcpl=make_fcpl(track_order=track_order, fs_strategy=fs_strategy,
    436                                fs_persist=fs_persist, fs_threshold=fs_threshold),
--> 437                                swmr=swmr)
    438 
    439             if isinstance(libver, tuple):

~/miniconda3/envs/h5py-dev/lib/python3.7/site-packages/h5py/_hl/files.py in make_fid(name, mode, userblock_size, fapl, fcpl, swmr)
    192         if swmr and swmr_support:
    193             flags |= h5f.ACC_SWMR_READ
--> 194         fid = h5f.open(name, flags, fapl=fapl)
    195     elif mode == 'r+':
    196         fid = h5f.open(name, h5f.ACC_RDWR, fapl=fapl)

h5py/h5f.pyx in h5py.h5f.open()

FileNotFoundError: [Errno 2] Unable to open file (unable to open file: name = 'woijkjer', errno = 2, error message = 'No such file or directory', flags = 0, o_flags = 0)
```

I've only exposed the errno and the message we construct from HDF5's error stack. OSError also has a `filename` field which is meant to be the name of the file, if relevant. We could try to fill that as well, but I haven't seen it used nearly as much, and it adds complexity (e.g. undoing the quoting from the message), so I've left it out for now.

I suspect this is technically an API breaking change, as errors that people have previously caught might now be a different type which isn't caught. However, the tests didn't fail, because the most obvious cases switch from the general OSError to more specific subclasses of that. So I think it's reasonable to put it in a 3.x release.

Closes #1133.

### DIFF
--- a/h5py/_errors.pyx
+++ b/h5py/_errors.pyx
@@ -141,9 +141,9 @@ cdef int set_exception() except -1:
     # so that the traceback doesn't point here.
 
     m = re.search(b'errno\s*=\s*(\d+)', desc_bottom)
-    if m:
-        # An errno was found in the message. Python can automatically create
-        # the appropriate OSError subclass (e.g. FileNotFoundError) from this.
+    if m and eclass is OSError:
+        # Python can automatically create an appropriate OSError subclass
+        # (e.g. FileNotFoundError) given the POSIX errno (e.g. ENOENT)
         errno = int(m.group(1))
         PyErr_SetObject(OSError, (errno, msg.decode('utf-8', 'replace')))
     else:
--- a/h5py/_errors.pyx
+++ b/h5py/_errors.pyx
@@ -10,7 +10,8 @@
 # Python-style minor error classes.  If the minor error code matches an entry
 # in this dict, the generated exception will be used.
 
-from cpython cimport PyErr_Occurred
+from cpython cimport PyErr_Occurred, PyErr_SetObject
+import re
 
 _minor_table = {
     H5E_SEEKERROR:      IOError,    # Seek failed
@@ -136,10 +137,17 @@ cdef int set_exception() except -1:
 
     msg = b"%b (%b)" % (bytes(desc).capitalize(), bytes(desc_bottom))
 
-    # Finally, set the exception.  We do this with the Python C function
+    # Finally, set the exception.  We do this with a Python C function
     # so that the traceback doesn't point here.
 
-    PyErr_SetString(eclass, msg)
+    m = re.search(b'errno\s*=\s*(\d+)', desc_bottom)
+    if m:
+        # An errno was found in the message. Python can automatically create
+        # the appropriate OSError subclass (e.g. FileNotFoundError) from this.
+        errno = int(m.group(1))
+        PyErr_SetObject(OSError, (errno, msg.decode('utf-8', 'replace')))
+    else:
+        PyErr_SetString(eclass, msg)
 
     return 1
 
--- a/h5py/_errors.pyx
+++ b/h5py/_errors.pyx
@@ -14,28 +14,28 @@ from cpython cimport PyErr_Occurred, PyErr_SetObject
 import re
 
 _minor_table = {
-    H5E_SEEKERROR:      IOError,    # Seek failed
-    H5E_READERROR:      IOError,    # Read failed
-    H5E_WRITEERROR:     IOError,    # Write failed
-    H5E_CLOSEERROR:     IOError,    # Close failed
-    H5E_OVERFLOW:       IOError,    # Address overflowed
-    H5E_FCNTL:          IOError,    # File control (fcntl) failed
+    H5E_SEEKERROR:      OSError,    # Seek failed
+    H5E_READERROR:      OSError,    # Read failed
+    H5E_WRITEERROR:     OSError,    # Write failed
+    H5E_CLOSEERROR:     OSError,    # Close failed
+    H5E_OVERFLOW:       OSError,    # Address overflowed
+    H5E_FCNTL:          OSError,    # File control (fcntl) failed
 
-    H5E_FILEEXISTS:     IOError,    # File already exists
-    H5E_FILEOPEN:       IOError,    # File already open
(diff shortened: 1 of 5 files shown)


takluyver:
It turns out the tests pass on HDF5 1.10 and 1.12, but there's a failure on 1.8 that shows how this could be a breaking change. The test checks that following an external link to a missing file gets KeyError, but on HDF5 1.8 it gives FileNotFoundError instead:

```
___________________ TestExternalLinks.test_exc_missingfile ____________________

self = <h5py.tests.test_group.TestExternalLinks testMethod=test_exc_missingfile>

    def test_exc_missingfile(self):
        """ KeyError raised when attempting to open missing file """
        self.f['ext'] = ExternalLink('mongoose.hdf5','/foo')
        with self.assertRaises(KeyError):
>           self.f['ext']

py37-test-deps\lib\site-packages\h5py\tests\test_group.py:737: 
h5py\_objects.pyx:54: in h5py._objects.with_phil.wrapper
    ???
h5py\_objects.pyx:55: in h5py._objects.with_phil.wrapper
    ???
py37-test-deps\lib\site-packages\h5py\_hl\group.py:288: in __getitem__
    oid = h5o.open(self.id, self._e(name), lapl=self._lapl)
h5py\_objects.pyx:54: in h5py._objects.with_phil.wrapper
    ???
h5py\_objects.pyx:55: in h5py._objects.with_phil.wrapper
    ???

>   ???
E   FileNotFoundError: [Errno 2] Unable to open object (unable to open file: name = 'C:\Users\VSSADM~1\AppData\Local\Temp\h5py-test_ll5knmh5\mongoose.hdf5', errno = 2, error message = 'No such file or directory', flags = 1, o_flags = 2)

h5py\h5o.pyx:190: FileNotFoundError
```

I think we should make that raise KeyError again, as it hasn't changed for newer HDF5 versions. But there are a lot of operations in HDF5 which can read/write data and potentially get a system error, including several places where it might try to open a new file (external links, external datasets, adding data with the family file driver).

I'm inclined to duct-tape over specific cases like the one above (to make it KeyError again), and otherwise surface the OSError wherever HDF5 gives us an errno. But we could also restrict the OSError behaviour to specific places (like when you use `h5py.File()` to explicitly open a file) or specific errnos (e.g. ENOSPC for when you're out of disk space). We could also only allow it where we would already have raised OSError - though we may then find cases where we should but don't.

takluyver:
I decided to only expose the errno in places where we were already raising OSError, which should avoid most breakage (it's still possible that code is broken if it handles e.g. FileNotFoundError differently from a general OSError, but that's unavoidable). If there are cases where we should be raising OSError but aren't, we'll need to notice and fix those inidividually.

I'm confused again about why coverage seems to have dropped. The main change codecov shows is a branch for szip not being taken, but looking at the CI logs, at least some of them definitely aren't skipping the szip test, and have higher coverage of `filters.py` than codecov shows. :confused:

takluyver:
@aragilar if you've got some time to look at a PR... :wink:  This one is a bit more involved than my other recent PRs, but I'd quite like to get it in for 3.2 - I think it will be a nice improvement to the exceptions we raise when things go wrong.

aragilar:
I think this can be merged. I was planning on seeing if I could get something that autogenerated a subclass of the older exception plus a more descriptive exception in the exception hierarchy, but I suspect that depends on me having time to work on it and it being more robust than our current guessing ;).

takluyver:
Thanks!

I've done it at the moment to only affect errors that were already `OSError`, and where the exception class changes it will only be to a subclass of OSError. That's a relatively conservative change, but I think it improves the cases we most often care about (getting a specific error when trying to open a file).

PR #1816, opened 2021-02-17
Document that object arrays for vlen are a workaround

takluyver:
Ragged data with vlen datatypes is probably never going to work as nicely in h5py as regular, rectangular data. This came up in #1548, and I'd like to add a gentle pointer in the docs to steer people away from using it if they can easily find an alternative. We added a similar warning in #1704 for wrapping Python file-like objects.

Closes gh-1548


aragilar:
Given this is a doc change, not sure why the coverage is dropping... merging anyway.

PR #1818, opened 2021-02-17
Fix read_direct & write_direct with integer indices

takluyver:
Thanks @bmerry for providing a simple test case for this, which I've included here.

`sel.array_shape` drops dimensions where the selection uses scalar indexing, which is what we want here. `sel.mshape` keeps those dimensions, because HDF5 has no concept of scalar indexing, so indexing `5` has to be handled more like `5:6`.

Along the way, I also renamed an exception class in the tests to squash an annoying warning (pytest was seeing the TestException class as a potential test).

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -969,7 +969,7 @@ class Dataset(HLObject):
             else:
                 dest_sel = sel.select(dest.shape, dest_sel)
 
-            for mspace in dest_sel.broadcast(source_sel.mshape):
+            for mspace in dest_sel.broadcast(source_sel.array_shape):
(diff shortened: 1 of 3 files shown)


takluyver:
- API docs failure appears to be a random network error. Re-running it now.
- Coverage 'drop' is similar to other recent PRs. I don't understand it, but I don't think it represents a real change in coverage.

Sometimes it seems like CI is as much work as the actual code. :disappointed:

aragilar:
Not sure why this has a coverage drop either (SZIP support seems to be what's being flagged, but it's not changing in this PR...). Given this is fixing bugs, merging.

takluyver:
Thanks @aragilar! I realised too late I forgot to put news entries in a couple of these PRs; I'll make a new little PR for those.

Issue #1822, opened 2021-02-22
Reading custom opaque types


kutoga:
To assist reproducing bugs, please include the following:
 * Operating System: Ubuntu 18.04
 * Python version: 3.6.9
 * Where Python was acquired: System python
 * h5py version: 3.1.0
 * HDF5 version: 1.12.0

I have a h5-file which contains an array of a custom opaque 2-byte datatype (`bfloat16`). The data is written by a non-python application and should now be read with h5py.

I try to read the data as 2-byte-blobs and then to convert them to a float32 numpy-array. Unfortunately, I could not find a way to just read the blobs (the conversion should then be trivial).

I always get this error:

```
h5py/_objects.pyx in h5py._objects.with_phil.wrapper()

h5py/_objects.pyx in h5py._objects.with_phil.wrapper()

/mnt/c/git/test/.venv/lib/python3.6/site-packages/h5py/_hl/dataset.py in __getitem__(self, args)
    571         mspace = h5s.create_simple(mshape)
    572         fspace = selection.id
--> 573         self.id.read(mspace, fspace, arr, mtype, dxpl=self._dxpl)
    574 
    575         # Patch up the output for NumPy

h5py/_objects.pyx in h5py._objects.with_phil.wrapper()

h5py/_objects.pyx in h5py._objects.with_phil.wrapper()

h5py/h5d.pyx in h5py.h5d.DatasetID.read()

h5py/_proxy.pyx in h5py._proxy.dset_rw()

h5py/_proxy.pyx in h5py._proxy.H5PY_H5Dread()

OSError: Can't read data (no appropriate function for conversion path)
```

Is there a possibility to register a custom function for reading opaque data? Alternatively it would also be okay to read the data as (u)int16 if this is possible.

Thanks

aragilar:
Do you have some example code? Incidentally, you should not store bfloats in an opaque type, HDF5's floating point support allows you to specify enough details such that you can store it natively, and h5py will do the automatic conversion to the smallest floating point type which can losslessly hold the data that your system supports.

kutoga:
Okay, I see. I'll check the documentation regarding custom floating point types.

A snippet of the C-code:
```c
hid_t h5_float_dtype = H5Tcreate(H5T_OPAQUE, float_size);
const herr_t status = H5Tset_tag(file->h5_float_dtype, "bfloat16");
```
Later the data is converted to a bfloat16 and written.

There is not yet too much python-code. Just some playground:
```python
[...]

PR #1823, opened 2021-02-22
Add test for storing & reading opaque (np.void) data

takluyver:
Looking at #1822, I realised there didn't seem to be a test for this.


aragilar:
Not sure what's up with codecov, but I can't see the reducing coverage ;). Thanks @takluyver, merging.

takluyver:
Thanks!  Yes, I'm puzzled as to why codecov keeps thinking my changes decrease coverage.

PR #1824, opened 2021-02-22
Document how to install for development

takluyver:
I've written up what I do for testing as I change h5py, and added a tiny script for convenience.

I also made a few other changes to the source installation section while I was looking at it.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,5 +1,6 @@
 include AUTHORS
 include api_gen.py
+include dev-install.sh
 include LICENSE
 include MANIFEST.in
 include pylintrc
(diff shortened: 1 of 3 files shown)


aragilar:
Minor question, why do the `setup.py build` separately, I can see maybe if you were adding `--in-place` it would make sense, but I've never really trusted `--in-place` not to subtlety break something? Otherwise this looks good to me.

takluyver:
When you do `pip install .`, it copies the files from the current folder to a temporary build folder before running `setup.py`. So if you always do that, I think you get a clean build each time, with no caches. But if you run `setup.py build` first, compiled modules go in the `build/` folder, only recompiling if they've changed. Then pip copies `build/` into its temporary folder, so it doesn't have to recompile anything.

It's not a particularly pretty solution, but it does mean most small changes can be installed in just a few seconds.

PR #1831, opened 2021-02-25
Fail fast in dev-install.sh when build fails

takluyver:
I forgot shell scripting 101... If `setup.py build` fails, it should stop immediately, rather than trying to install with pip, which makes the primary error less obvious.


takluyver:
Merging this myself since it's a simple change.

PR #1833, opened 2021-02-27
Document where to get filter plugins

### LINKED ISSUE #1606: Document how to use/link to hdf5plugin
[hdf5plugin](https://github.com/silx-kit/hdf5plugin) provides a number of additional compression filters. We should add these filters to those listed at https://h5py.readthedocs.io/en/stable/high/dataset.html#filter-pipeline, and document them appropriately.

This to me seems like the best solution for #611.

takluyver:
Inc. hdf5plugin (the Python package), HDF group's package of plugins, and the index of plugins.

Closes #1606.

### DIFF
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -261,6 +261,18 @@ dynamically loaded by the underlying HDF5 library. This is done by passing a
 filter number to :meth:`Group.create_dataset` as the ``compression`` parameter.
 The ``compression_opts`` parameter will then be passed to this filter.
 
+.. seealso::
+
+   `hdf5plugin <https://pypi.org/project/hdf5plugin/>`_
+     A Python package with several popular filters, including Blosc and LZ4,
+     for convenient use with h5py
+
+   `HDF5 Filter Plugins <https://portal.hdfgroup.org/display/support/HDF5+Filter+Plugins>`_
+     A collection of filters as a single download from The HDF Group
+
+   `Registered filter plugins <https://portal.hdfgroup.org/display/support/Filters>`_
+     The index of publicly announced filter plugins
+
 .. note:: The underlying implementation of the compression filter will have the
     ``H5Z_FLAG_OPTIONAL`` flag set. This indicates that if the compression
     filter doesn't compress a block while writing, no error will be thrown. The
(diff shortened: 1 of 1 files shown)


docs/high/dataset.rst
```diff
@@ -261,6 +261,18 @@ dynamically loaded by the underlying HDF5 library. This is done by passing a
 filter number to :meth:`Group.create_dataset` as the ``compression`` parameter.
 The ``compression_opts`` parameter will then be passed to this filter.
 
+.. seealso::
+
+   `hdf5plugin <https://pypi.org/project/hdf5plugin/>`_
+     A Python package with several popular filters, including Blosc and LZ4,
```
takluyver, line 267: Q: which filters to highlight from hdf5plugin? I think Blosc should certainly be mentioned, and I picked LZ4 as the other one I think I'd heard of before learning about HDF5. But maybe it's better to mention ZFP (a lossy filter) to show the breadth. Or maybe it's simpler just to list them all.

cc @kif @vasole @t20100

vasole:
I think ZFP is certainly worth to be mentioned.  It can do wonders with (lossy) compression of floating point data and it is a filter people should consider.

Bitshuffle+LZ4 was the reason we started to provide the module : to support data from some detectors mostly used at synchrotrons but the potential users are mostly already aware of the existence of hdf5plugin :)

If you want to provide some details then I would restrict to Blosc, LZ4 and ZFP as you had envisaged. If you are just going to mention them, then it does not take much more to mention all those already supported.

docs/high/dataset.rst
t20100: I'd mention Blosc as well and ZFP to have a lossy filter.
If you list it all, you probably want to do the same for HDF5 Filter Plugins, and we'll just have to think of updating `h5py` doc when we add a new filter to the zoo.

takluyver:
Thanks both! I'll add ZFP. I'm inclined to give examples rather than a complete list to reduce the risk of it getting out of date, though this isn't a strong preference if you want to argue for a full list.

vasole:
It is fine with me. The reduced list points people to the filters that are really worth to take a closer look at.

PR #1834, opened 2021-03-01
Prepare for 3.2 release

takluyver:
By running rever steps `authors`, `version_bump` and `changelog`, and manually tidying up the release notes.


aragilar:
Looks reasonable to me (the changes that don't seem to be mentioned based `git shortlog 3.1.0... | grep Merge | sort` seem to be CI/testing changes, which I don't think we need to call out in the changelog, but @takluyver definitely deserves thanks for those). I'm happy to merge this as is, but I'll leave this for a day or so for any other comments (or whenever @takluyver wants to merge this, as I suspect the actual release will be set by his availability ;)).

takluyver:
Thanks @aragilar ! Yup, I generally try to focus the release notes on changes that can affect people using h5py, though there is a tension between that and using release notes to acknowledge everyone's work.

I'll leave this a couple of days to give all the maintainers a chance to comment, or to press for getting in some change they think should be prioritised.

PR #1839, opened 2021-03-03
Deprecate setting default_file_mode, only allow setting to 'r'

takluyver:
This is step 2 of removing the config option added to ease the transition from 2.x to 3.x:

1. Setting the default to anything other than 'r' (which is the default since 3.0) warns (done in #1789 for h5py 3.2)
2. Setting the default to 'r' warns, and to anything else is an error (this PR)
3. Remove the default_file_mode option

We added the option (and indicated that the default would change) in 2.10 (released September 2019). The default changed to 'r' in 3.0 (October 2020). The warning for setting the default to modes other than 'r' was added in 3.2 (March 2021).

I'm provisionally marking this for 3.3, but I'm happy to leave it longer if people prefer. We could also start error-ing on modes other than r but leave the warning on setting the default to r for later.

### DIFF
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -315,7 +315,7 @@ class File(Group):
         else:
             raise RuntimeError('SWMR support is not available in HDF5 version {}.{}.{}.'.format(*hdf5_version))
 
-    def __init__(self, name, mode=None, driver=None,
+    def __init__(self, name, mode='r', driver=None,
                  libver=None, userblock_size=None, swmr=False,
                  rdcc_nslots=None, rdcc_nbytes=None, rdcc_w0=None,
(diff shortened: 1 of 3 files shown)


takluyver:
I'm aiming for 3.3 fairly soon. On the one hand, 3 months is fairly short for going from a warning to an error. On the other hand, the proper fix - adding an explicit mode to `h5py.File` calls wherever necessary - is simple and obvious, and people have had the best part of 2 years to make that change (since 2.10 warned that the default was changing). So I will merge this in a few more days if no-one objects.

tacaswell (approved):

tacaswell:
Given @aragilar' comment is https://github.com/h5py/h5py/issues/1909#issuecomment-863139182 I'm going to merge this.

PR #1840, opened 2021-03-03
Update cibuildwheel to 1.10.0

takluyver:
I noticed while preparing 3.2 that there have been a few releases of cibuildwheel since we configured it. I don't think we've had any major problems with the wheels, so I finished making that release with cibuildwheel 1.6. But now I'll try using the newer version.

Changelog: https://cibuildwheel.readthedocs.io/en/stable/changelog/


takluyver:
I'll merge this one on Monday if no-one objects.

PR #1842, opened 2021-03-03
Deprecate 'with dset.astype()'

takluyver:
h5py 3.0 introduced the ability to slice an astype wrapper directly (`ds.astype(np.int32)[:]`). I'd like to remove the old pattern of using astype() as a context manager with mutable state, and get rid of the thread-local storage it requires.

Edit: I am happy if we want to give people more time to update to 3.x before adding this warning. We released 3.0 in October 2020. The 3.3 release may be around 6 months after that, depending on how eager we are to do it.

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -17,12 +17,14 @@ except ImportError:
     from cached_property import cached_property
 import posixpath as pp
 import sys
+from warnings import warn
 
 from threading import local
 
 import numpy
 
 from .. import h5, h5s, h5t, h5r, h5d, h5p, h5fd, h5ds, _selector
+from ..h5py_warnings import H5pyDeprecationWarning
 from .base import HLObject, phil, with_phil, Empty, find_item_type
 from . import filters
 from . import selections as sel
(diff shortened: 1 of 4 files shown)


takluyver:
Add the warning for 3.3, or wait a while longer?

This is trickier than #1839. The sooner we start pushing people away from this, the sooner we can get rid of the thread-locals piece. But the only way to keep code compatible with h5py 2.x and 3.x is to use the older style (`with dset.astype(np.int32):`). This also means it's necessary for any code that still has to support Python 2.

I'm leaning towards leaving this a bit longer - maybe add the warning a year after h5py 3.0 was released.

tacaswell (dismissed):

tacaswell:
I'm sold that we should let people have a 12 month support window for older versions of h5py.  This should wait until Nov 2021 (which I presume will be h5py3.4).

aragilar:
Should we merge this now?

takluyver:
Yes, it has been a year since 3.0 was released with the new `arr = dset.astype(np.int32)[:10]` pattern, which was the delay we agreed on before warning. We can allow it with a warning for another 6 or 12 months, I guess.

PR #1844, opened 2021-03-04
Fix File.driver when built with ROS3 enabled

### LINKED ISSUE #1843: Test failures with ROS3 enabled
@satra, can you take a look?

Conda-forge tried to update to 3.2 (https://github.com/conda-forge/h5py-feedstock/pull/86) but got test failures. It appears that their HDF5 has the ROS3 driver enabled. All the tracebacks I've spotted end here:

```
    @property
    @with_phil
    def driver(self):
        """Low-level HDF5 file driver used to open file"""
        drivers = {h5fd.SEC2: 'sec2',
                   h5fd.STDIO: 'stdio',
                   h5fd.CORE: 'core',
                   h5fd.FAMILY: 'family',
                   h5fd.WINDOWS: 'windows',
                   h5fd.MPIO: 'mpio',
                   h5fd.MPIPOSIX: 'mpiposix',
                   h5fd.fileobj_driver: 'fileobj'}
        if ros3:
>           drivers[h5fd.ROS3] = 'ros3'
E           AttributeError: module 'h5py.h5fd' has no attribute 'ROS3'
```

satra:
closes #1843 

checking where enabling the ROS3 variable breaks the h5py build.


satra:
@takluyver - the good news is that i can successfully build and test the conda-forge build (locally using their build stack) by adding `ROS3 = H5FD_ROS3` into h5fd.pyx. 

however, adding that makes it fail in the CIs here for versions less than 1.10.6. so i tried to make that a conditional. however, when cython runs it seems that the variable HDF5_VERSION is None at that point. any suggestions?

takluyver:
Thanks for looking into this promptly!

I think you need to `include "config.pxi"` in the file where you check HDF5_VERSION, like here:

https://github.com/h5py/h5py/blob/57a57b34f0ae14897b71f28dee3575830944fcf9/h5py/h5pl.pyx#L18-L23

takluyver:
Thanks! Don't worry about codecov too much - it currently likes to tell us that every PR decreases coverage :shrug: 

I'll probably merge this & do a bugfix release tomorrow, if other maintainers don't get there first.

satra:
@takluyver - i'm testing the conda build now - will post here and clean up the PR.

satra:
@takluyver - just pushed a release note. the conda build worked with these patches.

news/ros3-driver-fix.rst
```diff
@@ -0,0 +1,30 @@
...
+Exposing HDF5 functions
+-----------------------
+
+* <news item>
+
+Bug fixes
+---------
+
+* Fix so that h5py builds cleanly on systems with and without HDF5 drivers with
+  ROS3 enabled.
```
takluyver, line 20:
```suggestion
* Fix :attr:`.File.driver` when the read-only S3 driver is available.
```

takluyver:
Thanks @satra . I'm just uploading a 3.2.1 release with this fix.

PR #1848, opened 2021-03-06
Use >= for numpy dep in SETUP_REQUIRES consistently

mgorny:
The numpy dep uses >= operator in RUN_REQUIRES but == in SETUP_REQUIRES.
This seems like a typo, so update it to use >= consistently.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,7 @@ SETUP_REQUIRES = [
     "Cython >=0.29.14; python_version=='3.8'",
     "Cython >=0.29.15; python_version>='3.9'",
 ] + [
-    f"numpy =={np_min}; python_version{py_condition}"
+    f"numpy >={np_min}; python_version{py_condition}"
     for np_min, py_condition in NUMPY_MIN_VERSIONS
 ]
 


mgorny:
NB: I'm not sure if this change requires a news entry.

takluyver:
Thanks, but no, it's not a typo. The idea is that we build with the oldest supported version of numpy, and then the resulting wheels work with newer versions. It's not an ideal solution, but currently it's the best option we're aware of. This comment tries to explain it:

https://github.com/h5py/h5py/blob/7f918857e2a63477038afccda518ebd3159c13b2/setup.py#L42-L45

The [oldest-supported-numpy](https://pypi.org/project/oldest-supported-numpy/) metapackage has more details about this idea.

mgorny:
While indeed that sounds like an interesting solution, it sucks a bit for end users building from source. I guess pip works around this somehow but for distribution maintainers, it means we have to keep patching the pin because `setup.py` now fails if another version of numpy is installed. Could you please consider moving this logic somewhere else in the build pipeline so that it doesn't affect end users?

takluyver:
For downstream packaging, you likely want to bypass all of pip & setuptools mechanisms for getting build dependencies. That's documented under 'development installation', and should be possible without patching it:

https://docs.h5py.org/en/stable/build.html#development-installation

aragilar:
@mgorny If there were an environment variable (e.g. `H5PY_DISTRO_REQ_BUILD`) which set `setup_requires` to `[]`, would that be sufficient for your purposes, or are you trying to extract/use the `setup_requires` information at build time. The latter is much more complicated to support, and based on our previous survey of the state of build backends, there's no backend which provides the needed hooks to solve this.

takluyver:
> If there were an environment variable (e.g. H5PY_DISTRO_REQ_BUILD) which set setup_requires to []

There is :smile: : `H5PY_SETUP_REQUIRES=0`. I accept we could document it better, but it is now mentioned in the docs.

aragilar:
@takluyver I probably should have checked the actual code first before commenting ;)

@mgorny Does `H5PY_SETUP_REQUIRES=0` cover what you need? I highly doubt we're going to be adding new build-time dependencies any time soon (I think pkgconfig was the most recent one, and we've had that for ages), so manually specifying the build dependencies should be fine (at worst, you may need to be careful with cython versions, as I know we had some issues building with versions somewhere in the range of 0.23-0.25 previously).

mgorny:
@takluyver, thank you. I feel stupid that I didn't notice it — though I have to admit it never occurred to me that such a thing would exist in the first place ;-).

@aragilar, yes, that works for me. Sorry about the trouble.

takluyver:
No worries. We're still navigating our way around the changing landscape of Python packaging, and also sometimes contributing to discussions about the packaging machinery.

PR #1854, opened 2021-03-10
Make file closing and get_obj_ids more robust

### LINKED ISSUE #1495: Error when garbage collecting from non-terminated generator
Versions: 
 * h5py=="any"

If you have an unterminated generator then a `SystemError` is thrown when the garbage collector runs at the end of a module. 
```
import h5py

def yield_groups(filename):
    with h5py.File(filename, 'r') as fh:
        for group in fh:
            yield group

filename = "file_with_10_groups.hdf5"
grp_generator = yield_groups(filename)
next(grp_generator)
```
Raises 
```
Exception ignored in: <generator object yield_groups at 0x7f3fa6690468>
  File "h5py_generator.py", line 6, in yield_groups
  File "venv/lib/python3.5/site-packages/h5py/_hl/files.py", line 461, in __exit__
  File "venv/lib/python3.5/site-packages/h5py/_hl/files.py", line 432, in close
  File "h5py/h5f.pyx", line 267, in h5py.h5f.get_obj_ids
  File "h5py/h5i.pyx", line 37, in h5py.h5i.wrap_identifier
SystemError: Parent module 'h5py' not loaded, cannot perform relative import
```
This is a pretty minor/niche bug. It seems that the `h5py` module is unloaded by the garbage collector before the file handle is closed, which means that the` __exit__()` on the `File` object cannot be called successfully. It only occurs for unterminated generators, with an open file handle, which are being called at the module level.  

This issue does not occur when if the generator is terminated (either by `grp_generator.close()` or `for i in grp_generator: pass`). It also does not happen if the generator call is in a function e.g.:
```
def print_groups():
    filename = "file_with_10_groups.hdf5"
    grp_generator = yield_groups(filename)
    grp = next(grp_generator)
    print(grp)

print_groups()
```
since the `__exit__()` on the file handler is called when the generator object goes out of scope, before it unloads the h5py library at the end of execution.

### LINKED ISSUE #1852: Issue when garbage collecting h5py objects during File.close
There is an issue when the garbage collector is run during File.close. Here is a script that reproduce the issue

```py
import h5py

N = 100
with h5py.File('foo.h5', 'w') as f:
    for i in range(N):
        f[str(i)] = []

for i in range(100):
    print("i =",i)
    with h5py.File('foo.h5', 'r') as f:
        refs = []
        for i in range(N):
            refs.append(f[str(i)].id)
        refs.append(refs)
        del refs  # <-- important to make it likely that gc of id refs happens during h5py.File.close
```

The traceback is

```pytb
i = 0
i = 1
  File "/Users/aaronmeurer/Documents/versioned-hdf5/test3.py", line 15, in <module>
    del refs  # <-- important to make it likely that gc of id refs happens during h5py.File.close
  File "/Users/aaronmeurer/Documents/h5py/h5py/_hl/files.py", line 493, in __exit__
    self.close()
  File "/Users/aaronmeurer/Documents/h5py/h5py/_hl/files.py", line 463, in close
    id_list = h5f.get_obj_ids(self.id, ~h5f.OBJ_FILE)
  File "h5py/h5f.pyx", line 277, in h5py.h5f.get_obj_ids
RuntimeError: Can't increment id ref count (can't locate ID)
```

The loop is there to make sure that the issue consistently reproduces. Sometimes a small change can appear to "fix" the issue because it changes when the garbage collector is run, but it will still occur after enough iterations.

The issue exists in h5py 2.10, 3.1, and the git master. 

The problem is that `get_obj_ids` gets a list of valid object IDs from `H5Fget_obj_ids`, but one of the object IDs becomes invalid in the middle of the function call, so `H5Iinc_ref` fails. 

A viable workaround is to either call `gc.collect()` before the file is closed (at the end of the `with` block), or to disable garbage collection in `File.close`, like

```patch
diff --git a/h5py/_hl/files.py b/h5py/_hl/files.py
index 60a24c0c..a0f1766e 100644
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -460,22 +460,27 @@ class File(Group):
 
                 # Close file-resident objects first, then the files.
                 # Otherwise we get errors in MPI mode.
-                id_list = h5f.get_obj_ids(self.id, ~h5f.OBJ_FILE)
-                file_list = h5f.get_obj_ids(self.id, h5f.OBJ_FILE)
-
-                id_list = [x for x in id_list if h5i.get_file_id(x).id == self.id.id]
-                file_list = [x for x in file_list if h5i.get_file_id(x).id == self.id.id]
-
-                for id_ in id_list:
-                    while id_.valid:
-                        h5i.dec_ref(id_)
-
-                for id_ in file_list:
-                    while id_.valid:
-                        h5i.dec_ref(id_)
-
-                self.id.close()
-                _objects.nonlocal_close()
+                import gc
+                gc.disable()
+                try:
+                    id_list = h5f.get_obj_ids(self.id, ~h5f.OBJ_FILE)
+                    file_list = h5f.get_obj_ids(self.id, h5f.OBJ_FILE)
+
+                    id_list = [x for x in id_list if h5i.get_file_id(x).id == self.id.id]
+                    file_list = [x for x in file_list if h5i.get_file_id(x).id == self.id.id]
+
+                    for id_ in id_list:
+                        while id_.valid:
+                            h5i.dec_ref(id_)
+
+                    for id_ in file_list:
+                        while id_.valid:
+                            h5i.dec_ref(id_)
+
+                    self.id.close()
+                    _objects.nonlocal_close()
+                finally:
+                    gc.enable()
 
     def flush(self):
         """ Tell the HDF5 library to flush its buffers.
```

Both of these appear to consistently make the issue go away. I don't know if that is a real "solution" or just a bandaid to the real underlying problem. I'm not familiar enough with the Python garbage collection, Cython, or HDF5 to know what a real fix should look like for this.

```

h5py    3.2.1
HDF5    1.10.6
Python  3.9.1 | packaged by conda-forge | (default, Jan 26 2021, 01:32:59)
[Clang 11.0.1 ]
sys.platform    darwin
numpy   1.20.1
cython (built with) 0.29.21
numpy (built against) 1.19.3
HDF5 (built against) 1.10.6
```

takluyver:
This now contains two related changes.

First, temporarily disable garbage collection in `h5py.h5f.get_obj_ids`, so it can't invalidate HDF5 IDs between us retrieving them and calling `H5Iinc_ref` to keep them alive. This fixes #1852.

However, `get_obj_ids` is still a fairly awkward thing to call from `File.close()` - it creates h5py ObjectID objects, which can cause problems at interpreter shutdown, and it increments HDF5 ref counts only to decrement them again. Now it also has to turn garbage collection off and on again. So I moved part of the `File.close` logic into a `FileID._close_open_objects` method, which calls the autogenerated bindings directly and avoids these complexities. This fixes #1495.

### DIFF
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -460,19 +460,8 @@ class File(Group):
 
                 # Close file-resident objects first, then the files.
                 # Otherwise we get errors in MPI mode.
-                id_list = h5f.get_obj_ids(self.id, ~h5f.OBJ_FILE)
-                file_list = h5f.get_obj_ids(self.id, h5f.OBJ_FILE)
-
-                id_list = [x for x in id_list if h5i.get_file_id(x).id == self.id.id]
-                file_list = [x for x in file_list if h5i.get_file_id(x).id == self.id.id]
-
-                for id_ in id_list:
-                    while id_.valid:
-                        h5i.dec_ref(id_)
-
-                for id_ in file_list:
-                    while id_.valid:
-                        h5i.dec_ref(id_)
+                self.id._close_open_objects(h5f.OBJ_LOCAL | ~h5f.OBJ_FILE)
+                self.id._close_open_objects(h5f.OBJ_LOCAL | h5f.OBJ_FILE)
 
                 self.id.close()
                 _objects.nonlocal_close()
--- a/h5py/h5f.pyx
+++ b/h5py/h5f.pyx
@@ -330,6 +330,25 @@ cdef class FileID(GroupID):
         self._close()
         _objects.nonlocal_close()
 
+    @with_phil
+    def _close_open_objects(self, int types):
+        # Used by File.close(). This avoids the get_obj_ids wrapper, which
+        # creates Python objects and increments HDF5 ref counts while we're
+        # trying to clean up. E.g. that can be problematic at Python shutdown.
+        cdef int count, i
+        cdef hid_t *obj_list = NULL
+
+        count = H5Fget_obj_count(self.id, types)
+        if count == 0:
+            return
+        obj_list = <hid_t*> emalloc(sizeof(hid_t) * count)
+        try:
+            H5Fget_obj_ids(self.id, types, count, obj_list)
+            for i in range(count):
+                while H5Iis_valid(obj_list[i]):
+                    H5Idec_ref(obj_list[i])
+        finally:
+            efree(obj_list)
 
     @with_phil
     def reopen(self):
--- a/h5py/h5f.pyx
+++ b/h5py/h5f.pyx
@@ -24,6 +24,7 @@ from .h5ac cimport CacheConfig
 from .utils cimport emalloc, efree
 
 # Python level imports
+import gc
 from . import _objects
 from ._objects import phil, with_phil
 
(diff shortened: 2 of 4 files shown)


asmeurer:
I can confirm that this also makes the issue go away for the other reproducers I'm aware of.

asmeurer:
Is there a rough timeline on when a version with this will be released?

takluyver:
Sorry for the delay. Things seem to have gone quiet, which hopefully means 3.2.1 is working OK for most people. But I'll try to do a new release some time soon. I would like to:

- Get #1889 in (convert bitfield data to unsigned ints by default)
- Do something in the vein of #1863 & #1865 (improving how virtual dataset creation works)
- Decide whether deprecations #1839 & #1842 go in 3.3 or wait for another release.

PR #1856, opened 2021-03-10
Raise ValueError when accessing a group like a dataset

### LINKED ISSUE #1829: Clearer error message when indexing a group like a dataset
HDF5 groups work like dictionaries in h5py. You use string keys (or special reference objects) to access items from them. If you think you're indexing a dataset but you've actually got a group, you get rather unhelpful exceptions like this:

```
In [7]: f[0]
AttributeError                            Traceback (most recent call last)
<ipython-input-7-d8711d29ed7d> in <module>
----> 1 f[0]

~/.local/lib/python3.9/site-packages/h5py/_hl/group.py in __getitem__(self, name)
    286                 raise ValueError("Invalid HDF5 object reference")
    287         else:
--> 288             oid = h5o.open(self.id, self._e(name), lapl=self._lapl)
    289 
    290         otype = h5i.get_type(oid)

~/.local/lib/python3.9/site-packages/h5py/_hl/base.py in _e(self, name, lcpl)
    198         else:
    199             try:
--> 200                 name = name.encode('ascii')
    201                 coding = h5t.CSET_ASCII
    202             except UnicodeEncodeError:

AttributeError: 'int' object has no attribute 'encode'
```

We should provide a more meaningful error message in cases like this.

Prompted by #1826

cydanil:
Hello,
This PR aims to close #1829 , a request for clearer error messages when accessing a group with non-encodable types.  

When a group is being accessed with something else, such as `...`, then a `ValueError` is raised.  
I've added some tests in a place that seemed reasonable, but I'd need confirmation.

Thanks!
Cyril

### DIFF
--- a/h5py/_hl/group.py
+++ b/h5py/_hl/group.py
@@ -284,6 +284,9 @@ class Group(HLObject, MutableMappingHDF5):
             oid = h5r.dereference(name, self.id)
             if oid is None:
                 raise ValueError("Invalid HDF5 object reference")
+        elif not isinstance(name, (bytes, str)):
+            raise TypeError("Accessing a group is done with bytes or str, "
+                            " not {}".format(type(name)))
         else:
             oid = h5o.open(self.id, self._e(name), lapl=self._lapl)
 
--- a/h5py/_hl/group.py
+++ b/h5py/_hl/group.py
@@ -284,11 +284,11 @@ class Group(HLObject, MutableMappingHDF5):
             oid = h5r.dereference(name, self.id)
             if oid is None:
                 raise ValueError("Invalid HDF5 object reference")
-        elif not isinstance(name, (bytes, str)):
+        elif isinstance(name, (bytes, str)):
+            oid = h5o.open(self.id, self._e(name), lapl=self._lapl)
+        else:
             raise TypeError("Accessing a group is done with bytes or str, "
                             " not {}".format(type(name)))
-        else:
-            oid = h5o.open(self.id, self._e(name), lapl=self._lapl)
 
         otype = h5i.get_type(oid)
         if otype == h5i.GROUP:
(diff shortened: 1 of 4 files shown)


takluyver:
Thanks Cyril! A couple of suggestions:

1. I think `TypeError` would be preferable for the exception class - after all, it's saying that something is the wrong type. You get a TypeError from `[]['a']` and `os.environ[2]`, for instance.
2. If the error message talks about groups, it should come from the group class, because the `_e` method is used by other code as well.
3. I'd probably do this with an `isinstance()` check, rather than trying to allow duck-typing. String types are pretty fundamental, and we already check for `str` and `bytes` specifically in various places in the code. I've just checked, and NumPy `np.str_` and `np.bytes_` objects also count as instances of `str` and `bytes` respectively.

takluyver:
Also, don't forget to [write a release note](https://docs.h5py.org/en/latest/contributing.html#write-a-release-note).

cydanil:
Thanks for your comments: I've now made the validation when accessing groups only and changed the `ValueError` to the more reasonable `TypeError`.

h5py/_hl/group.py
```diff
@@ -284,6 +284,9 @@ def __getitem__(self, name):
             oid = h5r.dereference(name, self.id)
             if oid is None:
                 raise ValueError("Invalid HDF5 object reference")
+        elif not isinstance(name, (bytes, str)):
+            raise TypeError("Accessing a group is done with bytes or str, "
+                            " not {}".format(type(name)))
         else:
             oid = h5o.open(self.id, self._e(name), lapl=self._lapl)
```
takluyver, line 291:
```suggestion
        elif isinstance(name, (bytes, str)):
            oid = h5o.open(self.id, self._e(name), lapl=self._lapl)
        else:
            raise TypeError("Accessing a group is done with bytes or str, "
                            " not {}".format(type(name)))
```

This is logically exactly the same, so it's not important, but I think it's slightly neater to group together the cases it can handle, and have the error in a final `else`. E.g. if we one day wanted to handle another type, we could just drop another `elif` in.

cydanil: Thanks!

takluyver:
Thank you! I'll merge this once the CI has passed (& possibly once I've slept :-)

PR #1858, opened 2021-03-11
Replace H5E_ATOM error category with H5E_ID

hyoklee:
The latest HDF5 will no longer have ATOM. It's replaced with ID.

See https://github.com/HDFGroup/hdf5/commit/c256612e0950f3389249aa3a151d66ce01704d89#diff-b3f4171cdf3ed0a754c3f5e3ddbb109716f6acf2248b20fa59f24d04c16d696f for the deatils.


takluyver:
We will need to support existing versions of HDF5 - with `ATOM` - for quite some time. Have a look at `api_types_hdf5.pxd` (among other files) to see how we can define things conditionally on different HDF5 versions.

takluyver:
PR #1875 has tackled the H5E_ID change in a compatible way.

I see this PR now contains a couple of other pieces, but it looks like they're things you're experimenting with and not close to being usable. Is it OK if we close this PR?

hyoklee:
> PR #1875 has tackled the H5E_ID change in a compatible way.
> 
> I see this PR now contains a couple of other pieces, but it looks like they're things you're experimenting with and not close to being usable. Is it OK if we close this PR?

Yes.

takluyver:
Thanks!

PR #1861, opened 2021-03-15
[DOC] - Update README link to the HDF5 website

TomDonoghue:
This is a quick update to the link to HDF5 which is in the README. 

The current link (https://support.hdfgroup.org/HDF5/) is to an old / no longer supported website. 
I think this link should now go to here: https://portal.hdfgroup.org/

Assuming this update to the link is desired, this PR does so.

### DIFF
--- a/.authors.yml
+++ b/.authors.yml
@@ -890,6 +890,8 @@
   github: timgates42
 - name: Thomas Donoghue
   email: tdonoghue.research@gmail.com
+  alternate_emails:
+  - tdonoghue@ucsd.edu
   num_commits: 1
   first_commit: 2021-03-15 16:31:00
   github: TomDonoghue
--- a/.authors.yml
+++ b/.authors.yml
@@ -892,7 +892,7 @@
   email: tdonoghue.research@gmail.com
   alternate_emails:
   - tdonoghue@ucsd.edu
-aliases:
+  aliases:
   - Tom Donoghue
   - Tom
   num_commits: 1
--- a/.authors.yml
+++ b/.authors.yml
@@ -888,3 +888,8 @@
   num_commits: 1
   first_commit: 2020-12-29 19:51:32
   github: timgates42
+- name: Thomas Donoghue
+  email: tdonoghue.research@gmail.com
+  num_commits: 1
+  first_commit: 2021-03-15 16:31:00
+  github: TomDonoghue
--- a/.authors.yml
+++ b/.authors.yml
@@ -892,6 +892,9 @@
   email: tdonoghue.research@gmail.com
   alternate_emails:
   - tdonoghue@ucsd.edu
+aliases:
+  - Tom Donoghue
+  - Tom
   num_commits: 1
   first_commit: 2021-03-15 16:31:00
   github: TomDonoghue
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks! I'm not sure the support portal landing page is the best target for a link just saying 'HDF5', though. Maybe something like one of these:

- https://www.hdfgroup.org/solutions/hdf5/
- https://portal.hdfgroup.org/display/HDF5/HDF5
- https://portal.hdfgroup.org/display/knowledge/What+is+HDF5

takluyver:
Oh, and the CI wants you to add yourself to the `.authors.yml` file.

TomDonoghue:
Thanks @takluyver for checking in - I have added myself to the authors list to address the CI checks. 

In terms of the best link, I totally agree that a more specific page would work well. All the ones you listed make sense to me - do you recommend a particular option? I'm not sure if there's a better option (in terms of, for example, what's likely to be a more stable link / or more useful page, etc).

takluyver:
I don't have a strong preference. I think the reason people might follow that link is if they've stumbled on h5py without knowing what HDF5 is, so it's tempting to put the 'what is HDF5' page... but I don't think it actually does a good job of explaining what HDF5 is. It's so much 'HDF5 can do everything!' marketing that it doesn't really tell you anything. Maybe the [Introduction to HDF5](https://portal.hdfgroup.org/display/HDF5/Introduction+to+HDF5) is a better starting point?

(CI is now complaining again because you used a different email address for the second commit, so it wants that in the authors list too. Sorry about this.)

TomDonoghue:
I chose to update the link to the HDF5 support page, because I think this is generally the most useful, in that it:
- has a quick description of HDF5 at the top of the page
- has obvious & direct links to further description, examples, and references manuals

This seems to me to be more useful than the more "marketing" pages. It's also effectively the landing page for the Introduction page you linked, which is the start of the tutorial linked on the left sidebar. Does this seem like a sensible landing page?

Also, oops on my email (the Github client added used an unexpected email). I checked the log properly from a local command line, and added my alternate email. It looks like CI still doesn't like it though... maybe because now I (annoyingly) have commits under two different emails? Not sure what I should do to address that.

.authors.yml
```diff
@@ -888,3 +888,10 @@
   num_commits: 1
   first_commit: 2020-12-29 19:51:32
   github: timgates42
+- name: Thomas Donoghue
+  email: tdonoghue.research@gmail.com
+  alternate_emails:
+  - tdonoghue@ucsd.edu
```
takluyver, line 894:
```suggestion
  - tdonoghue@ucsd.edu
aliases:
  - Tom Donoghue
  - Tom
```

I think this is what it's asking for. I'll apply the suggestion now and see if it makes the CI happy.

.authors.yml
```diff
@@ -888,3 +888,13 @@
...
+aliases:
+  - Tom Donoghue
+  - Tom
```
takluyver, line 897:
```suggestion
  aliases:
  - Tom Donoghue
  - Tom
```

Could anything better represent modern development than breaking YAML syntax by editing it in a web UI? :disappointed:

takluyver:
Thanks @TomDonoghue !

PR #1863, opened 2021-03-16
Check when assembling virtual dataset if number of selected points match

### LINKED ISSUE #1776: Check selection sizes when mapping data into a VirtualLayout
@kirienko reminded me about this. HDF5's virtual dataset machinery checks that each mapping has the same number of points in the source data as it covers in the virtual dataset. But this check only happens when you call `f.create_virtual_dataset()`. We should pre-check it when you do `vlayout[...] = vsrc[...]`, so errors occur closer to the code that caused them, making them easier to debug.

I've added the 'good first issue' tag, but it's a bit borderline. I think the necessary pieces are already there so it won't take much code, but it needs a decent understanding of virtual datasets to make sense of what's required.

takluyver:
Closes #1776.

I had tagged that 'good first issue', but no-one seemed eager to work on it, and I wanted the more informative error messages.

### DIFF
--- a/h5py/_hl/vds.py
+++ b/h5py/_hl/vds.py
@@ -65,6 +65,17 @@ def _convert_space_for_key(space, key):
         space.select_hyperslab(start, count, stride, block)
 
 
+def _sel_is_finite(s: h5s.SpaceID):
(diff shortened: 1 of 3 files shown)


takluyver:
(We're discussing a different possible approach for this on #1865)

PR #1873, opened 2021-04-12
TST: Fix skipping unsupported x86 datatypes with newer versions of numpy

mgorny:
Apparently newer versions of numpy (1.19.5 is the oldest I have tested)
no longer declare 'float128' and 'complex256' on 32-bit x86.  Instead,
they declare 'float96' and 'complex192'.  Rewrite the test for
unsupported types to account for the two new types, and handle missing
types gracefully.

This fixes the following test failure:

    E       AttributeError: module 'numpy' has no attribute 'float128'

plus, later on:

    E   TypeError: Illegal length 24 for complex dtype

### DIFF
--- a/h5py/tests/test_dtype.py
+++ b/h5py/tests/test_dtype.py
@@ -14,6 +14,8 @@ except ImportError:
 from .common import ut, TestCase
 
 UNSUPPORTED_LONG_DOUBLE = ('i386', 'i486', 'i586', 'i686', 'ppc64le')
+UNSUPPORTED_LONG_DOUBLE_TYPES = ('float96', 'float128', 'complex192',
+                                 'complex256')
 
 
 class TestVlen(TestCase):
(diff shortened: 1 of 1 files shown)


mgorny:
Actually, let me try something else.

h5py/tests/test_dtype.py
```diff
@@ -289,13 +291,14 @@ def test_float_round_tripping(self):
...
-                              if d not in (np.float128, np.complex256)}
-        else:
-            dtype_dset_map = {str(j): d
-                              for j, d in enumerate(dtypes)}
+            for x in UNSUPPORTED_LONG_DOUBLE_TYPES:
+                if hasattr(np, x):
+                    unsupported_types.append(getattr(np, x))
+        dtype_dset_map = {str(j): d
+                          for j, d in enumerate(dtypes)
+                          if d not in unsupported_types}
```
aragilar, line 301: I'm not sure why we need the additional list `unsupported_types`, I'd expect `if d not in UNSUPPORTED_LONG_DOUBLE_TYPES` should be sufficient (we could also make them sets rather than tuples, but not sure that would actually cause a speed up given the small number of iterms)?

aragilar:
Assuming the tests pass, I think this is fine to merge. Ideally we should have a x86-32 run, but given the CI is already overloaded...

As an aside @mgorny, what are you using x86-32 for (e.g. older system, low-power cpu, reduced memory usage)?

h5py/tests/test_dtype.py
mgorny: `UNSUPPORTED_LONG_DOUBLE_TYPES` are strings while `d` is the actual type.

mgorny:
> Assuming the tests pass, I think this is fine to merge. Ideally we should have a x86-32 run, but given the CI is already overloaded...
> 
> As an aside @mgorny, what are you using x86-32 for (e.g. older system, low-power cpu, reduced memory usage)?

I don't. I'm maintaining Python packages in Gentoo, and we still support and test the i686 target. While every time someone reports a bug specific to i686 I feel like we should stop doing that, I do realize that the majority of these bugs would also affect (still) more meaningful 32-bit targets like ARM.

h5py/tests/test_dtype.py
aragilar: Ah, ok.

mgorny:
Thanks!

aragilar:
That makes sense. Thankfully, `float96` is an x86-32 only feature (as far as I know, ARM longdouble is always the standard IEEE 64bit float, though it would be nice if the more powerful chips started adding IEEE 128-bit floats), so if you run into more problems, feel free to add more xfails (I was wondering if there was a use-case that would suggest prioritising x86-32). And thanks for the PR!

PR #1235, opened 2019-06-14
TST: Skip failing long double test on i386 for 2.10 release

aragilar:
This kinda fixes #1163, in that the test won't fail (we just skip it), and we can look at properly handing complex floats post 2.10.

I've tested this inside an inside a i386 container on an x86_64 system, so it probably should be fine.

PR #1874, opened 2021-04-12
TST: Skip problematic datatypes based on their length

mgorny:
Filter datatypes for tests based on their length rather than keeping
a list of illegal platform+type combinations.  This is more robust
than the previous solution and fixes the test failures with new numpy
versions (1.19.5 is the oldest I have tested) that declare 'float96'
and 'complex192' on i686 rather than 'float128' and 'complex256'.

This fixes the following test failure:

    E       AttributeError: module 'numpy' has no attribute 'float128'

plus, later on (again):

    E   TypeError: Illegal length 24 for complex dtype

### DIFF
--- a/h5py/tests/test_dtype.py
+++ b/h5py/tests/test_dtype.py
@@ -13,8 +13,6 @@ except ImportError:
 
 from .common import ut, TestCase
 
-UNSUPPORTED_LONG_DOUBLE = ('i386', 'i486', 'i586', 'i686', 'ppc64le')
-
 
 class TestVlen(TestCase):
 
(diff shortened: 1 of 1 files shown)


mgorny:
Ok, I see that the ppc64le is something different. Let's go with the earlier PR then.

aragilar:
This is failing on ppc—the skipped arches mix two different issues, one is that the Intel's 80 bit float on x86-32 isn't properly supported yet (https://github.com/h5py/h5py/pull/1235 added that skip), and there's a separate issue for ppc, which uses the double-double format (https://github.com/h5py/h5py/pull/1246). numpy on x86-32 has always had float96 if the 80-bit float is available, so I'm not sure what has changed there.

mgorny:
> This is failing on ppc—the skipped arches mix two different issues, one is that the Intel's 80 bit float on x86-32 isn't properly supported yet (#1235 added that skip), and there's a separate issue for ppc, which uses the double-double format (#1246). numpy on x86-32 has always had float96 if the 80-bit float is available, so I'm not sure what has changed there.

Well, when I tried running tests, it turned out that numpy on my i686 system doesn't have `float128` or `complex256`. When I updated the test to account for that, it turned out that `float96` and `complex192` types cause the same issue as addressed earlier.

mgorny:
PR #1873 should fix this without breaking ppc.

PR #1875, opened 2021-04-13
Fix h5py exceptions and new libver "v114"

ajelenak:
In this PR:

1. Fixes for translation of HDF5 error codes to Python exceptions for next HDF5-1.12.1 and HDF5-1.13 (develop). Includes the PR #1858 issue.
1. Support for the future HDF5 format version `H5F_LIBVER_V114` that is in the HDF5-develop.


takluyver:
Thanks! Do you know how close 1.14 is? Or how definite these APIs are? I think it makes sense for h5py to adapt to changes before an HDF5 final release, but I'd also rather not chase after every development change.

This also relates to the value of having a regular automated test of h5py with the development branch of HDF5 (#1604).

h5py/_errors.pyx
```diff
@@ -10,76 +10,145 @@
...
-    (H5E_SYM, H5E_CANTINIT):        ValueError, # Object already exists/1.8
-    (H5E_ARGS, H5E_BADTYPE):        ValueError, # Invalid location in file
-    (H5E_REFERENCE, H5E_CANTINIT):  ValueError, # Dereferencing invalid ref
-
-    # needed for 1.10.3+ to maintain compatibility with 1.10.{0,1,2}
-
-    # due to changes to H5F.c:H5Fstart_swmr_write
-    (H5E_FILE, H5E_CANTCONVERT):    ValueError, # Invalid file format
-  }
+IF HDF5_VERSION < (1, 13, 0):
```
takluyver, line 18: Is it possible to have these `IF` within the long lists, so we don't need to duplicate all the entries that aren't changing? I *think* Cython capitalised `IF` work like C preprocess `#ifdef`, so you can put them in places that wouldn't be valid for a Python `if` statement.

takluyver: (Where there's only a few things in the list, I think it's probably clearer to duplicate them, but for something this long it's hard to see what's different between the `IF` & `ELSE` branches)

epourmal:
> Thanks! Do you know how close 1.14 is? Or how definite these APIs are? I think it makes sense for h5py to adapt to changes before an HDF5 final release, but I'd also rather not chase after every development change.

@takluyver Release date for 1.14.0 is not established yet. We will be doing some work on the HDF5 develop branch before the release, finalizing VOL APIs and addressing performance regression that SWMR and VOL features introduced. Since we are running now daily tests against h5py master we would like h5py to have changes that will allow us to build and test h5py with the HDF5 under development. Said this, we will not be submitting any change to h5py :-) unless we know it is triggered by a change in HDF5 that will be present in 1.14.0 and later. 

Our goal is to have h5py version that will work with the HDF5 VOL connectors for example, [Async VOL)](https://github.com/hpc-io/vol-async).

Elena

h5py/_errors.pyx
ajelenak: I wanted to keep the content in sync with `_errors.pxd` and was easier to follow. One block for HDF5-1.13.0 and later, one for before, in both `_errors` files.

aragilar: I'd also prefer having the IF/ELSE only switch what was needed (in both files), and a comment about what changed (e.g. `H5E_ATOM` -> `H5E_ID`). That makes it easy to know what someone needs to search for in the rest of the code (when adding/removing support). The reference for the conditional compilation is https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#compile-time-definitions (I'm not sure through if those definitions are evaluated at the cython -> C transformation time, or if they actually get converted to C preprocessor macros, I think the former, but for h5py currently there's no distinction between the phases).

ajelenak:
I reorganized the changes in the `.pyx` file and tried to reduce the "noise" of  `IF...ELSE...` blocks. I could not find a way for Cython to accept `IF...ELSE...` inside a `ctypedef` so the `.pxd` file is unchanged.

ajelenak:
@takluyver @aragilar HDF5-1.12.1 will be released soon. Are the latest changes acceptable?

takluyver (commented):
It's still a bit annoying that we have to duplicate the whole list of error types just to replace one, but I agree that there doesn't seem to be any easy way round that in Cython.

h5py/_errors.pyx
```diff
@@ -59,27 +61,34 @@ _minor_table = {
...
+}
+
+IF HDF5_VERSION > (1, 12, 0):
+    _exact_table[(H5E_DATASET, H5E_CANTCREATE)] = ValueError  # bad param for dataset setup
+
+IF HDF5_VERSION < (1, 13, 0):
+    _minor_table["H5E_BADATOM"] = ValueError  # Unable to find atom information (already closed?)
+    _exact_table[(H5E_SYM, H5E_CANTINIT)] = ValueError  # Object already exists/1.8
+ELSE:
+    _minor_table["H5E_BADID"] = ValueError  # Unable to find ID information
```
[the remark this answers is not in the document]
ajelenak: Fixed.

h5py/h5p.pyx
```diff
@@ -1196,11 +1196,12 @@ cdef class PropFAID(PropInstanceID):
     def set_libver_bounds(self, int low, int high):
         """ (INT low, INT high)
 
-        Set the compatibility level for file format.  Legal values are:
+        Set the compatibility level for file format. Legal values are:
 
         - h5f.LIBVER_EARLIEST
         - h5f.LIBVER_V18 (HDF5 1.10.2 or later)
         - h5f.LIBVER_V110 (HDF5 1.10.2 or later)
+        - h5f.LIBVER_V112 (HDF5 1.11.4 or later)
```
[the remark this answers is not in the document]
ajelenak: It should, and done.

takluyver (approved):
Thanks @ajelenak . I'll give other maintainers a day to have a look, otherwise I'll merge this tomorrow.

aragilar (approved):

aragilar:
My only issue was the duplication, and if that's a cython issue, that's painful but I guess we need to accept that.

PR #1889, opened 2021-05-11
Add default cast from H5T_NATIVE_B8 to uint8

### LINKED ISSUE #1258: Default cast type of H5T_NATIVE_B8 bitfields (bool, int8, uint8)
As briefly discussed in #821, reading datasets which contain bitfields (H5T_NATIVE_B8) is not possible unless you explicitly state the dtype:

```
import h5py
h5_file = h5py.File('BigDataFile.h5', 'r')

dset = h5_file.get('/Base/GroupA')
dt = [('Time', '<f8'),
        ('SubsetA', [('DataA1', '?'), ('DataA2', '<f8')])]
with dset.astype(dt):
    data = dset[:]
```
whereas datasets without bitfields can be accessed as:

```
import h5py
h5_file = h5py.File('BigDataFile.h5', 'r')
dset = h5_file.get('/Base/GroupA')
data = dset[:]
```
Attempting to read a compound dataset in this way produces an error if it has at least one bitfield:
`TypeError: No NumPy equivalent for TypeBitfieldID exists`

Some of the suggestions from #821 were to automatically convert to bool, uint8 or int8. Imho I think it should convert to something to allow reading without explicitly stating the datatype, enabling most users to use the second code snippet.

If there was going to be a default, uint8 or int8 would achieve this and also not have any data loss compared to bool, although the original motivation for using bool in the pull request was PyTables interop. I'd suggest having uint8 - then it would work out of the box without requiring explicit casts, not have any data loss, and users who want it as something else can still do an explicit cast.

All comments are welcome :) I may be missing some understanding of h5py, so feel free to correct if there's anything wrong!

samuel-emrys:
This adds a default cast from H5T_NATIVE_B8 to uint8. The motivation for doing this to rectify an issue where reading mixed-type compound datasets containing a bitfield as follows:

```
import h5py
h5_file = h5py.File('BigDataFile.h5', 'r')
dset = h5_file.get('/Base/GroupA')
data = dset[:]
```

would result in an error message `TypeError: No NumPy equivalent for TypeBitfieldID exists`. To rectify this, explicit specification of the `dtype` was required:

```
import h5py
h5_file = h5py.File('BigDataFile.h5', 'r')

dset = h5_file.get('/Base/GroupA')
dt = [('Time', '<f8'), ('SubsetA', [('DataA1', '?'), ('DataA2', '<f8')])]
with dset.astype(dt):
    data = dset[:]
```

Knowledge and pre-specification of the schema may not always be possible, especially in large datasets. The change that this pull request introduces is the ability to import mixed-type compound datasets without requiring the user to know the schema of that dataset using the following method:

```
import h5py
h5_file = h5py.File('BigDataFile.h5', 'r')
dset = h5_file.get('/Base/GroupA')
data = dset[:]
```

Since the default cast is to `numpy.uint8`, this retains all of the original bit information necessary to perform an explicit cast to another type if required, for example `numpy.bool`.

Comments are welcome, this is my first time working with the h5py source code.

Relates to #821, Relates to #641, Closes #1258

### DIFF
--- a/h5py/_conv.pyx
+++ b/h5py/_conv.pyx
@@ -972,7 +972,8 @@ cpdef int unregister_converters() except -1:
     H5Tunregister(H5T_PERS_HARD, "boolenum2b8", -1, -1, boolenum2b8)
     H5Tunregister(H5T_PERS_HARD, "b82boolenum", -1, -1, b82boolenum)
 
-    H5Tunregister(H5T_PERS_HARD, "uint2bitfield", -1, -1, uint2bitfield)
-    H5Tunregister(H5T_PERS_HARD, "bitfield2uint", -1, -1, bitfield2uint)
+    # Pass an empty string to unregister all methods that use these functions
+    H5Tunregister(H5T_PERS_HARD, "", -1, -1, uint2bitfield)
+    H5Tunregister(H5T_PERS_HARD, "", -1, -1, bitfield2uint)
 
     return 0
--- a/h5py/_conv.pyx
+++ b/h5py/_conv.pyx
@@ -860,14 +860,14 @@ cdef herr_t boolenum2b8(hid_t src_id, hid_t dst_id, H5T_cdata_t *cdata,
     return 0
 
 # =============================================================================
-# B8 to UINT8 routines
+# BITFIELD to UINT routines
 
-cdef herr_t b82uint8(hid_t src_id, hid_t dst_id, H5T_cdata_t *cdata,
+cdef herr_t bitfield2uint(hid_t src_id, hid_t dst_id, H5T_cdata_t *cdata,
                      size_t nl, size_t buf_stride, size_t bkg_stride, void *buf_i,
                      void *bkg_i, hid_t dxpl) except -1:
     return 0
 
-cdef herr_t uint82b8(hid_t src_id, hid_t dst_id, H5T_cdata_t *cdata,
+cdef herr_t uint2bitfield(hid_t src_id, hid_t dst_id, H5T_cdata_t *cdata,
                      size_t nl, size_t buf_stride, size_t bkg_stride, void *buf_i,
                      void *bkg_i, hid_t dxpl) except -1:
     return 0
@@ -915,8 +915,29 @@ cpdef int register_converters() except -1:
     H5Tregister(H5T_PERS_HARD, "boolenum2b8", boolenum, H5T_NATIVE_B8, boolenum2b8)
     H5Tregister(H5T_PERS_HARD, "b82boolenum", H5T_NATIVE_B8, boolenum, b82boolenum)
 
-    H5Tregister(H5T_PERS_HARD, "uint82b8", H5T_NATIVE_UINT8, H5T_NATIVE_B8, uint82b8)
-    H5Tregister(H5T_PERS_HARD, "b82uint8", H5T_NATIVE_B8, H5T_NATIVE_UINT8, b82uint8)
+    H5Tregister(H5T_PERS_HARD, "uint82b8", H5T_STD_U8BE, H5T_STD_B8BE, uint2bitfield)
+    H5Tregister(H5T_PERS_HARD, "b82uint8", H5T_STD_B8BE, H5T_STD_U8BE, bitfield2uint)
+
+    H5Tregister(H5T_PERS_HARD, "uint82b8", H5T_STD_U8LE, H5T_STD_B8LE, uint2bitfield)
+    H5Tregister(H5T_PERS_HARD, "b82uint8", H5T_STD_B8LE, H5T_STD_U8LE, bitfield2uint)
+
+    H5Tregister(H5T_PERS_HARD, "uint162b16", H5T_STD_U16BE, H5T_STD_B16BE, uint2bitfield)
+    H5Tregister(H5T_PERS_HARD, "b162uint16", H5T_STD_B16BE, H5T_STD_U16BE, bitfield2uint)
+
+    H5Tregister(H5T_PERS_HARD, "uint162b16", H5T_STD_U16LE, H5T_STD_B16LE, uint2bitfield)
+    H5Tregister(H5T_PERS_HARD, "b162uint16", H5T_STD_B16LE, H5T_STD_U16LE, bitfield2uint)
+
+    H5Tregister(H5T_PERS_HARD, "uint322b32", H5T_STD_U32BE, H5T_STD_B32BE, uint2bitfield)
+    H5Tregister(H5T_PERS_HARD, "b322uint32", H5T_STD_B32BE, H5T_STD_U32BE, bitfield2uint)
+
+    H5Tregister(H5T_PERS_HARD, "uint322b32", H5T_STD_U32LE, H5T_STD_B32LE, uint2bitfield)
+    H5Tregister(H5T_PERS_HARD, "b322uint32", H5T_STD_B32LE, H5T_STD_U32LE, bitfield2uint)
+
+    H5Tregister(H5T_PERS_HARD, "uint642b64", H5T_STD_U64BE, H5T_STD_B64BE, uint2bitfield)
+    H5Tregister(H5T_PERS_HARD, "b642uint64", H5T_STD_B64BE, H5T_STD_U64BE, bitfield2uint)
+
+    H5Tregister(H5T_PERS_HARD, "uint642b64", H5T_STD_U64LE, H5T_STD_B64LE, uint2bitfield)
+    H5Tregister(H5T_PERS_HARD, "b642uint64", H5T_STD_B64LE, H5T_STD_U64LE, bitfield2uint)
 
     H5Tregister(H5T_PERS_SOFT, "vlen2str", vlstring, pyobj, vlen2str)
     H5Tregister(H5T_PERS_SOFT, "str2vlen", pyobj, vlstring, str2vlen)
@@ -951,7 +972,7 @@ cpdef int unregister_converters() except -1:
     H5Tunregister(H5T_PERS_HARD, "boolenum2b8", -1, -1, boolenum2b8)
     H5Tunregister(H5T_PERS_HARD, "b82boolenum", -1, -1, b82boolenum)
 
-    H5Tunregister(H5T_PERS_HARD, "uint82b8", -1, -1, uint82b8)
-    H5Tunregister(H5T_PERS_HARD, "b82uint8", -1, -1, b82uint8)
+    H5Tunregister(H5T_PERS_SOFT, "uint2bitfield", -1, -1, uint2bitfield)
+    H5Tunregister(H5T_PERS_SOFT, "bitfield2uint", -1, -1, bitfield2uint)
 
     return 0
--- a/h5py/_conv.pyx
+++ b/h5py/_conv.pyx
@@ -972,7 +972,7 @@ cpdef int unregister_converters() except -1:
     H5Tunregister(H5T_PERS_HARD, "boolenum2b8", -1, -1, boolenum2b8)
     H5Tunregister(H5T_PERS_HARD, "b82boolenum", -1, -1, b82boolenum)
 
-    H5Tunregister(H5T_PERS_SOFT, "uint2bitfield", -1, -1, uint2bitfield)
-    H5Tunregister(H5T_PERS_SOFT, "bitfield2uint", -1, -1, bitfield2uint)
+    H5Tunregister(H5T_PERS_HARD, "uint2bitfield", -1, -1, uint2bitfield)
+    H5Tunregister(H5T_PERS_HARD, "bitfield2uint", -1, -1, bitfield2uint)
 
     return 0
--- a/h5py/h5t.pyx
+++ b/h5py/h5t.pyx
@@ -159,6 +159,17 @@ STD_I16BE = lockid(H5T_STD_I16BE)
 STD_I32BE = lockid(H5T_STD_I32BE)
 STD_I64BE = lockid(H5T_STD_I64BE)
 
+# Bitfields
+STD_B8LE = lockid(H5T_STD_B8LE)
+STD_B16LE = lockid(H5T_STD_B16LE)
+STD_B32LE = lockid(H5T_STD_B32LE)
+STD_B64LE = lockid(H5T_STD_B64LE)
+
+STD_B8BE = lockid(H5T_STD_B8BE)
+STD_B16BE = lockid(H5T_STD_B16BE)
+STD_B32BE = lockid(H5T_STD_B32BE)
+STD_B64BE = lockid(H5T_STD_B64BE)
+
 # Unsigned integers
 STD_U8LE  = lockid(H5T_STD_U8LE)
 STD_U16LE = lockid(H5T_STD_U16LE)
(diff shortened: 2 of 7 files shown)


aragilar:
Quick comment as I haven't looked into this in detail: there was discussion around which style should be the default in some earlier issues and PRs, I'm not sure we came to any consensus?

samuel-emrys:
@aragilar you're right - no consensus was reached. It looked to me as though the discussion was leaning towards `uint8`, but the discussion seemed to have become stale. This is a blocking issue for me so I thought I'd be proactive and implement it that way, at least as a way of re-igniting a conversation about an implementation.

From my perspective, even though it may not be possible to retain injectivity due to the mismatch between types in HDF5 and numpy, a default cast to `uint8` retains all of the information in the original HDF type, allowing an explicit cast to the desired type if required, and also provides the ability to read mixed-type compound datasets (especially where the schema may not be known in advance/schemas are generated dynamically and in large quantity) - functionality not currently supported. Other solutions that seemed viable might be:

* Introducing new types upstream to numpy to address differences b/w numpy and HDF
* Context manager that allows casting of all of a specified type to another type
* Other?

I think these solutions are more about larger scope improvements to h5py generally though, so are probably more appropriate as their own issues for discussion if worth considering. To address #1258 specifically, I think that the default cast to uint8 is a good first step to bridge the gap in functionality here.

takluyver:
I think converting bitfields to uint by default makes sense. But if we're going to do it for one, should we do it for all of the standard bitfield types (8, 16, 32, 64 bit, big & little endian)? It would seem a bit odd to treat B8 differently from the other sizes.

takluyver:
We could also tag the dtype in `dt.metadata`, like we do with string dtypes, so data read from an HDF5 bitfield dataset can conveniently be used to create a new bitfield dataset. But I'm inclined not to do that: dtype metadata is not super obvious, so it's going to be confusing if h5py does something different with arrays x and y which appear to have the same dtype. Explicit is better than implicit.

samuel-emrys:
@takluyver I don't think there's an inconsistency unless one of the other bitfields is dealt with in a different way, which is distinct from not supporting those data types. Having said that, I have no objection to adding conversions for those types as well. To clarify what you're after, are you talking about converting `H5T_NATIVE_B16`, `H5T_NATIVE_B32`, and `H5T_NATIVE_B64` or the `STD` equivalents below:
* `H5T_STD_B8BE`
* `H5T_STD_B8LE`
* `H5T_STD_B16BE `
* `H5T_STD_B16LE`
* `H5T_STD_B32BE`
* `H5T_STD_B32LE`
* `H5T_STD_B64BE`
* `H5T_STD_B64LE`

If you mean the `STD` types - what behaviour would you expect when reading these? on a little endian system, attempting to read the value `00000010` value (represented as LE), which is stored in a `H5T_STD_B8BE` type as `01000000`, would you expect this to be stored in a `uint8` object as `0100000` (decimal 128) or `00000010` (decimal 2)? If it's just required to read these as represented, my understanding is that this is what you would get by reading `H5T_NATIVE_B8` anyway.

Additionally, I'm not _that_ familiar with h5py (I'm making these changes so I can start using it), but I've had a look through the low level API to see if there's a way that I can manually create a dataset with one of these types (i.e. `H5T_BITFIELD_B16`) so that I can confirm that the conversion is working. The method presently employed uses the `tables` library which writes `bool`s this way - this appears to be a one-off, so I'm not sure it's scalable to the larger bitfields. Is anybody able to provide some guidance on the best way to achieve this?

takluyver:
Short answer: I'd suggest all the standard bitfield types are converted in the same way as the unsigned integer types.

NumPy, like HDF5, is capable of holding data in either endianness. E.g. `np.dtype('>u4')` makes a big-endian uint32 (4 bytes) dtype. So we should be able to read all of the standard HDF5 data types with no actual conversion needed (it probably still needs a conversion *function* to satisfy HDF5, but that can be a no-op). I think it's also possible to register a conversion function for a whole family of types, so we don't have to register it 8 times over, though that might require a bit more fiddling in `_conv.pyx`.

I think the HDF5 NATIVE types are just aliases for whichever STD types match the platform's own endianness, so handling all of the STD types should make the NATIVE types work as well. But check that, because I might be wrong.

samuel-emrys:
@takluyver I've updated the PR to include conversion paths for all of the `STD` uint -> bitfield and bitfield -> uint conversion paths in a manner consistent with the way that integer associations are currently conducted.

I did try to register a conversion function for the class of types, without success. The documentation for [`H5Tregister`](https://support.hdfgroup.org/HDF5/doc/RM/RM_H5T.html#Datatype-Register) suggests:

> The path is specified by the source and destination datatypes src_id and dst_id. For soft conversion functions, only the class of these types is important. 

I interpreted that this might mean `H5Tregister` takes in the class type:

```cpp
H5Tregister(H5T_PERS_SOFT, "uint2bitfield", H5T_INTEGER, H5T_BITFIELD, uint2bitfield)
H5Tregister(H5T_PERS_SOFT, "bitfield2uint", H5T_BITFIELD, H5T_INTEGER, bitfield2uint)
```
This leads to runtime exceptions as `H5T_INTEGER` and `H5T_BITFIELD` are not recognised as datatypes (probably obviously, since they're classes). My only remaining theory is that in order to register a conversion function for a family of types is to register a `H5T_PERS_SOFT` conversion for the `NATIVE` datatypes, i.e.:

* `H5T_NATIVE_UINT` -> `H5T_NATIVE_B8`, `H5T_NATIVE_B16`, `H5T_NATIVE_B32`, `H5T_NATIVE_B64`

For two way definitions this leads to 8 (4*2) registration calls as opposed to the current 16 (4*2*2), which includes explicit conversions for each datatype size and for each endianness. Without a way to write these types (since we have no injectivity, we can only read - writing `np.uint8` will write `H5T_NATIVE_UINT8` not a `H5T_NATIVE_B8`) it's difficult to test that having only a soft conversion pathway would satisfactorily define the conversion pathways for differing endianness. I.e., will reading `H5T_STD_B8BE` if `H5T_STD_B8BE != H5T_NATIVE_B8` identify the correct conversion pathway?

To address this, I thought that being explicit would be better, and so I've defined `H5T_PERS_HARD` conversion pathways for all of the `STD` uint and bitfield types. Having said this, I still have the issue I raised in my last comment of not having a good way to write anything other than a `H5T_NATIVE_B8` to test this with.

Feedback is appreciated if there's a better way of going about this.

samuel-emrys:
@takluyver any further blockers on getting this merged in?

takluyver:
Sorry, it had just slipped out of my mind.

I think the "only the class of these types is important" in the HDF5 docs means that you still have to pass a datatype ID like `H5T_STD_B8BE`, but it will only look at the class of that, i.e. `H5T_BITFIELD`.

If you register a 'soft' conversion function between type families (e.g. bitfield -> uint), HDF5 will call it with an 'init' command with the details of the specific datatypes it wants to convert (e.g. bitfield 32 BE -> uint 32 BE), and at that point the function can indicate that it can't do that specific conversion. This is a bit of extra complexity inside the conversion function, but it should avoid having to register the same thing 8 times.

However, a few extra almost duplicated lines aren't a big deal, if you prefer to stick with registering 'hard' functions.

h5py/_conv.pyx
```diff
@@ -951,7 +972,7 @@ cpdef int unregister_converters() except -1:
     H5Tunregister(H5T_PERS_HARD, "boolenum2b8", -1, -1, boolenum2b8)
     H5Tunregister(H5T_PERS_HARD, "b82boolenum", -1, -1, b82boolenum)
 
-    H5Tunregister(H5T_PERS_HARD, "uint82b8", -1, -1, uint82b8)
-    H5Tunregister(H5T_PERS_HARD, "b82uint8", -1, -1, b82uint8)
+    H5Tunregister(H5T_PERS_SOFT, "uint2bitfield", -1, -1, uint2bitfield)
+    H5Tunregister(H5T_PERS_SOFT, "bitfield2uint", -1, -1, bitfield2uint)
```
takluyver, line 976: However, I think that the unregister calls should match the register ones, whether we go for hard or soft conversion functions.

news/1258-default-cast-b8.rst
```diff
@@ -0,0 +1,42 @@
+New features
+------------
+
+* H5T_NATIVE_B8 types will now be cast to ``numpy.uint8`` by default
```
takluyver, line 4: & don't forget to update the news entry now that it affects all bitfield types. :slightly_smiling_face:

samuel-emrys:
@takluyver I've implemented the suggested changes. Looks like azure pipelines had a failure installing hdf5 so the tests are failing. Couldn't see a way to re-run the failed tests, but you can see [here](https://travis-ci.com/github/samuel-emrys/h5py/builds/228059081) that they're passing in Travis CI.

h5py/_conv.pyx
```diff
@@ -951,7 +972,7 @@ cpdef int unregister_converters() except -1:
     H5Tunregister(H5T_PERS_HARD, "boolenum2b8", -1, -1, boolenum2b8)
     H5Tunregister(H5T_PERS_HARD, "b82boolenum", -1, -1, b82boolenum)
 
-    H5Tunregister(H5T_PERS_HARD, "uint82b8", -1, -1, uint82b8)
-    H5Tunregister(H5T_PERS_HARD, "b82uint8", -1, -1, b82uint8)
+    H5Tunregister(H5T_PERS_HARD, "uint2bitfield", -1, -1, uint2bitfield)
+    H5Tunregister(H5T_PERS_HARD, "bitfield2uint", -1, -1, bitfield2uint)
```
takluyver, line 976: Looking at the HDF5 source code, I think the name (the string here) needs to either match, or be a NULL pointer or an empty string to unregister all matches with the function:

https://github.com/HDFGroup/hdf5/blob/ce5f1ff1c502a1f341560254e4ed3b0dc65cfade/src/H5T.c#L2863-L2866

samuel-emrys: My read is that it probably needs to be an uninitialized string - an empty string will still evaluate to boolean `true`. This assumes that passing `-1` for both `src` and `dst` will effectively leave the `H5T_t` object pointers uninitialized. If they _are_ initialized, they will evaluate to `true` and the function won't be the matching criteria - I've tried to investigate this but I'm finding the code pretty difficult to interpret.

This is the first time I've written any cython, so would just like to run this by you before I push anything:

```python
cdef char* empty_str
H5T_unregister(H5T_PERS_HARD, empty_str, -1, -1, uint2bitfield)
```
Does this look right to you?

samuel-emrys: Is there a way to test that the way we're calling `H5Tunregister` is actually unregistering all of the conversion functions we intend to? If not, and I can't see any existing unit tests, perhaps it would be better to be explicit and just unregister the conversion functions for every name we have to be sure, since the name quality comparison precedes the `src` and `dst` initialisation. What are your thoughts @takluyver?

aragilar: Not as far as I know, last time I looked at adding tests around the type code, the information I could get from HDF5 was very limited.

takluyver: I think an empty string literal - `""` - should work, because it should be terminated by a null byte, and that's what the C code will look at (there are no strings in C, only pointers to a char, which may have some other chars after it). Or passing `NULL` should work as well (in the C code `name && *name && ...`, a null pointer should be false for `name`, and an empty string, a pointer to a null byte, should be false for `*name`).

Unfortunately, no, I don't think there's a great way to test these. The `unregister_converters` functions is only called when Python is exiting, as part of cleanup that's only sometimes important.

samuel-emrys: Yep, you're right. I was only checking `name` when setting `const char* name = ""`. `*name` does indeed return false. Alright, I'll opt for an empty string literal then :)

h5py/h5t.pyx
```diff
@@ -764,7 +766,39 @@ cdef class TypeBitfieldID(TypeID):
...
+
+        SGN_NONE
+            Unsigned
+
+        SGN_2
+            Signed 2's complement
+        """
+
+        # This will always return SGN_NONE to cast bitfield types to unsigned integer
+        return SGN_NONE
```
takluyver, line 783: I don't think we should have a method for this (the relevant HDF5 method [is documented as](https://confluence.hdfgroup.org/display/HDF5/H5T_GET_SIGN) "for an integer type"). We can just hardcode the `'u'` for unsigned in `py_dtype`.

h5py/tests/test_dtype.py
```diff
@@ -394,37 +394,77 @@ class TestB8(TestCase):
...
+            expected_default_cast_dtype=np.dtype([('x', ('u1', (3,)))])
+        )
+        self._test_b8(
+            arr1,
+            expected_default_cast_dtype=np.dtype([('x', ('u1', (3,)))]),
+            cast_dtype=np.dtype([('x', ('?', (3,)))]),
         )
 
-    def _test_b8(self, arr1, dtype=None):
+    def _test_b8(self, arr1, expected_default_cast_dtype, cast_dtype=None):
```
takluyver, line 467: While we're at it, could we have one test of a longer bitfield type (i.e. anything beyond 8-bit) as well?

news/1258-default-cast-b8.rst
```diff
@@ -0,0 +1,42 @@
...
+  .. code::
+
+     arr = dset[:]
+
+  Alternatively, these types can still be cast to ``numpy.bool`` explicitly:
+
+  .. code::
+
+     with dset.astype(numpy.bool):
+        arr = dset[:]
```
takluyver, line 17:
```suggestion
  Alternatively, 8-bit bitfields can still be cast to booleans explicitly:

  .. code::

     arr = dset.astype(numpy.bool_)[:]:
```

`numpy.bool` recently started giving a deprecation warning. And I'd eventually like to shuffle people away from using `astype()` as a context manager now that it can be used as a wrapper.

takluyver (commented):
Thanks! I think this is nearly there, just a few more things I spotted doing another review pass.

h5py/tests/test_dtype.py
samuel-emrys: This is something I've struggled to work out how to do - since we don't have reflection for bitfield types, my understanding is that we're only able to read them, not write them, since the matching `numpy.uint` equivalents will be written as `H5T_NATIVE_UINT` types. The way we circumvent this for the 8 bit case is to use `tables`, which writes it's bools as `H5T_NATIVE_B8`. I did some digging and couldn't find any information about other types that write as larger bitfields - are you able to provide any guidance on how we might be able to test the larger bitfield types?

takluyver: Aha, good point. You've added conversions both ways, so writing uint data into a bitfield dataset shouldn't be a problem, but creating it in the first place is a bit trickier. I think at present, we'd need to use the [low-level API](https://api.h5py.org/). Something like this (untested):

```python
f = h5py.File('foo.h5', 'w')

space = h5py.h5s.create_simple((10,))
dset_id = h5py.h5d.create(f.id, b'data', h5py.h5t.STD_B32LE, space)
dset = h5py.Dataset(dset_id)
```

(We should probably add the STD & NATIVE bitfield types to `h5t.pyx`, in the same way that the equivalent integer types are there. I've assumed that in the example)

h5py/h5t.pyx
samuel-emrys: Yep okay. I do return a constant `SGN_NONE` so that it will always be `'u'`- I thought this might be more "consistent" with the rest of the integer implementation, but I can see how it might just be confusing. I'll change it to just be `'u'`.

aragilar:
One other thing to do would to update https://docs.h5py.org/en/stable/faq.html for the new bitfields.

h5py/tests/test_dtype.py
```diff
@@ -492,6 +493,23 @@ def _test_b8(self, arr1, expected_default_cast_dtype, cast_dtype=None):
...
+            dset = h5py.Dataset(dset_id)
+            dset[:] = arr1
+
+        with h5py.File(path, 'r') as f:
+            dset = f['test']
+            arr2 = dset[:]
+            self.assertArrayEqual(
+                arr2,
+                arr1.astype(np.dtype(np.uint16), copy=False)
+            )
```
takluyver, line 511:
```suggestion
            self.assertArrayEqual(dset[:], arr1)
```

arr1 already has dtype uint16, so no need to convert it - unless I'm missing something?

takluyver:
Other than that, I think this is ready to merge. :-)

h5py/tests/test_dtype.py
```diff
@@ -492,6 +493,23 @@ def _test_b8(self, arr1, expected_default_cast_dtype, cast_dtype=None):
                 arr3 = dset[:]
             self.assertArrayEqual(arr3, arr1.astype(cast_dtype, copy=False))
 
+    def test_b16_uint16(self):
+        arr1 = np.array(np.arange(10), dtype=np.dtype(np.uint16))
```
takluyver, line 497:
```suggestion
        arr1 = np.arange(10, dtype=np.uint16)
```

takluyver:
Given the time where you are, and the minor changes I'm suggesting, I'm going to commit them, and merge if the CI still passes.

PR #1891, opened 2021-05-17
Implemented H5DOappend

bomber8013:
Implemented H5DOappend by adding the C function to the API list and adding append function to h5d.pyx. Implemented and passed tests and added contribution note.

### DIFF
--- a/h5py/tests/test_dataset.py
+++ b/h5py/tests/test_dataset.py
@@ -1020,6 +1020,27 @@ class TestLen(BaseDataset):
         self.assertEqual(dset.len(), 2 ** 33)
 
 
+class TestAppend(BaseDataset):
+    """
+        Feature: H5DOappend testing
+    """
+    def test_append_1(self):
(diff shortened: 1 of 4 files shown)


h5py/tests/test_dataset.py
```diff
@@ -1020,6 +1020,35 @@ def test_len_big(self):
         self.assertEqual(dset.len(), 2 ** 33)
 
 
+class TestAppend(BaseDataset):
+    """
+        Feature: H5DOappend testing
+    """
+    @ut.skipIf(
+        h5py.version.hdf5_version_tuple < (1, 10, 0),
+        "Reading non-existent label segfaults"
```
tacaswell, line 1029: Is this comment correct?

bomber8013:
Wow, apparently I'm pretty bad at following up. This was because H5DOappend was added in 1.10.0 of HDF5. I might need to change the test of why it's being skipped? Also I thought this feature was undesirable since we didn't want to add the optimized functions/library as a required component for h5py?

PR #1899, opened 2021-06-06
Fix undefined behavior in test-dtype.py

rainwoodman:
np.empty and np.arange can fill the memory with undefined bytes, which
can trigger undefined behavior errors under msan.

### DIFF
--- a/h5py/tests/test_dtype.py
+++ b/h5py/tests/test_dtype.py
@@ -195,7 +195,7 @@ class TestOffsets(TestCase):
 
     def test_aligned_data(self):
         dt = np.dtype('i4,f8,i2', align=True)
-        data = np.empty(10, dtype=dt)
+        data = np.zeros(10, dtype=dt)
 
         data['f0'] = np.array(np.random.randint(-100, 100, size=data.size),
                               dtype='i4')
@@ -234,7 +234,7 @@ class TestOffsets(TestCase):
         })
 
         self.assertTrue(dt.itemsize == itemsize)
-        data = np.empty(10, dtype=dt)
+        data = np.zeros(10, dtype=dt)
 
         # don't trust numpy struct handling, keep fields out of band in case content insertion is erroneous
         # yes... this has also been known to happen.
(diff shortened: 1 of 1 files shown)


aragilar:
I'm confused by msan complaining about `np.arange` (as either that's a msan bug or possible numpy issue).

For the `np.empty` case, if we are trying to read undefined bytes (rather than write to them), where's that in the code (as that likely means there's a bug in the compound dtype handling code)?

takluyver:
I agree it doesn't seem like `arange` should be a problem.

rainwoodman:
It appears that arange does not purge the buffer (It is from PyArray_New):
https://github.com/numpy/numpy/blob/290a0345c8a0eb33ba98172267e6d392110fe905/numpy/core/src/multiarray/ctors.c#L2982

I think this is related to long floating point numbers that do not occupy
the entire 16 bytes it claims (80bit vs 128bit). That was the case for an
HDF5 bug: https://github.com/HDFGroup/hdf5/pull/263

With the byte-width difference, I suspect a naive looking reg-to-memory
assignment may leave uninitialized bytes, which can leak information in
unintended ways.

tacaswell:
On one hand, it looks like we immediately write to the `np.empty` (rather than read as @aragilar notes), but I suspect that the msan is being maximally paranoid and noticing that there is a period of time when we have undefined values around (and in e.g. a threading case something could jump in and read it!) rather than concrete evidence that we are reading from uninitialized memory.

This seems harmless (makes the tests _marginally_ slower) and if it silences warnings it is worth merging.

tacaswell (approved):

rainwoodman:
Thanks for merging.

It is probably a good/fortifying practice to always zero/error-fill an output buffer before filling it with real-data.

The way msan noticed this is that the uninitialized bytes are consumed during IO. My explanation for this is that the uninitialized bytes are the gaps left in a 128bit float due to the 80bit precision "optimization".

Whatever existing values stored in these bytes can leak to the saved file and get passed along with the saved file. The paranoidal aspect is that the probability these bytes contain exploitable, and interesting information is low.

PR #1902, opened 2021-06-08
Improve errors when opening files in append mode

takluyver:
'Append mode' is an h5py addition which tries to open an existing file (read+write), and falls back to creating a new file. With this change, it will only create a new file when opening the file fails with FileNotFoundError (Posix errno 'ENOENT'). This should mean clearer errors if e.g. a read-only file exists. It depends on #1815, retrieving the errno from HDF5 error messages.

There is a minor risk of breaking code if it depends on getting specific errors which have changed. But we only raised specific subtypes of OSError from 3.2 anyway, so it's not a big risk.


takluyver:
Yeah, #1877 seemed to fix it for a while, but seemingly not entirely.

PR #1903, opened 2021-06-08
Use Ubuntu 20.04 on Azure pipelines

takluyver:
I just noticed a warning on the CI builds for #1902 that Ubuntu 16.04 would be removed in a few more months. This jumps to 20.04, the latest LTS version at present.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -114,7 +114,7 @@ jobs:
         HDF5_DIR: $(HDF5_CACHE_DIR)/$(HDF5_VERSION)
         H5PY_ENFORCE_COVERAGE: yes
     # do mpi tests
-      py37-deps-hdf51103-mpi:
+      py37-deps-hdf51105-mpi:
         python.version: '3.7'
         TOXENV: py37-test-mindeps-mpi4py
         HDF5_VERSION: 1.10.5
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -117,7 +117,7 @@ jobs:
       py37-deps-hdf51103-mpi:
         python.version: '3.7'
         TOXENV: py37-test-mindeps-mpi4py
-        HDF5_VERSION: 1.10.3
+        HDF5_VERSION: 1.10.5
         HDF5_DIR: $(HDF5_CACHE_DIR)/$(HDF5_VERSION)
         HDF5_MPI: ON
         CC: mpicc
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -8,7 +8,7 @@ variables:
 jobs:
 - job: 'static_checks'
   pool:
-    vmImage: ubuntu-16.04
+    vmImage: ubuntu-20.04
   variables:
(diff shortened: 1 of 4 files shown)


takluyver:
@aragilar the MPI build is failing to install mpi4py, with errors like this:

```
    /usr/lib/x86_64-linux-gnu/openmpi/include/mpi.h:322:57: error: expected expression before ‘_Static_assert’
      322 | #define THIS_SYMBOL_WAS_REMOVED_IN_MPI30(func, newfunc) _Static_assert(0, #func " was removed in MPI-3.0.  Use " #newfunc " instead.")
          |                                                         ^~~~~~~~~~~~~~
    /usr/lib/x86_64-linux-gnu/openmpi/include/mpi.h:1112:24: note: in expansion of macro ‘THIS_SYMBOL_WAS_REMOVED_IN_MPI30’
     1112 | #        define MPI_UB THIS_SYMBOL_WAS_REMOVED_IN_MPI30(MPI_UB, MPI_Type_create_resized);
          |                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    src/mpi4py.MPI.c:15086:36: note: in expansion of macro ‘MPI_UB’
    15086 |   __pyx_t_1 = (((__pyx_v_ob[0]) == MPI_UB) != 0);
          |                                    ^~~~~~
```

Any idea what we need to do here? We can obviously fall back to Ubuntu 18.04 and see if the problem disappears, but that's not very satisfying. I presume it works with newer platforms.

aragilar:
I suspect we need to bump the mpi4py version (see https://bitbucket.org/mpi4py/mpi4py/issues/115/cannot-build-against-openmpi-400). Based on the release dates at https://pypi.org/project/mpi4py/#history, 3.0.1 should be fine.

takluyver:
Same error. Should we just try 3.0.2? Feel free to push to my branch if you're still awake & interested. :slightly_smiling_face:

aragilar:
Yeah, lets try 3.0.2 (I've also fixed the requirement in `tox.ini` as it was out of date).

aragilar:
There doesn't appear to be a simple way to add the vmImage name to the cache tags, instead I'm bumping the index at the beginning (so that we do not use an old cached h5py built against an old ABI of openmpi).

azure-pipelines.yml
```diff
@@ -117,7 +117,7 @@ jobs:
       py37-deps-hdf51103-mpi:
```
takluyver, line 117:
```suggestion
      py37-deps-hdf51105-mpi:
```

takluyver (commented):
Aha, caching. It looks in the logs like there's an environment variable `ImageOS=ubuntu20`, but I can't see it documented anywhere. Perhaps we could use a file like `/etc/os-release`. But maybe it's easiest just to manually tweak it every few years when we switch to a newer platform.

Are you happy with this as it stands? I'm happy for it to be merged, give or take the naming tweak I've suggested. :slightly_smiling_face:

aragilar:
I'm happy with it, merging.

PR #1904, opened 2021-06-08
Use manylinux image from Github container registry instead of Dockerhub

takluyver:
Dockerhub is ending free builds, and this is one less service to manage.

See https://github.com/h5py/hdf5-manylinux/issues/1 for more discussion.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -28,7 +28,7 @@ jobs:
     vmImage: ubuntu-16.04
   variables:
     CIBW_BUILD: cp3[789]-manylinux_x86_64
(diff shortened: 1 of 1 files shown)


takluyver:
This appears to mean that the wheels are no longer manylinux1 compatible. I'm not sure exactly what changed, but presumably it's some difference in the manylinux2010 base image we're building on.

manylinux2010 is meant to be compatible with platforms released roughly since 2010, so this shouldn't be a problem.

tacaswell (approved):

tacaswell:
I added a note about changing the wheel version.

takluyver:
Thanks!

PR #1905, opened 2021-06-09
Improvements to assembling virtual datasets

### LINKED ISSUE #1776: Check selection sizes when mapping data into a VirtualLayout

### LINKED ISSUE #1865: Management of HDF5 dataspace identifiers for virtual dataset layouts
This issue was discovered when running the following code:
```python
raw_file = "rawfile.h5"

layout = h5py.VirtualLayout(shape=(16_000_000,), dtype='i4')
vsource = h5py.VirtualSource(raw_file, 'data', shape=(2,))

for i in range(8_000_000):
    layout[2*i:2*i+1] = vsource[0:1]
```
The `layout.sources` list will hold 16 million *active* (created but not yet closed) HDF5 dataspace identifiers after the *for* loop. The HDF5 library requires huge amounts of memory for these many dataspace identifiers, on the order of several tens of gigabytes.

Until this issue is addressed on the library side, would it be possible to modify h5py code to reduce the number of active dataspace identifiers for virtual dataset layouts? One possibility could be to defer creating dataspace identifiers until having to use them, like before the `PropDCID.set_virtual()` call in the [`make_new_virtual_dset()`](https://github.com/h5py/h5py/blob/cff8537cc74a4897e2c0dc523309be17bec06dd4/h5py/_hl/dataset.py#L159-L160) function. The `VirtualSource.__getitem__()` and `VirtualLayout.__setitem__()` methods would then need to store the required information only as Python objects that do not include any low-level h5py objects that represent HDF5 identifiers.

Summary of the h5py configuration:
```
h5py    3.2.1
HDF5    1.12.0
Python  3.9.2 | packaged by conda-forge | (default, Feb 21 2021, 05:02:20) 
[Clang 11.0.1 ]
sys.platform    darwin
numpy   1.20.1
cython (built with) 0.29.22
numpy (built against) 1.19.3
HDF5 (built against) 1.12.0
```

cc: @bmribler @bljhdf

takluyver:
The core of this is that VirtualLayout contains and modifies a Dataset Creation Property List (DCPL). This combines two goals: failing faster when virtual dataset mappings don't have the right number of points (#1776) and reducing memory use by not duplicating all the HDF5 identifiers to store them in our own list before adding them to an HDF5 property list.

There's a wrinkle: we want to translate source filenames matching the destination file to `'.'` (#1546), but with the existing API for creating virtual datasets doesn't know the destination filename until after we have made the mappings in VirtualLayout. There's a new option to give the name up front. If that isn't used, VirtualLayout keeps a set of source filenames, and will transfer the mappings to a new property list if necessary when creating the virtual dataset.

A new API falls out of this rather naturally:

```python
with f.build_virtual_dataset('virt', (10, 1000), np.uint32) as layout:
    layout[0] = h5py.VirtualSource('foo.h5', 'data', (1000,))
```

You get a `VirtualLayout` object at the start of the context, and leaving the context writes it into the file.

Closes #1865
Closes #1863
Closes #1776

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -142,34 +142,6 @@ def make_new_dset(parent, shape=None, dtype=None, data=None, name=None,
     return dset_id
 
 
-def make_new_virtual_dset(parent, shape, sources, dtype, name=None,
-                          maxshape=None, fillvalue=None):
-    """ Return a new low-level dataset identifier for a virtual dataset """
-
-    # create the creation property list
-    dcpl = h5p.create(h5p.DATASET_CREATE)
-    if fillvalue is not None:
-        dcpl.set_fill_value(numpy.array([fillvalue]))
-
-    if maxshape is not None:
-        maxshape = tuple(m if m is not None else h5s.UNLIMITED for m in maxshape)
-
-    virt_dspace = h5s.create_simple(shape, maxshape)
-
-    for vspace, fpath, dset, src_dspace in sources:
-        dcpl.set_virtual(vspace, fpath, dset, src_dspace)
-
-    if isinstance(dtype, Datatype):
-        # Named types are used as-is
-        tid = dtype.id
-    else:
-        dtype = numpy.dtype(dtype)
-        tid = h5t.py_create(dtype, logical=1)
-
-    return h5d.create(parent.id, name=name, tid=tid, space=virt_dspace,
-                      dcpl=dcpl)
-
-
 class AstypeWrapper(object):
     """Wrapper to convert data on reading from a dataset.
     """
--- a/h5py/_hl/vds.py
+++ b/h5py/_hl/vds.py
@@ -13,8 +13,13 @@
 
 from copy import deepcopy as copy
 from collections import namedtuple
+
+import numpy as np
+
+from .compat import filename_encode
+from .datatype import Datatype
 from .selections import SimpleSelection, select
-from .. import h5s, h5
+from .. import h5d, h5p, h5s, h5t, h5
 from .. import version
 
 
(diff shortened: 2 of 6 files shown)


h5py/_hl/dataset.py
```diff
@@ -142,34 +142,6 @@ def make_new_dset(parent, shape=None, dtype=None, data=None, name=None,
     return dset_id
 
 
-def make_new_virtual_dset(parent, shape, sources, dtype, name=None,
```
tacaswell, line 145: I have convinced myself that this was not previously publicly accessible :+1:

ajelenak:
For what it's worth, I really like this new VDS API. 😄

h5py/_hl/vds.py
```diff
@@ -156,17 +160,83 @@ class VirtualLayout(object):
...
+    def _source_file_name(src_filename, dst_filename) -> bytes:
+        src_filename = filename_encode(src_filename)
+        if dst_filename and (src_filename == filename_encode(dst_filename)):
+            # use relative path if the source dataset is in the same
+            # file, in order to keep the virtual dataset valid in case
+            # the file is renamed.
+            return b'.'
+        return filename_encode(src_filename)
+
+    def get_dcpl(self, dst_filename=None):
```
tacaswell, line 197: Should `dst_filename` be required?  If `None` makes in into `filename_encode` it will explode with 

```
In [10]: filename_encode(None)
<ipython-input-10-eade088af214> in <module>
----> 1 filename_encode(None)

~/.virtualenvs/sys39/lib/python3.9/site-packages/h5py/_hl/compat.py in filename_encode(filename)
     17     filenames in h5py for more information.
     18     """
---> 19     filename = fspath(filename)
     20     if sys.platform == "win32":
     21         if isinstance(filename, str):

TypeError: expected str, bytes or os.PathLike object, not NoneType
```

takluyver: Ah, yes, you're right. I was changing my mind about whether to require it or not.

h5py/_hl/vds.py
```diff
@@ -156,17 +160,83 @@ class VirtualLayout(object):
...
+        """Get the property list containing virtual dataset mappings
+
+        If the destination filename wasn't known when the VirtualLayout was
+        created, it is handled here.
+        """
+        dst_filename = filename_encode(dst_filename)
+        if (self.filename is None) and (dst_filename in self._src_filenames):
```
tacaswell, line 204: I'm not sure that the complexity of _sometimes_ knowing the destination file up front is worth it?

tacaswell: I'm also worried about bug reports where someone changes `layout.filename` after having created it and now has very mangled VDS if we thought we could use `b'.'` when `__setitem__` was called, but not when `make_dataset` was called.

Also, if I'm reading this right I think that if `parent.file.filename != layout.filename` we can also get mangled VDS?

takluyver: So part of the motivation is #1865, where @ajelenak complained about the memory usage of a large number of mappings, because both h5py & HDF5 store them. I think that's a somewhat pathological case, but it's not much extra complexity to handle when we do know the filename in advance. My use cases could plausibly have a few thousand mappings, and I know calls into HDF5 have some overhead, so I'd rather not copy everything across to a new plist if I don't have to.

I agree that there's scope for messing it up. How about if I make `_filename` and `_get_dcpl` private, and add a check that `dst_filename == self._filename` (unless `self._filename is None`)? The intention is that if you give it a filename up front, you've committed to that as the destination file.

takluyver: I guess we might also make the filename matching and translation to `'.'` optional - I'm not sure if that helps or just adds another dimension of complication.

tacaswell: I had not considered the hdf5 call overhead.

Should this check be a bit more complicated? I think we need to go through making a new plist in this case (where the layout did not pre-know its file name and the dest is in the sources) and in the case where it did know the file name, but it was different than `dst_filename` ultimately was.  If the name-the-layout-know is in `self._src_filenames` then we need to both convert some `b'.'` -> paths and if dst_filename is in `self.src_filenames` then we may get to convert some paths ot `b'.'` 

We probably should make `self.filename` private as well?

tacaswell: > I agree that there's scope for messing it up. How about if I make _filename and _get_dcpl private, and add a check that dst_filename == self._filename (unless self._filename is None)? The intention is that if you give it a filename up front, you've committed to that as the destination file.

I like that plan.

> I guess we might also make the filename matching and translation to '.' optional - I'm not sure if that helps or just adds another dimension of complication.

Maybe wait until someone complain about not wanting relative paths?

tacaswell: [deleted a comment which I wrote having not read @takluyver 's second paragraph 🤦🏻 ]

takluyver:
If there are no objections, I'll merge this tomorrow, and then work on getting version 3.3 out.

tacaswell (approved):

takluyver:
I managed to break some of my own code which used `layout.sources` - it's in that gray area of attributes that aren't documented, but are named as though they're public. Hopefully no-one else was doing that :crossed_fingers: .

tacaswell:
Sorry I missed that!

asmeurer:
This has also broken some of my code that uses layout.sources. 

In one place, it's somewhat my fault, because I intentionally used lower level code instead of VirtualSource to construct a VirtualLayout. The reason is that I found that `VirtualSource.__getitem__` calls `deepcopy(self)`, which can be a performance issue when creating a virtual dataset with lots of sources (not a serious one, but significant enough that it was worth doing the optimization). It would be nice it `VirtualSource.__getitem__` was faster.

In the second place, I used `dset.virtual_sources()` to reconstruct a new virtual dataset with the exact same layout except with a different source dataset. In the old code, I achieved this by modifying the dset_name, creating a new `VDSMap`, and adding it to a new VirtualLayout's `sources`. In the new code, I have to instead do `layout.dcpl.set_virtual(*new_vds_map)`. There's a minor annoyance here which is that `virtual_sources` returns a `VDSMap` with strings but `set_virtual` requires bytes.

In this second case, both my new and old methods are about equally complicated. However, I wonder if it would make sense to make a new method on VirtualLayout to directly add a VDSMap to it.

PR #1913, opened 2021-06-24
__len__ for wrappers in dataset.py

Mizzcoollizz:

### DIFF
--- a/h5py/tests/test_dataset.py
+++ b/h5py/tests/test_dataset.py
@@ -1472,8 +1472,7 @@ class TestAstype(BaseDataset):
     def test_astype_wrapper_len(self):
         dset = self.f.create_dataset('x', (100,), dtype='i2')
         dset[...] = np.arange(100)
-        arr = dset.astype('f4')[:]
-        self.assertEqual(100, len(arr))
+        self.assertEqual(100, len(dset.astype('f4')))
 
 
 class TestScalarCompound(BaseDataset):
--- a/h5py/tests/test_dataset.py
+++ b/h5py/tests/test_dataset.py
@@ -1151,6 +1151,10 @@ class TestStrings(BaseDataset):
         # latin-1 will decode it but give the wrong text
         self.assertNotEqual(ds.asstr('latin-1')[0], data)
 
+        # len of ds
+        self.assertEqual(10, len(ds.asstr()))
+
+
         # Array output
         np.testing.assert_array_equal(
             ds.asstr()[:1], np.array([data], dtype=object)
@@ -1463,6 +1467,14 @@ class TestAstype(BaseDataset):
         dset[...] = np.arange(100)
         arr = dset.astype('f4')[:]
         self.assertArrayEqual(arr, np.arange(100, dtype='f4'))
+
+    
+    def test_astype_wrapper_len(self):
+        dset = self.f.create_dataset('x', (100,), dtype='i2')
+        dset[...] = np.arange(100)
+        arr = dset.astype('f4')[:]
+        self.assertEqual(100, len(arr))
+
 
 class TestScalarCompound(BaseDataset):
 
--- a/h5py/tests/test_dataset.py
+++ b/h5py/tests/test_dataset.py
@@ -1278,6 +1278,9 @@ class TestCompound(BaseDataset):
             self.f['test'].fields('x')[:], testdata['x']
         )
 
+        # Check len() on fields wrapper
+        assert len(self.f['test'].fields('x')) == 16
+
 
 class TestSubarray(BaseDataset):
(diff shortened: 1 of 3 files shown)


news/wrappers_len.rst
```diff
@@ -0,0 +1,6 @@
+New features
+------------
+
+* __len__ function for AstypeWrapper
+* __len__ function for AsStrWrapper
+* __len__ function for FieldsWrapper
```
takluyver, line 6: The class names aren't really obvious in the public API - I'd probably describe them as `dset.astype()` etc.

takluyver (commented):
Bedankt!

Could I ask you to add a couple of little tests for these as well? Search for e.g. `test_asstr` to find relevant existing tests.

tacaswell:
Thank you for working on this @Mizzcoollizz !

The failure is still in `.authors.yml`, 

```
'elize.herrewijnen@gmail.com' does not have 'Elize' listed! Please add the name to 'elize.herrewijnen@gmail.com' in '.authors.yml'. For example:

- name: Elize
  email: elize.herrewijnen@gmail.com

Or add an alias:

- name: Some Body Else
  email: elize.herrewijnen@gmail.com
  aliases:
    - Elize
All CLI and import requirements met for authors activity
ERROR: InvocationError for command /home/vsts/work/1/s/.tox/rever/bin/rever check --activities authors,version_bump,changelog (exited with code 1)

```

It looks like it is failing due to a miss-match between the name you put in Authors (your full name) and the name you told git and is being parsed out of the commit (just your first name). 

----

Could you please add a test for this?   I know it seems trivial, but it will make sure we do not accidentally delete or refactor the functionality away in the future!

Are there any other things would would want to forward from dset?  Maybe `{shape, ndims, dims, chunks}` which are in the same category as `len`?  Would we also want to forward a majority of the API?  Not sure how far down that path we want to go (either _now_ or ever).

tacaswell:
Sorry you got simultaneous reviews!

takluyver:
@Mizzcoollizz just checking if you're planning to come back to this? I think tests were the main thing both @tacaswell and I asked for in our reviews. :slightly_smiling_face:

h5py/tests/test_dataset.py
```diff
@@ -1464,6 +1468,14 @@ def test_astype_wrapper(self):
         arr = dset.astype('f4')[:]
         self.assertArrayEqual(arr, np.arange(100, dtype='f4'))
 
+
+    def test_astype_wrapper_len(self):
+        dset = self.f.create_dataset('x', (100,), dtype='i2')
+        dset[...] = np.arange(100)
+        arr = dset.astype('f4')[:]
+        self.assertEqual(100, len(arr))
```
takluyver, line 1476: This isn't actually testing the method you added, because you're reading the data to a numpy array first. Try this:

```suggestion
        self.assertEqual(100, len(dset.astype('f4')))
```

takluyver:
Thanks! Can you also add a test for `FieldsWrapper`? You can find a relevant test to start from here:

https://github.com/h5py/h5py/blob/b5782032b85633157552e2cd31a3826edd8e7b49/h5py/tests/test_dataset.py#L1255-L1275

Also, unfortunately the author check is now seeing two different email addresses, and it wants to find both in `.authors.yml`.

takluyver:
I added a test for `len(dset.fields())`. Unfortunately, the CI is still failing on the author information. It's getting confused because your commits have two different names (with & without your surname) and two different emails (gmail & kpn).

If you're happy for all of those details to be recorded in Github, the easiest thing is just add them all to `.authors.yml`, using `aliases` for names and `alternate_emails` for extra emails. If you'd rather not record some of them, we'll need to rewrite some commits, which is possible but a bit fiddly. If you want that but aren't sure how to do it, let me know and I can squash your commits together into a single one with whichever name & email you prefer.

takluyver:
OK, we've got rid of the authors check that was causing these problems.

As things stand, your two email addresses (gmail & kpn) will still be recorded in the commit history (`git log`). If you'd prefer not to have one of these recorded, let me know soon and I'll clean it up. Otherwise, I'll assume this is OK.

Mizzcoollizz:
> OK, we've got rid of the authors check that was causing these problems.
> 
> As things stand, your two email addresses (gmail & kpn) will still be recorded in the commit history (`git log`). If you'd prefer not to have one of these recorded, let me know soon and I'll clean it up. Otherwise, I'll assume this is OK.

Hi! Thank you! Could you remove the kpn email? That email address doesn't exist and I'm not sure where it came from...

takluyver:
No problem.

takluyver:
Thanks for your patience :slightly_smiling_face:

Issue #1920, opened 2021-07-06
Creation of file with flag="a" and driver="mpio" not working


sbacchio:
Hello,

I'm running in a issue introduced by PR #1902 (https://github.com/h5py/h5py/pull/1902/files#diff-6515aaa459980029aa5ae782298c82cacb68f8f6f3218a9f0419f7148bd80f30L208).
When opening a non existing file with `File(filename, "a", driver="mpio", comm=comm)` `OSError` gets raised.
This is because `mpi4py` manages the opening and it raises `OSError` instead of `FileNotFoundError`.

Here an error log
```
  /opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/site-packages/h5py/_hl/files.py:211: in make_fid
      with File(filename, "a", driver="mpio", comm=comm) as h5f:
  /opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/site-packages/h5py/_hl/files.py:444: in __init__
      fid = make_fid(name, mode, userblock_size,
  /opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/site-packages/h5py/_hl/files.py:211: in make_fid
      fid = h5f.open(name, h5f.ACC_RDWR, fapl=fapl)
  h5py/_objects.pyx:54: in h5py._objects.with_phil.wrapper
      fid = h5f.open(name, h5f.ACC_RDWR, fapl=fapl)
  h5py/_objects.pyx:54: in h5py._objects.with_phil.wrapper
      ???
      ???
  h5py/_objects.pyx:55: in h5py._objects.with_phil.wrapper
  h5py/_objects.pyx:55: in h5py._objects.with_phil.wrapper
      ???
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
  
      ???
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
  
  >   ???
  E   OSError: Unable to open file (MPI_ERR_NO_SUCH_FILE: no such file or directory)
  
  h5py/h5f.pyx:100: OSError
  >   ???
  E   OSError: Unable to open file (MPI_ERR_NO_SUCH_FILE: no such file or directory)
  
  h5py/h5f.pyx:100: OSError
[...]

Issue #1916, opened 2021-07-01
"core" file driver slightly broken in 3.3.0 with `backing_store=False`


paulmueller:
The usage of the "core" driver is not consistent from version 3.2.1 to 3.3.0.

While this worked in 3.2.1:

```python
import h5py
h5 = h5py.File(name="nonexistentfile.h5", driver="core", backing_store=False, mode="a")
```

... it doesn't in 3.3.0:

```
 /opt/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/site-packages/h5py/_hl/files.py:447: in __init__
    swmr=swmr)
/opt/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/site-packages/h5py/_hl/files.py:211: in make_fid
    fid = h5f.open(name, h5f.ACC_RDWR, fapl=fapl)
h5py/_objects.pyx:54: in h5py._objects.with_phil.wrapper
    ???
h5py/_objects.pyx:55: in h5py._objects.with_phil.wrapper
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   OSError: Unable to open file (unable to open file)

h5py/h5f.pyx:100: OSError
```

 In 3.3.0, you have to specify `mode="w"` (which also works in 3.2.1).

```python
import h5py
h5 = h5py.File(name="nonexistentfile.h5", driver="core", backing_store=False, mode="w")
```

Note that the file "nonexistentfile.h5" does not exist.

The [docs](https://docs.h5py.org/en/stable/high/file.html#file-drivers) are not clear as to why `mode="a"` should not work.

Additional information
----------------------------

 * Operating System: Ubuntu 20.04
 * Python version: 3.8.5
 * Where Python was acquired: apt
 * h5py version: 3.3.0
 * HDF5 version: 1.12.0

```
Summary of the h5py configuration
---------------------------------

h5py    3.2.1
HDF5    1.12.0
Python  3.8.5 (default, May 27 2021, 13:30:53) 
[GCC 9.3.0]
sys.platform    linux
sys.maxsize     9223372036854775807
numpy   1.20.2
cython (built with) 0.29.22
numpy (built against) 1.17.5
HDF5 (built against) 1.12.0
```

aragilar:
@paulmueller Can you confirm the issue isn't with different HDF5 versions?

[...]

Issue #1901, opened 2021-06-07
opening a file in mode 'a' fails


alexlenail:
The docs say

    >>> f = h5py.File('myfile.hdf5','r')

The file name may be a byte string or unicode string. Valid modes are:

    ========  ================================================
     r        Readonly, file must exist (default)
     r+       Read/write, file must exist
     w        Create file, truncate if exists
     w- or x  Create file, fail if exists
     a        Read/write if exists, create otherwise
    ========  ================================================

But opening a file in mode `'a'` causes `OSError: Unable to create file (file exists)` if the file exists. 

<details>
h5py    3.2.1
HDF5    1.12.0
Python  3.7.10 (default, Feb 27 2021, 02:19:57)
[Clang 12.0.0 (clang-1200.0.32.29)]
sys.platform    darwin
sys.maxsize     9223372036854775807
numpy   1.18.1
cython (built with) 0.29.22
numpy (built against) 1.14.5
HDF5 (built against) 1.12.0
</details>

takluyver:
Do you have permission to open the file in write mode? I suspect this is a problem with reporting the error - it's failing to open it, then falling back to trying to create it.

alexlenail:
I do have permission. It seems like my mistake may have been that the file might have already been open, with a different handle, elsewhere in my notebook. If that explains it, please close this issue -- unless you think it would make sense to catch that specific error case.

takluyver:
That might do it, although normally HDF5 is OK with opening the same file twice in the same process (the library coordinates access internally). I've opened PR #1902 to improve how errors are handled with append mode.

Issue #940, opened 2017-10-23
Incorrect error message when opening corrupt file


pletnes:
Minimal working example:
```python
import h5py
h5py.File('./corrupt-file.hdf', 'a')
```

Corrupt file example can be downloaded here, if necessary.
https://www.dropbox.com/s/7gttf9h8s0mbxkw/corrupt-file.hdf?dl=0

Version info:
```ipython
In [7]: print(h5py.version.info)
Summary of the h5py configuration
---------------------------------
h5py    2.7.0
HDF5    1.8.17
Python  3.5.4 |Anaconda, Inc.| (default, Oct 13 2017, 11:22:58) 
[GCC 7.2.0]
sys.platform    linux
sys.maxsize     9223372036854775807
numpy   1.12.1
```

Tried to open a file in 'a' mode. After the fact, I have found that the file is corrupt. The error message is incorrect, and I get a 'core dumped', which should not happen (python should exit normally). Seems like some edge case to do with interpreting error codes when opening files. The error message is 'file exists', but this should be okay when opening with `h5py.File(mode='a')`.
```
Traceback (most recent call last):
  File "/home/paul/anaconda3/envs/ecopy/lib/python3.5/site-packages/h5py/_hl/files.py", line 113, in make_fid
    fid = h5f.open(name, h5f.ACC_RDWR, fapl=fapl)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper (/home/ilan/minonda/conda-bld/h5py_1490028461908/work/h5py/_objects.c:2846)
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper (/home/ilan/minonda/conda-bld/h5py_1490028461908/work/h5py/_objects.c:2804)
  File "h5py/h5f.pyx", line 78, in h5py.h5f.open (/home/ilan/minonda/conda-bld/h5py_1490028461908/work/h5py/h5f.c:2123)
OSError: Unable to open file (File signature not found)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/paul/PycharmProjects/ecopy/ecopy/apps/uhivis/UHIvis.py", line 164, in select_file
    self.loadData(os.path.join(self.head, file))
  File "/home/paul/PycharmProjects/ecopy/ecopy/apps/uhivis/UHIvis.py", line 182, in loadData
    dfBands = self.specImg.default_bands
[...]

PR #1922, opened 2021-07-07
Except OSError for certain drivers when creating a file with "a"

sbacchio:
This PR is in response to issues #1916 and #1920 and partially reverts the changes introduced by PR #1902

### DIFF
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -209,7 +209,18 @@ def make_fid(name, mode, userblock_size, fapl, fcpl=None, swmr=False):
         # existing one (ACC_EXCL)
         try:
             fid = h5f.open(name, h5f.ACC_RDWR, fapl=fapl)
-        except FileNotFoundError:
+        # Not all drivers raise FileNotFoundError (commented those that do not)
+        except FileNotFoundError if fapl.get_driver() in (
+            h5fd.SEC2,
+            # h5fd.STDIO,
+            # h5fd.CORE,
+            h5fd.FAMILY,
+            h5fd.WINDOWS,
+            # h5fd.MPIO,
+            # h5fd.MPIPOSIX,
+            h5fd.fileobj_driver,
+            h5fd.ROS3D if ros3 else -1,
+        ) else OSError:
             fid = h5f.create(name, h5f.ACC_EXCL, fapl=fapl, fcpl=fcpl)
     else:
         raise ValueError("Invalid mode; must be one of r, r+, w, w-, x, a")
--- a/h5py/tests/test_file.py
+++ b/h5py/tests/test_file.py
@@ -784,7 +784,6 @@ class TestMPI(object):
         """ Testing creation of file with append """
         from mpi4py import MPI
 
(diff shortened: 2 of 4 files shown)


h5py/tests/test_file.py
```diff
@@ -762,6 +780,15 @@ def test_mpio(self, mpi_file_name):
             assert f
             assert f.driver == 'mpio'
 
+    def test_mpio_append(self, mpi_file_name):
+        """ Testing creation of file with append """
+        from mpi4py import MPI
+
+        if os.path.exists(mpi_file_name): os.remove(mpi_file_name)
```
takluyver, line 787: I believe `mpi_file_name` is guaranteed not to exist when the test starts (@aragilar ?). And if not, I suspect that this check doesn't work reliably in MPI - if >1 process realises the file exists, they'll race to delete it.

takluyver:
Thanks! Don't forget to write a [release note](https://docs.h5py.org/en/stable/contributing.html#write-a-release-note).

It looks some of the CI jobs are getting a pre-release version of Cython which is failing to build h5py. :crossed_fingers: that's an easy fix...

h5py/tests/test_file.py
sbacchio: Ok I checked and it is unique. I will remove that.

h5py/_hl/files.py
```diff
@@ -209,7 +209,18 @@ def make_fid(name, mode, userblock_size, fapl, fcpl=None, swmr=False):
         # existing one (ACC_EXCL)
         try:
             fid = h5f.open(name, h5f.ACC_RDWR, fapl=fapl)
-        except FileNotFoundError:
+        # Not all drivers raise FileNotFoundError (commented those that do not)
+        except FileNotFoundError if fapl.get_driver() in (
```
tacaswell, line 213: I am 50/50 that it is worth the complexity to make the exception slightly stricter.

tacaswell (approved):

tacaswell:
tried power cycling the PR to re-triggere the jobs, but not sure if they work on the branch commit or the synthetic merge commit.  If the first, this will require a rebase @sbacchio .

-----


Thank you for fix this and sorry we let it slip by.

h5py/_hl/files.py
takluyver: There have been a few issues filed through the years where the overly broad `except` obscured the original, meaningful error. At a quick look, that includes permission errors (#1901), file corruption (#940) and locking (#712, #780). Since the default driver does give us an errno, I'm convinced that it's worth dealing with some extra complexity to be specific when possible.

If you don't like the hardcoded list of drivers, another way we could do this is to check the exception, and if there's no errno, assume it's probably FileNotFound (like we assumed all OSErrors were before 3.3):

```python
except OSError as e:
    if (e.errno is not None) and (e.errno != errno.ENOENT):
        raise
    fid = h5f.create(name, h5f.ACC_EXCL, fapl=fapl, fcpl=fcpl)
```

PR #1923, opened 2021-07-07
Convert cpdef local variables to cdef

takluyver:
cpdef is meaningless here, because they're local variables, not exposed outside those functions. Hopefully this fixes the Cython failures on PR #1922.


tacaswell (approved):

tacaswell:
I have this exact same diff sitting on my computer uncommitted.

Issue #1682, opened 2020-09-29
Wheel support for linux aarch64


odidev:
**Summary**
Installing h5py on aarch64 via pip using command "pip3 install h5py" tries to build wheel from source code

**Problem description**
h5py doesn't have wheel for aarch64 on PyPI repository. So, while installing h5py via pip on aarch64, pip builds wheel for same resulting in it takes more time to install h5py. Making wheel available for aarch64 will benefit aarch64 users by minimizing h5py installation time.

**Expected Output**
Pip should be able to download h5py wheel from PyPI repository rather than building it from source code.

@h5py-team, please let me know if I can help you building wheel/uploading to PyPI repository. I am curious to make h5py wheel available for aarch64. It will be a great opportunity for me to work with you.

takluyver:
I've been working recently on automating wheel builds as part of this repository (#1674), but this doesn't include ARM. I'm not aware of free CI services which give us the option to run on aarch64.

I kind of agree with [this comment](https://github.com/pypa/manylinux/issues/84#issuecomment-623064775) from one of the NumPy maintainers. Just because the manylinux definition now supports more architectures, does not mean it's reasonable to expect every open-source project to provide precompiled wheels for each architecture. You still have the option to install from source, and you may be able to use more specific repositories like [piwheels](https://www.piwheels.org/project/h5py/).

odidev:
@takluyver , thank you for your quick response. CI services like travis gives us the option to run aarch64 builds. For azure-pipelines, we can use qemu to run an aarch64 build on x86 machine. I am able to successfully run azure-pipeline using qemu for few packages for aarch64 platform.

Adding aarch64 wheel will benefit aarch64 users to install the packages quickly.

Please let me know your feedback and suggestions regarding this.

aragilar:
I'm wary of depending on qemu (or any emulator) for producing wheels, as there may be subtle incompatibilities (I know the linux distros require being able to build natively to avoid this), and because it will likely be quite slow to build/test.

If there was an index of locations/indices (which would include piwheels) to get custom build wheels for specific hardware/software (maybe run by PyPA or some other trusted org), I think we'd link to that in the installation docs (we can link to pywheels directly, but it would be nice if there was a central place to link to so we don't end up with a lots of links to dead/unmaintained lists).

rhenwood-arm:
[...]

PR #1924, opened 2021-07-07
Changes to create and publish h5py arm64 wheel

tbbharaj:
This PR addresses the issue to build arm64 wheels. To address this issue, I created a github action to create and publish  arm64 wheels

If there are any questions/comments - I'd be happy fix things as needed.

Thank you
Tanveen


takluyver:
That's really very slow, even if we didn't rebuild HDF5 before making each wheel.

If we're doing ARM wheels, I think we need to build them on one of the CI services that offer real ARM machines (Drone, Circle & Travis are the ones we identified on #1682), rather than in an emulator.

There was also a move to build ARM wheels only at release time (https://github.com/MacPython/h5py-wheels/pull/12), but the upload there is currently broken. So fixing that would be another option.

aragilar:
Is there a reason this is using auditwheel manually? It would seem simpler and more maintainable to me to follow https://cibuildwheel.readthedocs.io/en/stable/setup/#travis-ci, and reuse https://github.com/h5py/hdf5-manylinux (adding a second dockerfile), meaning we're not going to get out of sync with the other platforms?

tbbharaj:
> Is there a reason this is using auditwheel manually? It would seem simpler and more maintainable to me to follow https://cibuildwheel.readthedocs.io/en/stable/setup/#travis-ci, and reuse https://github.com/h5py/hdf5-manylinux (adding a second dockerfile), meaning we're not going to get out of sync with the other platforms?

Sure, will work on changes to use cibuildwheels

tbbharaj:
Thank you for your feedbacks.  
1. Updated PR to build wheels using travis-ci via cibuildwheels. 
2. For upload I updated [.travis.yml](https://github.com/tbbharaj/h5py/blob/tbbharaj/arm64-wheel/.travis.yml#L78) to upload wheels to [anaconda.org](https://anaconda.org/multibuild-wheels-staging/). This is similar to how scipy, matplotlib-plot etc are doing. Ref: [scipy-wheels](https://github.com/MacPython/scipy-wheels/blob/master/.travis.yml#L172).  
We will need secret tokens to be created on your end for upload on anaconda to take place successfully.
I tested the upload process using my test org: [test-upload](https://anaconda.org/test-upload/h5py/files)
Successful builds: https://travis-ci.com/github/tbbharaj/h5py/jobs/523940441
Once the secure token and required env variables (`SCIPY_WHEELS_NIGHTLY_ACCESS` and `MULTIBUILD_WHEELS_STAGING_ACCESS`) are created on your end, I will update the .travis.yml with correct secure key.
3. Also based on feedbacks received, I created this [PR](https://github.com/h5py/hdf5-manylinux/pull/2) to create aarch64 docker image with hdf5 installed. Once merged, I will update .travis.yml to use this image via `CIBW_MANYLINUX_AARCH64_IMAGE` instead of installing the dependency here via [script](https://github.com/tbbharaj/h5py/blob/tbbharaj/arm64-wheel/ci/get_hdf5_arm64.sh)

Looking forward to inputs.
Thank you

.travis.yml
```diff
@@ -19,6 +19,7 @@ cache:
 env:
   global:
     - HDF5_CACHE_DIR=$HOME/.cache/hdf5
+    - secure: j7516m880cagQvknX5hkyucFo/VImiNLwmTUfpWs+naUo6+yzyI/9qH4/4ba6woY42tZNO9gZ9p6o7nL/UDSr+ChT9tWBBqbAh3gQWqWonDDzvym3EOTdHXyERd1ecgYtAg02jAcrbgiNaIG+5BbB31eiRrRdEz+XE1Lfqz37xair1pM+ZIOZ2E9ZFecmYWrLMn60pwcTWQI4alFLZU5F+rauSHL6pnBorLLv3HewbBPXsuE5W90Nep8S01aTNPWFGYde6hXhH1E/HOM5u3mAdWBzGNgbqsOd58KEm/U9MS/qDAFJyPbyRPIxGrZB2UcCvRgWND6F5Kgvbav3Qvd6RyncULnf/mrKFR8ic9gbIO9tepf16d4EEs+tFHNlXiqk9ByglUEHgphTs7O1JOPFceI1O9pBk3RPzqzC10Wn3BwfPScYd6kHLDBxCCHMdcDOBV2/Gb5XYrwkhMPnDQLi5Tzo6TM9meYgBktuY5tyATc6Q7e/+hl5IasaRI5KmBZl8sQxY1bP6FcMxXdMZ//8sBbMRogAl+3Z2kbzaIBvl9ZutMMJlXeTUt3E4ixEnfkxe9xSBhkBwuP7MdSEQn8ndCKZ6dV9cLxmODM8DgkW7RjQ6BwAdYwjYGc2UIpAE4iabJZiXwPdqVH7R52DwVmk6pzYdfZFJQfDMJ8PglLpes=
```
takluyver, line 22: I assume this contains the upload tokens? Can it go under the env settings for the specific aarch64 job, so it's not visible to the other job?

.travis.yml
```diff
@@ -32,17 +33,57 @@ matrix:
...
+      virt: vm
+      group: edge
+      arch: arm64
+      env:
+        - TOXENV=py37-test-deps
+        - HDF5_VERSION=1.10.5
+        - HDF5_DIR=$HDF5_CACHE_DIR/$HDF5_VERSION
+        - H5PY_ENFORCE_COVERAGE=yes
+        - CIBW_BUILD: cp3[789]-*
+        - CIBW_BEFORE_BUILD: ./ci/get_hdf5_arm64.sh
```
takluyver, line 47: For x86_64 wheels, we make our own Docker image with HDF5 added on top of the manylinux images, so we don't need to rebuild HDF5 on every CI run:

https://github.com/h5py/hdf5-manylinux

We should probably do something similar for this, though I imagine it won't be quite as straightforward. Let me know what you think. I can create a `hdf5-manylinux-arm64` repo if that's useful.

.travis.yml
```diff
@@ -32,17 +33,57 @@ matrix:
...
+  # multibuild-wheels-staging
+  #
+  - if [[ $CIBW_ARCHS == aarch64 && "$TRAVIS_EVENT_TYPE" == "cron" ]]; then
+          ANACONDA_ORG="scipy-wheels-nightly";
+          TOKEN=${SCIPY_WHEELS_NIGHTLY_ACCESS};
+    fi;
+    if [[ $CIBW_ARCHS == aarch64 && "$TRAVIS_EVENT_TYPE" == "push" ]]; then
+          ANACONDA_ORG="multibuild-wheels-staging";
+          TOKEN=${MULTIBUILD_WHEELS_STAGING_ACCESS};
+    fi;
```
takluyver, line 85: Is this something we're free to use? It looks like only a handful of packages are uploading to either of those orgs frequently. :thinking: 

@ogrisel, I see your picture on both Anaconda organisations. Is it reasonable for h5py's CI to post ARM wheels there, or should we find our own space to do that?

.travis.yml
tbbharaj: I created this PR https://github.com/h5py/hdf5-manylinux/pull/2 to make docker image with HDF5 added on top of the manylinux images for aarch64 but if we need separate repo like you mentioned (`hdf5-manylinux-arm64`) I can add changes for aarch64 to this new repo

.travis.yml
tbbharaj: Yes I can do that and update PR

.travis.yml
tbbharaj: I believe that's free, I created an org to test wheel upload for my local testing which was free. https://anaconda.org/test-upload/h5py/files

.travis.yml
tbbharaj: This key is encrypted access token for `${ANACONDA_ORG}` against this repo. When Travis-CI runs, this causes the `TOKEN` environment variable to contain the API key above.
Ref: https://github.com/matthew-brett/multibuild

tbbharaj:
Updated PR to use docker image `ghcr.io/h5py/manylinux2014_aarch64-hdf5`. To use anaconda for wheels upload, I will need to update `secure` with correct value

.travis.yml
takluyver: I looked into this, and it appears that anaconda.org has a 3 GB limit per organisation. We're looking at 3 packages somewhere between 4 MB and 10 MB each (I'm not quite sure why the aarch64 wheels you've produced appear to be much bigger than the x86_64 wheels). That seems like we'll fill it up before very long. Although the scipy-wheels-nightly repo seems to have a lot of files in, so maybe that has an exception? :confused: 

The way Azure Pipelines & similar services work is to discard older artifacts after some number of days or number of later CI runs. I wonder if it's reasonable to put files on something like https://transfer.sh/ to achieve something similar, or whether that's abusing a free service.

tbbharaj: What are your thoughts on using [twine](https://twine.readthedocs.io/en/latest/) for uploading wheels. 
Ref: https://cibuildwheel.readthedocs.io/en/stable/deliver-to-pypi/
I created a test script to upload wheels to [test.pypi](https://test.pypi.org/project/tbbharaj-h5py/3.3.3/#files
) 

https://github.com/tbbharaj/h5py/blob/tbbharaj/upload-wheels/.travis.yml#L85
Builds: https://travis-ci.com/github/tbbharaj/h5py/builds/233562261

tbbharaj: @takluyver any concerns with using twine?

takluyver: Yes, maybe using test.pypi.org is the way to go. It appears Tensorflow even uploads [`tf-nightly` packages] to the real PyPI, and they're huge (hundreds of MB for each package), so maybe that's OK.

takluyver: Just had a chat with @tacaswell; we'll look again at using the scipy-wheels-nightly repo on anaconda.org.

tbbharaj: > Just had a chat with @tacaswell; we'll look again at using the scipy-wheels-nightly repo on anaconda.org.

Thanks @takluyver let me know whatever works best, I am happy to make required changes. Thanks again!

tbbharaj:
@takluyver Following up on this PR... have any inputs on using scipy-wheels-nightly vs test.pypi?

takluyver:
Sorry for the delay. Olivier, who manages the scipy-wheels-nightly group on anaconda.org, has got back to me, so it looks like we're going to be using that. I'm waiting for him to add me to the group.

I guess we'll need to ensure each uploaded version has a unique version number, by incorporating a commit hash or a date or something. Ideally, when we tag a release version, it will build with that plain version number, so we can download the wheels and upload them to PyPI. Getting that all working may involve a bit of fiddling.

takluyver:
Olivier has given me access; I've made a token and added it to Travis (through the web interface) as an environment variable called `ANACONDA_ORG_UPLOAD_TOKEN`.

Looking at how other projects do it, I guess we want a cron build for the nightly upload, plus uploads on tags.

tbbharaj:
> Olivier has given me access; I've made a token and added it to Travis (through the web interface) as an environment variable called `ANACONDA_ORG_UPLOAD_TOKEN`.
> 
> Looking at how other projects do it, I guess we want a cron build for the nightly upload, plus uploads on tags.

Thanks @takluyver I will work on requested changes to upload it to anaconda.org

tbbharaj:
@takluyver Updated the PR to upload wheels to anaconda.org. Currently I am appending timestamp to wheels name
e.g. `h5py-3.3.0-cp37-cp37m-manylinux_2_17_aarch64.manylinux2014_aarch64.whl` to `h5py-3.3.0-cp37-cp37m-20211004151033-manylinux_2_17_aarch64.manylinux2014_aarch64.whl`. Let me know if you have further suggestions. Thanks!

takluyver:
Up to you (both) - if this works on Travis, I'm happy enough to merge it, once the question about the wheel naming is figured out. But if the ARM job becomes excessively flaky or slow, we may just disable it again until someone can fix it.

.travis.yml
```diff
@@ -32,17 +31,44 @@ matrix:
...
+        - CIBW_BUILD: cp3[789]-*
+        - CIBW_MANYLINUX_AARCH64_IMAGE: ghcr.io/h5py/manylinux2014_aarch64-hdf5
+        - CIBW_ARCHS: aarch64
 
-before_install:
-  # - export PATH=/usr/lib/ccache:$PATH
-  - ccache -s
+  before_install:
+    # - export PATH=/usr/lib/ccache:$PATH
+    - ccache -s
```
takluyver, line 51: This doesn't look quite right - the indentation puts it under `matrix:` . I think `before_install` should be at the same level as `install`.

takluyver:
Some of that unreliability showing up on the last but one build :confused: 

```
$ python --version
/home/travis/.travis/functions: line 109: /home/travis/virtualenv/python3.7.9/bin/python: Too many open files in system
$ pip --version
/home/travis/.travis/functions: line 109: /home/travis/virtualenv/python3.7.9/bin/pip: Too many open files in system
/home/travis/.travis/functions: cannot make pipe for command substitution: Too many open files in system
/home/travis/.travis/functions: cannot make pipe for command substitution: Too many open files in system
$ ccache -s
/home/travis/.travis/functions: line 109: /usr/bin/ccache: Too many open files in system
/home/travis/.travis/functions: cannot make pipe for command substitution: Too many open files in system
The command "ccache -s" failed and exited with 126 during .

Your build has been stopped.
/home/travis/.travis/functions: cannot make pipe for command substitution: Too many open files in system
```

takluyver:
Close-reopen to build with oldest-supported-numpy now in master - I think this should speed things up a bit by not needing to build numpy.

takluyver:
That did seem to speed up building the wheels a bit.

Thanks for all your work on this! I'll ping you if there are problems with it.

PR #1944, opened 2021-08-17
fix `_fast_reader`

### LINKED ISSUE #1943: `_fast_reader` doesn't work?
I was investigating dataset reading performance and noticed that the "fast" reading path in [`h5py/_hl/dataset.py:Dataset.__getitem__`](https://github.com/h5py/h5py/blob/7b19cc34d0ae9b9e7ae7d41d8818fc836a69471d/h5py/_hl/dataset.py#L700) was never actually taken for some reason.

For example:

<details>
<summary><code>test.py</code></summary>

```python
import h5py
import numpy as np

# Create a test dataset
with h5py.File("test.h5", "w") as f:
    f["data"] = np.ones(1024 * 1024, dtype=np.uint8)

# Read the data back
with h5py.File("test.h5") as f:
    f["data"][...]
    # Attempt to use the "fast" path manually
    h5py._selector.Reader(f["data"].id)
```

</details>

The fast path is not taken because the `self._fast_reader` property attempts to create a `h5py._selector.Reader`, which fails with
```python
  File "/tmp/tmp.h371s1cdem/test.py", line 12, in <module>
    h5py._selector.Reader(f["data"].id)
  File "h5py/_selector.pyx", line 312, in h5py._selector.Reader.__cinit__
TypeError: an integer is required
```
This exception is caused by [`PyArray_IsNativeByteOrder(np_dtype.byteorder)` in `h5py/_selector.pyx`](https://github.com/h5py/h5py/blob/7b19cc34d0ae9b9e7ae7d41d8818fc836a69471d/h5py/_selector.pyx#L312).

The problem is that `np_dtype.byteorder` is a python string (one of `"<"`, `">"`, `"="`, `"s"` or `"|"`), but `PyArray_IsNativeByteOrder` expects its argument to be a `char` (integer) and not a string. See [`numpy/core/include/numpy/ndarraytypes.h`](https://github.com/numpy/numpy/blob/9e1c34eda4569a96fba07ed56fb971dedbb438fa/numpy/core/include/numpy/ndarraytypes.h#L1773).

---

So unless I'm missing something, `_fast_reader` is currently never used. I think, that the correct way to check for native byte order is to cast the byteorder string to its character ordinal like this:

<details>
<summary><code>patch</code></summary>

```diff
diff --git a/h5py/_selector.pyx b/h5py/_selector.pyx
index bb88caa5..be42b95e 100644
--- a/h5py/_selector.pyx
+++ b/h5py/_selector.pyx
@@ -309,7 +309,7 @@ cdef class Reader:
         h5_stored_datatype = typewrap(H5Dget_type(self.dataset))
         np_dtype = h5_stored_datatype.py_dtype()
         self.np_typenum = np_dtype.num
-        self.native_byteorder = PyArray_IsNativeByteOrder(np_dtype.byteorder)
+        self.native_byteorder = PyArray_IsNativeByteOrder(ord(np_dtype.byteorder))
         self.h5_memory_datatype = py_create(np_dtype)
 
     cdef ndarray make_array(self, hsize_t* mshape):
```

</details>

I've tested the read speed with the following script:

<details>
<summary><code>speed.py</code></summary>

```python
import h5py
import numpy as np
from timeit import timeit

# Create a test dataset
with h5py.File("test.h5", "w") as f:
    f["data"] = np.ones(1024 * 1024, dtype=np.uint8)

# Read the data back
with h5py.File("test.h5") as f:
    print(timeit(lambda: f["data"][...], number=64*1024))
```

</details>

And I get `18.643` seconds without the patch and `13.009` seconds with it.

If I haven't missed anything, I'll submit a Pull Request with the above patch.

---

Related PRs and people:
 - #1428 Original Reader impl.
 - #1730 Native byteorder check.
 - #1731 Refactor to use `PyArray_IsNativeByteOrder`.
 - @takluyver

<details>
<summary>System info</summary>

 * Operating System: Manjaro Linux (Kernel 5.10)
 * Where Python was acquired: system Python
 * Version info:
   
   ```
   h5py    3.3.0
   HDF5    1.12.0
   Python  3.9.6 (default, Jun 30 2021, 10:22:16) 
   [GCC 11.1.0]
   sys.platform    linux
   numpy   1.20.3
   cython (built with) 0.29.24
   numpy (built against) 1.19.3
   HDF5 (built against) 1.12.0
   ```

</details>

RuRo:
Fixes #1943.

### DIFF
--- a/h5py/_selector.pyx
+++ b/h5py/_selector.pyx
@@ -309,7 +309,7 @@ cdef class Reader:
         h5_stored_datatype = typewrap(H5Dget_type(self.dataset))
         np_dtype = h5_stored_datatype.py_dtype()
         self.np_typenum = np_dtype.num
-        self.native_byteorder = PyArray_IsNativeByteOrder(np_dtype.byteorder)
+        self.native_byteorder = PyArray_IsNativeByteOrder(ord(np_dtype.byteorder))
         self.h5_memory_datatype = py_create(np_dtype)
 
     cdef ndarray make_array(self, hsize_t* mshape):
--- a/h5py/tests/common.py
+++ b/h5py/tests/common.py
@@ -130,7 +130,7 @@ class TestCase(ut.TestCase):
             assert np.all(dset[...] == arr[...]), \
                 "Arrays are not equal (dtype %s) %s" % (arr.dtype.str, message)
 
-    def assertNumpyBehavior(self, dset, arr, s):
+    def assertNumpyBehavior(self, dset, arr, s, skip_fast_reader=False):
         """ Apply slicing arguments "s" to both dset and arr.
 
         Succeeds if the results of the slicing are identical, or the
@@ -144,11 +144,23 @@ class TestCase(ut.TestCase):
         except Exception as e:
             exc = type(e)
 
+        s_fast = s if isinstance(s, tuple) else (s,)
+
         if exc is None:
             self.assertArrayEqual(dset[s], arr_result)
+
+            if not skip_fast_reader:
+                self.assertArrayEqual(
+                    dset._fast_reader.read(s_fast),
+                    arr_result,
+                )
         else:
             with self.assertRaises(exc):
                 dset[s]
+
+            if not skip_fast_reader:
+                with self.assertRaises(exc):
+                    dset._fast_reader.read(s_fast)
 
 NUMPY_RELEASE_VERSION = tuple([int(i) for i in np.__version__.split(".")[0:2]])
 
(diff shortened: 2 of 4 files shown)


takluyver:
Thanks! While we're looking at this, can I ask you to add a simple test so we'd catch if something like this happened again? It should be enough to make a simple dataset and then check that `dset._fast_reader.read(np.s_[:])` gives an appropriate array.

Also, don't forget to add a [release note](https://docs.h5py.org/en/stable/contributing.html#write-a-release-note) about it.

takluyver:
Thanks, this looks good. :-)

PR #1945, opened 2021-08-18
Build Windows & Mac wheels with HDF5 1.12.1

takluyver:
Linux wheels are already using 1.12.1, from our [hdf5-manylinux](https://github.com/h5py/hdf5-manylinux) Docker images.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -40,7 +40,7 @@ jobs:
   pool:
     vmImage: vs2017-win2016
   variables:
-    HDF5_VERSION: 1.12.0
+    HDF5_VERSION: 1.12.1
     HDF5_DIR: $(HDF5_CACHE_DIR)/$(HDF5_VERSION)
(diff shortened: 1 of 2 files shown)


aragilar:
Not sure how we're doing news/changelogs (rever?), but we may want to add something so we don't forget me made this change :smiley:.

With respect to the actual upgrade, my general impression is 1.12 is always better than 1.10 (vs. the 1.10 from 1.8 upgrade, where I think there was a performance decrease, not that I noticed anything personally), so upgrading for the next release makes sense to me.

takluyver:
Yup, news entries are managed by rever; I'll add one in a moment. :smiley:

takluyver:
We'd already upgraded to 1.12.0 for the last release, so :crossed_fingers: there's no reason not to go for 1.12.1.

takluyver:
Thanks. I will hopefully get some time to work on preparing a release later today.

Issue #1950, opened 2021-08-24
Wrapping hdf5 v1.12.1 H5Pset_file_locking


t20100:
libhdf5 v1.12.1 provides a function to toggle file locking instead of using the `HDF5_USE_FILE_LOCKING` environment variable:
- [H5Pset_file_locking](https://portal.hdfgroup.org/display/HDF5/H5P_SET_FILE_LOCKING) 
- [H5Pget_file_locking](https://portal.hdfgroup.org/display/HDF5/H5P_GET_FILE_LOCKING)

It would be great to have access to this through h5py since at the same time libhdf5 changed the behaviour of the `HDF5_USE_FILE_LOCKING` environment variable and it's no longer possible to use it to workaround file locking on a per file basis.

related to #1722

takluyver:
Adding it to the low-level API should be simple and uncontroversial, if you want to make a PR for that (there are [docs on how to add HDF5 functions](https://docs.h5py.org/en/stable/contributing.html#adding-a-function-from-the-hdf5-c-api)).

Whether & how to expose it in the high-level API might be a bit more controversial, so I'd suggest leaving that to a separate PR. But people have problems with file locking often enough that it's probably reasonable.

t20100:
OK, I got something that starts to work, I'll prepare 2 PRs then.

takluyver:
Closed by #1951 (low-level methods) and #1952 (high-level option for `File` class).

PR #1951, opened 2021-08-25
Add HDF5 1.12.1 file locking low-level API to PropFAID

t20100:
This PR adds low level wrappers and minimal tests for: 
- [H5Pset_file_locking](https://portal.hdfgroup.org/display/HDF5/H5P_SET_FILE_LOCKING) 
- [H5Pget_file_locking](https://portal.hdfgroup.org/display/HDF5/H5P_GET_FILE_LOCKING)

It is enabled for HDF5 >= 1.12.1.

Actually, those 2 functions are also available in HDF5 v1.10.7, but not in v1.12.0 and I don't see how to express this in `api_functions.txt`.

Related to #1950


t20100:
Added support for HDF5 1.10.x >= 1.10.7

takluyver:
Thanks!

Issue #1864, opened 2021-03-23
Opening a file with "w" which is already open elsewhere deletes the file's contents and errors


ivirshup:
Initially reported: https://github.com/theislab/anndata/issues/522

If I open an hdf5 file in one process (process A) as "r", then open the same path in another process (process B) as "w" the contents of the file are deleted, process B errors, and an empty file is left in it's place

### Process A

```python
In [1]: import h5py
   ...: f = h5py.File("tmp.h5", "w")
   ...: f["X"] = [1, 2, 3]
   ...: f.close()
   ...: 
   ...: g = h5py.File("tmp.h5")
```

### Process B

```python
In [1]: import h5py
   ...: f = h5py.File("tmp.h5", "w")
---------------------------------------------------------------------------
BlockingIOError                           Traceback (most recent call last)
<ipython-input-1-a87c05b11692> in <module>
      1 import h5py
----> 2 f = h5py.File("tmp.h5", "w")

/usr/local/lib/python3.8/site-packages/h5py/_hl/files.py in __init__(self, name, mode, driver, libver, userblock_size, swmr, rdcc_nslots, rdcc_nbytes, rdcc_w0, track_order, fs_strategy, fs_persist, fs_threshold, **kwds)
    440             with phil:
    441                 fapl = make_fapl(driver, libver, rdcc_nslots, rdcc_nbytes, rdcc_w0, **kwds)
--> 442                 fid = make_fid(name, mode, userblock_size,
    443                                fapl, fcpl=make_fcpl(track_order=track_order, fs_strategy=fs_strategy,
    444                                fs_persist=fs_persist, fs_threshold=fs_threshold),

/usr/local/lib/python3.8/site-packages/h5py/_hl/files.py in make_fid(name, mode, userblock_size, fapl, fcpl, swmr)
    199         fid = h5f.create(name, h5f.ACC_EXCL, fapl=fapl, fcpl=fcpl)
    200     elif mode == 'w':
--> 201         fid = h5f.create(name, h5f.ACC_TRUNC, fapl=fapl, fcpl=fcpl)
    202     elif mode == 'a':
    203         # Open in append mode (read/write).

h5py/_objects.pyx in h5py._objects.with_phil.wrapper()

h5py/_objects.pyx in h5py._objects.with_phil.wrapper()

h5py/h5f.pyx in h5py.h5f.create()

[...]

PR #1952, opened 2021-08-25
Add `File` argument to select HDF5 file locking behaviour

t20100:
This PR builds on PR  #1951 and proposes the possibly more controversial integration of file locking in the high level API.

This PR proposes to add a `locking` argument to the already long list of `h5py.File` arguments in order to select the file locking behaviour.

It turned out that there is currently 3 file locking behaviours provided by HDF5 v1.12.1 (and 1.10.7) that can be set through the `HDF5_USE_FILE_LOCKING` env. var. and `H5Pset_file_locking` function:
- Disable file locking: `HDF5_USE_FILE_LOCKING=FALSE` and `H5Pset_file_locking(use_file_locking=False, ignore_when_disabled=False)`
- Enable file locking: `HDF5_USE_FILE_LOCKING=TRUE` and `H5Pset_file_locking(use_file_locking=True, ignore_when_disabled=False)`
- Enable file locking but ignores error if HDF5 detects that file locking was disable on the file system: `HDF5_USE_FILE_LOCKING=BEST_EFFORT` and `H5Pset_file_locking(use_file_locking=True, ignore_when_disabled=True)`

I support all 3 (+HDF5 default) by accepting `None` (default), `bool` and `str` as `locking` value , I'd take any better solution.
Maybe only providing `BEST_EFFORT` (`locking=True`) and `FALSE` (locking=`False`) is enough? In this case, `locking` would be `True`, `False` or `None`.

If the `locking` argument (or something similar) does not make it into `File`, it would be good to at least have it in `make_fapl`.
In this case, unless I misunderstood something, it would be possible to create a `File` object by giving it a `fid` created with `make_fid`, `make_facl` and `make_fcpl`, though less convenient.

related to #1950

### DIFF
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -108,7 +108,7 @@ def registered_drivers():
     return frozenset(_drivers)
 
 
-def make_fapl(driver, libver, rdcc_nslots, rdcc_nbytes, rdcc_w0, **kwds):
+def make_fapl(driver, libver, rdcc_nslots, rdcc_nbytes, rdcc_w0, locking=None, **kwds):
     """ Set up a file access property list """
     plist = h5p.create(h5p.FILE_ACCESS)
 
@@ -131,6 +131,19 @@ def make_fapl(driver, libver, rdcc_nslots, rdcc_nbytes, rdcc_w0, **kwds):
     if rdcc_w0 is not None:
         cache_settings[3] = rdcc_w0
     plist.set_cache(*cache_settings)
+
+    if locking is not None:
+        if hdf5_version < (1, 12, 1):
+            raise ValueError("HDF version 1.12.1 or greater required for file locking.")
+
+        if locking in ("false", False):
+            plist.set_file_locking(False, ignore_when_disabled=False)
+        elif locking in ("true", True):
+            plist.set_file_locking(True, ignore_when_disabled=False)
+        elif locking == "best-effort":
+            plist.set_file_locking(True, ignore_when_disabled=True)
+        else:
+            raise ValueError("Unsupported locking value: %s" % locking)
 
     if driver is None or (driver == 'windows' and sys.platform == 'win32'):
         # Prevent swallowing unused key arguments
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -133,8 +133,9 @@ def make_fapl(driver, libver, rdcc_nslots, rdcc_nbytes, rdcc_w0, locking=None, *
     plist.set_cache(*cache_settings)
 
     if locking is not None:
-        if hdf5_version < (1, 12, 1):
-            raise ValueError("HDF version 1.12.1 or greater required for file locking.")
+        if hdf5_version < (1, 12, 1) and (hdf5_version[:2] != (1, 10) or hdf5_version[2] < 7):
+            raise ValueError(
+                "HDF version >= 1.12.1 or 1.10.x >= 1.10.7 required for file locking.")
 
         if locking in ("false", False):
             plist.set_file_locking(False, ignore_when_disabled=False)
@@ -426,7 +427,7 @@ class File(Group):
             None                Use HDF5 defaults
             Warning: The HDF5_USE_FILE_LOCKING environment variable can override
             this parameter.
-            Only available with HDF5 >= 1.12.1.
+            Only available with HDF5 >= 1.12.1 or 1.10.x >= 1.10.7.
         Additional keywords
             Passed on to the selected file driver.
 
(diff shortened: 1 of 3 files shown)


h5py/tests/test_file.py
```diff
@@ -840,6 +841,59 @@ def test_swmr_mode_consistency(self):
...
+
+
+    with h5py.File(fname, mode="w", locking=True) as f:
+        f.flush()
+
+        # Opening again with locking is expected to fail
+        assert not open_in_subprocess(fname, mode="r", locking=True)
+
+        # Opening again without locking is expected to work
+        assert open_in_subprocess(fname, mode="r", locking=False) is True
```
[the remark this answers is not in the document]
t20100: There were 2 issues here, first an issue with escaping `\` in the filename and then indeed a file lock.
It seems Windows locks the file when opened in write mode but not in read mode, so inversing the order made the test pass.

t20100:
> the difference between True & best-effort is just about whether it errors if it can't lock the file, or carries on.
Yes, "best-effort" ignores errors if it detects file locking is disabled on the file system.

- Added support for HDF5 1.10.x >= 1.10.7
- Fixed test with Windows: let's see if CI is happy with it
- I kept locking="true"/True/"false"/False/"best-effort"/None. Not nice, but consistent with values accepted by the env. var. `HDF5_USE_FILE_LOCKING` (`TRUE`/`FALSE`/`BEST_EFFORT`)

t20100:
Rebased to master since PR #1951 was merged.

t20100:
I reworked the tests:
- Moved file locking tests in a class to remove duplicated `skipif`
- For the test with a subprocess: Patched `sys.path` to make sure to use the same `h5py` in the subprocess. This test works but is probably a bit fragile due to the use of subprocess.
- Added some test to get 100% coverage: This requires patching  `h5py._hl.files.hdf5_version` to fake an unsupported HDF5 version (if you don't want to enforce 100% coverage, it is maybe best to remove this).

t20100:
BTW, still hesitate to remove `"true"` and `"false"` and only keep boolean for those options.

takluyver:
Sorry, I've left this a while without looking at it.

Naming: I think what you've got now is good. I like that the string names correspond to what you can set in the environment variable, and I agree that if we're accepting true/false as strings, we should accept the booleans as well.

> This requires patching h5py._hl.files.hdf5_version to fake an unsupported HDF5 version (if you don't want to enforce 100% coverage, it is maybe best to remove this).

I don't want to enforce 100% coverage. :slightly_smiling_face: And yes, I think leaving some error branches uncovered is better than doing weird monkeypatching to pretend we have a different version of HDF5.

(If we really wanted to test it, we could have a test that's skipped on *newer* HDF5 - the CI includes jobs with old enough HDF5 to test it. But I don't think it's worth worrying about, to be honest.)

h5py/_hl/files.py
```diff
@@ -418,6 +442,10 @@ def __init__(self, name, mode='r', driver=None,
             raise ValueError(
                 "h5py was built without ROS3 support, can't use ros3 driver")
 
+        if locking is not None and hdf5_version < (1, 12, 1) and (
+                hdf5_version[:2] != (1, 10) or hdf5_version[2] < 7):
+            raise ValueError("HDF version >= 1.12.1 or 1.10.x >= 1.10.7 required for file locking.")
```
takluyver, line 447:
```suggestion
            raise ValueError("HDF version >= 1.12.1 or 1.10.x >= 1.10.7 required for file locking options.")
```

Just to make it slightly clearer - older versions already did file locking, they just didn't have this way to control it.

h5py/tests/test_file.py
```diff
@@ -840,6 +842,91 @@ def test_swmr_mode_consistency(self):
...
+
+            process = subprocess.run(
+                [
+                    sys.executable,
+                    "-c",
+                    f"""
+import sys
+sys.path.insert(0, '{escaped_h5py_import_dir}')
+import h5py
+f = h5py.File('{escaped_filename}', mode='{mode}', locking={locking})
```
takluyver, line 915:
```suggestion
sys.path.insert(0, {h5py_import_dir!r})
import h5py
f = h5py.File({filename!r}, mode={mode!r}, locking={locking})
```

`!r` is formatting shorthand for `repr()`, which will do the necessary escaping & quoting to make a string literal, which is what we want here:

```
>>> a = "C:\\foo"
>>> f"{a}"
'C:\\foo'
>>> f"{a!r}"
"'C:\\\\foo'"
```

t20100:
I removed the test monkey-patching hdf5version to test errors with unsupported version of libhdf5.

t20100:
Fixed issue with `f{...!r}` usage and parameter being `pathlib.Path`.

takluyver:
Thanks!

tdegeus:
Thanks for this! However, from the doctoring (which is not rendered on readthedocs I think) I could not really figure out if this option is what I need. 
I need to lock a file against any write while having it open. Should I do:
```python
with h5py.File("mydata.h5", "a", locking=True) as file:
    ...
```
?

takluyver:
Locking files (`locking=True`) is HDF5's default behaviour - these options are mostly about skipping locking where people find it's a problem.

What you see might be a bug, either in HDF5 itself (see e.g. #1864) or with the filesystem which you're writing to (locking and network filesystems don't always go well together).

Issue #1947, opened 2021-08-22
Numpy / H5py types sometimes behave as expected, but depends on order


rtpavlovsk21:
Numpy and h5py data types are often compatible, however there are nuances. It seems like the typing for ```==``` will produce mask arrays in the case of the input being a scalar and a h5py dataset. However if the operands are swapped ```==``` seems to return an unexpected result. Below are tests to show this behavior in a self-contained tests.py.

To assist reproducing bugs, please include the following:
 * Linux 5.10.41-1-MANJARO  SMP PREEMPT Fri May 28 19:10:32 UTC 2021 x86_64 GNU/Linux
 * platform linux -- Python 3.9.5, pytest-6.2.4,
 * Where Python was acquired: pacman / pip
 * h5py version 3.1.0
 * hdf5 1.12.0-2
 * See the output of tests.py below

`h5py.version.info` contains the needed versions, which can be displayed by
```bash
Summary of the h5py configuration
---------------------------------

h5py    3.1.0
HDF5    1.12.0
Python  3.9.5 (default, May 24 2021, 12:50:35) 
[GCC 11.1.0]
sys.platform    linux
sys.maxsize     9223372036854775807
numpy   1.21.0
cython (built with) 0.29.21
numpy (built against) 1.19.3
HDF5 (built against) 1.12.0

```

```python tests.py
# coding: utf-8
import h5py
import numpy as np
import logging

fname = 'mydatafile.h5'

def create_dataset():
    f = h5py.File(fname, 'w')
    dset = f.create_dataset("test", (100,1))
    f.close()

def test_convert_to_numpy_test_commutative():
    '''
    Conversion of h5py datasets to numpy array returns correct result
    '''
    print('%s %s'%('h5py.__version__',h5py.__version__))
    create_dataset()

    data = h5py.File(fname)
    dets = data['test']
    udets = np.unique(dets)

    type(dets)
    dets = np.asarray(dets)
    for i in udets:
        assert np.sum(i==dets) == np.sum(dets==i)


    type(dets)

def test_commutative_h5py():
    '''
    h5py dataset has suprising commutative asymmetry when not converted to numpy types explicitly
    '''
    print('%s %s'%('h5py.__version__',h5py.__version__))
    data = h5py.File(fname)
    dets = data['test']
    udets = np.unique(dets)

    for i in udets:
        assert np.sum(i == dets) ==  np.sum(dets == i)

    for i in udets:
[...]

PR #1954, opened 2021-08-26
PR to make Numpy / h5py comparisons commutative in __eq__ and __neq__

rtpavlovsk21:
Numpy / h5py types are often compatible, this pull request extends the ```==``` and ```!=``` functionality. In the Numpy special case this enables masking operations  with Numpy arrays and H5py datasets, so that they are commutative.

This pull request is related to the issue https://github.com/h5py/h5py/issues/1947.

Before opening a pull request, please:

[x] Run simple static checks with `tox -e pre-commit`
[o] Run the tests with e.g. `tox -e py37-test-deps`
[x] Run the tests with e.g. `tox -e py39-test-deps`

### DIFF
--- a/h5py/_hl/base.py
+++ b/h5py/_hl/base.py
@@ -348,7 +348,7 @@ class HLObject(CommonStateObject):
     def __eq__(self, other):
         if hasattr(other, 'id'):
             return self.id == other.id
-        return False
+        return NotImplemented
 
     @with_phil
     def __ne__(self, other):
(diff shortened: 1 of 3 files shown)


takluyver:
codecov reporting is a bit inconsistent, unfortunately - it likes to randomly say coverage has dropped on pull requests. I think it's something to do with the fact that we have a lot of test jobs on several CI services - possibly it's not waiting for all the jobs to finish before it reports the coverage.

If you can figure out how to fix it, that's great, but don't spend too much time on it. :wink:

PR #1958, opened 2021-09-07
Disable dataset timestamps by default

### LINKED ISSUE #1953: Disable writing dataset timestamps by default?
This is a bit speculative, so I won't push it if there doesn't seem to be a consensus.

The option to disable writing timestamps (`group.create_dataset(..., track_times=False)`) appears to have been there since 2013 (#271), based on a request from 2011 (#225). It's useful because, in some circumstances, people want running the same code to produce an identical file. There are much bigger challenges for reproducibility, so an option you can simply turn off isn't the end of the world, but it's an extra little annoyance for people to deal with (e.g. [on Stackoverflow](https://stackoverflow.com/questions/16019656/hdf5-file-h5py-with-version-control-hash-changes-on-every-save), and issue #1919).

I just tried to actually inspect some timestamps (to work out how they behave with `group.copy()`), and it took a while to work out how to do so.

- I can't find how to make any of the HDF5 command line tools show timestamps (h5ls, h5dump, h5stat).
- The HDF5 C API favours [H5Oget_info](https://portal.hdfgroup.org/display/HDF5/H5O_GET_INFO3) as the way to do this, but none of the time fields are exposed in h5py (even in the low-level). The HDF5 docs say that of 4 timestamps, only ctime is implemented.
- `h5py.h5g.get_objinfo().mtime` appears to work, although the [HDF5 C function is deprecated](https://portal.hdfgroup.org/display/HDF5/H5G_GET_OBJINFO) in favour of H5Oget_info. Only datasets appear to get a timestamp by default, and you'd need the low-level API to enable it for groups, so it's weird to use an H5G API for datasets.

I can't see any issues asking for us to expose the timestamps on `h5py.h5o.ObjInfo`, or to expose any kind of timestamp in the high-level API. I also did a couple of searches on the HDF forum, without finding much. I get a strong impression that very few people have any use for HDF5 object timestamps.

If more people want timestamps disabled than want to use timestamps, should we disable writing them by default? It would still be possible to enable them by passing `track_times=True` if someone does want them.

takluyver:
Closes #1953 (see discussion there).

I've allowed for passing `track_times=None` explicitly, with the same behaviour as not passing it at all. I'm not sure anyone actually needs that, but it's possible some wrapper library has None as its own default, and it's easy to handle it.

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -38,7 +38,7 @@ MPI = h5.get_config().mpi
 def make_new_dset(parent, shape=None, dtype=None, data=None, name=None,
                   chunks=None, compression=None, shuffle=None,
                   fletcher32=None, maxshape=None, compression_opts=None,
-                  fillvalue=None, scaleoffset=None, track_times=None,
+                  fillvalue=None, scaleoffset=None, track_times=False,
                   external=None, track_order=None, dcpl=None,
                   allow_unknown_filter=False):
     """ Return a new low-level dataset identifier """
@@ -115,7 +115,7 @@ def make_new_dset(parent, shape=None, dtype=None, data=None, name=None,
 
     if track_times in (True, False):
         dcpl.set_obj_track_times(track_times)
-    elif track_times is not None:
+    else:
         raise TypeError("track_times must be either True or False")
     if track_order is True:
         dcpl.set_attr_creation_order(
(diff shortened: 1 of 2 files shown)


takluyver:
If there are no objections in a couple more days, I'll merge this.

tacaswell (approved):

Issue #1962, opened 2021-09-09
Feature request: consider adding `Group.__reversed__`


BvB93:
Back in python 3.8 the `__reversed__` method was added to dictionary and its various keys/values/items-views, a change made possible by the fact that dictionaries were made ordered one release earlier. I noticed that this method is lacking from `h5py.Group` though, dispite the fact that they are very much ordered.

Is there a specific reason for it's absence, and, if not, would there be any interest in adding `__reversed__`?

Looking at relevant `GroupIter` class, it should be fairly trivial to implement a reversed iterator by calling `GroupIter.names.reverse()` after populating the relevant list. Afterwards you could just iterate through the `GroupIter` instance as usual.

https://github.com/h5py/h5py/blob/8a22294e27be0678f755d66b288b2d97b258b633/h5py/h5g.pyx#L110-L132

aragilar:
The dictionary-like interfaces in h5py are set by https://docs.python.org/3/library/collections.abc.html#collections.abc.MutableMapping. It appears that `__reversed__` isn't on that class (which is why it's not defined on `Group`). I've got no objection to support being added, but given the ordering is defined by the file, I'd personally be cautious about depending on it.

takluyver:
I'm  on the fence about this. By default, groups are unordered (we iterate over them in name order, but that's not inherent to the group itself), and in this case I would generally prefer not having `__reversed__`. But you can enable tracking order when you create a group. And if we iterate in a predictable order even for unordered groups, I guess there's no actual *problem* with being able to reverse that order.

BvB93:
> By default, groups are unordered (we iterate over them in name order, but that's not inherent to the group itself)

To me this does sound like there is always a well defined iteration order, whether this is based on insertion- or alphabetical ordering. Its deterministic nature should, at the very least, allow for a sensible `__reversed__` implementation.

PR #1963, opened 2021-09-10
Add the `__reversed__` method to `Group` and its mapping views

BvB93:
closes https://github.com/h5py/h5py/issues/1962

Prompted by the recent introduction of `dict.__reversed__`, this PR adds the `__revered__` method 
to `Group`, `GroupID` and the various mapping views.
The implementation makes heavy usage of the existing `__iter__` code (see the `GroupIter` class).

### DIFF
--- a/h5py/tests/test_group.py
+++ b/h5py/tests/test_group.py
@@ -436,21 +436,29 @@ class TestTrackOrder(BaseGroup):
     def test_track_order(self):
         g = self.f.create_group('order', track_order=True)  # creation order
         self.populate(g)
-        self.assertEqual(list(g),
-                         [str(i) for i in range(100)])
+
+        ref = [str(i) for i in range(100)]
+        self.assertEqual(list(g), ref)
+        self.assertEqual(list(reversed(g)), list(reversed(ref)))
 
     def test_no_track_order(self):
         g = self.f.create_group('order', track_order=False)  # name alphanumeric
         self.populate(g)
-        self.assertEqual(list(g),
-                         sorted([str(i) for i in range(100)]))
+
+        ref = [str(i) for i in range(100)]
+        ref.sort()
+        self.assertEqual(list(g), ref)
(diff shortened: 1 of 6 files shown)


tacaswell (approved):
I think there are use-cases where this would be valuable (e.g. a group you add data sets to over time and track the order.  Being able to say "go back in time" vs "go forward in time" is useful) and very little harm in having this in other cases.

tacaswell:
I'm hoping codecov is currently failing because it is still waiting for all of the other CI jobs to report logs back.

h5py/tests/test_group.py
```diff
@@ -436,37 +436,51 @@ def populate(self, g):
...
+        self.assertEqual(list(reversed(g)), list(reversed(ref)))
 
     def test_no_track_order(self):
         g = self.f.create_group('order', track_order=False)  # name alphanumeric
         self.populate(g)
-        self.assertEqual(list(g),
-                         sorted([str(i) for i in range(100)]))
+
+        ref = [str(i) for i in range(100)]
+        ref.sort()
```
takluyver, line 449:
```suggestion
        ref = sorted([str(i) for i in range(100)])
```

`sorted()` is [documented](https://docs.python.org/3/library/functions.html#sorted) as returning a list, and I think the more functional style is clearer.

news/reversed.rst
```diff
@@ -0,0 +1,9 @@
+New features
+------------
+
+The ``__reversed__`` method has been added to the following iterables:
```
takluyver, line 4:
```suggestion
The built-in ``reversed()`` function now works with the following iterables:
```

Release notes are meant for users, so the focus is on what you can do rather than the details of special methods.

takluyver (commented):
This looks good. My suggestions are minor, so I'll commit them and merge if CI passes again.

tacaswell:
Thank you @BvB93 !  I think this was your first PR to h5py 🎉 Hopefully we will hear from you again!

PR #1964, opened 2021-09-13
Drop calls to readline, compatibility with ancient IPython in completer code

takluyver:
IPython 1.0 was released in 2013. We definitely don't need to maintain the completer code for versions older than that.

[IPython 5.0](https://ipython.readthedocs.io/en/stable/whatsnew/version5.html#ipython-5-0) moved away from readline for the terminal interface, so since then the `readline.set_completer_delims` calls have been doing nothing. I'm not sure they were right even before that (modifying the behaviour of readline globally...). Importing readline on Windows requires the pyreadline package, now unmaintained, so getting rid of this makes things a bit easier for Windows users.

### DIFF
--- a/h5py/ipy_completer.py
+++ b/h5py/ipy_completer.py
@@ -45,30 +45,10 @@ from ._hl.attrs import AttributeManager
 from ._hl.base import HLObject
 
 
-try:
-    # >=ipython-1.0
-    from IPython import get_ipython
-except ImportError:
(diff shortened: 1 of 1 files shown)


takluyver:
I already discussed this with @tacaswell and we agreed on the idea, so I'll merge it. We can always do fixes later if a problem is spotted.

Issue #1965, opened 2021-09-17
Support HDF5 library page buffering


ajelenak:
h5py currently supports HDF5 [file space management](https://portal.hdfgroup.org/display/HDF5/File+Space+Management) but not [page buffering](https://portal.hdfgroup.org/display/HDF5/Page+Buffering). The latter has the potential to improve the overall performance with carefully tailored file space page size when creating a file and the appropriately sized page buffer when reading it.

I propose to add these HDF5 functions: `H5Pset_file_space_page_size`, `H5Pget_file_space_page_size`, `H5Pset_page_buffer_size`, and `H5Pget_page_buffer_size`. All of them are available since HDF5-1.10.1, the same as the file space management functions.

If this is acceptable, I will create a PR where we can discuss high-level implementation details.

tacaswell:
Seems good to me!

Would this helped with #1960 ?

ajelenak:
Not directly. Page buffer holds "raw" bytes, prior they are turned into HDF5 objects or their data. It could help avoid reading the data from the file again if such datasets are opened later.

ajelenak:
I also want to add these two HDF5 functions: [H5Fget_page_buffering_stats](https://docs.hdfgroup.org/hdf5/v1_12/group___h5_f.html#title27) and [H5Freset_page_buffering_stats](https://docs.hdfgroup.org/hdf5/v1_12/group___h5_f.html#title35).

takluyver:
Closed by #1967

PR #1967, opened 2021-09-24
Enable HDF5 page buffering

ajelenak:
This PR implements the page buffering feature explained in #1965. The added low-level API functions are:
* page buffer size set and get;
* file space page size set and get (this was somehow missed when `fs_strategy` & friends were implemented);
* reset and get page buffer statistics.

Setting file space page or page buffer size are added to `File.__init__()` as named arguments and then passed to `make_fcpl()` and `make_fapl()`, respectively.

Get/reset page buffer statistics are not added to the high-level API because this is not something of interest to typical users. Page buffer statistics is returned as a named tuple of two named tuples. An example:

    PageBufferStats(
        meta=MetaStats(accesses=155, hits=147, misses=8, evictions=8, bypasses=0), 
        raw=RawStats(accesses=98, hits=20, misses=31, evictions=28, bypasses=47)
    )

I'll add tests once the above additions are finalized.

### DIFF
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -108,7 +108,8 @@ def registered_drivers():
     return frozenset(_drivers)
 
 
-def make_fapl(driver, libver, rdcc_nslots, rdcc_nbytes, rdcc_w0, **kwds):
+def make_fapl(driver, libver, rdcc_nslots, rdcc_nbytes, rdcc_w0, page_buf_size,
+              min_meta_keep, min_raw_keep, **kwds):
     """ Set up a file access property list """
     plist = h5p.create(h5p.FILE_ACCESS)
 
@@ -132,6 +133,10 @@ def make_fapl(driver, libver, rdcc_nslots, rdcc_nbytes, rdcc_w0, **kwds):
         cache_settings[3] = rdcc_w0
     plist.set_cache(*cache_settings)
 
+    if page_buf_size:
+        plist.set_page_buffer_size(int(page_buf_size), int(min_meta_keep),
+                                   int(min_raw_keep))
+
     if driver is None or (driver == 'windows' and sys.platform == 'win32'):
         # Prevent swallowing unused key arguments
         if kwds:
@@ -150,7 +155,8 @@ def make_fapl(driver, libver, rdcc_nslots, rdcc_nbytes, rdcc_w0, **kwds):
     return plist
 
 
-def make_fcpl(track_order=False, fs_strategy=None, fs_persist=False, fs_threshold=1):
+def make_fcpl(track_order=False, fs_strategy=None, fs_persist=False,
+              fs_threshold=1, fs_page_size=None):
     """ Set up a file creation property list """
     if track_order or fs_strategy:
         plist = h5p.create(h5p.FILE_CREATE)
@@ -171,6 +177,8 @@ def make_fcpl(track_order=False, fs_strategy=None, fs_persist=False, fs_threshol
                 raise ValueError("Invalid file space strategy type")
 
             plist.set_file_space_strategy(fs_strat_num, fs_persist, fs_threshold)
+            if fs_page_size and fs_strategy == 'page':
+                plist.set_file_space_page_size(int(fs_page_size))
     else:
         plist = None
     return plist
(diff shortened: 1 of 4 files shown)


h5py/h5f.pyx
```diff
@@ -582,3 +583,42 @@ cdef class FileID(GroupID):
...
+                unsigned int evictions[2]
+                unsigned int bypasses[2]
+
+            H5Fget_page_buffering_stats(
+                self.id, &accesses[0], &hits[0], &misses[0], &evictions[0], &bypasses[0])
+            PageBufStats = namedtuple('PageBufferStats', ['meta', 'raw'])
+            MetaStats = namedtuple(
+                'MetaStats', ['accesses', 'hits', 'misses', 'evictions', 'bypasses'])
+            RawStats = namedtuple(
+                'RawStats', ['accesses', 'hits', 'misses', 'evictions', 'bypasses'])
```
aragilar, line 618: Can these be moved to the top-level scope? It would be cleaner and avoid defining them on each call.

ajelenak: Sure, done in #d6a868f.

h5py/h5f.pyx
```diff
@@ -81,6 +82,14 @@ IF HDF5_VERSION >= (1, 10, 1):
     FSPACE_STRATEGY_AGGR = H5F_FSPACE_STRATEGY_AGGR
     FSPACE_STRATEGY_NONE = H5F_FSPACE_STRATEGY_NONE
 
+    # Used in FileID.get_page_buffering_stats()
+    PageBufStats = namedtuple('PageBufferStats', ['meta', 'raw'])
+    MetaStats = namedtuple(
+        'MetaStats', ['accesses', 'hits', 'misses', 'evictions', 'bypasses'])
+    RawStats = namedtuple(
+        'RawStats', ['accesses', 'hits', 'misses', 'evictions', 'bypasses'])
```
takluyver, line 90: I think we only need one type here, as they've got the same fields, right?

h5py/_hl/files.py
```diff
@@ -396,6 +405,9 @@ def __init__(self, name, mode='r', driver=None,
             "aggregate"  Aggregators, VFD
             "none"       VFD
             If None use HDF5 defaults.
+        fs_page_size
+            File space page size in bytes. Only used when fs_strategy="page". If
+            None use HDF5 defaults.
```
takluyver, line 410:
```suggestion
            None use HDF5 defaults (4096 bytes).
```

If you need to ensure `page_buf_size` is at least this big, what the default is is fairly important.

h5py/_hl/files.py
```diff
@@ -404,12 +416,24 @@ def __init__(self, name, mode='r', driver=None,
             The smallest free-space section size that the free space manager
             will track.  Only allowed when creating a new file.  The default
             value is 1.
+        page_buf_size
+            Page buffer size in bytes. Only allowed in "r" or "r+" modes and for
+            HDF5 files created with fs_strategy="page". Must be a power of two
+            value and greater or equal than the file space page size when
+            creating the file. It is not used by default.
```
takluyver, line 423: This seems like fairly specific conditions. If you use a value smaller than the file space page size, or the file wasn't created with `fs_strategy="page"`, does HDF5 throw an error, or does it just quietly not use the page buffer?

takluyver: (Just found in the HDF5 docs that it will error)

h5py/_hl/files.py
```diff
@@ -443,6 +467,9 @@ def __init__(self, name, mode='r', driver=None,
             if fs_strategy and mode not in ('w', 'w-', 'x'):
                 raise ValueError("Unable to set file space strategy of an existing file")
 
+            if page_buf_size and mode not in ('r', 'r+'):
```
takluyver, line 470: Does it really make sense to only set this for these modes? It sounds from the HDF5 docs like the page buffer also applies to writing, and you can read data in any mode.

takluyver:
I think it would be good to have at least a brief test of creating a suitable file, and reopening it with the page buffer enabled.

Don't forget to [write a release note](https://docs.h5py.org/en/stable/contributing.html#write-a-release-note).

h5py/_hl/files.py
takluyver: We might also want to document that 512 is the minimum.

h5py/h5f.pyx
ajelenak: Replaced `MetaStats` and `RawStats` with one `PageStats`.

h5py/_hl/files.py
ajelenak: Add the default value.

h5py/_hl/files.py
ajelenak: Correct, I was given misleading info. Removed this check.

takluyver:
Thanks!

PR #1976, opened 2021-10-15
FIX: memory leak in fast-reader Reader.read

### LINKED ISSUE #1975: Memory leak when reading from an hdf5 dataset
OS
NAME="Ubuntu"
VERSION="20.04.3 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.3 LTS"
VERSION_ID="20.04"

h5py    3.4.0
HDF5    1.12.1
Python  3.8.9 (default, Oct 11 2021, 09:08:45) 
[GCC 9.3.0]
sys.platform    linux
numpy   1.21.2
cython (built with) 0.29.24
numpy (built against) 1.17.5
HDF5 (built against) 1.12.1

Code to reproduce the issue (memory is leaking, please track the currently used RAM):

```python
import h5py 
import numpy as np 
 
with h5py.File("leak.h5py", "w") as f: 
    f["dataset"] = np.random.rand(1000, 1000) 

with h5py.File("leak.h5py", "r") as f: 
    while True: 
        f["dataset"][0]

```

Am I doing something wrong? Perhaps, reading the data?

Will appreciate any help!

tacaswell:
Previously we were creating but never destroying memory space objects.  This
changes the code to reuse the `SpaceID` class which will take care of making
the c level lifeccyle track the Python level life cycle.

This bug was exposed by fixing a different bug in #1944 that prevented this
code from ever being run.

Closes #1975

@takluyver @aragilar I think we need to get this in and do a 3.4.1 or pull the fast reader code out and do a 3.4.1

### DIFF
--- a/h5py/_selector.pyx
+++ b/h5py/_selector.pyx
@@ -359,11 +359,13 @@ cdef class Reader:
             buf = PyArray_DATA(arr)
 
             mspace = H5Screate_simple(self.selector.rank, mshape, NULL)
-            H5Dread(self.dataset, self.h5_memory_datatype.id, mspace,
-                    self.selector.space, H5P_DEFAULT, buf)
-
         finally:
             efree(mshape)
+
+        try:
+            H5Dread(self.dataset, self.h5_memory_datatype.id, mspace,
+                    self.selector.space, H5P_DEFAULT, buf)
+        finally:
             H5Sclose(mspace)
 
         if arr.ndim == 0:


ajelenak:
Why not just `H5Sclose(mspace)`? `mspace` is local to that method.

tacaswell:
@ajelenak no reason other than not knowing better, using `H5Sclose` is even marginally faster!

```
Elapsed 2.80 at 357628.596 read/sec  # third commit
```

h5py/_selector.pyx
```diff
@@ -359,11 +359,12 @@ cdef class Reader:
...
             mspace = H5Screate_simple(self.selector.rank, mshape, NULL)
+            H5Dread(self.dataset, self.h5_memory_datatype.id, mspace,
+                    self.selector.space, H5P_DEFAULT, buf)
+
         finally:
             efree(mshape)
-
-        H5Dread(self.dataset, self.h5_memory_datatype.id, mspace,
-                self.selector.space, H5P_DEFAULT, buf)
+            H5Sclose(mspace)
```
takluyver, line 367: If there's an exception before we create mspace, I think this will fail, because it will get something that's not a space ID. I think we should either have another try/finally wrapped around `H5Dread`, or initialise it to 0, and only call H5Sclose if it's non-zero.

takluyver:
Thanks for taking the time to test the performance! I didn't expect it to make that big a difference.

I'm going to bed shortly, but I can probably pick this up tomorrow and prepare a 3.4.1 release if we're all happy.

takluyver:
I've pushed a fix for my last concern. This doesn't seem to impact timing (@tacaswell's script gives slightly over 6 seconds either way with my laptop, and the variation from re-running it is bigger than any difference before & after my commit.

I'll wait for one of you to have a look at this before I merge it and do a new release.

takluyver:
@tacaswell @aragilar I was waiting for one of you to have a quick look before I merge this and prepare a release. :slightly_smiling_face:

Issue #1984, opened 2021-10-27
pytest-mpi required for non-mpi HDF5 ?


mboisson:
I am building h5py on top of a serial version of HDF5. I don't expect it to use or work MPI. Yet, pytest-mpi seems required: 

```
```11152 Loading library to get build settings and version: /home/mboisson/.local/easybuild/software/2020/avx2/Compiler/gcc9/hdf5/1.12.1/lib/libhdf5.so
11153 ********************************************************************************
11154                        Summary of the h5py configuration
11155
11156 HDF5 include dirs: [
11157   '/home/mboisson/.local/easybuild/software/2020/avx2/Compiler/gcc9/hdf5/1.12.1/include'
11158 ]
11159 HDF5 library dirs: [
11160   '/home/mboisson/.local/easybuild/software/2020/avx2/Compiler/gcc9/hdf5/1.12.1/lib'
11161 ]
11162      HDF5 Version: (1, 12, 1)
====>>>>. 11163       MPI Enabled: False
11164  ROS3 VFD Enabled: False
11165  Rebuild Required: False
11166
11167 ********************************************************************************
11168 Executing cythonize()
=====>>>>11169 ERROR: Missing required plugins: pytest-mpi``` 
``` 

Why is that ?

takluyver:
Closing this again; we've got rid of the special support for `setup.py test` entirely. We recommend using `tox` if you want to build h5py & test it in one go, or `pytest --pyargs h5py` if you want to run the tests after installing it.

PR #1986, opened 2021-10-29
Remove setup.py test

aragilar:
#1984 highlighted that users still try to run setup.py test, this removes our custom test command.

A possible alternative would be to throw an error instead (I'm not sure if there are systems which try to detect `setup.py test` and erroring would cause those systems to break).


takluyver (approved):
I didn't realise we had this custom command there...

PR #1973, opened 2021-10-07
Set up CI for Python 3.10

takluyver:
This probably isn't going to work straight away, but I thought I'd get started on this before the inevitable bug reports roll in.

PR #1987, opened 2021-10-29
Build Python 3.10 wheels for Mac

takluyver:
This was removed from PR #1973 because it wasn't working. Unless it mysteriously starts working again, I'm probably going to leave it to Mac users to figure out, because I don't know how to investigate.


henryiii:
This should be fixed as of NumPy 1.21.4.

takluyver:
That does seem to work; thanks!

henryiii:
It's been updated in oldest-supported-numpy too. ;)

PR #1988, opened 2021-11-01
DOC: Update details on unicode handling for filenames

takluyver:
Various details were out of date, I've updated them and tried to clarify some bits.

cc @aragilar - I think you wrote this section a few years back. :slightly_smiling_face:

### DIFF
--- a/docs/high/file.rst
+++ b/docs/high/file.rst
@@ -252,44 +252,46 @@ Filenames on different systems
 
 Different operating systems (and different file systems) store filenames with
 different encodings. Additionally, in Python there are at least two different
-representations of filenames, as encoded bytes (via str on Python 2, bytes on
-Python 3) or as a unicode string (via unicode on Python 2 and str on Python 3).
-The safest bet when creating a new file is to use unicode strings on all
-systems.
+representations of filenames, as encoded ``bytes`` or as a Unicode string
+(``str`` on Python 3).
+
+h5py's high-level interfaces always return filenames as ``str``, e.g.
+:attr:`File.filename`. h5py accepts filenames as either ``str`` or ``bytes``.
+In most cases, using Unicode (``str``) paths is preferred, but there are some
+caveats.
+
+.. note::
+
+   HDF5 handles filenames as bytes (C ``char *``), and the h5py :doc:`lowlevel`
+   matches this.
 
 macOS (OSX)
 ...........
 macOS is the simplest system to deal with, it only accepts UTF-8, so using
-unicode paths will just work (and should be preferred).
+Unicode paths will just work (and should be preferred).
 
 Linux (and non-macOS Unix)
 ..........................
-Unix-like systems use locale settings to determine the correct encoding to use.
-These are set via a number of different environment variables, of which ``LANG``
-and ``LC_ALL`` are the ones of most interest. Of special interest is the ``C``
-locale, which Python will interpret as only allowing ASCII, meaning unicode
-paths should be pre-encoded. This will likely change in Python 3.7 with
-https://www.python.org/dev/peps/pep-0538/, but this will likely be backported by
-distributions to earlier versions.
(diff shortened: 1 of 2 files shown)


docs/high/file.rst
```diff
@@ -252,44 +252,46 @@ Filenames on different systems
...
+.. note::
+
+   HDF5 handles filenames as bytes (C ``char *``), and the h5py :doc:`lowlevel`
+   matches this.
 
 macOS (OSX)
 ...........
 macOS is the simplest system to deal with, it only accepts UTF-8, so using
-unicode paths will just work (and should be preferred).
+Unicode paths will just work (and should be preferred).
```
takluyver, line 271: I suspect this 'just works' only holds for Apple's own filesystems - if Macs can mount an NFS share or an ext4 hard drive, then sooner or later you'll have to contend with filenames that aren't valid UTF-8. But I suspect that's not very common, and I don't know exactly what the right answer is in that case (same as for other Unix-y systems?), so I've left this for now.

docs/high/file.rst
```diff
@@ -252,44 +252,46 @@ Filenames on different systems
...
+encoding is used to convert to and from unicode; on most modern systems this
+will be UTF-8 by default (especially since Python 3.7, with :pep:`538`).
 
-To summarise, use unicode strings where possible, but be aware that sometimes
-using encoded bytes may be necessary to read incorrectly encoded filenames.
+Passing Unicode paths will mostly work, and Unicode paths from system
+functions like ``os.listdir()`` should always work. But if there are filenames
+that aren't in the expected encoding (e.g. on a network filesystem or a
+removable drive, or because something is misconfigured), you may want to handle
+them as bytes.
```
takluyver, line 283: I reworded this line a few times. You never* *need* to use `bytes`, because you can represent any bytes as surrogateescape in Unicode strings, and `os.fsencode()` will do the right thing. But I don't want to get into explaining that here, so I've gone with 'may want to'.

(* OK, almost never. If your locale encoding is not ASCII compatible, it doesn't hold. But that's extremely weird, so anyone affected probably either knows how to work around it or is having a very bad day trying to use Python 3 at all.)

docs/high/group.rst
```diff
@@ -158,11 +158,11 @@ link resides.
...
 .. note::
 
-    How the filename is processed is operating system dependent, it is
-    recommended to read :ref:`file_filenames` to understand potential limitations on
-    filenames on your operating system. Note especially that Windows is
-    particularly susceptible to problems with external links, due to possible
-    encoding errors and how filenames are structured.
+    The filename is stored in the file as bytes, normally UTF-8 encoded.
+    In most cases, this should work reliably, but problems are possible if a
+    file created on one platform is accessed on another. Older versions of HDF5
```
takluyver, line 163: I think cross-platform issues are the only way this could show up now - e.g. if you make an external link on Linux using a non-UTF-8 locale, then copy the files to a Windows machine in a way that correctly converts the filenames to Unicode, HDF5 will try to interpret the name from the link as UTF-8.

docs/high/file.rst
aragilar: I *think* there's some level of enforcement of this within Python's stdlib on MacOS, or at least locales play less of a role. The more Unix-y setups (with NFS/ext4 etc.) I'm not sure about.

aragilar:
This looks good to me (it's nice to see that all the improvements we've been watching have now been merged for a while). I suspect any remaining unicode issues will also be encountered without the use of h5py (or HDF5), and users will need to work out how best to operate python in that environment (likely by talking with their sysadmins).

I'll merge this tomorrow (in case anyone else wants to comment), unless someone else wants to merge this.

takluyver:
Thanks James! I think this should be uncontroversial, so I'll merge it now. We can always refine it in other PRs. :slightly_smiling_face:

PR #1990, opened 2021-11-02
Deprecation warning on getting config.default_file_mode

takluyver:
I just noticed that while we deprecated setting this attribute several releases ago, there was no warning on getting it. I'd like to get rid of it entirely, so I've added a warning for getting it as well. There's no obvious reason you'd need to get it, but that doesn't mean no-one does...

### DIFF
--- a/h5py/h5.pyx
+++ b/h5py/h5.pyx
@@ -67,7 +67,6 @@ cdef class H5PYConfig:
         self._t_name = b'TRUE'
         self._bytestrings = ByteStringContext()
         self._track_order = False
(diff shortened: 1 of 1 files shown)


aragilar:
Nice catch @takluyver. I'd guess we'd remove the default in 12 months (3.8 maybe?).

takluyver:
I might have a slightly shorter deprecation for this, because we were already warning about setting it, and I imagine there's almost no code that gets it without setting it. But there's also not much pressure to remove it - it's just a few lines of unused code.

PR #1941, opened 2021-08-12
Remove rever authors activity

takluyver:
Keeping this working has proven to be a headache for new contributors.
Git log and Github both show who has contributed commits to h5py.

Closes gh-1940

PR #1991, opened 2021-11-02
Remove section about authors list from contributor docs

takluyver:
We removed the authors list in #1941, but the contributor docs still tell people to add their names.


aragilar:
Looks like the appveyor-github connection failed once, so that can be ignored. Merging.

Issue #1981, opened 2021-10-21
`pip install h5py` tries to install an old version of numpy


dstansby:
In a clean conda environment, running the commands below fails, because `h5py` tries to install numpy 1.19.3 which is not compatible with python 3.10. I'm not sure why this is the case - this block implies that `h5py` should on enforce a minimum version, instead of installing an exact version of numpy https://github.com/h5py/h5py/blob/fb9989a700e60cd1c1dec20e329218d449e039cf/setup.py#L28-L39

```python
conda install python=3.10
pip install numpy # Installs numpy==1.21.3
pip install h5py
```
Output of final command:
```python
$ pip install h5py --verbose
Using pip 21.2.4 from /Users/dstansby/miniconda3/envs/py310/lib/python3.10/site-packages/pip (python 3.10)
Collecting h5py
  Using cached h5py-3.5.0.tar.gz (384 kB)
  Running command /Users/dstansby/miniconda3/envs/py310/bin/python /private/var/folders/3j/f7vtq4q1397bxjn19wftn_fm0000gn/T/pip-standalone-pip-nestuxec/__env_pip__.zip/pip install --ignore-installed --no-user --prefix /private/var/folders/3j/f7vtq4q1397bxjn19wftn_fm0000gn/T/pip-build-env-3u2eo2pe/overlay --no-warn-script-location --no-binary :none: --only-binary :none: -i https://pypi.org/simple -- setuptools wheel
  Collecting setuptools
    Using cached setuptools-58.2.0-py3-none-any.whl (946 kB)
  Collecting wheel
    Using cached wheel-0.37.0-py2.py3-none-any.whl (35 kB)
  Installing collected packages: wheel, setuptools
  Successfully installed setuptools-58.2.0 wheel-0.37.0
  Installing build dependencies ... done
  Running command /Users/dstansby/miniconda3/envs/py310/bin/python /Users/dstansby/miniconda3/envs/py310/lib/python3.10/site-packages/pip/_vendor/pep517/in_process/_in_process.py get_requires_for_build_wheel /var/folders/3j/f7vtq4q1397bxjn19wftn_fm0000gn/T/tmp9i7s2cwc
  Getting requirements to build wheel ... done
[...]

PR #1995, opened 2021-11-08
Change the version control on setup

jackwangshanghai:
Feels like the Version control has some problem on setup. There is a mismatch with the run requirement.


neutrinoceros:
This is intentional. Build-time dependencies are pinned to oldest possible version to maximise forward/backward compatibility. This is because code compiled with an old version of numpy is guaranteed to be compatible with more recent versions used at runtime, but not the other way around.
See for instance the discussion in #1981

takluyver:
@neutrinoceros covered it; thanks!

PR #2003, opened 2021-11-16
Explicit configuration for Readthedocs

takluyver:
I just noticed RTD builds have been failing (old version of Sphinx was being used alongside a new docutils. This should fix it, by upgrading to the latest Sphinx.

I've also enabled pull request integration on RTD, so we should get a status for this.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,13 @@
+version: 2  # Required
+
+build:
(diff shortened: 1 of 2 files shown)


takluyver:
Merging so the 3.6 release notes can be built & published. :slightly_smiling_face:

PR #2004, opened 2021-11-17
Refine how aarch64 wheels are named for upload to anaconda.org

takluyver:
When pushing a tag, upload a wheel without inserting a build tag in the name; we can collect these and conveniently re-upload them to PyPI.

When adding a build tag for wheels built by cron jobs, use the git commit hash instead of a timestamp. This should avoid uploading a duplicate file if there are no changes since the last one - the upload will fail because the file already exists. On the other hand, it means the build tags are no longer ordered. We may be able to do something smarter with `git describe` if that's a concern.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,4 +70,4 @@ after_success:
   # ANACONDA_ORG_UPLOAD_TOKEN is a secret token
   # used in Travis CI config, originally
   # generated at anaconda.org for scipy-wheels-nightly (cron jobs).
-    - if [[ $CIBW_ARCHS == aarch64 && "$TRAVIS_EVENT_TYPE" == "cron" ]]; then bash ci/upload_wheels.sh; fi
+    - if [[ $CIBW_ARCHS == aarch64 ]]; then bash ci/upload_wheels.sh; fi
--- a/ci/upload_wheels.sh
+++ b/ci/upload_wheels.sh
@@ -7,12 +7,14 @@ if [[ "$TRAVIS_EVENT_TYPE" != "cron" && -z "$TRAVIS_TAG" ]] ; then
 fi
 
 # rename wheels if not on a tag
-# inserting short commit hash as build tag in wheel filename
+# inserting commit count & hash from git describe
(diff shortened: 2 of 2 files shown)


takluyver:
Thinking about it some more, I think it would be valuable to preserve the sort order.

aragilar:
This looks good to me. Did we want to start using something like setuptools-scm or versioneer, which would give a more detailed version on every dev build?

takluyver:
Thanks!

I tend to dislike the approach of things like setuptools_scm, because in some scenarios they end up doing things like running subprocesses or scanning metadata on import, which seems like a bad idea.

As h5py needs a build step anyway, though, maybe there's a nice way to stamp a dynamically generated version number into it then. But adding that to the existing build scripts doesn't sound much fun. I'm already thinking of nicking the meson machinery from scipy when they've got it working nicely, so maybe then...

aragilar:
Ah, good point, I hadn't really dug into how they work (I can see subprocesses failing in odd ways leading to weird import errors being a pain to debug, and guessing versions from metadata is...), I've been using versioneer which bakes in the VCS variables from git, so at least standard installs are static strings (and I've kinda gone off develop mode, since half the time it doesn't work with the code I'm working on...). I'm hoping the meson stuff succeeds, it'd be nice to not have to faff around with where HDF5 is installed, and which MPI to use...

takluyver:
Seems to be working as I hoped. The latest [Travis build log](https://app.travis-ci.com/github/h5py/h5py/jobs/549698917) says:

```
Uploading file "scipy-wheels-nightly/h5py/3.6.0/h5py-3.6.0-5_g050c8fc-cp37-cp37m-manylinux_2_17_aarch64.manylinux2014_aarch64.whl"
Distribution already exists. Please use the -i/--interactive or --force or --skip options or `anaconda remove scipy-wheels-nightly/h5py/3.6.0/h5py-3.6.0-5_g050c8fc-cp37-cp37m-manylinux_2_17_aarch64.manylinux2014_aarch64.whl
```

And this doesn't fail the build. :slightly_smiling_face:

PR #2014, opened 2021-12-10
Use pre-commit.ci for pre-commit CI

takluyver:
https://pre-commit.ci/ is a CI service specific to `pre-commit`. It's meant to be faster than running pre-commit in regular CI (because it can do shared caching), and it can also commit changes directly rather than failing until contributors make them.

I've enabled it for this repo, so it should run on this PR.

PR #2016, opened 2021-12-10
Build HDF5 with Visual Studio 2019 on Azure CI

### LINKED ISSUE #2015: Building HDF5 broken on Azure Windows CI, Windows 2019 image
I noticed on #2014 that all the Windows builds failed to build HDF5 itself, with this error:

```
CMake Error at CMakeLists.txt:2 (project):
  Generator

    Visual Studio 15 2017 Win64

  could not find any instance of Visual Studio.
```

I think this is the first time it has tried to rebuild HDF5 since #1985 - it has been retrieving it from cache, and hiding the problem.

I think the problem is that we're specifying to build with Visual Studio 2017 (`HDF5_VSVERSION: "15-64"`), but the `windows-2019` image has Visual Studio 2019.

VS 2017 [is documented as](https://pythondev.readthedocs.io/windows.html#python-and-visual-studio-version-matrix) the version used to build Python releases on python.org, and the recommended version for building Python yourself. This note [in the packaging guide](https://packaging.python.org/en/latest/guides/packaging-binary-extensions/#binary-extensions-for-windows) also hints at sticking with 2017:

> One caveat: if you use Visual Studio 2019 or later, your extension will depend on an “extra” file, VCRUNTIME140_1.dll, in addition to the VCRUNTIME140.dll that all previous versions back to 2015 depend on. This will add an extra requirement to using your extension on versions of CPython that do not include this extra file. To avoid this, you can add the compile-time argument `/d2FH4-`. Recent versions of Python may include this file.

The 'embeddable' 64-bit Python builds available on python.org seem to contain that `_1.dll` file since round about 3.8.3 (May 2020). IDK if the Python.org installers have it, or Anaconda, or any other ways people might get Python on Windows. :shrug: 

We can't really go back to Windows 2016, because that [will be removed in March](https://github.com/actions/virtual-environments/issues/4312). It's [possible to install VS 2017](https://github.com/actions/virtual-environments/issues/4312#issuecomment-987320937) on the newer image, but it apparently takes about 10 minutes to do so, so that's really a last resort unless there's some way to cache it. :disappointed:

takluyver:
Closes #2015 :crossed_fingers:

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -42,7 +42,7 @@ jobs:
   variables:
     HDF5_VERSION: 1.12.1
     HDF5_DIR: $(HDF5_CACHE_DIR)/$(HDF5_VERSION)
-    HDF5_VSVERSION: "15-64"
+    HDF5_VSVERSION: "16-64"
     CIBW_BUILD: cp3{7,8,9,10}-win_amd64
   steps:
     - template: ci/azure-pipelines-wheels.yml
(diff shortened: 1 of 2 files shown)


aragilar:
Look like everything worked (I had a quick look at the logs, and I don't see anything concerning)!

@takluyver If you want to rebase #2014 and merge it, that's cool with me.

PR #2024, opened 2021-12-23
MAINT: use super() as described by PEP 3135

mwtoews:
This PR refactors `super()` to be simpler, as described by [PEP 3135](https://www.python.org/dev/peps/pep-3135/) and approved in 2007 for Python 3.0.

The "old" convention (e.g. `super(C, self)`) was only required to support Python 2.

### DIFF
--- a/examples/swmr_multiprocess.py
+++ b/examples/swmr_multiprocess.py
@@ -24,7 +24,7 @@ from multiprocessing import Process, Event
 
 class SwmrReader(Process):
     def __init__(self, event, fname, dsetname, timeout = 2.0):
-        super(SwmrReader, self).__init__()
+        super().__init__()
         self._event = event
         self._fname = fname
         self._dsetname = dsetname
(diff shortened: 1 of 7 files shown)


tacaswell (approved):

tacaswell:
Thanks!  I suspect that there is a whole lot of py2 related code still in the code base that we have never put in the effort to strip out.

To some degree I think there is a question if we _want_ to strip it out (it is working why risk putting bugs in?), but I think in this case it is clearly a) safe b) makes the code simpler

takluyver:
Close/reopen to re-run the CI with a fix in master.

To echo @tacaswell, I think this is fine, but I'm wary of large scale 'cleanup' in general - the benefits are pretty minimal, it can introduce subtle bugs, and it can cause unnecessary merge conflicts, making more work for other people. If it ain't broke, sometimes it's best just to leave it.

PR #2031, opened 2022-01-07
Try to avoid NumPy 1.21.5 on ppc64le build

takluyver:
Hopefully fix Travis CI builds.


takluyver:
That seemed to work. Merging so CI passes on master again.

PR #2032, opened 2022-01-07
Don't build HDF5 tests in CI

takluyver:
This will hopefully speed up CI slightly when it builds HDF5 from source (rather than using the cache).


takluyver:
Some nice speedups, especially on Windows (times for the 'Ensure HDF5' step in Azure CI):

image

But it seems that both Ubuntu test jobs for HDF5 1.12 failed with a segfault in test_lzf. I don't understand how disabling tests would affect that, but two jobs failing on the same line can't be coincidence.

https://github.com/h5py/h5py/blob/e5c14919440dac857ed7e415ad0f08306445b3bd/h5py/tests/test_dataset.py#L595

PR #1240, opened 2019-06-17
Fix non-linux build.

hameerabbasi:
Fixes #1237.

PR #2038, opened 2022-01-18
Fix dll path on Cygwin.

DWesl:
Cygwin uses [a `cyg${libname}-${SOVERSION}.dll` naming scheme for shared libraries](https://cygwin.com/cgi-bin2/package-cat.cgi?file=x86_64%2Flibhdf5_200%2Flibhdf5_200-1.12.1-1&grep=libhdf5), rather than the lib${libname}.so.${SOVERSION} of linux or the ${libname}.dll that is apparently used on Windows.  This patch assumes the shared library name is used by `dlopen` or similar, because [the import library used for normal linking uses a `lib${libname}.dll.a` naming scheme](https://cygwin.com/cgi-bin2/package-cat.cgi?file=x86_64%2Flibhdf5-devel%2Flibhdf5-devel-1.12.1-1&grep=libhdf5).

Fixes #1240 for use on Cygwin.

### DIFF
--- a/setup_configure.py
+++ b/setup_configure.py
@@ -200,14 +200,16 @@ class HDF5LibWrapper:
         if sys.platform.startswith('darwin'):
             default_path = 'libhdf5.dylib'
             regexp = re.compile(r'^libhdf5.dylib')
-        elif sys.platform.startswith('win') or \
-            sys.platform.startswith('cygwin'):
+        elif sys.platform.startswith('win'):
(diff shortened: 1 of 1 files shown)


takluyver:
Close/reopen to re-run tests.

I don't really know anything about Cygwin, but it sounds like you do, and this shouldn't affect anything other than cygwin, so I'll just merge this (assuming the CI passes now).

DWesl:
Thank you

Issue #2034, opened 2022-01-08
Setting data alignment at File creation


hmaarrfk:
I'm hoping to add data alignment strategies to certain files. This can be achieved by adding parameters to `make_fapl` in:

https://github.com/h5py/h5py/blob/master/h5py/_hl/files.py#L111

Using `set_alignment`. We would need to add 2 parameters, the threshold, and the alignment interval.

One would need to add the lines
```
def make_fapl(...., threshold=1, alignment_interval=1):
    fapl.set_alignment(threshold, alignment_interval)
```
and provide additional parameters to the `File` constructor.

Would this be an acceptable PR to make?

Thank you!

References: https://support.hdfgroup.org/HDF5/doc/UG/FmSource/08_TheFile_favicon_test.html

To assist reproducing bugs, please include the following:
 * Operating System (e.g. Windows 10, MacOS 10.11, Ubuntu 16.04.2 LTS, CentOS 7): Linux
 * Python version (e.g. 2.7, 3.5): 3.9
 * Where Python was acquired (e.g. system Python on MacOS or Linux, Anaconda on
   Windows): conda-forge Linux
 * h5py version (e.g. 2.6): 3.6.0
 * HDF5 version (e.g. 1.8.17): 1.12.1

takluyver:
That seems reasonable enough to me - we've already gone for lots of keyword parameters for `File()`. @tacaswell, are you OK with adding a couple more?

I'd probably call them `alignment_threshold` (to make it clear what threshold it is) and `alignment` (following the name HDF5 uses).

hmaarrfk:
There might be some other configuration that would be desirable for performance reasons, cache size, and whatnot.

Do you wan tto make the fapl and fack overwrite able instead? not very high level, but gives the most control

tacaswell:
So long as they are un-coupled (that is the user can correctly set just a sub-set) I'm :+1: on adding as many extra keyword-only arguements to `File` as we need to expose everything in facl.

If there are a bunch of the inputs that _only_ make sense when the file is created (rather than just opened), what do people think of adding some `classmethod` constructors instead of adding to the `__init__` methods?  It might require a bit of jiggery-pokery, but I think we could make it work without having to close and then re-open the file...

hmaarrfk:
@tacaswell great.

This is an access parameter, so I believe it is valid both at creation time and at open time.

PR #2040, opened 2022-01-18
Support setting the data alignment within the HDF5 file

hmaarrfk:
Closes: https://github.com/h5py/h5py/issues/2034

Questions I have:
- [ ] May I simply cross-reference the HDF5 documentation? I feel like rewording is more dangerous than simply pointing the user to the right place.

TODO: add tests
- [x] Create a test that without alignment will create a datasets, then check their id offsets to ensure that at least one dataset is misaligned.
- [x] Set the property list, and then create 10 datasets, ensure they are all aligned if they are above the threshold.
- [x] Create datasets that are under the threshold, ensure at least one of them is misaligned.

The reason for the "at least one" is that one may get lucky with the alignment, though it is very improbable they they get lucky many times in a row.

Release note sample:

> New parameters are provided to the ``File`` constructor to control data alignment via the file access property lists. The new parameters, `alignment_threshold` and `alignment_interval` are passed to ``H5Pset_alignment`` during file creation and opening.

### DIFF
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -109,7 +109,8 @@ def registered_drivers():
 
 
 def make_fapl(driver, libver, rdcc_nslots, rdcc_nbytes, rdcc_w0, locking,
-              page_buf_size, min_meta_keep, min_raw_keep, **kwds):
+              page_buf_size, min_meta_keep, min_raw_keep,
+              alignment_threshold, alignment_interval, **kwds):
     """ Set up a file access property list """
     plist = h5p.create(h5p.FILE_ACCESS)
 
@@ -123,6 +124,7 @@ def make_fapl(driver, libver, rdcc_nslots, rdcc_nbytes, rdcc_w0, locking,
         # we default to earliest
         low, high = h5f.LIBVER_EARLIEST, h5f.LIBVER_LATEST
     plist.set_libver_bounds(low, high)
+    plist.set_alignment(alignment_threshold, alignment_interval)
 
     cache_settings = list(plist.get_cache())
     if rdcc_nslots is not None:
@@ -354,7 +356,8 @@ class File(Group):
     def __init__(self, name, mode='r', driver=None, libver=None, userblock_size=None, swmr=False,
                  rdcc_nslots=None, rdcc_nbytes=None, rdcc_w0=None, track_order=None,
                  fs_strategy=None, fs_persist=False, fs_threshold=1, fs_page_size=None,
-                 page_buf_size=None, min_meta_keep=0, min_raw_keep=0, locking=None, **kwds):
+                 page_buf_size=None, min_meta_keep=0, min_raw_keep=0, locking=None,
+                 alignment_threshold=1, alignment_interval=1, **kwds):
         """Create a new file object.
 
         See the h5py user guide for a detailed explanation of the options.
@@ -449,6 +452,19 @@ class File(Group):
             Warning: The HDF5_USE_FILE_LOCKING environment variable can override
             this parameter.
             Only available with HDF5 >= 1.12.1 or 1.10.x >= 1.10.7.
+
+        alignment_threshold
+            Together with ``alignment_interval``, this property ensures that
+            any file object greater than or equal in size to the alignement
+            threshold (in bytes) will be aligned on an address which is a
+            multiple of alignment interval.
+
+        alignment_interval
+            This property should be used in conjunction with
+            ``alignment_threshold``. See the description above. For more
+            details, see
+            https://portal.hdfgroup.org/display/HDF5/H5P_SET_ALIGNMENT
+
         Additional keywords
             Passed on to the selected file driver.
         """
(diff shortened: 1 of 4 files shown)


hmaarrfk:
I don't think the static errors are due to this PR
```
Warning, treated as error:
/home/vsts/work/1/s/docs/whatsnew/2.5.rst:19:hardcoded link '***/pull/551' could be replaced by an extlink (try using ':pr:`551`' instead)
ERROR: InvocationError for command /home/vsts/work/1/s/.tox/docs/bin/sphinx-build -W -b html -d /home/vsts/work/1/s/.tox/docs/tmp/doctrees . /home/vsts/work/1/s/.tox/docs/tmp/html (exited with code 2)
checkreadme create: /home/vsts/work/1/s/.tox/checkreadme

```

takluyver:
> May I simply cross-reference the HDF5 documentation? I feel like rewording is more dangerous than simply pointing the user to the right place.

I'd like the docstring to have a brief summary of what the parameters mean, so people can get a sense of it without having to copy and paste a link to some external site. Something like:

> alignment_threshold: Objects this size or larger will be placed at multiples of alignment_interval in the file. Mainly useful for MPI.

It should also be added to the [reference docs](https://docs.h5py.org/en/stable/high/file.html#reference) in `docs/high/file.rst`. Here the summary can be accompanied with a link to the HDF5 docs. I think this is the right page now: https://portal.hdfgroup.org/display/HDF5/H5P_SET_ALIGNMENT

([This landing page](https://support.hdfgroup.org/HDF5/doc/RM/RM_H5Front.html) says that the support.hdfgroup.org version of the reference manual is no longer maintained).

> I don't think the static errors are due to this PR

Nope, should be fixed now (#2049).

h5py/_hl/files.py
```diff
@@ -449,6 +452,19 @@ def __init__(self, name, mode='r', driver=None, libver=None, userblock_size=None
...
+            Together with ``alignment_interval``, this property ensures that
+            any file object greater than or equal in size to the alignement
+            threshold (in bytes) will be aligned on an address which is a
+            multiple of alignment interval.
+
+        alignment_interval
+            This property should be used in conjunction with
+            ``alignment_threshold``. See the description above. For more
+            details, see
+            https://support.hdfgroup.org/HDF5/doc/RM/RM_H5P.html#Property-SetAlignment
```
takluyver, line 466:
```suggestion
            https://portal.hdfgroup.org/display/HDF5/H5P_SET_ALIGNMENT
```

The [support.hdfgroup.org](https://support.hdfgroup.org/) pages are no longer maintained.

h5py/tests/test_file_alignment.py
```diff
@@ -0,0 +1,106 @@
...
+                for i in range(1000):
+                    dataset = h5file.create_dataset(
+                        dataset_name(i), shape, dtype='uint8')
+                    # Assign data so that the dataset is instantiated in
+                    # the file
+                    dataset[...] = i
+                    if not is_aligned(dataset, offset=alignment_interval):
+                        raise RuntimeError(
+                            "All data was expected to be aligned to "
+                            f"{alignment_interval}.")
```
takluyver, line 72:
```suggestion
                    assert is_aligned(dataset, offset=alignment_interval):
```

Doesn't really matter, but it keeps the test code shorter, and it's nice to be able to scan through tests for `assert` to see what is being checked.

takluyver (commented):
Thanks! Don't forget to add the parameters to the reference docs (`docs/high/file.rst`) as well.

h5py/tests/test_file_alignment.py
```diff
@@ -0,0 +1,103 @@
...
+                           alignment_interval=alignment_interval) as h5file:
+                # Create up to 1000 datasets
+                # They are all expected to be aligned
+                for i in range(1000):
+                    dataset = h5file.create_dataset(
+                        dataset_name(i), shape, dtype='uint8')
+                    # Assign data so that the dataset is instantiated in
+                    # the file
+                    dataset[...] = i
+                    assert is_aligned(dataset, offset=alignment_interval):
```
hmaarrfk, line 69:
```suggestion
                    assert is_aligned(dataset, offset=alignment_interval)
```

hmaarrfk:
> Thanks! Don't forget to add the parameters to the reference docs (docs/high/file.rst) as well.

Sorry i misunderstood. I presumed that most of it was autogenerated with sphinx. 

I hope the paragraph helps

h5py/tests/test_file_alignment.py
hmaarrfk: Thank you for the suggestion!

h5py/_hl/files.py
```diff
@@ -123,6 +124,7 @@ def make_fapl(driver, libver, rdcc_nslots, rdcc_nbytes, rdcc_w0, locking,
         # we default to earliest
         low, high = h5f.LIBVER_EARLIEST, h5f.LIBVER_LATEST
     plist.set_libver_bounds(low, high)
+    plist.set_alignment(alignment_threshold, alignment_interval)
```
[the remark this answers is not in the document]
hmaarrfk: However, my thoughts were the same as yours, that the likeliness of their override was minor.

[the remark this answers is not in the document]
takluyver: Yup, I'm happy enough with this. I just figured that I'd leave a note here in case someone in 10 years has to try to work out what happened. :laughing:

Issue #2039, opened 2022-01-18
Provide driver for H5FD_DIRECT


hmaarrfk:
Would you accept patches to expose the `H5FD_DIRECT` driver?

It seems like it was added in 1.8.0.

https://support.hdfgroup.org/HDF5/doc/RM/H5P/H5Pset_fapl_direct.htm

I've got a prototype working but i likely need to add some tests.

For ease of installation, a few months back, i got the DIRECT driver enabled in conda forge:
https://github.com/conda-forge/hdf5-feedstock/pull/160

This provides great benefits to performance if coupled with:
https://github.com/h5py/h5py/issues/2034
(I owe you a PR for this)

```
Summary of the h5py configuration
---------------------------------

h5py    3.6.0
HDF5    1.12.1
Python  3.9.9 (main, Dec 29 2021, 07:47:36) 
[GCC 9.4.0]
sys.platform    linux
sys.maxsize     9223372036854775807
numpy   1.22.0
cython (built with) 0.29.26
numpy (built against) 1.22.0
HDF5 (built against) 1.12.1
```

PR #2041, opened 2022-01-18
Expose DIRECT VFD to Linux Users that have the required support

hmaarrfk:
Closes: https://github.com/h5py/h5py/issues/2039

TODO: 
- [x] Add tests modeled after SEC2 driver
- [x] Go through the other references to the SEC2 driver I found to ensure that all locations are met

<details> <summary>References to SEC2 driver found</summary>

```
I found the following references to sec2, but I haven't had a chance to replicate them for the direct driver

./lzf/example.c:    Opened "test_lzf.hdf5" with sec2 driver.
./docs/faq.rst:sec2                                Standard optimized driver                   Default on UNIX/Windows
./docs/high/file.rst:    'sec2'
./docs/whatsnew/3.0.rst:  file driver that supports it, not only the sec2 driver.
./h5py/tests/test_h5f.py:    def test_descriptor_sec2(self):
./h5py/tests/test_h5f.py:        dn_tmp = tempfile.mkdtemp('h5py.lowtest.test_h5f.TestFileID.test_descriptor_sec2')
./h5py/tests/test_h5f.py:            with File(fn_h5, driver='sec2', mode='x') as f:
./h5py/tests/test_file.py:    def test_sec2(self):
./h5py/tests/test_file.py:        fid = File(self.mktemp(), 'w', driver='sec2')
./h5py/tests/test_file.py:        self.assertEqual(fid.driver, 'sec2')
./h5py/tests/test_file.py:        fid = File(self.mktemp(), 'a', driver='sec2')
./h5py/tests/test_file.py:        self.assertEqual(fid.driver, 'sec2')
./h5py/tests/test_file2.py:            return _drivers['sec2'](plist)


The following areas were touched up to duplicate the SEC2 code for the DIRECT code


./docs/high/file.rst:        for the current platform. On UNIX, this is the H5FD_SEC2 driver;
./docs_api/h5fd.rst:.. data:: SEC2
./h5py/_hl/files.py:            h5fd.SEC2,
./h5py/_hl/files.py:        drivers = {h5fd.SEC2: 'sec2',
./h5py/h5fd.pyx:SEC2 = H5FD_SEC2
./h5py/h5p.pyx:        Select the "section-2" driver (h5fd.SEC2).
./h5py/h5p.pyx:        - h5fd.SEC2
./h5py/api_types_hdf5.pxd:  hid_t H5FD_SEC2

./h5py/_hl/files.py:    'sec2': lambda plist, **kwargs: plist.set_fapl_sec2(**kwargs),
./h5py/_hl/files.py:        drivers = {h5fd.SEC2: 'sec2',
./h5py/_hl/files.py:            recommended), 'core', 'sec2', 'direct', 'stdio', 'mpio', 'ros3'.
./h5py/h5p.pyx:    def set_fapl_sec2(self):
./h5py/h5p.pyx:        H5Pset_fapl_sec2(self.id)
./h5py/api_functions.txt:  herr_t    H5Pset_fapl_sec2(hid_t fapl_id)
```

</details>


hmaarrfk:
@h5py I think I got the approach down. Do feel free to comment if my approach is totally off track.

I'm thinking of modelling the tests after the tests for the SEC2 driver.

For clarity, the DIRECT driver only exists for Linux. I think the detection code should be pretty generic. While we only test *with* DIRECT VFD, I think it should be OK since we have OSX and Windows builds that are missing DIRECT VFD support.

hmaarrfk:
Hmm, I take it back, there are also some system installed HDF5 versions, those might have DIRECT VFD enabled. Not sure.

takluyver:
I'm happy enough with exposing an additional driver when building against libhdf5 that supports it.

Whether to enable it in our pre-built packages (i.e. in `ci/get_hdf5_if_needed.sh`) is a bigger question. I'd like to understand why HDF5 has it disabled by default - is it experimental? does it have extra dependencies? could it be dangerous in some scenario? Or some other reason? E.g. we support the ROS3 driver, but we don't enable it in our pre-built packages on PyPI, because we don't want to bundle C libraries that do network access.

hmaarrfk:
I did not notice that the library you build in the CI step gets shipped.

To be honest, I think the main difference is that on Linux (maybe Mac, I'm not sure Mac supports it) the local file is open with `O_DIRECT`. i think there are also a few different access patterns for the data like the caching and whatnot. But if you are using the direct driver you typically want your data to be aligned to avoid all that.

hmaarrfk:
Apparently MAC does not have the same mechanism. only Linux.

https://stackoverflow.com/a/2307055

takluyver:
Those sound like good reasons not to make it the default driver, but none of them sound much like reasons not to even build it. :thinking: 

@epourmal do you know why HDF5 does not include the 'direct' file driver in a build with the default options? Is there some serious downside to it? Is it so rarely wanted that it makes sense to skip building it most of the time?

hmaarrfk:
Personally, I can say that with `O_DIRECT` i can saturate any NVMe RAID array (up to 10GB/s tested -- big B = byte, just so that there is no confusion). With the SEC driver, I can only reach 1GB/s. To each their own, but I think that the DIRECT driver can be very useful

ci/get_hdf5_if_needed.sh
```diff
@@ -18,25 +18,28 @@ else
...
+    if [[ "${HDF5_VERSION%.*}" = "1.12" ]]; then
+        ENABLE_BUILD_MODE="--enable-build-mode=production"
     fi
 
-    if [ -f $HDF5_DIR/lib/$lib_name ]; then
-        echo "using cached build"
-    else
+    # if [ -f $HDF5_DIR/lib/$lib_name ]; then
+    #     echo "using cached build"
+    # else
```
takluyver, line 29:
```suggestion
    if [ -f $HDF5_DIR/lib/$lib_name ]; then
        echo "using cached build"
    else
```

You shouldn't need to comment this out - the cache key in the CI config includes a reference to this script, so if the script changes it won't use the cache:

https://github.com/h5py/h5py/blob/9d625e7270d29275b5610febdc4b0be213871ab7/ci/azure-pipelines-steps.yml#L15

ci/get_hdf5_if_needed.sh
```diff
@@ -18,25 +18,28 @@ else
...
+        ./configure --prefix $HDF5_DIR \
+            ${EXTRA_MPI_FLAGS} \
+            ${ENABLE_DIRECT_VFD} \
+            ${ENABLE_BUILD_MODE}
         make -j $(nproc)
         make install
         popd
         popd
-    fi
+    # fi
```
takluyver, line 44:
```suggestion
    fi
```

ci/get_hdf5_if_needed.sh
```diff
@@ -18,25 +18,28 @@ else
         lib_name=libhdf5.dylib
     else
         lib_name=libhdf5.so
+        ENABLE_DIRECT_VFD="--enable-direct-vfd"
```
takluyver, line 21:
```suggestion
        # Test with the direct file driver on Linux. This setting does not affect the
        # HDF5 bundled in Linux wheels - that is built into a Docker image from
        # a separate repository.
        ENABLE_DIRECT_VFD="--enable-direct-vfd"
```

Just to make it clear what goes where. :slightly_smiling_face:

h5py/_hl/files.py
```diff
@@ -233,6 +237,9 @@ def make_fid(name, mode, userblock_size, fapl, fcpl=None, swmr=False):
         # Not all drivers raise FileNotFoundError (commented those that do not)
         except FileNotFoundError if fapl.get_driver() in (
             h5fd.SEC2,
+            # 2022/01/19 hmaarrfk
+            # TODO: I'm not sure if this actually raises FileNotFoundError
+            h5fd.DIRECT if direct_vfd else -1,
```
takluyver, line 242: Can you check? It should just be a matter of opening a nonexistant file with this driver in mode `'r'` or `'r+'` to see what error you get.

h5py/api_functions.txt
```diff
@@ -22,6 +22,8 @@
 #
 # Function lines are C function signatures, optionally preceded by:
 #   MPI     If present, function requires an MPI-aware build of h5py
+#   ROS3    If present, function requires a ROS3 enabled build of h5py
+#   DIRECT_VFD If present, function requires an DIRECT VFD enabled build of h5py
```
takluyver, line 26: I think these should all say 'build of HDF5'

h5py/h5p.pyx
```diff
@@ -1154,6 +1182,7 @@ cdef class PropFAID(PropInstanceID):
         - h5fd.MPIO
         - h5fd.MULTI
         - h5fd.SEC2
+        - h5fd.DIRECT  (if available)
         - h5fd.STDIO
```
takluyver, line 1186: I guess ROS3 should also be in this list, and it was just an oversight that it's not.

takluyver (commented):
This mostly looks good, thanks. Don't forget to add some tests using the direct driver.

ci/get_hdf5_if_needed.sh
hmaarrfk: Thank you!

ci/get_hdf5_if_needed.sh
hmaarrfk: Thank you for explaining this to me.

h5py/_hl/files.py
hmaarrfk:
```
import h5py
h5py.File("sdkfjsl", "r", driver="direct")

FileNotFoundError: [Errno 2] Unable to open file (unable to open file, errno = 2, error message = 'No such file or directory')
```
Seems to confirm that this is the correct error to catch. Thank you!

hmaarrfk:
I had a chance to go through the comments you left. Let me see if i have time to create the test cases.

h5py/h5p.pyx
hmaarrfk: I added it.

h5py/api_functions.txt
hmaarrfk: Thank you!

h5py/h5p.pyx
```diff
@@ -1130,6 +1130,37 @@ cdef class PropFAID(PropInstanceID):
...
+            Select the "direct" driver (h5fd.DIRECT).
+
+            Parameters:
+                hid_t fapl_id       IN: File access property list identifier
+                size_t alignment    IN: Required memory alignment boundary
+                size_t block_size   IN: File system block size
+                size_t cbuf_size    IN: Copy buffer size
+
+            Properites with value of 0 indicate that the HDF5 library should
+            choose the value.
```
hmaarrfk, line 1147: Here is the reference in HDF5's source that will populate non-zero values
https://github.com/HDFGroup/hdf5/blob/bf27121b5d01939c819ac8b5b96101a16175c4de/src/H5FDdirect.c#L352

hmaarrfk:
kind ping to see if there is anything else that is needed on this front.

aragilar:
@hmaarrfk This looks good to me, but I'll let @takluyver look at the tests before merging.

One minor thing: I'm not familiar with the direct VFD subsystem, so maybe it's obvious to its users, but exposing your comments about the parameters that you added in one of the tests more broadly would be nice (e.g. in documentation or docstrings). That doesn't need to be in this PR though.

hmaarrfk:
> but exposing your comments about the parameters that you added

Right, I'm not an expert in it either, still just benchmarking the performance improvements. I would rather point to official documentation at this stage, instead of proliferating my bad understanding.

hmaarrfk:
And of course, thank you for looking it over!

takluyver:
Thanks @hmaarrfk, this looks good.

takluyver:
I forgot to ask for a news entry. @hmaarrfk, could I ask you to make a little follow up PR for that? Just copy `news/TEMPLATE.rst` to a new file and modify the relevant section to briefly describe this.

PR #2042, opened 2022-01-22
Create new attributes directly

### LINKED ISSUE #1385: [BUG] track_order = True causes `RuntimeError: Can't rename attribute (record is already in B-tree)`
 * Operating System: Windows/Linux Ubuntu/MacOS-X
 * Python version: 3.5+
 * Where Python was acquired: Anaconda
 * h5py version: 2.9.0+
 * HDF5 version: 1.10.4
 * The full traceback/stack trace shown (if it appears):
```
  File "h5py_bug.py", line 9, in <module>
    file.attrs[str(i)] = i
  File "/home/evandervelden/anaconda3/envs/py3/lib/python3.7/site-packages/h5py/_hl/attrs.py", line 100, in __setitem__
    self.create(name, data=value)
  File "/home/evandervelden/anaconda3/envs/py3/lib/python3.7/site-packages/h5py/_hl/attrs.py", line 217, in create
    h5a.rename(self._id, self._e(tempname), self._e(name))
  File "h5py/h5a.pyx", line 118, in h5py.h5a.rename
RuntimeError: Can't rename attribute (record is already in B-tree)
```

----

This bug has been reported quite a few times already (e.g., #1180, #1185 and others probably), but all of them are reporting that the bug solely appears when building `h5py` against HDF5 1.10.5.
However, I can cause that bug to appear perfectly fine in HDF5 1.10.4 (and probably other versions as well), as long as the `track_order` parameter can be set to True (which was added in `h5py` 2.9.0).

The following code will reproduce this error on Python 3.5/3.6/3.7; on Windows/Linux Ubuntu/MacOS-X:
```python
import h5py
h5py.get_config().track_order = True

with h5py.File('test.hdf5', 'a') as file:
    for i in range(20):
        print(i)
        file.attrs[str(i)] = i
```

After setting an attribute 8 times, setting the 9th will raise the error as shown above.
It does not matter how the file was opened; what the attributes are; what their types are; what is already contained in the HDF5-file; and so on.
It also does not matter how many attributes are set at once; setting the 9th will always trigger the error.
For example, the following will do the exact same thing:
```python
import h5py
h5py.get_config().track_order = True

for i in range(20):
    with h5py.File('test.hdf5', 'a') as file:
        print(i)
        file.attrs[str(i)] = i
```

Attempting to execute either of these code snippets twice in a row (of course making sure that `mode='a'`), will cause the error immediately upon setting the first attribute.

Exception to the above is that when the code snippet is executed on an already existing HDF5-file that was created with `track_order = False`, it does not raise the error.
However, obviously, it will not keep track of the insertion order of any new attributes either.

Hopefully this helps.

kmuehlbauer:
Create new attributes directly instead via temporary attribute with subsequent rename. This fixes overwriting attributes if ``track_order=True``.

Fixes #1385

### DIFF
--- a/news/create-new-attrs-directly.rst
+++ b/news/create-new-attrs-directly.rst
@@ -1,0 +1,31 @@
+New features
+------------
+
+* <news item>
+
+Deprecations
+------------
+
+* <news item>
+
+Exposing HDF5 functions
+-----------------------
+
+* <news item>
+
+Bug fixes
+---------
+
+* New attributes are created directly instead via a temporary
+attribute with subsequent renaming. This fixes overwriting
+attributes with ``track_order=True``.
+
+Building h5py
+-------------
+
+* <news item>
+
+Development
+-----------
+
+* <news item>
--- a/news/create-new-attrs-directly.rst
+++ b/news/create-new-attrs-directly.rst
@@ -16,9 +16,9 @@ Exposing HDF5 functions
 Bug fixes
 ---------
 
-* New attributes are created directly instead via a temporary
-attribute with subsequent renaming. This fixes overwriting
-attributes with ``track_order=True``.
+* New attributes are created directly, instead of via a temporary
+  attribute with subsequent renaming. This fixes overwriting
+  attributes with ``track_order=True``.
 
 Building h5py
 -------------
--- a/h5py/_hl/attrs.py
+++ b/h5py/_hl/attrs.py
@@ -190,8 +190,17 @@ class AttributeManager(base.MutableMappingHDF5, base.CommonStateObject):
 
             # This mess exists because you can't overwrite attributes in HDF5.
             # So we write to a temporary attribute first, and then rename.
-
-            tempname = uuid.uuid4().hex
+            # see issue 1385
+            # if track_order is enabled new attributes (which exceed the
+            # max_compact range, 8 is default) cannot be created as temporary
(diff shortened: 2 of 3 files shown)


kmuehlbauer:
Seems that I need to skip the test on some runs?

kmuehlbauer:
This is ready for a first review. Thanks for consideration.

kmuehlbauer:
Feeling stupid not seeing, that this just shifts the problem to the next overwrite. Sorry for the noise.

kmuehlbauer:
> Is this still worth doing (at the cost of code complexity) to save some (possible) file operations in the case where there is only one writing of the attribute?

Yes, that would indeed reduce overhead for attribute creation. Shall I reopen?

news/create-new-attrs-directly.rst
```diff
@@ -0,0 +1,31 @@
...
+-----------------------
+
+* <news item>
+
+Bug fixes
+---------
+
+* New attributes are created directly instead via a temporary
+attribute with subsequent renaming. This fixes overwriting
+attributes with ``track_order=True``.
```
takluyver, line 21:
```suggestion
* New attributes are created directly, instead of via a temporary
  attribute with subsequent renaming. This fixes overwriting
  attributes with ``track_order=True``.
```

takluyver:
Thanks!

PR #2051, opened 2022-02-10
Collapse nested try/except in attrs.py

takluyver:
I realised the second level of try/except/else in this code was unnecessary - the `else` block can be simplified to just putting the next bit of code below the first bit inside the `try`.

This should have no visible effect in the API, so I haven't added a news entry.


aragilar:
Much cleaner! Thanks @takluyver.

PR #2053, opened 2022-02-11
better error message when building from source

### LINKED ISSUE #2052: suggestion: better error message when building
   Windows)
 * h5py version (e.g. 2.6)
 * HDF5 version (e.g. 1.8.17)

`h5py.version.info` contains the needed versions, which can be displayed by
```
python -c 'import h5py; print(h5py.version.info)'
```
where `python` should be substituted for the path to python used to install
`h5py` with.

---

```
      Building h5py requires pkg-config unless the HDF5 path is explicitly specified
      error: pkg-config probably not installed: FileNotFoundError(2, 'No such file or directory')
```

It could be helpful if this error message included the actual variable to be set, for example: "set `HDF5_DIR=/usr/local/` or to your installation of HDF5"

ngam:
Fixes #2052

### DIFF
--- a/setup_configure.py
+++ b/setup_configure.py
@@ -135,7 +135,9 @@ class BuildConfig:
             if os.name != 'nt':
                 print(
                     "Building h5py requires pkg-config unless the HDF5 path "
-                    "is explicitly specified", file=sys.stderr
+                    "is explicitly specified using the environment variable HDF5_DIR. "
+                    "For more information and details, "
+                    "see https://docs.h5py.org/en/stable/build.html#custom-installation", file=sys.stderr
                 )
                 raise
 


tacaswell (approved):

ngam:
Btw, I didn't check myself --- do you have a CI job for building from source with the hdf5 library in a different non-default directory?

ngam:
For reference, building from source works very well in my experience --- it's jus that the error message could be a little more informative

ngam:
I've never worked with codecov, how does this make sense 😮 🤒 😨 

https://app.codecov.io/gh/h5py/h5py/compare/2053/changes

tacaswell:
Our codecov tends to get early false-negatives because we have so many CI jobs that we get full coverage only when they are _all_ in, but if 1 fails you do not want to not get the covarage report because it is waiting for a result that is never going to come.

tacaswell:
Thank you @ngam !  Congratulations on your first merged PR to h5py 🎉   Hopefully we will hear from you again!

ngam:
Thank you, I will do my best ❤️

PR #2065, opened 2022-03-14
Add support for Apple Silicon wheels

### LINKED ISSUE #2008: Mac ARM Chip Support?
I am trying to install via pip on macOS 11, on an M1 (ARM) chip. It looks like there's no built-in wheel yet. What might be the timeline for making this available? Otherwise, please advise on how to replace this manually.

Can I use home-brew to install hdf5? Then, how would I specify the path to this so installation would work?

matham:
Fixes #2008.

Changes:

* Compiling mac arm64 and x86_64 binaries, including hdf5 and its dependencies as individual wheels for each platform.
* Bump mac vm version from 10.15 to 11 (required to build arm64).
* Build zlib/gzip.
* To support arm64 cross-compiling I had to apply the patches from [here](https://github.com/conda-forge/hdf5-feedstock/commit/2cb83b63965985fa8795b0a13150bf0fd2525ebd) (thanks conda-forge). Neither 1.13 nor 1.12 support cross-compiling just arm64.

This is based on my PR for pytables: https://github.com/PyTables/PyTables/pull/944.


matham:
So it builds successfully for python 3.7-3.9, but fails for 3.10 with:

```
clang -Wno-unused-result -Wsign-compare -Wunreachable-code -fno-common -dynamic -DNDEBUG -g -fwrapv -O3 -Wall -g -arch arm64 -arch x86_64 -DH5_USE_18_API -DNPY_NO_DEPRECATED_API=0 -I./h5py -I/Users/runner/work/1/cache/hdf5/1.13.1/include -I/Users/runner/work/1/s/lzf -I/Users/runner/work/1/s/lzf/lzf -I/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pip-build-env-aem1bgc4/overlay/lib/python3.10/site-packages/numpy/core/include -I/Library/Frameworks/Python.framework/Versions/3.10/include/python3.10 -c /Users/runner/work/1/s/h5py/h5t.c -o build/temp.macosx-10.9-universal2-3.10/Users/runner/work/1/s/h5py/h5t.o
2022-03-14T05:34:29.4167250Z   /Users/runner/work/1/s/h5py/h5t.c:18401:20: error: use of undeclared identifier 'h5py_size_n256'
2022-03-14T05:34:29.4167960Z       __pyx_v_size = h5py_size_n256;
2022-03-14T05:34:29.4168280Z                      ^
2022-03-14T05:34:29.4271390Z   /Users/runner/work/1/s/h5py/h5t.c:18410:21: error: use of undeclared identifier 'h5py_offset_n256_real'
2022-03-14T05:34:29.4272200Z       __pyx_v_off_r = h5py_offset_n256_real;
2022-03-14T05:34:29.4272540Z                       ^
2022-03-14T05:34:29.4373580Z   /Users/runner/work/1/s/h5py/h5t.c:18419:21: error: use of undeclared identifier 'h5py_offset_n256_imag'
2022-03-14T05:34:29.4374320Z       __pyx_v_off_i = h5py_offset_n256_imag;
2022-03-14T05:34:29.4374660Z                       ^
2022-03-14T05:34:29.4565600Z   /Users/runner/work/1/s/h5py/h5t.c:19027:20: warning: comparison of integers of different signs: 'int' and 'size_t' (aka 'unsigned long') [-Wsign-compare]
2022-03-14T05:34:29.4566960Z     if (((__pyx_t_12 > __pyx_t_9) != 0)) {
2022-03-14T05:34:29.4567320Z           ~~~~~~~~~~ ^ ~~~~~~~~~
2022-03-14T05:34:29.4568600Z   /Users/runner/work/1/s/h5py/h5t.c:19047:40: warning: comparison of integers of different signs: 'size_t' (aka 'unsigned long') and 'int' [-Wsign-compare]
2022-03-14T05:34:29.4570610Z     __pyx_t_11 = ((__pyx_v_member_offset > __pyx_cur_scope->__pyx_v_dt->elsize) != 0);
2022-03-14T05:34:29.4571340Z                    ~~~~~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2022-03-14T05:34:29.8337130Z   2 warnings and 3 errors generated.
2022-03-14T05:34:29.8469670Z   error: command '/usr/bin/clang' failed with exit code 1
2022-03-14T05:34:30.2033420Z   Building wheel for h5py (pyproject.toml): finished with status 'error'
2022-03-14T05:34:30.2045630Z   ERROR: Failed building wheel for h5py
```

It's similar as https://github.com/h5py/h5py/pull/1973, but that was for the test env, so I'm not sure why even the build fails!? And why suddenly now?

ajelenak:
Is there any specific reason to use HDF5-1.13.1? The v1.13 releases are experimental, not official. Next official release after 1.12 will be 1.14.

matham:
>Did you experience any problem with the 1.12.1 version?

Hmm I don't remember why I thought that 1.12 didn't support universal binaries. I [just tried](https://dev.azure.com/h5pyappveyor/h5py/_build/results?buildId=1377&view=logs&j=f01ad03c-2855-5abe-9868-a5864177a0ee&t=24dd722c-cc9a-576b-9d63-a316a3349c97) with 1.12 and used `file` at the end of `get_hdf5_if_needed.sh` and this is what it prints:

```
+ file /Users/runner/work/1/cache/hdf5/1.12.1/lib/libhdf5.200.dylib /Users/runner/work/1/cache/hdf5/1.12.1/lib/libhdf5.a /Users/runner/work/1/cache/hdf5/1.12.1/lib/libhdf5.dylib /Users/runner/work/1/cache/hdf5/1.12.1/lib/libhdf5.la /Users/runner/work/1/cache/hdf5/1.12.1/lib/libhdf5.settings /Users/runner/work/1/cache/hdf5/1.12.1/lib/libhdf5_hl.200.dylib /Users/runner/work/1/cache/hdf5/1.12.1/lib/libhdf5_hl.a /Users/runner/work/1/cache/hdf5/1.12.1/lib/libhdf5_hl.dylib /Users/runner/work/1/cache/hdf5/1.12.1/lib/libhdf5_hl.la
/Users/runner/work/1/cache/hdf5/1.12.1/lib/libhdf5.200.dylib:    Mach-O 64-bit dynamically linked shared library x86_64
/Users/runner/work/1/cache/hdf5/1.12.1/lib/libhdf5.a:            current ar archive random library
/Users/runner/work/1/cache/hdf5/1.12.1/lib/libhdf5.dylib:        Mach-O 64-bit dynamically linked shared library x86_64
/Users/runner/work/1/cache/hdf5/1.12.1/lib/libhdf5.la:           libtool library file, ASCII text
/Users/runner/work/1/cache/hdf5/1.12.1/lib/libhdf5.settings:     ASCII text, with very long lines
/Users/runner/work/1/cache/hdf5/1.12.1/lib/libhdf5_hl.200.dylib: Mach-O 64-bit dynamically linked shared library x86_64
/Users/runner/work/1/cache/hdf5/1.12.1/lib/libhdf5_hl.a:         current ar archive random library
/Users/runner/work/1/cache/hdf5/1.12.1/lib/libhdf5_hl.dylib:     Mach-O 64-bit dynamically linked shared library x86_64
/Users/runner/work/1/cache/hdf5/1.12.1/lib/libhdf5_hl.la:        libtool library file, ASCII text
```

But previous builds with 1.13 didn't have this issue because at least for py 3.8-3.9 the wheels were built, now it fails as soon as it tries to create the arm64 part of the wheel because the compiled hdf5 doesn't include arm64.

matham:
>So, it seems HDF5-1.12.1 may not help here. The question remains whether using an experimental but working HDF5 release is suitable for official h5py wheels.

I'm gonna try getting it to work on 1.13.1 and let y'all decide whether you want to merge it before 1.14 is released.

matham:
What would be the potential issues? Is it that the api would be incompatible, or that the files would be potentially unreadable in future hdf5 releases? If the former, as long as we don't actually use the newer api it should be ok? Or there may be potential instabilities?

ajelenak:
https://www.hdfgroup.org/2021/12/hdf5-1-13-0-introducing-experimental-releases/

takluyver:
At a brief look:

- I'm definitely against bundling an experimental version of HDF5. Other people may disagree, but my view is that it either needs to work with 1.12 or wait for 1.14.
- I think I remember that wheels built on one version of Mac work only on that version and newer, so we want to build on the oldest Mac OS we can easily use. I'm very much not a Mac expert, though.
- The tests seem to fail because deflate support (zlib) is not getting built in correctly
- I'm not keen on making the simple `get_hdf5_if_needed` script massively more complicated (switching to cmake based on the version, calling brew, downloading dependencies...)
- The changes to complex256 support seem like a mess.

I'd like to see Mac-on-ARM support following the same pattern we already use for Linux on different architectures: find a CI platform that runs that architecture natively, and use that to do the build and the testing. I think having smaller platform specific (x86_64 and arm64) wheels is preferable to having 'universal' binaries. Most people will get the appropriate wheel chosen by pip anyway, so the convenience of a 'universal' binary doesn't really matter. NumPy has also gone for separate wheels (looking at 1.22.3 on PyPI).

aragilar:
I agree with @takluyver, and would add there's likely going to be a long tail of users still using x86 Macs on older MacOS versions (due to them not being able to upgrade—I doubt apple is going to support x86 long term). Keeping the wheel separate means we can treat x86 and arm differently, and adapt more easily to how those different platforms evolve.

matham:
>I'm definitely against bundling an experimental version of HDF5. Other people may disagree, but my view is that it either needs to work with 1.12 or wait for 1.14.

That does make sense and I agree we should wait until 1.14 as users won't get much of a benefit from 1.13 if their files are not forward compatible.

For now, my plan is to try to add separate arm64/x86 wheels and get back to universal when 1.14 comes out.

>I think I remember that wheels built on one version of Mac work only on that version and newer, so we want to build on the oldest Mac OS we can easily use. I'm very much not a Mac expert, though.

That's not quite the case. We control that with the `MACOSX_DEPLOYMENT_TARGET` variable, even if the OS it is built on is newer. You need the newer OS anyway to build arm64 wheels, I believe.

>I'm not keen on making the simple get_hdf5_if_needed script massively more complicated (switching to cmake based on the version, calling brew, downloading dependencies...)

I'm not sure if you can build arm64 without cmake, or at least the sense I got is that this is where hdf5 is headed. But I'd love to be corrected as I wouldn't mind sticking to just configure. But I don't think cmake was supported for other versions!? 

About brew, I guess there's no reason to use in that script, it can be called instead before the script is run.

About dependencies, I can see why downloading deps may complicate things, but how did you get them before? From the OS or is it bundled in the repo?

>The changes to complex256 support seem like a mess.

I tried following the `config.pxi` pattern. I couldn't add the variable checking whether we're on apple/arm64 there because you can't use cython's `DEF` like this as it's run during the cythonizing stage, not at the c compile stage. So the c header is meant as a config for things that need to be written in c and available from there.


>I'd like to see Mac-on-ARM support following the same pattern we already use for Linux on different architectures: find a CI platform that runs that architecture natively, and use that to do the build and the testing. I think having smaller platform specific (x86_64 and arm64) wheels is preferable to having 'universal' binaries. Most people will get the appropriate wheel chosen by pip anyway, so the convenience of a 'universal' binary doesn't really matter. NumPy has also gone for separate wheels (looking at 1.22.3 on PyPI).

The problem with this approach is that currently there's no CI that provides apple silicon support, except for paying ones. But nothing like github actions etc. And the [github issue about this](https://github.com/actions/virtual-environments/issues/2187) is going nowhere. Apple is very controlling about this. E.g. to use aws you need to pay immediately for 24-hr usage, even if you only need 20 min. But there are others (outside the US!?) that are more flexible. Anyway, that is currently not easily possible.

So, you'll need the changes in this PR to be able to build (cross compile) arm64 either way. The only additional changes required for universal wheels is the added `config.h` file, I think.

I looked at numpy, and they did have universal wheels until two releases ago, but I think there was some issue with it so they dropped it. Right now they are [looking into](https://github.com/numpy/numpy/pull/20787) whether it's worth their effort to get it working. But, for h5py, I think supporting universal wheels vs arm64/x86 will require very little additional effort I think once 1.14 comes out. But that's upto you to decide.


To explain why I needed universal wheels I'll paste what I wrote on the numpy issue:

Universal wheels are needed for app developers that package apps using only github CI. E.g. we at Kivy (a python GUI framework) provide our libraries as universal wheels as well as a packaging mechanism to bundle it all into a dmg. If there are no universal wheels or binaries then you need a M1 to build the app for the M1 and a x86 to build the app for x86. If there are universal wheels then you can build both into one dmg supporting both architectures.

Given the absence of CI for M1, I'm trying to get the universal wheels.

aragilar:
With respect to `complex256`, I think it work be better to map over the float-choice logic to complex floats as well (which should be done in a different PR).

Maybe I missed this, but does the lack of M1 CI implies we're building wheels without testing them?

Also, personally I'd be concerned about apps which rely on including our wheels. Whilst HDF5 is probably not a major target (and we avoid enabling ROS3), we haven't been updating our wheels when HDF5 releases bugfixes (and I don't we don't have the bandwidth to do so). I'd suggest that apps which depend on us should build HDF5 separately (so that they can upgrade if needed), and then build h5py on top of that (and given the HDF5 ABI is relatively stable, it should be reasonable to upgrade HDF5 without needing to rebuild h5py, like the linux distros do).

matham:
Ok, after some further investigation I have given up on the idea of universal wheels and instead focused just on building arm64 and x86 wheels for the python versions that support them. Ultimately to make universal2 wheels, `delocate` now has a fuse option that can just merge the two wheels, but packaging libraries can add that functionality and it doesn't have to be added here.

To be able to cross compile on x86 for arm64 I had to apply [these patches](https://github.com/conda-forge/hdf5-feedstock/commit/2cb83b63965985fa8795b0a13150bf0fd2525ebd) from conda-forge. Neither 1.12  not 1.13 would otherwise cross-compile just arm64.

Are these patches acceptable? Currently I download them and apply them from the original commit, but I could add them to the repo if that is desired. I brought this up on [HDF5's repo](https://github.com/HDFGroup/hdf5/issues/566), but I'm not sure when cross-compiling will be fixed. They indicated on the issue that they are fixing it, hopefully by 1.14. Otherwise, to support it, for newer versions of HDF5, we'd probably need to upgrade the patches (or just wait until conda-forge does it 😁).

I tested the wheels from [this build](https://dev.azure.com/h5pyappveyor/h5py/_build/results?buildId=1401&view=results) on a Scaleaway M1 I spun up:

```py
Desktop % python
Python 3.9.10 (main, Jan 15 2022, 11:40:53)
[Clang 13.0.0 (clang-1300.0.29.3)] on darwin
>>> import platform
>>> platform.uname()
uname_result(system='Darwin', node=...', release='21.1.0', version='Darwin Kernel Version 21.1.0: Wed Oct 13 17:33:24 PDT 2021; root:xnu-8019.41.5~1/RELEASE_ARM64_T8101', machine='arm64')
>>> import h5py
>>> f = h5py.File("mytestfile.hdf5", "w")
>>> dset = f.create_dataset("mydataset", (100,), dtype='i')
>>> dset.name
'/mydataset'
>>> f
<HDF5 file "mytestfile.hdf5" (mode r+)>
>>> dset
<HDF5 dataset "mydataset": shape (100,), type "<i4">
```

If these patches are ok, then I think this PR is ready.

matham:
>With respect to complex256, I think it work be better to map over the float-choice logic to complex floats as well (which should be done in a different PR).

I removed this change from the current changes.

>Maybe I missed this, but does the lack of M1 CI implies we're building wheels without testing them?

Sadly yes. This is currently where all projects that build for arm64 are because there no VMs available, unless you pay or have a private M1 runner set up on your M1. However, I think this is better than not building at all, even if it's not tested. And I did a simple test to see that works on a M1 with the compiled.

>Also, personally I'd be concerned about apps which rely on including our wheels.

I understand the concern, but for most people using h5py in a python app/project this should not really be a concern. I know for me it's not really a concern not having cutting edge hdf5, as long as I can easily download a wheel containing hdf5 that is not ancient. 

Consider that most python users wouldn't even know how to go about compiling hdf5 or setup their system to be able to do that. While it's easy enough for them the follow a guide on how to pip install wheels and make it into a app, having to compile individual packages when making an app would probably not something they'd be inclined to do. Even I, who know how to do that, wouldn't find that convenient and just having a wheel that installs it, even if hdf5 is like 2 years old is good enough.

takluyver:
Thanks. I'm still not very keen on the changes in the build script.

- Can the bulk of these additions go in a new script (something like `configure_hdf5_mac_arm64.sh`) which is called from the existing one as needed?
- Do we really need to build gzip? I'd expect zlib to be sufficient to link HDF5 against.
- Do we really need all those environment variables? Even the ones that seem to be about Fortran? We shouldn't be building anything in Fortran (and if we are, we should probably disable it).
- Are all those variables documented somewhere, or has someone figured them out with trial and error?
- If we need to patch HDF5, let's commit the patches in this repo rather than curl-ing them in the CI job. Again, I don't think we should need the two that seem to be Fortran specific.

matham:
>Can the bulk of these additions go in a new script (something like configure_hdf5_mac_arm64.sh) which is called from the existing one as needed?

Done. I named it `configure_hdf5_mac.sh` because I moved some overall mac specific config there e.g. zlib building, which is also for x86. Let me know if the structure is ok.

>Do we really need to build gzip? I'd expect zlib to be sufficient to link HDF5 against.

Not sure why I thought I needed it, I removed it.

>Do we really need all those environment variables? Even the ones that seem to be about Fortran? We shouldn't be building anything in Fortran (and if we are, we should probably disable it).

I just copied it all over from conda-forge, but I removed what seemed fortran specific now.

>Are all those variables documented somewhere, or has someone figured them out with trial and error?

I'm really not sure, but I trust the conda maintainer as they seemed to understand the variables. Very likely they did trial and error, figuring out what config choked on, getting their correct value from cross-compiling, and then using that value. But, hdf5 said they'll fix this so we won't eventually need this.

>If we need to patch HDF5, let's commit the patches in this repo rather than curl-ing them in the CI job. Again, I don't think we should need the two that seem to be Fortran specific.

Done

matham:
Looks like pre-commit is trying to fix the patch files, so I had to disable write access to the branch so I could revert and then figure out how to exclude the files from pre-commit.

If I may suggest, pre-commit may be more helpful if it simply shows the issues and blocks merging until it's all addressed, rather than trying to fix it itself as that will cause the sudden commits to the branch, which could be inconvenient like in this case.

ci/configure_hdf5_mac.sh
```diff
@@ -0,0 +1,49 @@
...
+    export CC="/usr/bin/clang"
+    export CXX="/usr/bin/clang"
+    export CFLAGS="$CFLAGS -arch $arch"
+    export CPPFLAGS="$CPPFLAGS -arch $arch"
+    export CXXFLAGS="$CXXFLAGS -arch $arch"
+}
+
+
+function build_zlib() {
+    ZLIB_VERSION="1.2.11"
```
takluyver, line 12:
```suggestion
    ZLIB_VERSION="1.2.12"
```

Looks like it was just released the other day. :slightly_smiling_face:

setup_build.py
```diff
@@ -134,38 +151,29 @@ def run(self):
...
-            'ros3': bool(config.ros3),
-            'direct_vfd': bool(config.direct_vfd),
-            'version': config.hdf5_version,
-            'complex256_support': hasattr(numpy, 'complex256'),
-            'numpy_version': numpy.__version__,
-            'cython_version': cython_version,
-        }
-        write_if_changed(config_file, s)
+        write_if_changed(localpath('h5py', 'config.pxi'), s)
+        write_if_changed(localpath('h5py', 'config.pxi', root=build_path), s)
```
takluyver, line 176: Why do we now need to write it in the build folder as well, when we didn't before? I'm not saying it's wrong, I'd just like to understand it a bit.

azure-pipelines.yml
```diff
@@ -51,16 +51,30 @@ jobs:
 
 - job: 'macos_wheels'
   pool:
-    vmImage: macOS-10.15
+    vmImage: macOS-11
   variables:
     MACOSX_DEPLOYMENT_TARGET: 10.9
     HDF5_VERSION: 1.12.1
+    HDF5_ROS3: '0'
+    HDF5_DIRECT_VFD: '0'
```
takluyver, line 59: These 2 environment variables are new and only affect the h5py build, right? It might be clearer to use `H5PY_` as the prefix, to make it clear that they're not controlling the feature in HDF5 itself.

takluyver:
Thanks, this is looking promising. Have you downloaded the wheels from Azure and tried installing one manually to check that it's getting built OK?

azure-pipelines.yml
matham: Fixed

ci/configure_hdf5_mac.sh
matham: Fixed

setup_build.py
matham: So... we don't actually need this to make it build on the CI. I had added this earlier when I thought the config file may not have been copied, but the issue was something else.

The reason I left it, is that I think my changes is the correct approach. I had seen issues in two other [projects](https://github.com/kivy/kivy/pull/2934) I contributed to caused by the build-time generated script not being correctly copied or copied too early before it had the correct values. So I figured it's better to be safe than sorry and left the change there. But I can remove it if needed.

matham:
>Have you downloaded the wheels from Azure and tried installing one manually to check that it's getting built OK?

I just spun up a new M1 on scaleaway and tested it with homebrew python 3.9 on the latest changes. and it seemed to work:

```shell
Desktop % pip install h5py-3.6.0-cp39-cp39-macosx_11_0_arm64.whl
Processing ./h5py-3.6.0-cp39-cp39-macosx_11_0_arm64.whl
Collecting numpy>=1.14.5
  Downloading numpy-1.22.3-cp39-cp39-macosx_11_0_arm64.whl (12.8 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 12.8/12.8 MB 89.3 MB/s eta 0:00:00
Installing collected packages: numpy, h5py
Successfully installed h5py-3.6.0 numpy-1.22.3
```

```shell
Desktop % python
Python 3.9.12 (main, Mar 26 2022, 15:44:31)
[Clang 13.1.6 (clang-1316.0.21.2)] on darwin
>>> import platform
>>> platform.uname()
uname_result(system='Darwin', node='...', release='21.1.0', version='Darwin Kernel Version 21.1.0: Wed Oct 13 17:33:24 PDT 2021; root:xnu-8019.41.5~1/RELEASE_ARM64_T8101', machine='arm64')
>>>
```

And then running through the test items after installing pytest and pytest-mpi:

```shell
Desktop % python -c "import sys; print('64 bit?', sys.maxsize > 2**32)"
64 bit? True
Desktop % python -c "from h5py.version import info; print(info)"

h5py    3.6.0
HDF5    1.12.1
Python  3.9.12 (main, Mar 26 2022, 15:44:31)
[Clang 13.1.6 (clang-1316.0.21.2)]
sys.platform    darwin
numpy   1.22.3
cython (built with) 0.29.28
numpy (built against) 1.19.3
HDF5 (built against) 1.12.1
 Desktop % python -m pytest --pyargs h5py
============================================= test session starts =============================================

platform darwin -- Python 3.9.12, pytest-7.1.1, pluggy-1.0.0
rootdir: /Users/m1/Desktop
plugins: mpi-0.6
collected 703 items

venv/lib/python3.9/site-packages/h5py/tests/test_attribute_create.py ......                             [  0%] 
venv/lib/python3.9/site-packages/h5py/tests/test_attrs.py ..........................                    [  4%] 
venv/lib/python3.9/site-packages/h5py/tests/test_attrs_data.py ....................                     [  7%] 
venv/lib/python3.9/site-packages/h5py/tests/test_base.py ...........                                    [  8%] 
venv/lib/python3.9/site-packages/h5py/tests/test_big_endian_file.py ..                                  [  9%] 
venv/lib/python3.9/site-packages/h5py/tests/test_completions.py ..                                      [  9%] 
venv/lib/python3.9/site-packages/h5py/tests/test_dataset.py .s......................................... [ 15%] 
..........................s............................................................................ [ 30%] 
...................                                                                                     [ 33%] 
venv/lib/python3.9/site-packages/h5py/tests/test_dataset_getitem.py ................................... [ 37%] 
.................................................................                                       [ 47%] 
venv/lib/python3.9/site-packages/h5py/tests/test_dataset_swmr.py sssss..........                        [ 49%] 
venv/lib/python3.9/site-packages/h5py/tests/test_datatype.py ..                                         [ 49%] 
venv/lib/python3.9/site-packages/h5py/tests/test_dimension_scales.py ....s.................             [ 52%] 
venv/lib/python3.9/site-packages/h5py/tests/test_dims_dimensionproxy.py .                               [ 52%] 
venv/lib/python3.9/site-packages/h5py/tests/test_dtype.py ..................ssssss.                     [ 56%] 
venv/lib/python3.9/site-packages/h5py/tests/test_errors.py ...                                          [ 56%] 
venv/lib/python3.9/site-packages/h5py/tests/test_file.py ..................s.....ssss.................. [ 63%] 
............ssss.....s.                                                                                 [ 66%] 
venv/lib/python3.9/site-packages/h5py/tests/test_file2.py .........s........                            [ 69%] 
venv/lib/python3.9/site-packages/h5py/tests/test_file_alignment.py ...                                  [ 69%] 
venv/lib/python3.9/site-packages/h5py/tests/test_file_image.py ..                                       [ 69%] 
venv/lib/python3.9/site-packages/h5py/tests/test_filters.py .s...s                                      [ 70%] 
venv/lib/python3.9/site-packages/h5py/tests/test_group.py ............................................. [ 77%] 
............................................                                                            [ 83%] 
venv/lib/python3.9/site-packages/h5py/tests/test_h5.py .....                                            [ 84%] 
venv/lib/python3.9/site-packages/h5py/tests/test_h5d_direct_chunk.py ....                               [ 84%] 
venv/lib/python3.9/site-packages/h5py/tests/test_h5f.py .s.....                                         [ 85%] 
venv/lib/python3.9/site-packages/h5py/tests/test_h5o.py .                                               [ 85%] 
venv/lib/python3.9/site-packages/h5py/tests/test_h5p.py ...........                                     [ 87%] 
venv/lib/python3.9/site-packages/h5py/tests/test_h5pl.py ......                                         [ 88%] 
venv/lib/python3.9/site-packages/h5py/tests/test_h5t.py ...                                             [ 88%] 
venv/lib/python3.9/site-packages/h5py/tests/test_objects.py ...                                         [ 89%] 
venv/lib/python3.9/site-packages/h5py/tests/test_selections.py .....                                    [ 89%] 
venv/lib/python3.9/site-packages/h5py/tests/test_slicing.py ......................................      [ 95%] 
venv/lib/python3.9/site-packages/h5py/tests/test_vds/test_highlevel_vds.py ...........                  [ 96%] 
venv/lib/python3.9/site-packages/h5py/tests/test_vds/test_lowlevel_vds.py ....                          [ 97%] 
venv/lib/python3.9/site-packages/h5py/tests/test_vds/test_virtual_source.py ..................          [100%]

======================================= 675 passed, 28 skipped in 2.52s =======================================
```

setup_build.py
takluyver: Thanks! I'd say that if we can't identify a concrete problem with the code as it was before, let's leave it alone and not add to it (i.e. roll back these changes).

At some point, we're probably going to need to overhaul how h5py is built. I'm hoping that we can draw on either [scipy's work to use meson](https://github.com/rgommers/scipy/issues/22), or the [scikit-build overhaul](https://iscinumpy.dev/post/scikit-build-proposal/) for this before our setuptools based build breaks too badly. Until then, I'd like a minimum of changes and a minimum of complexity. :slightly_smiling_face:

matham: Ok, done!

takluyver:
Thanks, this looks good to me. I'll leave it a day or two in case the other h5py maintainers want to check it over, but if there are no more comments then I'll merge it.

takluyver:
@matham after I merged this, the build failed on master for making MacOS wheels for x86-64. Here's the error message:

```
  CCLD     H5make_libsettings
  CCLD     H5detect
LD_LIBRARY_PATH="$LD_LIBRARY_PATH`echo  |                  \
		sed -e 's/-L/:/g' -e 's/ //g'`"                               \
	 ./H5make_libsettings  H5lib_settings.c  ||                               \
	    (test $HDF5_Make_Ignore && echo "*** Error ignored") ||          \
	    (rm -f H5lib_settings.c ; exit 1)
dyld: Library not loaded: libz.so.1.2.12
  Referenced from: /private/tmp/hdf5-1.12.1/src/./H5make_libsettings
  Reason: image not found
/bin/sh: line 1: 22266 Abort trap: 6           LD_LIBRARY_PATH="$LD_LIBRARY_PATH`echo  |                  	sed -e 's/-L/:/g' -e 's/ //g'`" ./H5make_libsettings H5lib_settings.c
make[2]: *** [H5lib_settings.c] Error 1
```

And here's the full log: https://dev.azure.com/h5pyappveyor/h5py/_build/results?buildId=1422&view=logs&j=0ad4bf5c-a1c6-59db-420d-7a6f497f570b&t=aca53dbd-2d58-55a9-ea01-6022315be503

Could you take a look?

matham:
Weird that it failed on master but not PR. I'll take a look ASAP.

NAThompson:
@takluyver : Is this sufficient cause to do a new release? I would love to have this available immediately.

takluyver:
I think it's probably about time we did a new release in any case. HDF5 1.12.2 also just came out, but one of the patches from this PR conflicts with that (see #2087). I'd like to get that resolved and then make a release.

PR #2067, opened 2022-03-21
Add a news stub for the Direct VFD

hmaarrfk:
@takluyver this release note is a little hard since you indicated that you didn't want to add the `--enable-direct-vfd` to your pypi releases. But here is my attempt.

https://github.com/h5py/h5py/pull/2041#issuecomment-1073773166
https://github.com/conda-forge/hdf5-feedstock/blob/main/recipe/build.sh#L12

### DIFF
--- a/news/direct-vfd-stubs.rst
+++ b/news/direct-vfd-stubs.rst
@@ -1,0 +1,36 @@
+New features
+------------
+
+* Provide the ability to use the Direct Virtual File Driver (VFD) from
+  HDF5 (Linux only).
+  If the Direct VFD driver is present at the time of compilation, users can use the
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks, that looks pretty good. It looks like conda-forge users will get this without recompiling, so I might word the last bit something like:

> The pre-built h5py packages on PyPI don't include the direct driver. Other packages such as conda might include it, or you can build h5py from source against an HDF5 build with the direct driver enabled.

(And for the record, I'm not set against enabling it in our wheels - I'd just like to understand why it's disabled by default in HDF5, so we can assess whether that reason applies to us.)

takluyver:
Thanks!

PR #2076, opened 2022-03-29
Use zero-ed memory when reading data

### LINKED ISSUE #455: Reads from unallocated chunks may return uninitialized memory
Unlike contiguous datasets, when reading a chunked dataset with no fill value defined, HDF5 does not fill the buffer with zeros.  THG says this is the intended behavior. Does not affect datasets created with h5py, as we manually set the chunk allocation time to get around this.

Should explicitly initialize the read buffer to zeros (np.ndarray -> np.zeros).

See also https://groups.google.com/forum/#!topic/h5py/PgmJ-R7qxqI

### LINKED ISSUE #2069: h5 file content stays in memory after close() and is overwritten
* Windows 10, CentOS 7
 * Python 3.8.12 (on both)
 * Anaconda (on both)
 * h5py version: 3.6.0 on Windows and 2.10.0 on CentOS 7
 * HDF5 version: 1.10.6 on Windows, 1.8.12 on CentOS 7

I am using h5py to read a specific dataset from a file produced by an external code, written in Fortran. This dataset is a simple array of instants in time, from 0 up to 3. In my python script I need to manipulate this array, for example I pass it to "np.argwhere" to look for some specific time index. When I do this and I read again the dataset, the first value of the dataset changes.
Here you can find a minimum working example. 

````
import h5py as h5
import numpy as np
from copy import deepcopy as cp
# I read the file
fh5 = h5.File("C:\\Users\\Desktop\\output.h5", "r")
times = fh5['distributions']['timeDistr'][()]
fh5.close()
# the file seems to be closed
print(fh5)

print(times[0:3])

times[0] = 10
print(times[0:3])

# here I mimic some manipulations of the original dataset
ttt = []
for t in times:
    ttt.append(t)
ttt = cp(np.asarray(ttt))

tmp2 = list([ttt+42]) # this affects the dataset
tmp3 = list(np.asarray(tmp2)+10) # this does not affect the dataset

# I read again the dataset
fh52 = h5.File("C:\\Users\\Desktop\\output.h5", "r")
newtimes = fh52['distributions']['timeDistr'][()]
fh52.close()

print(newtimes[0:3])
````
The file "output.h5" is attached here as a zip file.
[output.zip](https://github.com/h5py/h5py/files/8335475/output.zip)

In this example I manipulate the dataset after dumping it into an array. When I read again the file _ex novo_, the dataset is changed! The first value is the only to be affected, it becomes 52 in this case, which is the results of the operations ``times[0] = 10`` and ``tmp2 = list([ttt+42])``. This is akward, especially because ``list([ttt+42])`` is assigned to another variable which has nothing to do with the dataset.

How can it be possible? Did I something extremely stupid? I never saw something like this, it seems completely crazy to me!

Thank you so much for your help, 
Nicolò

takluyver:
Avoid exposing values from memory when reading from a chunked dataset with missing chunks and no fill value. This is essentially the same fix attempted in #456.

Back then, [Andrew found that](https://groups.google.com/g/h5py/c/PgmJ-R7qxqI/m/RMk6-iEfngEJ) "Unfortunately it looks like this approach doesn't work for object-based dtypes; it fills them with Python int "0"'s rather than None." This is still the case - `np.zeros(3, dtype=object)` has zeros where `np.empty(3, dtype=object)` has None. But I'm not sure when this would matter. The only cases I can see where h5py would create an object array to read into are when converting vlen data or HDF5 references, and in both cases it seems that the conversion machinery fills the array with the expected objects anyway. The tests pass, at least on my machine.

For small reads, this might be marginally slower, but for large reads there's likely to be no penalty, because the OS can do [virtual memory tricks](https://vorpus.org/blog/why-does-calloc-exist/) to avoid really allocating the memory.

Closes #455
Closes #2069

### DIFF
--- a/h5py/_hl/attrs.py
+++ b/h5py/_hl/attrs.py
@@ -73,7 +73,7 @@ class AttributeManager(base.MutableMappingHDF5, base.CommonStateObject):
             shape = attr.shape + subshape   # (5, 3)
             dtype = subdtype                # 'f'
 
-        arr = numpy.ndarray(shape, dtype=dtype, order='C')
+        arr = numpy.zeros(shape, dtype=dtype, order='C')
         attr.read(arr, mtype=htype)
 
         string_info = h5t.check_string_dtype(dtype)
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -569,7 +569,7 @@ class Dataset(HLObject):
     @with_phil
     def fillvalue(self):
         """Fill value for this dataset (0 by default)"""
-        arr = numpy.ndarray((1,), dtype=self.dtype)
+        arr = numpy.zeros((1,), dtype=self.dtype)
         self._dcpl.get_fill_value(arr)
         return arr[0]
 
@@ -749,7 +749,7 @@ class Dataset(HLObject):
             if mshape is None:
                 # 0D with no data (NULL or deselected SCALAR)
                 return Empty(new_dtype)
-            out = numpy.empty(mshape, dtype=new_dtype)
+            out = numpy.zeros(mshape, dtype=new_dtype)
             if out.size == 0:
                 return out
 
@@ -764,7 +764,7 @@ class Dataset(HLObject):
             # Check 'is Ellipsis' to avoid equality comparison with an array:
             # array equality returns an array, not a boolean.
             if args == () or (len(args) == 1 and args[0] is Ellipsis):
-                return numpy.empty(self.shape, dtype=new_dtype)
+                return numpy.zeros(self.shape, dtype=new_dtype)
 
         # === Scalar dataspaces =================
 
@@ -772,9 +772,9 @@ class Dataset(HLObject):
             fspace = self.id.get_space()
             selection = sel2.select_read(fspace, args)
             if selection.mshape is None:
-                arr = numpy.ndarray((), dtype=new_dtype)
+                arr = numpy.zeros((), dtype=new_dtype)
             else:
-                arr = numpy.ndarray(selection.mshape, dtype=new_dtype)
+                arr = numpy.zeros(selection.mshape, dtype=new_dtype)
             for mspace, fspace in selection:
                 self.id.read(mspace, fspace, arr, mtype)
             if selection.mshape is None:
@@ -787,9 +787,9 @@ class Dataset(HLObject):
         selection = sel.select(self.shape, args, dataset=self)
 
         if selection.nselect == 0:
-            return numpy.ndarray(selection.array_shape, dtype=new_dtype)
+            return numpy.zeros(selection.array_shape, dtype=new_dtype)
 
-        arr = numpy.ndarray(selection.array_shape, new_dtype, order='C')
+        arr = numpy.zeros(selection.array_shape, new_dtype, order='C')
 
         # Perform the actual read
         mspace = h5s.create_simple(selection.mshape)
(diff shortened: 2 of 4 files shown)


tacaswell:
We also hand back object arrays for variable length strings.  It may be worth gating this on type?


[1] https://docs.h5py.org/en/stable/special.html#variable-length-strings

aragilar:
Should there be a test checking that HDF5 does not override the 0s with junk (or does HDF5 know which bits to copy and which bits to not copy to)?

takluyver:
At least with my test - creating a chunked dataset of vlen strings with no data, checking that `ds.id.get_storage_size()` returns 0, and reading it back with `ds[:]` - the conversion machinery still runs and fills the output array with empty string objects. Intuitively, it feels like this ought to be something we can rely on - even missing data should be passed to the conversion function as zeros.

I'm reluctant to gate it on type without understanding why it matters - and without being able to write a test case for it.

> Should there be a test checking that HDF5 does not override the 0s with junk (or does HDF5 know which bits to copy and which bits to not copy to)?

I'm not sure I follow. We want HDF5 to overwrite the zeros, this is just about what happens when it doesn't. If it writes junk into the output buffer on reading, that would be a bug in HDF5, and I think our test suite should largely assume HDF5 works as intended.

aragilar:
I was thinking if you had a dataset created three empty chunks, and only write to the second chunk, reading the first and second chunks may cause the first chunk to have junk instead of 0s (from HDF5, not from an uninitialised numpy buffer), unless HDF5 knows the first hasn't been written, and can simply skip it (which is what I thought #2069 was about)? I don't know if HDF5 knows to skip unwritten chunks, or if it blindly copies what it was told?

takluyver:
With the default settings, HDF5 will only allocate chunks in the file when they're written, so there wouldn't be any first chunk to copy junk from.

You can affect this with [H5Pset_alloc_time](https://confluence.hdfgroup.org/display/HDF5/H5P_SET_ALLOC_TIME) and [H5Pset_fill_time](https://confluence.hdfgroup.org/display/HDF5/H5P_SET_FILL_TIME). But if there's any scenario where H5Dread writes junk data into the output array, I think that would be a bug in HDF5. I've never seen any suggestion that the user is meant to figure out which chunks contain data before reading, so I believe it's up to HDF5 to keep track of this.

takluyver:
I'm still somewhat uneasy about whatever the issue was back in 2014, but I think I'm inclined to release it with this change, and see if any issues come up.

Do you think it's worth writing a test for this? I can reproduce it with a freshly created dataset using the low-level API, but the failure mode depends on implementation details of both NumPy (reusing memory) and HDF5, so a passing test might not mean much. I can isolate it from NumPy by using `read_direct` to pass in an array, but then it's not testing the bit this is actually fixing. :shrug:

tacaswell:
I think it is worth putting in a test that atleast documents how to make a file that is susceptible to this bug and then reading it back.   Even if there is a chance it will pass by chance, the next time this comes up we will at least have a starting point for how to force it back to failing.

aragilar:
It's good that we don't need to manage the chunk validity. Maybe if we write a parameterised test which checks that strings and other variable length types return the values we expect, then we'll catch whether some other (possibly apparently unrelated) would break that, and gain insight into Andrew's comments about his earlier attempt.

Might be overkill, but should we allow users to set the initial backing array value, without modifying the dataset fill value? I can see being able to set the fill value to NaN could be quite useful.

tacaswell:
> Might be overkill, but should we allow users to set the initial backing array value, without modifying the dataset fill value? 

I'm wary of inventing too much new API over what libhdf gives us.  I think if users really want to control this they have to use `read_direct`?

takluyver:
Yeah, I wouldn't try to do anything for that. Especially as it's hard to be sure that HDF5 won't write its chunk value (e.g. if a chunk is partly filled, you'll probably get the fill value for the rest of that chunk).

takluyver:
I've added a test, and verified that if I undo the change in `_selector.pyx`, it fails.

tacaswell (approved):

PR #2079, opened 2022-04-01
Fix for #1847: Convert boolean arrays into index arrays for fancy slicing

### LINKED ISSUE #1750: fancy indexing via a mix of NumPy array and list not working for v3.1.0
The fancy indexing using a mix of NumPy boolean array and an explicit list does not seem to work in the latest h5py version 3.1.0. The detailed trackback is shown at the bottom.

The exact same commands work in h5py version 2.10. 

In h5py version 3.1.0, fancy indexing using NumPy boolean array only or using explicit lists only also works.

 * Operating System (MacOS 10.14)
 * Python version (3.7 & 3.9)
 * Where Python was acquired (Anaconda on MacOS)
 * h5py version (3.1)
 * HDF5 version (1.10.6)

```python
(insar) yunjunz:~>$ python
Python 3.7.8 | packaged by conda-forge | (default, Jul 31 2020, 02:37:09) 
[Clang 10.0.1 ] on darwin
>>> import numpy as np
>>> import h5py
>>> arr = np.arange(100).reshape((10,10))
>>> with h5py.File('test.h5', 'w') as f:
...     dset = f.create_dataset("MyDataset", data=arr)
...     result = dset[arr[:,0] > 20, :]
...     result.shape
... 
  File "<stdin>", line 3, in <module>
  File "/Users/yunjunz/tools/miniconda3/envs/insar/lib/python3.7/site-packages/h5py/_hl/dataset.py", line 777, in __getitem__
  File "/Users/yunjunz/tools/miniconda3/envs/insar/lib/python3.7/site-packages/h5py/_hl/selections.py", line 82, in select
  File "h5py/_selector.pyx", line 272, in h5py._selector.Selector.make_selection
  File "h5py/_selector.pyx", line 183, in h5py._selector.Selector.apply_args
TypeError: Indexing arrays must have integer dtypes```

### LINKED ISSUE #1847: Indexing arrays must have integer dtypes - Why? This was not the case before
h5py version: 3.1.0 (bug not present in 2.10.0)

During the transition from 3.10 to 2.10.0 my code stopped working due to an error that can be easily reproduced using the following code:

```
import h5py
import numpy as np

mask = np.array((0,1,0), dtype=bool)
d = np.arange(6).reshape((3,2))
print(d)
print(d[mask,:])  # works and prints [2 3] (middle row)
f = h5py.File("temp.h5py", "w")
f.create_dataset("temp", data=d)
print(f["temp"][:,:][mask,:])  # workaround, because the code below:
print(f["temp"][mask,:])  # does not work: TypeError: Indexing arrays must have integer dtypes
```

### PR DESCRIPTION
**Pull request by dguest**
This fixes #1847, where boolian indexing arrays weren't supported.

The following would crash

```python
import h5py
import numpy as np

mask = np.array((0,1,0), dtype=bool)
d = np.arange(6).reshape((3,2))
f = h5py.File("temp.h5py", "w")
f.create_dataset("temp", data=d)
print(f["temp"][mask,:])  # does not work: TypeError: Indexing arrays must have integer dtypes
```

The fix was very simple: I ported @takluyver's suggestion into the function that throws this exception, and had it convert bool arrays to indices using `array.nonzero()[0]`. I doubt this is the most elegant solution but it saves new users from having to learn a weird workaround for an inconsistency with numpy.

Fixes #1750


dguest:
I was having issues getting `tox` to work so I just pushed this. I'd hoped I'd get lucky, but seems not. I'll try to figure out how to run the failing test locally.

h5py/_selector.pyx
```diff
@@ -179,6 +179,12 @@ cdef class Selector:
                     a = np.asarray(a)
                 if a.ndim != 1:
                     raise TypeError("Only 1D arrays allowed for fancy indexing")
+                # Fix for issue #1847, not sure this is optimal but it works
+                if np.issubdtype(a.dtype, bool):
+                    if a.size != l:
+                        raise TypeError("Boolian index array must match dataset size")
```
takluyver, line 185:
```suggestion
                        raise TypeError("Boolean index array must match dataset size")
```

news/fix-for-issue-1847.rst
```diff
@@ -0,0 +1,4 @@
+Bug fixes
+---------
+
+* Fixed issue #1847: fancy indexing with boolian arrays was broken.
```
takluyver, line 4:
```suggestion
* Fixed issue #1847: fancy indexing with boolean arrays was broken.
```

h5py/_selector.pyx
```diff
@@ -179,6 +179,12 @@ cdef class Selector:
                     a = np.asarray(a)
                 if a.ndim != 1:
                     raise TypeError("Only 1D arrays allowed for fancy indexing")
+                # Fix for issue #1847, not sure this is optimal but it works
```
takluyver, line 182: I think this is probably fine. It's exactly what the code [was doing in h5py 2.x](https://github.com/h5py/h5py/blob/2.10.0/h5py/_hl/selections.py#L355-L358) before it broke, and I can't see an obvious better way.

h5py/_selector.pyx
```diff
@@ -179,6 +179,12 @@ cdef class Selector:
...
+                if np.issubdtype(a.dtype, bool):
```
takluyver, line 183:
```suggestion
                if a.dtype.kind == 'b':
```

This check seemed to be going wrong on some older numpy versions in CI. I don't know what the best way to write it is, but this suggestion is the same way we check for bool arrays elsewhere in the code: https://github.com/h5py/h5py/blob/3051c76afa73ad0c72d2c7c6724fe3617d2fefd8/h5py/_hl/selections.py#L62

h5py/_selector.pyx
dguest: should I remove this line?

dguest: I'm going to remove this.

dguest:
I'm not sure how [the error in this pipeline](https://dev.azure.com/h5pyappveyor/h5py/_build/results?buildId=1426&view=logs&j=0ad4bf5c-a1c6-59db-420d-7a6f497f570b&t=aca53dbd-2d58-55a9-ea01-6022315be503&l=8195) could be related to my MR
```
2022-04-01T10:17:47.8681040Z LD_LIBRARY_PATH="$LD_LIBRARY_PATH`echo  |                  \
2022-04-01T10:17:47.8699460Z 		sed -e 's/-L/:/g' -e 's/ //g'`"                               \
2022-04-01T10:17:47.8731980Z 	 ./H5make_libsettings  H5lib_settings.c  ||                               \
2022-04-01T10:17:47.8742650Z 	    (test $HDF5_Make_Ignore && echo "*** Error ignored") ||          \
2022-04-01T10:17:47.8752030Z 	    (rm -f H5lib_settings.c ; exit 1)
2022-04-01T10:17:48.0105240Z dyld: Library not loaded: libz.so.1.2.12
2022-04-01T10:17:48.0159980Z   Referenced from: /private/tmp/hdf5-1.12.1/src/./H5make_libsettings
2022-04-01T10:17:48.0170850Z   Reason: image not found
2022-04-01T10:17:48.0205860Z /bin/sh: line 1: 25618 Abort trap: 6           LD_LIBRARY_PATH="$LD_LIBRARY_PATH`echo  |                  	sed -e 's/-L/:/g' -e 's/ //g'`" ./H5make_libsettings H5lib_settings.c
2022-04-01T10:17:48.0207420Z make[2]: *** [H5lib_settings.c] Error 1
2022-04-01T10:17:48.0208210Z make[2]: *** Waiting for unfinished jobs....
2022-04-01T10:17:58.2470400Z make[1]: *** [all] Error 2
2022-04-01T10:17:58.2535150Z make: *** [all-recursive] Error 1
2022-04-01T10:17:58.2667660Z ##[error]Bash exited with code '2'.
```

takluyver:
Thanks! Don't worry, you didn't cause the CI failure, that's related to #2065 which I just merged. For some reason it passed in the PR branch, but then failed in master. I've asked the author of that PR to take a look.

The changes here look fine to me, but I'll wait to merge them until we've got the CI working properly.

tacaswell:
@dguest Could you also add a test?

dguest:
Sure, I'll look into a test soon. Might not be today.

takluyver:
Hmm, now another test seems to be failing:

```
_______________________ Test1DFloat.test_mask_wrongsize ________________________

self = <h5py.tests.test_dataset_getitem.Test1DFloat testMethod=test_mask_wrongsize>

    def test_mask_wrongsize(self):
        """ we require the boolean mask shape to match exactly """
>           self.dset[np.ones((2,), dtype='bool')]

py37-test-deps-tables/lib/python3.7/site-packages/h5py/tests/test_dataset_getitem.py:519: 
h5py/_objects.pyx:54: in h5py._objects.with_phil.wrapper
    ???
h5py/_objects.pyx:55: in h5py._objects.with_phil.wrapper
    ???
py37-test-deps-tables/lib/python3.7/site-packages/h5py/_hl/dataset.py:714: in __getitem__
    return self._fast_reader.read(args)
h5py/_selector.pyx:355: in h5py._selector.Reader.read
    ???

>   ???
E   IndexError: boolean index did not match indexed array

h5py/_selector.pyx:184: IndexError
```

dguest:
Ok, I had changed the exception to an `IndexError` for consistency with numpy here, but then keeping internal consistency in h5py might mean changing `TypeError` to `IndexError` in other places as well. 

What do the maintainers what to do here? I'm going back to raising a `TypeError` for now.

dguest:
Also should I move this test to `test_dataset_getitem`?

takluyver:
Thanks, I think `test_dataset_getitem` is probably the right place for it.

Yes, that's a tough one. The issue is that we have a separate special case for when you index with only a boolean array, as a mask for the dataset, and that already raises TypeError:

https://github.com/h5py/h5py/blob/50b6dc1c9208614c3dbba5ecc5f4d7607525fcd7/h5py/_hl/selections.py#L62-L65

I'd say internal consistency and stability win here, so we should stick to TypeError, as you've currently implemented. It's a shame not to match NumPy, though. I'll give it a while for others to chime in, though (cc @tacaswell @aragilar).

tacaswell:
We could do:

```python

class TypeIndexError(TypeError, IndexError):
    ...
```

which matches both.

dguest:
Ok I've moved the test into `test_dataset_getitem`. If everyone agrees to @tacaswell's suggestion I can have a go at implementing it, but I'm not sure how much is going to unravel if we start pulling on this thread. Given that, I might opt to leave the `TypeError`s in place for this MR.

takluyver:
Using multiple inheritance is clever. I think I'd keep it simple, though, leave them as TypeError for now, and maybe make a note that if we ever do h5py 4.0, we should change that to IndexError to match NumPy.

takluyver:
OK, I'm going to merge this with the exception as TypeError. If we want to do the TypeIndexError trick, we can always do it in a separate PR.

dguest:
Thanks a lot for merging this! I'm curious about the release plans:
 - There's not any chance of a "bugfix" release, i.e. h5py 3.6.1, is there? There is at least one library ([tensorflow](https://github.com/tensorflow/tensorflow/blob/25022463cef81cab811a781acaf8eb9eeb3d6cd3/tensorflow/tools/ci_build/release/requirements_common.txt#L9) that pins on 2.6.
 - When do we expect to tag 3.7.0? This seems to be the only issue on the milestone, is there more to do there?

takluyver:
We should probably do 3.7 soon - it's been a few months since 3.6, and there are a bunch of improvements queued up. Mac ARM wheels were recently added, which I think quite a few people will be happy about.

I'm generally against doing separate bugfix releases - it's extra work to maintain, and we're usually pretty careful about breaking changes, so it should be fine to upgrade to the latest feature release. It's not unusual these days for projects to pin specific versions but have some service like dependabot updating the pins to ensure that it works with new versions. And tensorflow's setup.py (which pip will use) [doesn't pin h5py](https://github.com/tensorflow/tensorflow/blob/25022463cef81cab811a781acaf8eb9eeb3d6cd3/tensorflow/tools/pip_package/setup.py#L84).

dguest:
ok, so things are better than I thought, thanks for correcting me and sorry for the noise.

takluyver:
No worries! I'm not saying you're completely wrong, there are different ways of getting tensorflow, and I don't know what the file you linked applies to. But my guess from a quick look is that it aims to be compatible with the latest h5py. :slightly_smiling_face:

dguest:
Nagging again about pushing a new release to pypi. 

I have a half-dozen distributed colleagues who just started working with h5py (and numpy!) and are running into this bug, I'd rather tell them to update than have them all implement the workaround.

PR #2080, opened 2022-04-02
Fix Mac x86 wheel hdf5 zlib build failure

matham:
After merging https://github.com/h5py/h5py/pull/2065, which built all the wheels successfully, master failed building mac x86 wheels.

I suspect the issue is that hdf5 1.12.1 does not build with latest zlib (1.2.12). However, why did it not fail in the PR? I think because when we added the changes switching zlib version from 1.2.11 to 1.2.11, the PR didn't rebuild hdf5 but used the previously cached hdf5 build. The reason is that I didn't add the new `configure_hdf5_mac.sh` to the hash.

This add it to the hash to force a rebuild. x86 should still fail. If it does, then I'll downgrade zlib to 1.2.11 and if does build (and does not use the cache), we'll know that was the cause and will have to wait for a newer hdf5 release before we can use the latest zlib.

### DIFF
--- a/ci/azure-pipelines-steps.yml
+++ b/ci/azure-pipelines-steps.yml
@@ -12,7 +12,7 @@ steps:
     ${{ if eq(parameters.platform, 'windows') }}:
       key: 0 | HDF5 | "$(Agent.OS)" | "$(Agent.OSArchitecture)" | "$(HDF5_VERSION)" | "$(HDF5_VSVERSION)" | "$(HDF5_MPI)" | ci/get_hdf5_win.py
     ${{ if ne(parameters.platform, 'windows') }}:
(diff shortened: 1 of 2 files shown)


matham:
Yeah, the [first run failed like master](https://dev.azure.com/h5pyappveyor/h5py/_build/results?buildId=1427&view=logs&j=0ad4bf5c-a1c6-59db-420d-7a6f497f570b&t=aca53dbd-2d58-55a9-ea01-6022315be503). But downgrading zlib now [builds](https://dev.azure.com/h5pyappveyor/h5py/_build/results?buildId=1429&view=logs&j=0ad4bf5c-a1c6-59db-420d-7a6f497f570b&t=aca53dbd-2d58-55a9-ea01-6022315be503) (and didn't reuse cache).

takluyver:
Thanks! It's a shame the newer version doesn't work, because it appears that it [fixes a security issue](https://github.com/madler/zlib/issues/605), but if it doesn't, it doesn't. I'll open an issue to try to help us remember to upgrade this.

PR #2082, opened 2022-04-03
Fix build building MacOS zlib 1.2.12

### LINKED ISSUE #2081: Upgrade bundled zlib to 1.2.12 wheels once HDF5 is compatible
In #2080, @matham found that HDF5 1.12.1 doesn't build with zlib 1.2.11. This is unfortunate, because there's a security issue fixed in zlib 1.2.12. Once there's a new release of HDF5, we should remember to upgrade the bundled zlib wherever we can - certainly this applies to at least the Mac wheels.

For Linux, it appears that we don't bundle zlib - it's now whitelisted in auditwheel as something you can assume will be on the system (pypa/auditwheel#152, pypa/auditwheel#334). So that's not our problem. :smiley: 

For Windows, we get zlib from [this nuget package](https://www.nuget.org/packages/zlib-msvc-x64/), which is still on 1.2.11. Then again, 1.2.12 has only been out a few days, so it may just be a matter of waiting.

matham:
Fixes #2081. 

The issue with zlib 1.2.12 was due to https://github.com/madler/zlib/issues/615. This fixes it in the same way fixed in zlib.

### DIFF
--- a/ci/azure-pipelines-wheels.yml
+++ b/ci/azure-pipelines-wheels.yml
@@ -7,9 +7,9 @@ steps:
     # Increment the first number to clear the cache if you change something
     # not captured by the key, e.g. dependencies.
     ${{ if eq(parameters.platform, 'windows') }}:
-      key: 0 | HDF5 | "$(Agent.OS)" | "$(Agent.OSArchitecture)" | "$(HDF5_VERSION)" | "$(HDF5_VSVERSION)" | ci/get_hdf5_win.py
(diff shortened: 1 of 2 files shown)


matham:
Weird, even though `ci/configure_hdf5_mac.sh` was changed, the cache was used.

matham:
Phew, the problem was that the hash key was not updated in all the files, only in https://github.com/h5py/h5py/blob/50b6dc1c9208614c3dbba5ecc5f4d7607525fcd7/ci/azure-pipelines-steps.yml#L12=.

matham:
Ok, done.

PR #2087, opened 2022-04-29
Test and make wheels with HDF5 1.12.2

takluyver:
Linux wheels should also use HDF5 1.12.2, because I updated it in the Docker images we use to build them: https://github.com/h5py/hdf5-manylinux/commit/8a6616972278f6b72de01a52ce25384d33da32fe

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -40,7 +40,7 @@ jobs:
   pool:
     vmImage: windows-2019
   variables:
-    HDF5_VERSION: 1.12.1
+    HDF5_VERSION: 1.12.2
     HDF5_DIR: $(HDF5_CACHE_DIR)/$(HDF5_VERSION)
     HDF5_VSVERSION: "16-64"
     CIBW_BUILD: cp3{7,8,9,10}-win_amd64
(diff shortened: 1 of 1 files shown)


takluyver:
@matham one of the patches you added appears to be failing to apply on HDF5 1.12.2 - could you take a look?

matham:
Ok, I will look into it.

takluyver:
I just noticed that the HDF Group's announcement of 1.12.2 includes:

> New supported platforms added is the macOS Apple M1 11.6 Darwin 20.6.0 arm64 with Apple clang version 12.0.5.

I don't know if that means anything should be easier - I don't know if support includes cross compiling.

matham:
Some PRs related to cross-compiling did recently get merged, but it doesn't yet support cross-compiling (https://github.com/HDFGroup/hdf5/issues/566#issuecomment-1111428129, https://github.com/HDFGroup/hdf5/issues/1203).

I have been giving it a little time to see whether the conda folks update the patches, because I'm a little weary about messing around with it, not fully understanding the details. But if there's no updates soon I'll update (or remove) the patches as needed.

takluyver:
Thanks!

PR #2090, opened 2022-05-05
Update File signature and argument descriptions in docs

sfalmo:
The h5py user guide lacks some arguments of the `File` constructor, especially the useful `locking` option (added in version 3.5). To make users aware of these options and to be consistent with the source code doc strings, the user guide should be updated.


takluyver:
This looks good, thanks!

PR #2091, opened 2022-05-06
Update eiger_use_case.py

marcocamma:
make example self consistent by creating files...


takluyver:
Thanks, this looks fine other than the style things in my suggestion, so I'll commit that and then merge it.

Issue #2055, opened 2022-02-14
H5Pset_efile_prefix is missing


mkitti:
I noticed that a low-level interface (and thus high-level interface) is missing for [`H5Pset_efile_prefix`](https://portal.hdfgroup.org/display/HDF5/H5P_SET_EFILE_PREFIX), which sets the prefix for external datasets. It also allows external datasets to be made relative to location of the HDF5 file. External datasets should be differentiated from external links, which are references to other HDF5 files.

The function was mentioned in https://github.com/h5py/h5py/issues/945#issuecomment-366472956 but ultimately not implemented in https://github.com/h5py/h5py/pull/1000.

There are similar low-level functions for [links ](https://api.h5py.org/h5p.html#h5py.h5p.PropLAID.set_elink_prefix) and [virtual ](https://api.h5py.org/h5p.html#h5py.h5p.PropDAID.set_virtual_prefix) datasets. Those correspond to [H5Pset_elink_prefix](https://portal.hdfgroup.org/display/HDF5/H5P_GET_ELINK_PREFIX) and [H5Pset_virtual_prefix](https://portal.hdfgroup.org/display/HDF5/H5P_SET_VIRTUAL_PREFIX).

aragilar:
I don't see any reason why we wouldn't accept a pull request add low-level wrappers around them. I'm less sure how best to expose them at the high level, but that can be discussed once there are low-level wrappers.

PR #2092, opened 2022-05-06
Add H5Pset_efile_prefix, H5Pget_efile_prefix

mkitti:
The primary purpose of this pull request is to add `H5Pget_efile_prefix` and `H5Pset_efile_prefix` to the low level API and expose this functionality to the high level API in `create_dataset` and `require_dataset`.
* Some buffers for reading the prefix were not initialized. This pull request initializing the first byte to NUL so that an empty string is correctly returned when the prefixes are not set.

Fixes https://github.com/h5py/h5py/issues/2055

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -37,7 +37,8 @@ def make_new_dset(parent, shape=None, dtype=None, data=None, name=None,
                   chunks=None, compression=None, shuffle=None,
                   fletcher32=None, maxshape=None, compression_opts=None,
                   fillvalue=None, scaleoffset=None, track_times=False,
-                  external=None, track_order=None, dcpl=None,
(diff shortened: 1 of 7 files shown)


aragilar:
This looks reasonable enough to me, but it would be good to work out why this is failing on Windows.

mkitti:
The error sounds like HDFFV-9716 and seems to only occur when using HDF5 1.8

https://github.com/HDFGroup/hdf5/blob/779ab44aa03aa351f01bddbfc4a565aebdcc15ae/release_docs/HISTORY-1_10.txt#L5591-L5600

mkitti:
Pinging @aragilar. Are there any other pending issues with this pull request?

aragilar:
Looks good to me, thanks for chasing down the Windows issue, and for the PR.

PR #2099, opened 2022-05-18
Bump to 1.12.2 and regenerate osx patches

matham:
Fixes the issues from https://github.com/h5py/h5py/pull/2087.

Closees #2087


matham:
Looks like I just needed to update the patch as the actual code didn't change much.

I'm guessing the cross compilation changes may not get into HDF5 until the next 1.14 release!?

azure-pipelines.yml
```diff
@@ -124,7 +124,7 @@ jobs:
       py39-deps-hdf51120:
```
takluyver, line 124:
```suggestion
      py39-deps-hdf51122:
```

takluyver:
Thanks @matham !

Issue #2100, opened 2022-05-18
MPI based h5py not compatible with python3.10 


jorgensd:
The version of mpi4py required by h5py [https://github.com/h5py/h5py/blob/master/setup.py#L48](3.0.3) [does not](https://github.com/mpi4py/mpi4py/blob/3.0.3/setup.py#L77) have Python 3.10 support.
Later versions of mpi4py supports Python 3.10, see: https://pypi.org/project/mpi4py/

Ubuntu 22.04 LTS is shipped with python 3.10 as default.

 * Ubuntu 22.04 LTS
 * Python version 3.10
 * Ubuntu package manager
 *  h5py version (e.g. 3.6.0)
 * HDF5 version (e.g. 1.10.7)
 * Minimal Docker environment:
```bash
FROM ubuntu:22.04

WORKDIR /tmp

ARG MPI="mpich"

RUN export DEBIAN_FRONTEND=noninteractive && \
    apt-get -qq update && \
    apt-get -yq --with-new-pkgs -o Dpkg::Options::="--force-confold" upgrade && \
    apt-get -y install \
	python3-dev \
    libhdf5-${MPI}-dev \
    lib${MPI}-dev \
    pkg-config \
    python3-pip && \
    apt-get -y install \
    git  && \
    apt-get clean && \
    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*


RUN HDF5_MPI="ON" HDF5_DIR="/usr/lib/x86_64-linux-gnu/hdf5/mpich/" CC=mpicc pip3 install --no-cache-dir --no-binary=h5py h5py 
```
* Trace :
Gives a very long error message, so I've only added the beginning, where it is clear that the issue is due to the restriction 

```bash
 Collecting h5py
   Downloading h5py-3.6.0.tar.gz (384 kB)
      ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 384.2/384.2 kB 70.6 MB/s eta 0:00:00
   Installing build dependencies: started
   Installing build dependencies: finished with status 'done'
   Getting requirements to build wheel: started
   Getting requirements to build wheel: finished with status 'done'
   Installing backend dependencies: started
   Installing backend dependencies: still running...
   Installing backend dependencies: still running...
   Installing backend dependencies: finished with status 'error'
   error: subprocess-exited-with-error

   × pip subprocess to install backend dependencies did not run successfully.
 │ exit code: 1
[...]

PR #2101, opened 2022-05-20
workaround mpi4py packaging issues on py3.10

julesghub:
Added a fix for #2100. 
Not the prettiest logic statements but this works with my py3.10 setup and should be compatible with python versions <3.10. Simple tests run fine, unfortunately `tox` testing produces errors with only py3.10.
I've added an entry note in the news folder.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,8 +46,7 @@ if setup_configure.mpi_enabled():
     RUN_REQUIRES.append('mpi4py >=3.0.2')
     SETUP_REQUIRES.append("mpi4py ==3.0.2; python_version<'3.8'")
     SETUP_REQUIRES.append("mpi4py ==3.0.3; python_version=='3.8.*'")
-    SETUP_REQUIRES.append("mpi4py ==3.0.3; python_version=='3.9.*'")
-    SETUP_REQUIRES.append("mpi4py ==3.1.3; python_version>='3.10'")
+    SETUP_REQUIRES.append("mpi4py ==3.1.0; python_version>='3.9'")
 
 # Set the environment variable H5PY_SETUP_REQUIRES=0 if we need to skip
 # setup_requires for any reason.
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,9 @@ SETUP_REQUIRES = []
 if setup_configure.mpi_enabled():
     RUN_REQUIRES.append('mpi4py >=3.0.2')
     SETUP_REQUIRES.append("mpi4py ==3.0.2; python_version<'3.8'")
-    SETUP_REQUIRES.append("mpi4py ==3.0.3; python_version>='3.8'")
+    SETUP_REQUIRES.append("mpi4py ==3.0.3; python_version=='3.8.*'")
+    SETUP_REQUIRES.append("mpi4py ==3.0.3; python_version=='3.9.*'")
+    SETUP_REQUIRES.append("mpi4py ==3.1.3; python_version>='3.10'")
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -45,7 +45,9 @@
 if setup_configure.mpi_enabled():
     RUN_REQUIRES.append('mpi4py >=3.0.2')
     SETUP_REQUIRES.append("mpi4py ==3.0.2; python_version<'3.8'")
-    SETUP_REQUIRES.append("mpi4py ==3.0.3; python_version>='3.8'")
+    SETUP_REQUIRES.append("mpi4py ==3.0.3; python_version=='3.8.*'")
+    SETUP_REQUIRES.append("mpi4py ==3.0.3; python_version=='3.9.*'")
+    SETUP_REQUIRES.append("mpi4py ==3.1.3; python_version>='3.10'")
```
tacaswell, line 50: This is the latest version version of mpi4py, is it really the minimum version we can use?

rbeucher: https://pypi.org/project/mpi4py/3.1.0/ 
Official support for python 3.10 (and 3.9 actually) was introduced with mpi4py 3.1.0
Python 3.9 is not listed as supported for version 3.0.3

@julesghub

rbeucher:
```
    SETUP_REQUIRES.append("mpi4py ==3.0.3; python_version>='3.8'")
    SETUP_REQUIRES.append("mpi4py ==3.1.0; python_version>='3.9'")
```
should be enough.

julesghub: Ok, I've set  it to use mpi4py 3.1.0 for py39 and py310.

tacaswell:
I changed the issue title to be more accurate.

This does seem to be the minimal change and the pragmatic thing to do but it is rather annoying.  This seems to be the interaction of two well intentioned things interacting badly to make h5py not install:

1. mpi4py shipping cython output in their sdist.  On one hand, it means you can skip having cython as a build/install time dependency, but the down side is exactly this.  Cython makes use of a bunch of details of the cpython implementation and is very sensitive to changes upstream (there are a handful of changes to CPython that break cython every minor release) and this leaves sdists that are uninstallable.
2. agressively pinning versions in setup.py for building forward-compatible wheels which is not really needed for someone building for their own purposes. 

Between now and the next release, I believe you can work around the problem by installing mpi4py from their repo (which will re-cythonize).

```
python -m pip install git+https://github.com/mpi4py/mpi4py@3.0.3
```

tacaswell (approved):

rbeucher:
Thanks @tacaswell @julesghub 
I completely agree. This is annoying.
I also think that this deserves a minor release as soon as possible.

tacaswell:
@takluyver How would you feel about doing a bug-fix release with only this change?

julesghub:
@tacaswell  - Thanks for accepting the change and totally agree a bug-fix release is needed for the py3.10 user base.

takluyver:
I never like doing 'this specific change' releases - there are a bunch of other changes that people are keen to see in a release, and I've been meaning to make one. So I'd like to do a regular 3.7 release rather than a special 3.6.1.

I'd been hoping to also update to HDF5 1.12.2, but there were issues with the OS X ARM thing. It looks like #2099 aims to resolve that, so I'll take a look at that today, and hopefully get a release out.

takluyver:
I'm getting the release notes for 3.7 ready, see #2103.

takluyver:
3.7 is out on PyPI

PR #2105, opened 2022-05-28
setup: Fix hdf5 dll search in mingw

Biswa96:
This fixes the following error in mingw environment:
Loading library to get build settings and version: hdf5.dll
error: Could not find module 'hdf5.dll' (or one of its dependencies).
Try using the full path with constructor syntax.

The error is fixed by checking if the python executable
is compiled in MSVC environment using sys.version. If not
then search for libhdf5-0.dll file which is the name of
hdf5 shared library in mingw environment.

### DIFF
--- a/setup_configure.py
+++ b/setup_configure.py
@@ -224,8 +224,12 @@ class HDF5LibWrapper:
             default_path = 'libhdf5.dylib'
             regexp = re.compile(r'^libhdf5.dylib')
         elif sys.platform.startswith('win'):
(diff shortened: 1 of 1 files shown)


takluyver:
Like with cygwin (PR #2038), I know basically nothing about mingw, it's not in our CI, and I don't really want to add it to CI (we've got a messy combination of platforms, architectures, and versions of HDF5, Python and NumPy to test already). So I'm happy to merge this, but without any particular promise that building or using h5py on mingw will keep working.

Issue #2110, opened 2022-06-04
Fatal Python error: Aborted when importing h5py Windows wheels


adamjstewart:
In [TorchGeo](https://github.com/microsoft/torchgeo), we use h5py to load sample HDF5 files in our unit tests. These tests are run on CI for Linux/macOS/Python. Since the new 3.7.0 release, our tests have been failing on Windows (for example, [here](https://github.com/microsoft/torchgeo/runs/6739917905?check_suite_focus=true)). I believe there may be an issue with the h5py 3.7.0 wheels on Windows. I don't have easy access to a Windows machine to test this, but I'm guessing something as simple as importing h5py will fail on Windows, at least with Python 3.10.

 * Operating System: Microsoft Windows Server 2022 (10.0.20348)
 * Python version: 3.10.4
 * Source: setup-python GitHub Action
 * h5py version: 3.7.0
 * HDF5 version: whatever comes with the 3.7.0 Windows wheels
 * The full traceback/stack trace shown:
 ```
Fatal Python error: Aborted
Current thread 0x000013c4 (most recent call first):
  File "<frozen importlib._bootstrap>", line 241 in _call_with_frames_removed
  File "<frozen importlib._bootstrap_external>", line 1184 in exec_module
  File "<frozen importlib._bootstrap>", line 688 in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1006 in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 1027 in _find_and_load
  File "<frozen importlib._bootstrap>", line 241 in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1078 in _handle_fromlist
  File "C:\Miniconda\envs\test\lib\site-packages\h5py\__init__.py", line 56 in <module>
  File "<frozen importlib._bootstrap>", line 241 in _call_with_frames_removed
  File "<frozen importlib._bootstrap_external>", line 883 in exec_module
  File "<frozen importlib._bootstrap>", line 688 in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1006 in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 1027 in _find_and_load
  File "C:\Miniconda\envs\test\lib\site-packages\_pytest\outcomes.py", line 288 in importorskip
  File "D:\a\torchgeo\torchgeo\tests\datasets\test_so2sat.py", line 18 in <module>
[...]

PR #2044, opened 2022-02-05
Fix fillvalue str

kmuehlbauer:
Builds on #1635, closes #941

Based on the work of @aragilar and the discussion in #941 I've added the necessary changes to get vlen string fillvalues working.

Update:

There was one test failing, but it looks like the `dtype` forwarding to `set_fill_value` make that working too. So I adapted that test. Meanwhile this is settled to prior behvior.

The changes in this PR allow string fillvalues to work for vlen strings as well as fixed size strings regardless of the encoding. One source of misunderstanding for the user could be that `bytes` and `ascii` strings have the same internal representation and will not roundtrip (since `.asstr` can't be facilitated). `fillvalue` will be returned as `bytes`. 

The following works with the current implementation.

```python
with h5py.File(f"test_xxx.h5", "w") as f:   
    dt = h5py.string_dtype()
    fill_value = 'bár'
    ds = f.create_dataset('x', (10,), dtype=dt, fillvalue=fill_value)
    assert f['x'][0] == fill_value.encode("utf-8")
    assert f['x'].asstr()[0] == fill_value
    assert f['x'].fillvalue == fill_value.encode("utf-8")

with h5py.File(f"test_xxx.h5", "w") as f:   
    dt = h5py.string_dtype("ascii")
    fill_value = b'bar'
    ds = f.create_dataset('x', (10,), dtype=dt, fillvalue=fill_value)
    assert f['x'][0] == fill_value
    assert f['x'].asstr()[0] == fill_value.decode()
    assert f['x'].fillvalue == fill_value

with h5py.File(f"test_xxx.h5", "w") as f:   
    dt = h5py.string_dtype(length=10)
    fill_value = 'bár'
    ds = f.create_dataset('x', (10,), dtype=dt, fillvalue=fill_value)
    assert f['x'][0] == fill_value.encode("utf-8")
    assert f['x'].asstr()[0] == fill_value
    assert f['x'].fillvalue == fill_value.encode("utf-8")

with h5py.File(f"test_xxx.h5", "w") as f:   
    dt = h5py.string_dtype("ascii", length=10)
    fill_value = b'bar'
    ds = f.create_dataset('x', (10,), dtype=dt, fillvalue=fill_value)
    assert f['x'][0] == fill_value
    assert f['x'].asstr()[0] == fill_value.decode("ascii")
    assert f['x'].fillvalue == fill_value
```

Issue #941, opened 2017-10-24
Setting fillvalue for datasets with variable length string dtype


shoyer:
This does not appear to work:
```
In [34]: f = h5py.File('varlenstr.h5')

In [35]: vlen_str = h5py.special_dtype(vlen=str)

In [36]: f.create_dataset('foo', (2,), dtype=vlen_str, fillvalue='bar')
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-36-ace807cb8186> in <module>()
----> 1 f.create_dataset('foo', (2,), dtype=vlen_str, fillvalue='bar')

~/conda/envs/xarray-netcdf4-py36/lib/python3.6/site-packages/h5py/_hl/group.py in create_dataset(self, name, shape, dtype, data, **kwds)
    106         """
    107         with phil:
--> 108             dsid = dataset.make_new_dset(self, shape, dtype, data, **kwds)
    109             dset = dataset.Dataset(dsid)
    110             if name is not None:

~/conda/envs/xarray-netcdf4-py36/lib/python3.6/site-packages/h5py/_hl/dataset.py in make_new_dset(parent, shape, dtype, data, chunks, compression, shuffle, fletcher32, maxshape, compression_opts, fillvalue, scaleoffset, track_times)
    119     if fillvalue is not None:
    120         fillvalue = numpy.array(fillvalue)
--> 121         dcpl.set_fill_value(fillvalue)
    122
    123     if track_times in (True, False):

h5py/_objects.pyx in h5py._objects.with_phil.wrapper (/Users/ilan/minonda/conda-bld/h5py_1496887972496/work/h5py/_objects.c:2846)()

h5py/_objects.pyx in h5py._objects.with_phil.wrapper (/Users/ilan/minonda/conda-bld/h5py_1496887972496/work/h5py/_objects.c:2804)()

h5py/h5p.pyx in h5py.h5p.PropDCID.set_fill_value (/Users/ilan/minonda/conda-bld/h5py_1496887972496/work/h5py/h5p.c:5073)()

h5py/h5t.pyx in h5py.h5t.py_create (/Users/ilan/minonda/conda-bld/h5py_1496887972496/work/h5py/h5t.c:17938)()

h5py/h5t.pyx in h5py.h5t.py_create (/Users/ilan/minonda/conda-bld/h5py_1496887972496/work/h5py/h5t.c:17897)()

TypeError: No conversion path for dtype: dtype('<U3')
```

[...]

PR #2111, opened 2022-06-10
Add tests for fixed length strings with fillvalues

aragilar:
I suspect the support for adding vlen strings may have broken fixed length
strings. These tests do the same thing as the vlen strings, but use
fixed length dtypes.

This may be related to #2110 (in that I've seen segfaults), but possibly not (in that I'm seeing issues on linux).

### DIFF
--- a/h5py/h5p.pyx
+++ b/h5py/h5p.pyx
@@ -547,6 +547,14 @@ cdef class PropDCID(PropOCID):
         if string_info is not None and string_info.length is None:
             tid = py_create(value.dtype, logical=1)
             ret = H5Pget_fill_value(self.id, tid.id, &c_ptr)
+            if c_ptr == NULL:
+                # If the pointer is NULL (either the value did not get changed,
+                # or maybe the 0 length string, it's unclear currently), if
+                # PyBytes_FromString is called on the pointer, we get a
+                # segfault. If we set the value to empty bytes, then we
+                # shouldn't segfault.
+                value[0] = b""
+                return
             fill_value = c_ptr
             value[0] = fill_value
             return
--- a/h5py/tests/test_dataset.py
+++ b/h5py/tests/test_dataset.py
@@ -1141,6 +1141,15 @@ class TestStrings(BaseDataset):
         self.assertEqual(string_info.encoding, 'ascii')
         self.assertEqual(string_info.length, 10)
 
+    def test_fixed_bytes_fillvalue(self):
+        """ Vlen bytes dataset handles fillvalue """
+        dt = h5py.string_dtype(encoding='ascii', length=10)
+        fill_value = b'bar'
+        ds = self.f.create_dataset('x', (100,), dtype=dt, fillvalue=fill_value)
+        self.assertEqual(self.f['x'][0], fill_value)
+        self.assertEqual(self.f['x'].asstr()[0], fill_value.decode())
+        self.assertEqual(self.f['x'].fillvalue, fill_value)
+
     def test_fixed_utf8(self):
         dt = h5py.string_dtype(encoding='utf-8', length=5)
         ds = self.f.create_dataset('x', (100,), dtype=dt)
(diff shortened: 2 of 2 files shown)


aragilar:
It looks like https://github.com/h5py/h5py/blob/master/h5py/h5p.pyx#L551 can sometime cause a segfault when `c_ptr` is `NULL`. It's unclear to me why `c_ptr` is `NULL` only some of the time (maybe we should be checking whether there is a fill value using https://github.com/h5py/h5py/blob/master/h5py/h5p.pyx#L558...), as even with the same file, I get segfaults only in some cases...

@adamjstewart @p-e-s Are you able to check if my fix in https://github.com/h5py/h5py/pull/2111/commits/735c2b959c3d08d7bd309c10e781433bed59150e fixes the issues you're seeing?

adamjstewart:
Not really sure how to test because I don't have easy access to Windows other than GitHub actions and I don't know how to use h5py on Windows without a pre-compiled HDF5 stuffed into a wheel.

aragilar:
https://artprodsu6weu.artifacts.visualstudio.com/Ad2f448a4-b018-4734-be80-3759b5f99ba9/347aa760-c0ad-4566-98fe-98d2281065ee/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL2g1cHlhcHB2ZXlvci9wcm9qZWN0SWQvMzQ3YWE3NjAtYzBhZC00NTY2LTk4ZmUtOThkMjI4MTA2NWVlL2J1aWxkSWQvMTQ4Ni9hcnRpZmFjdE5hbWUvd2luZG93c193aGVlbHM1/content?format=zip are the windows wheels from this build.

adamjstewart:
Is there a link to get just a single wheel?

p-e-s:
Same for me, i would need a wheel for Ubuntu to be able to test with my issue.

takluyver:
Close/reopen to re-run tests

takluyver:
OK, Travis didn't re-run the tests, but the failure there is not an issue with this.

The added tests look fine to me. The pointer stuff involved in the Cython code makes my head hurt a bit - a NULL check probably makes sense, but I can't tell if it's exactly correct. Is this note [in the HDF5 docs](https://portal.hdfgroup.org/display/HDF5/H5P_GET_FILL_VALUE) relevant?

> If the fill value is undefined, i.e., set to NULL in the property list, H5P_GET_FILL_VALUE will return an error. [H5P_FILL_VALUE_DEFINED](https://portal.hdfgroup.org/display/HDF5/H5P_FILL_VALUE_DEFINED) should be used to check for this condition before H5P_GET_FILL_VALUE is called.

It would still be good if someone with access to a Windows machine could see if the wheels on Azure for this PR make any difference to the issue seen there.

aragilar:
The tests are somewhat unrelated: they didn't fail when I expected them to, but they're checking that we roundtrip fine anyway. I expect the automated wrapper around `H5Pget_fill_value` to check the return value and throw an exception, so I don't think we need to check the return value again (but I'm not entirely sure about that). I know `NULL`s can appear, and `PyBytes_FromString` causes segfaults when passed `NULL` (it's unclear to me why it doesn't check, possibly it's an optimisation and there's a safe function which should be called).

Helveg:
I'm a willing Windows volunteer. Steps to execute?

takluyver:
Thanks! The build has been cleaned up now - I'll close and reopen the PR to launch it again, and then give some brief instructions.

takluyver:
OK, the new build page on Azure pipelines is https://dev.azure.com/h5pyappveyor/h5py/_build/results?buildId=1512&view=results (also accessible from the 'checks' links on this PR. Once it's built (in 15 minutes or so), there will be a link saying something like "4 artifacts". Follow that, and download the Windows wheels as a zip file.

Unpack the zip file and identify the wheel for the version of Python you're testing. E.g. for Python 3.9, it will have `cp39` in the filename. Use pip to install this file (you may want to do this in a new virtualenv or conda env - up to you). Then try to do whatever was causing an issue on Windows with the released version, and see if it's working now.

Helveg:
Ok, I am not part of the issue report so I don't know:

>  Then try to do whatever was causing an issue on Windows with the released version, and see if it's working now.

Any idea?

takluyver:
#2110 appears to describe an issue just importing it, with h5py 3.7.0. But if you don't see the issue to start with, then you won't really be able to tell if this fixes it. It's clearly not failing in all cases on Windows, as we've got Windows CI that's working. :confused:  Maybe this is unrelated to the Windows issue on import.

@aragilar should we add a test for the case that was (might have been?) failing, where you retrieve an unset fill value from a dataset with fixed-length string dtype?

aragilar:
Yep, I thought the tests were testing the error I saw, but apparently not, been busy with other things, but planning to come back to this soon.

asmeurer:
FWIW, in versioned-hdf5 we did experience the segfault from #2110 on, which this PR fixes. However, either this PR or that one did change the behavior of the fillvalue for fixed length string datasets, which we've had to work around here https://github.com/deshaw/versioned-hdf5/pull/235. Happy to provide more details and simplified reproducers if it helps.

takluyver:
With #2132 just merged (containing the same change to the actual h5py code), this is now only adding some tests, so I'll merge this as well.

@asmeurer you mentioned some change in behaviour for fill values for vlen strings. #2044 (in h5py 3.7) fixed supplying a fill value for vlen strings, which had been impossible for a long time (#941). Tests were added as part of that, and they're still passing. If something has got broken recently, could you open a new issue with an example?

PR #2113, opened 2022-06-13
Investigating hanging on Travis CI

takluyver:
This is probably not meant to be merged, I'm just poking Travis CI to figure out why builds are hanging.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,7 +63,7 @@ script:
     - tox $TOX_OPTS
 
 after_success:
-    #- python ci/upload_coverage.py
+    - python ci/upload_coverage.py
 
   # trigger an upload to the shared ecosystem
   # infrastructure at: https://anaconda.org/scipy-wheels-nightly
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,7 +63,7 @@ script:
     - tox $TOX_OPTS
 
 after_success:
-    - python ci/upload_coverage.py
+    #- python ci/upload_coverage.py
 
   # trigger an upload to the shared ecosystem
   # infrastructure at: https://anaconda.org/scipy-wheels-nightly
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,7 +59,7 @@ install:
     - if [[ $CIBW_ARCHS == aarch64 ]]; then python -m pip install cibuildwheel; fi
 
 script:
-    - if [[ $CIBW_ARCHS == aarch64 ]]; then python3 -m cibuildwheel --output-dir wheelhouse; fi
+    #- if [[ $CIBW_ARCHS == aarch64 ]]; then python3 -m cibuildwheel --output-dir wheelhouse; fi
     - tox $TOX_OPTS
 
 after_success:
--- a/ci/upload_coverage.py
+++ b/ci/upload_coverage.py
@@ -8,7 +8,7 @@ from os import chdir, listdir, environ
 from pathlib import Path
 import platform
 from pprint import pprint
-from subprocess import run, PIPE, CalledProcessError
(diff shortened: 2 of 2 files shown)


takluyver:
OK, I think I've figured it out. The codecov uploader gets an upload URL from codecov which is either on S3 (according to the logs) or Google Cloud (according to https://github.com/seantis/pytest-codecov/blob/master/src/pytest_codecov/codecov.py ). It tries to upload the coverage XML file to there with a PUT request. This appears to be blocked on the arm64 machines, in a way that causes the `.connect()` call on the socket to hang rather than failing. I might have a wild guess that this is part of some effort to mitigate cryptocurrency mining on free CI resources (so once again, crypto is the reason we can't have nice things :angry: ).

I thought about moving to the new codecov uploader, but if my guess above is correct, this wouldn't fix it - a connection to either S3 or google cloud is not going to work from typescript if it fails from Python. And the new uploader doesn't have an arm64 binary anyway (https://github.com/codecov/uploader/issues/523 ).

I suspect there's no way round that, so all we can really do is ignore the failure, or skip codecov altogether on the arm64 jobs. :disappointed: 

-----

Traceback:

```
    Pinging Codecov...
    Uploading to S3...
  File "/home/travis/virtualenv/python3.7.13/bin/codecov", line 8, in <module>
    sys.exit(main())
  File "/home/travis/virtualenv/python3.7.13/lib/python3.7/site-packages/codecov/__init__.py", line 1177, in main
    "Content-Encoding": "gzip",
  File "/home/travis/virtualenv/python3.7.13/lib/python3.7/site-packages/codecov/__init__.py", line 301, in retry_upload
    res = request_method(url, **kwargs)
  File "/home/travis/virtualenv/python3.7.13/lib/python3.7/site-packages/requests/api.py", line 130, in put
    return request("put", url, data=data, **kwargs)
  File "/home/travis/virtualenv/python3.7.13/lib/python3.7/site-packages/requests/api.py", line 59, in request
    return session.request(method=method, url=url, **kwargs)
  File "/home/travis/virtualenv/python3.7.13/lib/python3.7/site-packages/requests/sessions.py", line 587, in request
    resp = self.send(prep, **send_kwargs)
  File "/home/travis/virtualenv/python3.7.13/lib/python3.7/site-packages/requests/sessions.py", line 701, in send
    r = adapter.send(request, **kwargs)
  File "/home/travis/virtualenv/python3.7.13/lib/python3.7/site-packages/requests/adapters.py", line 499, in send
    timeout=timeout,
  File "/home/travis/virtualenv/python3.7.13/lib/python3.7/site-packages/urllib3/connectionpool.py", line 710, in urlopen
    chunked=chunked,
  File "/home/travis/virtualenv/python3.7.13/lib/python3.7/site-packages/urllib3/connectionpool.py", line 386, in _make_request
    self._validate_conn(conn)
  File "/home/travis/virtualenv/python3.7.13/lib/python3.7/site-packages/urllib3/connectionpool.py", line 1040, in _validate_conn
    conn.connect()
  File "/home/travis/virtualenv/python3.7.13/lib/python3.7/site-packages/urllib3/connection.py", line 358, in connect
    self.sock = conn = self._new_conn()
  File "/home/travis/virtualenv/python3.7.13/lib/python3.7/site-packages/urllib3/connection.py", line 175, in _new_conn
    (self._dns_host, self.port), self.timeout, **extra_kw
  File "/home/travis/virtualenv/python3.7.13/lib/python3.7/site-packages/urllib3/util/connection.py", line 85, in create_connection
    sock.connect(sa)
KeyboardInterrupt
```

takluyver:
I've rebased the changes to work around this as #2114.

PR #2114, opened 2022-06-14
Add timeout for uploading coverage data to codecov

takluyver:
This means the `upload_coverage.py` script fails rather than hanging. The call to `codecov` gets a maximum of 30 seconds to complete, which should be plenty.

See my investigation in #2113 for details of what I think is going on.


takluyver:
I'm going to merge this so we have CI working again; it doesn't touch anything in the actual h5py library.

PR #2116, opened 2022-06-26
Validate `maxshape` in `Group.require_dataset`

### LINKED ISSUE #2018: Exact shape matching hinders `require_dataset` usecase
What exactly is the usecase of `require_dataset` if one must go check the exact size of it to prevent these errors:

```
TypeError: Shapes do not match (existing (1, 3) vs new (0, 3))
```

In order to be able to satisfy that constraint you have to go check the existence of the dataset and retrieve its shape, what purpose could a call to `require_dataset` even serve after you've already checked existence?

I'd expect to be able to use `require_dataset` to retrieve or create similar to `require_group`:

```
ds = grp.require_dataset("name", some=args, that=only, matter=for_create)
safely_use(ds)
```

Helveg:
This PR validates the `maxshape` over `shape` in `Group.require_dataset` so that resizable datasets can be required without knowing their exact shape on resizable axes. closes #2018

PS: Sorry for the live testing, can't run `tox` due to an issue with my WSL setup.

### DIFF
--- a/h5py/_hl/group.py
+++ b/h5py/_hl/group.py
@@ -20,7 +20,7 @@ from .compat import filename_decode, filename_encode
 from .. import h5, h5g, h5i, h5o, h5r, h5t, h5l, h5p, h5s, h5d
 from . import base
 from .base import HLObject, MutableMappingHDF5, phil, with_phil
-from . import dataset
+from . import datasetu
 from . import datatype
 from .vds import vds_support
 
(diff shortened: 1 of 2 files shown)


takluyver:
The code looks OK to me, thanks. Can I ask you to update the docs - both the docstring and the reference in [`dataset.rst`](https://github.com/h5py/h5py/blob/master/docs/high/dataset.rst) - and [add a news entry](https://docs.h5py.org/en/stable/contributing.html#write-a-release-note) describing the change?

Helveg:
I updated the docstring, but I don't see which parts of `dataset.rst` involve any arguments of `Group.require_dataset`.

takluyver:
Gah, it seems like Ctrl-Enter on Github means 'comment & close' today. I hope that's a bug, or I'm going to be unintentionally closing a lot of things.

takluyver:
Thanks, I reran it and it passed now.

takluyver:
Thanks @Helveg !

PR #2119, opened 2022-07-05
fix regression: python range not accepted as FancySelection

### LINKED ISSUE #2118: regression: python range not accepted as FancySelection
I have an old project that relies on using python's built-in ranges as indices/selections.
For h5py versions < 3.0 this worked without an issue but since 3.0 (including latest master 50dfb96) it causes the reported crash (see end of the issue).

I believe the breaking change happened in #1526 - [here](https://github.com/h5py/h5py/commit/474ef53d2698f6fa8988df2a2ec99525863bc8c4#diff-a09e4221f2b07266b57d60d5bfd924721adbf123e85a8dd4222b4c5079898de6L78-R85) to be precise.
Previously, an internal exception was triggered and `args[0] == range(..)` was parsed by a `FancySelection`.
Now, the parsing is done by the `Selector` which [does not account for ranges anywhere](https://github.com/h5py/h5py/blob/474ef53d2698f6fa8988df2a2ec99525863bc8c4/h5py/_selector.pyx#L64-L206).
This is not listet as a breaking change in the [changelog](https://docs.h5py.org/en/latest/whatsnew/3.0.html#breaking-changes-deprecations).

A quick fix would be to do something like:
```diff
diff --git a/h5py/_selector.pyx b/h5py/_selector.pyx
index 1b06e75e..83c400fe 100644
--- a/h5py/_selector.pyx
+++ b/h5py/_selector.pyx
@@ -169,8 +169,8 @@ cdef class Selector:
                 continue
 
             # [[0, 2, 10]] - list/array of indices ('fancy indexing')
-            if isinstance(a, (list, np.ndarray)):
-                if isinstance(a, list) and len(a) == 0:
+            if isinstance(a, (list, range, np.ndarray)):
+                if isinstance(a, (list, range)) and len(a) == 0:
                     a = np.asarray(a, dtype=np.intp)
                 else:
                     a = np.asarray(a)
```
<hr>

 * OS: Debian Buster
 * Python version 3.7.3 (from system package manager)
 * h5py version 3.0.0 and 3.7.0
 * HDF5 version 1.10.4+repack-10 (from package manager)
 * The test file `test.py` (adapted from [docs for Datasets](https://docs.h5py.org/en/3.0.0/high/dataset.html#reading-writing-data)) and traceback:

```python
import h5py

print(h5py.__version__)

f = h5py.File('mytestfile.hdf5', 'w')
dset = f.create_dataset("MyDataset", (10,10,10), 'f')
dset[range(0, 4)]
```

```python
>>> python3 test.py 
3.0.0
  File "test.py", line 7, in <module>
    dset[range(0, 4)]
  File "/usr/local/lib/python3.7/dist-packages/h5py/_hl/dataset.py", line 774, in __getitem__
  File "/usr/local/lib/python3.7/dist-packages/h5py/_hl/selections.py", line 82, in select
  File "h5py/_selector.pyx", line 269, in h5py._selector.Selector.make_selection
  File "h5py/_selector.pyx", line 210, in h5py._selector.Selector.apply_args
TypeError: Simple selection can't process range(0, 4)
 ```

### PR DESCRIPTION
**Pull request by ezavod**
Fixes #2118 and adds a testcase.

### DIFF
--- a/h5py/tests/test_selections.py
+++ b/h5py/tests/test_selections.py
@@ -105,6 +105,10 @@ class TestSelection(BaseSelection):
         st = sel.select((10,), list([1,2,3]), dset)
         self.assertIsInstance(st, sel.FancySelection)
 
+        # args is range, return a FancySelection
+        st = sel.select((10,), range(1, 4), dset)
+        self.assertIsInstance(st, sel.FancySelection)
+
         # args is a Boolean mask, return a PointSelection
         st1 = sel.select((5,), np.array([True,False,False,False,True]), dset)
         self.assertIsInstance(st1, sel.PointSelection)
--- a/h5py/_selector.pyx
+++ b/h5py/_selector.pyx
@@ -171,9 +171,9 @@ cdef class Selector:
 
                 continue
 
-            # [[0, 2, 10]] - list/array of indices ('fancy indexing')
-            if isinstance(a, (list, np.ndarray)):
-                if isinstance(a, list) and len(a) == 0:
+            # [[0, 2, 10]] - list/range/array of indices ('fancy indexing')
+            if isinstance(a, (list, range, np.ndarray)):
+                if isinstance(a, (list, range)) and len(a) == 0:
                     a = np.asarray(a, dtype=np.intp)
                 else:
                     a = np.asarray(a)
--- a/h5py/tests/test_dataset_getitem.py
+++ b/h5py/tests/test_dataset_getitem.py
@@ -37,6 +37,7 @@
         MultiBlockSlice
         Indexing
         Index list
+        Index range
         Boolean mask
         Field names
 """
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

File: h5py/tests/test_selections.py [hunk 1]
```diff
@@ -105,6 +105,10 @@ def test_selection(self):
         st = sel.select((10,), list([1,2,3]), dset)
         self.assertIsInstance(st, sel.FancySelection)
 
+        # args is range, return a FancySelection
+        st = sel.select((10,), range(1, 4), dset)
+        self.assertIsInstance(st, sel.FancySelection)
```
ezavod, line 110: @tacaswell Is this sufficient to check that this matches the behavior of numpy?

tacaswell: No, I think this is an internal detail.  The thing we want to track is that h5py does what ever numpy does.  There is a bunch of tests in `test_dataset_getitem.py` that test different things tat can be thrown at `__getitem__` and checks that the behavior matches numpy.

I think this probably should also get tests for more than one dimension slicing.

ezavod: Thanks for pointing me to the right file because I missed these tests. It'd be too simple otherwise.

aragilar:
Adding this here so the next time we look here we're not confused why we didn't merge this: we're still arguing if whether is this the right fix (or should we do something with sequences) on #2118.

Issue #2126, opened 2022-07-30
Support for configuring chunk cache per dataset


ajelenak:
Configuring dataset chunk cache is currently only supported at the file level, which means the same cache settings apply to all datasets in a file. This may cause excessive memory consumption for large number of datasets. I propose to allow configuring chunk cache per dataset as well. These settings override the file-level ones.

These three methods: `Dataset.open_dset()`, `Group.require_dataset()`, and `Group.create_dataset()`, should get `rdcc_nslots`, `rdcc_nbytes`, and `rdcc_w0` keyword arguments. Then modify `Dataset.make_new_dset()` accordingly.

Is this proposal acceptable? If yes, I will create a PR to track code changes.

tacaswell:
Yes, seems reasonable to me.

ajelenak:
Closed via #2127.

PR #2127, opened 2022-08-01
Support chunk cache configuration per dataset

ajelenak:
As explained in #2126, this PR allows chunk cache configuration for each dataset. This is achieved via `Group.create_dataset()` for new datasets or `Group.require_dataset()` for already existing datasets. The relevant documentation was updated and there is a new test for both of these cases.

Working on the documentation, I noticed that currently the chunk cache explanation is in the [File section](https://docs.h5py.org/en/stable/high/file.html#chunk-cache). I think a more obvious place should be in the Dataset section.

### DIFF
--- a/h5py/_hl/group.py
+++ b/h5py/_hl/group.py
@@ -283,8 +283,6 @@ class Group(HLObject, MutableMappingHDF5):
             except KeyError:
                 dset = self[name]
                 raise TypeError("Incompatible object (%s) already exists" % dset.__class__.__name__)
-            except Exception:
-                raise
 
             if shape != dset.shape:
                 if "maxshape" not in kwds:
--- a/h5py/_hl/group.py
+++ b/h5py/_hl/group.py
@@ -39,13 +39,11 @@ class Group(HLObject, MutableMappingHDF5):
                 raise ValueError("%s is not a GroupID" % bind)
             super().__init__(bind)
 
-
     _gcpl_crt_order = h5p.create(h5p.GROUP_CREATE)
     _gcpl_crt_order.set_link_creation_order(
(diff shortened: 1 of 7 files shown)


h5py/_hl/group.py
```diff
@@ -258,7 +283,7 @@ def require_dataset(self, name, shape, dtype, exact=False, **kwds):
             except KeyError:
                 dset = self[name]
                 raise TypeError("Incompatible object (%s) already exists" % dset.__class__.__name__)
-            except:
+            except Exception:
                 raise
```
aragilar, line 287: Why are we catching and reraising exceptions here, and is switching to capturing only some now safe (as not everything subclasses from `Exception`)? Can we just delete these two lines?

aragilar:
This looks good to me, though I'm slightly concerned about the exception catching above (it's probably nothing to worry about though).

h5py/_hl/group.py
ajelenak: Explicitly stating `Exception` is considered a better practice. This will not capture only _SystemExit_, _KeyboardInterrupt_, and _GeneratorExit_.

I am okay with removing this except clause as it seems redundant.

ajelenak: Removed it in https://github.com/h5py/h5py/pull/2127/commits/35413a4ba7e447c7a85cbe390e3e87b1273c215c.

ajelenak:
I've just added a news item for this PR. Anything else needed before merge?

ajelenak:
I will merge this on 2022-09-16 since there are no outstanding change requests or unaddressed comments.

PR #2132, opened 2022-08-14
Test default fill values for some common types

aragilar:
This continues on from #2111, and the first commit actually segfaults like I expect it to.

Adding my string pointer check fixes the segfault (the tests are currently failing as I'm returning bytes from the function, but expecting strings—we need to decide which is the correct default and either change the test or the code to match). The changes in #2111 should also be included, but those tests didn't segfault, hence why I've opened this as a new PR (so its easy to confirm that both the fillvalues segfault, and the additional null pointer check fixes it).

### DIFF
--- a/h5py/h5p.pyx
+++ b/h5py/h5p.pyx
@@ -547,6 +547,14 @@ cdef class PropDCID(PropOCID):
         if string_info is not None and string_info.length is None:
             tid = py_create(value.dtype, logical=1)
             ret = H5Pget_fill_value(self.id, tid.id, &c_ptr)
+            if c_ptr == NULL:
+                # If the pointer is NULL (either the value did not get changed,
+                # or maybe the 0 length string, it's unclear currently), if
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks @aragilar !

> the tests are currently failing as I'm returning bytes from the function, but expecting strings—we need to decide which is the correct default 

Fill values are dataset specific, aren't they? Attributes are always fully allocated, so there's no fill value for unallocated space, IIRC.

If that is the case, then I'd say returning bytes is correct, because that matches what you get when reading from the dataset. We could add `dset.asstr().fillvalue` to get the fill value as str, but I wouldn't say it's high priority, because it's not much shorter than `dset.fillvalue.decode()`.

aragilar:
I've switched the tests to expect `b''`, which I guess makes sense as if we should be decoding `fillvalues`, we should be doing it in the high level code, not in the low-level wrapper.

takluyver:
Thanks @aragilar

PR #2134, opened 2022-08-18
Fix issues with vlen fields and compound types

### LINKED ISSUE #596: Multiple special_type(vlen) in single dtype (in attribute) => python crash
Hi everyone, 
this is my first post, so please be gentle =)
The issue arises when creating a dtype with multiple special_dtypes(vlen=str/np.X) next to each other.
Python simply crashes with something like:

> **\* Error in `python3': free(): invalid next size (fast): 0x00000000015f4c00 ***

I tried it on multiple machines (with python34):
Worked: Windows7_64 with hdf5-1.8.14 and h5py-2.5
Failed: Windows7_64 with hdf5-1.8.13 or hdf5-1.8.15 and h5py-2.5, Debian Jessie with hdf5-1.8.13 or 1.8.14 and h5py-2.4 or h5py-2.5

Here some code to provoke the error:

``` python
import numpy as np
import h5py as h5
print(h5.version.hdf5_version)
print(h5.version.version)

vls = h5.special_dtype(vlen=str)
vli = h5.special_dtype(vlen=np.uint64)
dt = np.dtype([('a', vls), ('b', vli)])
data = np.empty(2, dtype=dt)
data[0]['a'] = 'asdasd'
data[1]['a'] = 'asdasdasdasd'
data[0]['b'] = np.array(np.arange(100))
data[1]['b'] = np.array((0,))
f = h5.File('test.h5')
ds = f.require_dataset('aaa', shape=(100,), dtype=np.uint64)
ds.attrs['asd'] = data
ds.attrs.create('asda', data, shape=(2,), dtype=dt)
f.flush()
```

I saw that there were other recent issues regarding vlen data, maybe they are related?
Cheers,
Marius

**Edit**: Error only appears only when creating attributes!

### LINKED ISSUE #623: Segmentation Fault when reading VLEN data
I was very happy to learn that variable-length datatype support was recently added to `h5py`, since this makes HDF5 a viable format for our data. I'd like to be able to to create data structures in C++ and then read them out with `h5py`.

Unfortunately, right now I'm getting a segfault when I try to read the files with `h5py`. It seems to happen consistently on both OSX and Ubuntu, and only seems to be an issue when I include VLEN data within compound datatypes. More details are provided on the [mailing list](https://groups.google.com/d/topic/h5py/WLQlEr-VDtk/discussion). Unfortunately I've been unable to reproduce this with `h5py` alone, since I'm [also having trouble writing out files with the problematic structure](http://stackoverflow.com/questions/32669875/does-h5py-support-arbitrary-nested-compound-datatypes). 

I'd consider this a bug, since I don't usually expect python libraries to segfault when reading in data files.

Unfortunately we're dead in the water (or reverting to JSON) unless we can get this working. I've tried `PyTables`, but it flat refuses to read VLEN data. Since there are proposals to use `h5py` as the backend for `PyTables`, it seems like getting VLEN data working with `h5py` is the best option.

### LINKED ISSUE #2083: Error when accessing empty entries in vlen dataset
In our application, we use h5py to store arrays of variable-length numpy structured arrays, where often some of the entries are empty. In h5py version 2.10, the behavior was that accessing an element of an h5py Dataset that had no data would return a zero-length numpy array, with the correct structured dtype.

However, in the latest version (I am using 3.6.0), accessing a dataset entry with no data (whether initialized with an empty array or not) causes a ValueError. The relevant issue in our project is here: https://github.com/py4dstem/py4DSTEM/issues/262 . A minimal example is below. 

```python
import h5py
import numpy as np

dtype = np.dtype([('a',np.float64),('b',np.float64)])

f = h5py.File("vlen_test.h5",'a')
h5dt = h5py.vlen_dtype(dtype)
dset = f.create_dataset('vlen_test',(4,),dtype=h5dt)

empty_arr = np.zeros((0,),dtype=dtype)
dset[0] = empty_arr

print(dset[0])
```

Running this yields this error:
```
----> 1 print(dset[0])

File h5py/_objects.pyx:54, in h5py._objects.with_phil.wrapper()

File h5py/_objects.pyx:55, in h5py._objects.with_phil.wrapper()

File ~/opt/miniconda3/envs/new4d/lib/python3.8/site-packages/h5py/_hl/dataset.py:793, in Dataset.__getitem__(self, args, new_dtype)
    791 mspace = h5s.create_simple(selection.mshape)
    792 fspace = selection.id
--> 793 self.id.read(mspace, fspace, arr, mtype, dxpl=self._dxpl)
    795 # Patch up the output for NumPy
    796 if arr.shape == ():

File h5py/_objects.pyx:54, in h5py._objects.with_phil.wrapper()

File h5py/_objects.pyx:55, in h5py._objects.with_phil.wrapper()

File h5py/h5d.pyx:192, in h5py.h5d.DatasetID.read()

File h5py/_proxy.pyx:141, in h5py._proxy.dset_rw()

File h5py/_conv.pyx:660, in h5py._conv.vlen2ndarray()

File h5py/_conv.pyx:726, in h5py._conv.conv_vlen2ndarray()

ValueError: Cannot create cython.array from NULL pointer
```

Accessing `dset[1]`, which hasn't been initialized at all, gives the same error. 

Output of `h5py.version.info`:
```
h5py    3.6.0
HDF5    1.12.1
Python  3.8.13 | packaged by conda-forge | (default, Mar 25 2022, 06:05:47)
[Clang 12.0.1 ]
sys.platform    darwin
numpy   1.22.3
cython (built with) 0.29.24
numpy (built against) 1.17.3
HDF5 (built against) 1.12.1
```
and I am using macOS 12.2.1 on x86_64

### LINKED ISSUE #2085: Issue reading H5 with strings inside compound inside vlen dataset
I am trying to read a H5 file that another of my programs has written in CPP. My CPP program stores randomly structured vlen/compound datasets. I have noticed that whenever my structures include a STRING element in a COMPOUND datatype in a VLEN datatype, then reading the file again with h5py fails on reading the dataset with a segmentation fault. I have attached a simplified H5 file with a 1x1 dataset that shows this behaviour (in my configuration at least) - see test.h5.txt - to be renamed to test.h5.
  [test.h5.txt](https://github.com/h5py/h5py/files/8475016/test.h5.txt)
The following python code triggers a segmentation fault: 
```
f = h5py.File('test.h5', 'r')
print(f.keys())
img = f['image']
img[0]
```
Generates the following output:
```
 <KeysViewHDF5 ['image']>
Segmentation fault (core dumped)
```

However, i am able to read the file using h5dump and custom CPP programs. Here is an output of h5dump:
```
h5dump test.h5 
HDF5 "test.h5" {
GROUP "/" {
   DATASET "image" {
      DATATYPE  H5T_COMPOUND {
         H5T_STRING {
            STRSIZE H5T_VARIABLE;
            STRPAD H5T_STR_NULLTERM;
            CSET H5T_CSET_ASCII;
            CTYPE H5T_C_S1;
         } "deviceRole";
         H5T_STD_U64LE "frameId";
         H5T_VLEN { H5T_STD_U8LE} "data";
         H5T_STD_I64LE "timestamp_ns";
         H5T_STD_U64LE "height";
         H5T_STD_U64LE "width";
         H5T_STD_U64LE "bitDepth";
         H5T_STD_U64LE "format";
         H5T_STD_U64LE "encoding";
         H5T_VLEN { H5T_COMPOUND {
            H5T_STRING {
               STRSIZE H5T_VARIABLE;
               STRPAD H5T_STR_NULLTERM;
               CSET H5T_CSET_ASCII;
               CTYPE H5T_C_S1;
            } "name";
            H5T_STRING {
               STRSIZE H5T_VARIABLE;
               STRPAD H5T_STR_NULLTERM;
               CSET H5T_CSET_ASCII;
               CTYPE H5T_C_S1;
            } "value";
         }} "additionalInformation";
      }
      DATASPACE  SIMPLE { ( 1 ) / ( H5S_UNLIMITED ) }
      DATA {
      (0): {
            "TOTO",
            4294967295,
            (0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0),
            140734018973984,
            18446744071819729358,
            21975,
            0,
            0,
            65535,
            ({
                  "TOTOTOTOTOOT",
                  "TATATATATAT"
               }, {
                  "TYTYTYYTYTYTYTT",
                  "TUTUTUTUTUTTUUTUT"
               })
         }
      }
   }
}
}
```

Note that when my vlen array named "additionalInformation" is empty, I seem to be having the issue mentioned in #2083, when it is not empty, I have the error mentioned above.

Here is my system configuration:
 * Operating System : Ubuntu 20.04.2 
 * Python version : 3.8
 * Where Python was acquired : pip
 * h5py version : h5py==3.6.0
 * HDF5 version libhdf5-103/focal,now 1.10.4+repack-11ubuntu1 amd64 [installed,automatic]
 <KeysViewHDF5 ['image']>
Segmentation fault (core dumped)

h5py    3.6.0
HDF5    1.12.1
Python  3.8.10 (default, Mar 15 2022, 12:22:08) 
[GCC 9.4.0]
sys.platform    linux
numpy   1.22.3
cython (built with) 0.29.24
numpy (built against) 1.17.3
HDF5 (built against) 1.12.1

johnstairs:
Fixing some issues (#596, #623, #2083, #2085), all related to segmentation faults or buffer overruns when reading and writing compound types with vlen fields of compound types.

They all boil down to two specific problems:
1. The background buffer passed to `H5Tconvert` when writing attributes was of the wrong size. It needs to be the size of the destination data.
2. When converting a vlen to ndarray, when the inner data is compound, we need to provide a background buffer to the conversion function. This means that we can now support vlens of compounds that in turn contain vlens of compounds.

Also updated the `TestCase.assertArrayEqual` method to handle comparisons across "object" types (where memberwise `==` is not supported).

One (perhaps minor) unexpected behavior that remains is extra padding around ndarray fields that are read from an HDF5 file. 
``` python
with h5py.File("x.h5", mode="w") as f:
    dt = np.dtype([('a', h5py.vlen_dtype(np.int32)),
                    ('b', h5py.vlen_dtype(np.int32))])

    data = np.array((np.array(list(range(1, 5)), dtype=np.int32),
                    np.array(list(range(8, 10)), dtype=np.int32)), dtype=dt)

    f.attrs['x'] = data
    out = f.attrs['x']
    print(data.dtype)
    print(out.dtype)
```
```
[('a', 'O'), ('b', 'O')]
{'names': ['a', 'b'], 'formats': ['O', 'O'], 'offsets': [0, 16], 'itemsize': 32}
```
In this example, there is 8 bytes of padding after both `a` and `b` fields when reading the attribute. This is because of the vlen datatype requires 16 bytes in HDF5, and it looks like we use the same offsets for the in-memory data as the persisted types. We can automatically remove this padding (when reading attributes and datasets), but this broke a number of unit tests where we verify that the alignment is round-tripped, so I left the behavior as-is.

-----

Closes #596
Closes #623
Closes #2083
Closes #2085

### DIFF
--- a/h5py/_conv.pyx
+++ b/h5py/_conv.pyx
@@ -702,7 +702,6 @@ cdef int conv_vlen2ndarray(void* ipt,
         npy_intp dims[1]
         void* data
         void* back_buf = NULL
-        cdef char[:] buf
         cnp.ndarray ndarray
         PyObject* ndarray_obj
         vlen_t in_vlen0
@@ -727,16 +726,12 @@ cdef int conv_vlen2ndarray(void* ipt,
     # For simple dtypes, we can use SimpleNewFromData, but types like
     # string & void need a size specified, so this function can't be used.
     # Additionally, Cython doesn't expose NumPy C-API functions like NewFromDescr,
-    # so we fall back on the Python frombuffer() function. Finally, that function does
-    # not support object arrays, so in that case we copy directly to the underlying buffer
-    # of a new ndarray.
+    # so we fall back on copying directly to the underlying buffer
+    # of a new ndarray for other types.
 
     if elem_dtype.kind in b"biufcmMO":
         # type_num is enough to create an array for these dtypes
         ndarray = cnp.PyArray_SimpleNewFromData(1, dims, elem_dtype.type_num, data)
-    elif elem_dtype.kind != b'V':
-        buf = <char[:itemsize * size]> data
-        ndarray = np.frombuffer(buf, dtype=elem_dtype)
     else:
         ndarray = np.empty(size, dtype=elem_dtype)
         memcpy(PyArray_DATA(ndarray), data, itemsize * size)
--- a/h5py/_conv.pyx
+++ b/h5py/_conv.pyx
@@ -718,9 +718,12 @@ cdef int conv_vlen2ndarray(void* ipt,
         data = realloc(data, itemsize * size)
 
     if needs_bkg_buffer(intype.id, outtype.id):
-        back_buf = malloc(H5Tget_size(outtype.id)*size)
+        back_buf = emalloc(H5Tget_size(outtype.id)*size)
 
-    H5Tconvert(intype.id, outtype.id, size, data, back_buf, H5P_DEFAULT)
+    try:
+        H5Tconvert(intype.id, outtype.id, size, data, back_buf, H5P_DEFAULT)
+    finally:
+        free(back_buf)
 
     # We need to use different approaches to creating the ndarray with the converted
     # data depending on the destination dtype.
@@ -861,7 +864,7 @@ cdef int conv_ndarray2vlen(void* ipt,
         PyBuffer_Release(&view)
 
         if needs_bkg_buffer(intype.id, outtype.id):
-            back_buf = malloc(H5Tget_size(outtype.id)*len)
+            back_buf = emalloc(H5Tget_size(outtype.id)*len)
 
         H5Tconvert(intype.id, outtype.id, len, data, back_buf, H5P_DEFAULT)
 
--- a/h5py/_conv.pyx
+++ b/h5py/_conv.pyx
@@ -22,11 +22,12 @@ from .h5t cimport H5PY_OBJ, typewrap, py_create, TypeID, H5PY_PYTHON_OPAQUE_TAG
 from libc.stdlib cimport realloc
 from libc.string cimport strcmp
 from .utils cimport emalloc, efree
+from ._proxy cimport needs_bkg_buffer
 cfg = get_config()
 
 # Initialization of numpy
 cimport numpy as cnp
-from numpy cimport npy_intp, NPY_WRITEABLE, NPY_C_CONTIGUOUS, NPY_OWNDATA
+from numpy cimport npy_intp, NPY_WRITEABLE, NPY_C_CONTIGUOUS, NPY_OWNDATA, PyArray_DATA
 cnp._import_array()
 import numpy as np
 
@@ -700,6 +701,7 @@ cdef int conv_vlen2ndarray(void* ipt,
         int flags = NPY_WRITEABLE | NPY_C_CONTIGUOUS | NPY_OWNDATA
         npy_intp dims[1]
         void* data
+        void* back_buf = NULL
         cdef char[:] buf
         cnp.ndarray ndarray
         PyObject* ndarray_obj
@@ -714,17 +716,30 @@ cdef int conv_vlen2ndarray(void* ipt,
     itemsize = H5Tget_size(outtype.id)
     if itemsize > H5Tget_size(intype.id):
         data = realloc(data, itemsize * size)
-    H5Tconvert(intype.id, outtype.id, size, data, NULL, H5P_DEFAULT)
+
+    if needs_bkg_buffer(intype.id, outtype.id):
+        back_buf = malloc(H5Tget_size(outtype.id)*size)
+
+    H5Tconvert(intype.id, outtype.id, size, data, back_buf, H5P_DEFAULT)
+
+    # We need to use different approaches to creating the ndarray with the converted
+    # data depending on the destination dtype.
+    # For simple dtypes, we can use SimpleNewFromData, but types like
+    # string & void need a size specified, so this function can't be used.
+    # Additionally, Cython doesn't expose NumPy C-API functions like NewFromDescr,
+    # so we fall back on the Python frombuffer() function. Finally, that function does
+    # not support object arrays, so in that case we copy directly to the underlying buffer
+    # of a new ndarray.
 
     if elem_dtype.kind in b"biufcmMO":
         # type_num is enough to create an array for these dtypes
         ndarray = cnp.PyArray_SimpleNewFromData(1, dims, elem_dtype.type_num, data)
-    else:
-        # dtypes like string & void need a size specified, so can't be used with
-        # SimpleNewFromData. Cython doesn't expose NumPy C-API functions
-        # like NewFromDescr, so we'll construct this with a Python function.
+    elif elem_dtype.kind != b'V':
         buf = <char[:itemsize * size]> data
         ndarray = np.frombuffer(buf, dtype=elem_dtype)
+    else:
+        ndarray = np.empty(size, dtype=elem_dtype)
+        memcpy(PyArray_DATA(ndarray), data, itemsize * size)
 
     PyArray_ENABLEFLAGS(ndarray, flags)
     ndarray_obj = <PyObject*>ndarray
@@ -827,22 +842,29 @@ cdef int conv_ndarray2vlen(void* ipt,
         size_t len, nbytes
         PyObject* buf_obj0
         Py_buffer view
+        void* back_buf = NULL
+    try:
+        buf_obj0 = buf_obj[0]
+        ndarray = <cnp.ndarray> buf_obj0
+        len = ndarray.shape[0]
+        nbytes = len * max(H5Tget_size(outtype.id), H5Tget_size(intype.id))
 
-    buf_obj0 = buf_obj[0]
-    ndarray = <cnp.ndarray> buf_obj0
-    len = ndarray.shape[0]
-    nbytes = len * max(H5Tget_size(outtype.id), H5Tget_size(intype.id))
+        data = emalloc(nbytes)
 
-    data = emalloc(nbytes)
+        PyObject_GetBuffer(ndarray, &view, PyBUF_INDIRECT)
+        PyBuffer_ToContiguous(data, &view, view.len, b'C')
+        PyBuffer_Release(&view)
 
-    PyObject_GetBuffer(ndarray, &view, PyBUF_INDIRECT)
-    PyBuffer_ToContiguous(data, &view, view.len, b'C')
-    PyBuffer_Release(&view)
+        if needs_bkg_buffer(intype.id, outtype.id):
+            back_buf = malloc(H5Tget_size(outtype.id)*len)
 
-    H5Tconvert(intype.id, outtype.id, len, data, NULL, H5P_DEFAULT)
+        H5Tconvert(intype.id, outtype.id, len, data, back_buf, H5P_DEFAULT)
 
-    in_vlen[0].len = len
-    in_vlen[0].ptr = data
+        in_vlen[0].len = len
+        in_vlen[0].ptr = data
+
+    finally:
+        free(back_buf)
 
     return 0
 
--- a/h5py/_conv.pyx
+++ b/h5py/_conv.pyx
@@ -701,6 +701,7 @@ cdef int conv_vlen2ndarray(void* ipt,
         int flags = NPY_WRITEABLE | NPY_C_CONTIGUOUS | NPY_OWNDATA
         npy_intp dims[1]
         void* data
+        char[:] buf
         void* back_buf = NULL
         cnp.ndarray ndarray
         PyObject* ndarray_obj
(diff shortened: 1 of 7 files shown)


h5py/_conv.pyx
```diff
@@ -714,17 +716,30 @@ cdef int conv_vlen2ndarray(void* ipt,
...
+    # of a new ndarray.
 
     if elem_dtype.kind in b"biufcmMO":
         # type_num is enough to create an array for these dtypes
         ndarray = cnp.PyArray_SimpleNewFromData(1, dims, elem_dtype.type_num, data)
-    else:
-        # dtypes like string & void need a size specified, so can't be used with
-        # SimpleNewFromData. Cython doesn't expose NumPy C-API functions
-        # like NewFromDescr, so we'll construct this with a Python function.
+    elif elem_dtype.kind != b'V':
```
takluyver, line 737: I'm a bit confused here - the comment suggests that the `else` branch you're adding is for object arrays, which I can see the sense behind, but object dtypes appear to have `dt.kind == 'O'` rather than 'V'?

h5py/tests/common.py
```diff
@@ -94,9 +95,9 @@ def assertSameElements(self, a, b):
             if not match:
                 raise AssertionError("Item '%s' appears in b but not a" % x)
 
-    def assertArrayEqual(self, dset, arr, message=None, precision=None):
+    def assertArrayEqual(self, dset, arr, message=None, precision=None, ignore_alignment=False):
```
takluyver, line 98: Nitpick: could we flip the sense of the parameter, so the default is something like `check_alignment=True`, and you pass False to skip it. I prefer True to mean the positive sense.

h5py/_conv.pyx
johnstairs: Yeah, I agree this is confusing. The 'V' data type is actually what numpy uses for structured types (compounds), but if you try to use `np.frombuffer` with 'V' you get the error message "cannot create an OBJECT array from memory buffer".
I've removed the `np.frombuffer` code path since I could not manage to write a test that reached it.

johnstairs:
I should mention that this PR does not add support for writing attributes of scalar vlen types (as opposed to arrays of vlens) and similarly directly nested vlens (like vlen(vlen(int))) remain unsupported.

johnstairs:
Hi @takluyver, wondering if you've had a chance to look at this some more and whether you think the approach here is the right one. Thanks!

hansenms:
@takluyver any chance we could get this in or get some comments on it?

dchansen:
@takluyver This would really help us in ismrmrd-python.

takluyver:
Sorry, I will try to have another look at this, there's just consistently more things to do than time to do them. :slightly_smiling_face:

h5py/_conv.pyx
takluyver: I still care about the `np.frombuffer` thing, because the purpose of that is to avoid allocating a new chunk of memory and copying the data across.

> I've removed the np.frombuffer code path since I could not manage to write a test that reached it.

If I haven't missed something in your code, you can hit it by creating a dataset where each entry holds a variable number of void elements:

```
>>> vdt = h5py.vlen_dtype(np.dtype('V4'))
>>> ds = f.create_dataset('v4', dtype=vdt, shape=10)
>>> ds[0] = np.array([b'abcd', b'1234'], dtype='V4')  # Write some data
>>> ds[0]  # Read it back, triggering conversion
```

You can do something similar with numpy string data (which means bytes, in Python 3 terms):

```
>>> ds2 = f.create_dataset('s3', dtype=vdt2, shape=10)
>>> ds2[0] = np.array([b'abc', b'123'], dtype='S3')
>>> ds[0]
```

I think compound dtypes also use a 'V' type, and it looks like there are tests already in the repo that cover that.

> if you try to use np.frombuffer with 'V' you get the error message "cannot create an OBJECT array from memory buffer".

I've just tried, and I get that message with dtype `'O'`, which is what I'd expect, but not with a V dtype (these need a size, like `'V4'`)

h5py/_conv.pyx
```diff
@@ -714,17 +715,26 @@ cdef int conv_vlen2ndarray(void* ipt,
     itemsize = H5Tget_size(outtype.id)
     if itemsize > H5Tget_size(intype.id):
         data = realloc(data, itemsize * size)
-    H5Tconvert(intype.id, outtype.id, size, data, NULL, H5P_DEFAULT)
+
+    if needs_bkg_buffer(intype.id, outtype.id):
+        back_buf = malloc(H5Tget_size(outtype.id)*size)
```
takluyver, line 720: Do we need to free this buffer again, like in the other function you modified below?

Also, let's use our `emalloc` wrapper, which is already used elsewhere in this module. This takes care of raising a Python MemoryError if we can't get memory.

h5py/_conv.pyx
```diff
@@ -714,17 +716,38 @@ cdef int conv_vlen2ndarray(void* ipt,
...
+        # which can't be used with SimpleNewFromData.
+        # Cython doesn't expose NumPy C-API functions
         # like NewFromDescr, so we'll construct this with a Python function.
         buf = <char[:itemsize * size]> data
         ndarray = np.frombuffer(buf, dtype=elem_dtype)
+    else:
+        # Compound dtypes containing object fields: frombuffer() refuses these,
+        # so we'll fall back to allocating a new array and copying the data in.
+        ndarray = np.empty(size, dtype=elem_dtype)
+        memcpy(PyArray_DATA(ndarray), data, itemsize * size)
```
takluyver, line 750: One thing I'm not clear on: if we copy from `data` to a new location in memory, is our function responsible for freeing `data`? I'm guessing it is, because we already call `realloc(data, ...)` above, which can free the old pointer.

johnstairs: @takluyver, I'm really sorry for letting this sit around for so long. I think you are right. HDF5 allocates the v-len buffer and we are responsible for freeing it when we are no longer using it, which is the case here.

hartsw:
Is there anything I can do to help this along?  Mostly testing.  :P

I have some HDF5 files with this problem that we're using h5py 2.x for and it would be nice to move to something from the past 5 years.

takluyver:
Thanks for the reminder; I've got round to taking another look at this, and I think I'm happy with it now. Thanks @johnstairs for digging into the cluster of issues, figuring out the common cause, and having patience with my limited time for reviewing it.

takluyver:
As it's been a while, I'm just going to close & reopen so that the tests run again with a fresh merge before I actually merge it properly.

PR #2135, opened 2022-08-27
See if bumping travis python versions fixes travis

aragilar:
Both travis builds seem to be failing with errors that look like they could be caused by running on an old python version. Let's see if bumping to 3.10 fixes that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: python
-dist: xenial
+dist: jammy
 
 notifications:
   email: false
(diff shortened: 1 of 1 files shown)


takluyver:
It doesn't seem to have fixed it, sadly, and 3.10 doesn't even seem to be there for ppc64le :confused: .

takluyver:
I think the issue is the Python versions inside the Docker image used to build aarch64 wheels. cibuildwheel must have upgraded to a version that tries to build wheels with Python 3.11. But we have a custom Docker image, based on the manylinux2014 image and adding HDF5 to it, so we needed to rebuild that (there's a button [on this page](https://github.com/h5py/hdf5-manylinux/actions/workflows/build-image.yml) to do so). That's [in progress](https://github.com/h5py/hdf5-manylinux/actions/runs/3053416831/jobs/4924010386) now, so hopefully after that the Travis builds will work again.

We should probably pin the cibuildwheel version to prevent issues like this.

takluyver:
Yup, that seems to have done it (I reran the failing Travis job on PR #2148).

Issue #2133, opened 2022-08-18
Document how to use ros3 with anonymous access


jreadey:
It can be a little confusing to setup ros3 without authentication to access public read files.  If you set the aws_region kwargs, the library will try to authenticate even if the secret_id and secret_key are null.  See cell 2 in: https://github.com/HDFGroup/hdflab_examples/blob/master/Tutorial/10-Fancy_Indexing.ipynb

I'd suggest adding this to the docs: "for anonymous access, don't set aws_region, secret_id, and secret_key".

Also, it may be a mystery how to set the path since ros3 doesn't support s3 uri's.  You need to use:  f"http://s3.{region}.amazonaws.com/{bucket}/{s3key}"

takluyver:
I see that currently h5py sets `authenticate = 1` if any of aws_region, secret_id or secret_key are passed. Would it make sense for the code to check that either all three of those fields have a non-empty value, or none of them do? It doesn't sound like any of them are optional.

jreadey:
Right, authenticate should be set only if all three are set.  Otherwise the library will throw a "Invalid ros3 config (Inconsistent authentication information)" error.

ajelenak:
Submitted a PR to fix ros3 authentication: #2136.

ajelenak:
> Also, it may be a mystery how to set the path since ros3 doesn't support s3 uri's. You need to use: f"[http://s3.{region}.amazonaws.com/{bucket}/{s3key}](http://s3.%7Bregion%7D.amazonaws.com/%7Bbucket%7D/%7Bs3key%7D)"

I would propose to support this in h5py code. The `urllib.parse` package is standard so parsing of S3 URIs will not be a problem.

ajelenak:
This can be closed now since PRs #2136 and #2140 have been merged.

PR #2136, opened 2022-08-27
Fix when to activate authentication for ros3 driver

ajelenak:
AWS authentication will be used only when AWS S3 bucket's region and both AWS access keys are specified. See: #2133.


aragilar:
This looks good to me, I think any URL parsing can be left to a separate PR. Thanks @ajelenak!

PR #2137, opened 2022-08-28
DOC: improve formatting of the locking parameter to File

tacaswell:
Also adds a show target to the docs MakeFile

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -175,3 +175,6 @@ pseudoxml:
 	$(SPHINXBUILD) -b pseudoxml $(ALLSPHINXOPTS) $(BUILDDIR)/pseudoxml
 	@echo
 	@echo "Build finished. The pseudo-XML files are in $(BUILDDIR)/pseudoxml."
+
+show:
+	@python -c "import webbrowser; webbrowser.open_new_tab('file://$(shell pwd)/_build/html/index.html')"
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -177,4 +177,4 @@ pseudoxml:
 	@echo "Build finished. The pseudo-XML files are in $(BUILDDIR)/pseudoxml."
 
 show:
-	@python -c "import webbrowser; webbrowser.open_new_tab('file://$(shell pwd)/_build/html/index.html')"
+	@python -m webbrowser -t "file://$(shell pwd)/_build/html/index.html"
--- a/docs/high/file.rst
+++ b/docs/high/file.rst
@@ -465,10 +465,19 @@ Reference
     :param min_raw_keep: Minimum percentage of raw data to keep in the page
             buffer before allowing pages containing raw data to be evicted.
             Applicable only if ``page_buf_size`` is set. Default value is zero.
-    :param locking: The file locking behavior. One of False (or "false"), True
-            (or "true"), "best-effort", or None. Warning: The
-            HDF5_USE_FILE_LOCKING environment variable can override this
-            parameter. Only available with HDF5 >= 1.12.1 or 1.10.x >= 1.10.7.
+    :param locking: The file locking behavior. One of:
+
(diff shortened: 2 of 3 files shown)


tacaswell:
@ajelenak Where is a good place to link to in the hdf5 docs to explain what the "stakes" of this locking actually are?

ajelenak:
@tacaswell I asked @derobins to respond on that issue.

docs/high/file.rst
```diff
@@ -465,10 +465,19 @@ Reference
...
-    :param locking: The file locking behavior. One of False (or "false"), True
-            (or "true"), "best-effort", or None. Warning: The
-            HDF5_USE_FILE_LOCKING environment variable can override this
-            parameter. Only available with HDF5 >= 1.12.1 or 1.10.x >= 1.10.7.
+    :param locking: The file locking behavior. One of:
+
+            - False (or "false") --  Disable file locking
+            - True (or "true")   --  Enable file locking
+            - "best-effort"      --  Enable file locking but ignore some errors
+            - None               --  Use HDF5 defaults
```
takluyver, line 473: Is it worth saying what the default is (locking enabled, as far as I know)? I know it could change, but it's a minor annoyance when docs tell me that the default is the default and I have to look elsewhere to find what it actually is.

derobins: Default is currently "best-effort" and should remain so until I rip the scheme out.

docs/Makefile
```diff
@@ -175,3 +175,6 @@ pseudoxml:
 	$(SPHINXBUILD) -b pseudoxml $(ALLSPHINXOPTS) $(BUILDDIR)/pseudoxml
 	@echo
 	@echo "Build finished. The pseudo-XML files are in $(BUILDDIR)/pseudoxml."
+
+show:
+	@python -c "import webbrowser; webbrowser.open_new_tab('file://$(shell pwd)/_build/html/index.html')"
```
takluyver, line 180: The webbrowser module has a [documented](https://docs.python.org/3/library/webbrowser.html) `-m` CLI. Hopefully I've got the quoting right here:

```suggestion
	@python -m webbrowser -t "file://$(shell pwd)/_build/html/index.html"
```

docs/high/file.rst
takluyver: Thanks! Out of interest, what do you mean about ripping the scheme out? Is that the best-effort scheme specifically, or the whole locking scheme? If the latter, will it be replaced with anything?

derobins: I'd like to rip out the entire file locking scheme since it seems to cause more problems than it solves. There would be no replacement scheme, so users would just have to be careful about accessing files.

Keep in mind that the configure option is from 1.10.7 and 1.12.1 and later. There is a matrix here:

https://github.com/HDFGroup/hdf5/blob/develop/doc/file-locking.md

takluyver: Thanks! We mention the version restriction just below this list.

I tried recently, and it seems that IBM Spectrum Scale (aka GPFS) doesn't share flock locks across nodes, whereas lockf/fcntl locks are shared. So it's already possible to have the same HDF5 file open for writing on multiple nodes simultaneously, even where the lock calls work.

Maybe there's an option that a sysadmin can set somewhere in GPFS to adjust this behaviour; I'm just going on what happens in one of our GPFS filesystems.

tacaswell: so the task are :
 - add the default
 - note it may be removed in the future
 - add the link  https://github.com/HDFGroup/hdf5/blob/develop/doc/file-locking.md

derobins: Also, FWIW, the .md file will eventually be incorporated in the Doxygen-based HDF5 user guide (included in the repo). I'll probably keep the developer-level bits around in the MD file, though.

PR #2140, opened 2022-08-31
Support HDF5 file s3:// locations for ros3 driver

ajelenak:
This PR is adding support for s3:// file locations when using the ros3 driver. This is done in the h5py code since the HDF5 library's ROS3 VFD currently does not support this. s3:// style is much more common. This functionality was also mentioned in #2133.

### DIFF
--- a/h5py/tests/test_file.py
+++ b/h5py/tests/test_file.py
@@ -1011,15 +1011,33 @@ f = h5py.File({str(filename)!r}, mode={mode!r}, locking={locking})
 
 # unittest doesn't work with pytest fixtures (and possibly other features),
 # hence no subclassing TestCase
+@pytest.mark.skipif(h5py.version.hdf5_version_tuple < (1, 10, 6)
+                    or not h5.get_config().ros3,
+                    reason="ros3 driver not available")
 class TestROS3:
-    @pytest.mark.skipif(h5py.version.hdf5_version_tuple < (1, 10, 6)
-                        or not h5.get_config().ros3,
-                        reason="ros3 file operations were added in HDF5 1.10.6+")
     def test_ros3(self):
         """ ROS3 driver and options """
 
         with File("https://dandiarchive.s3.amazonaws.com/ros3test.hdf5", 'r',
                   driver='ros3') as f:
+            assert f
+            assert 'mydataset' in f.keys()
+            assert f["mydataset"].shape == (100,)
+
+    def test_ros3_s3_fails(self):
+        """ROS3 exceptions for s3:// location"""
+        with pytest.raises(ValueError,
+                           match='AWS region required for s3:// location'):
+            File('s3://fakebucket/fakekey', 'r', driver='ros3')
+
+        with pytest.raises(ValueError,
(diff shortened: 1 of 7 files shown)


ajelenak:
I added two tests for s3:// locations when ros3 driver is available. Will merge this PR on 2022-09-23 if there are no additional comments.

h5py/tests/test_file.py
```diff
@@ -1024,6 +1024,24 @@ def test_ros3(self):
...
+        """ROS3 exceptions for s3:// location"""
+        with pytest.raises(ValueError,
+                           match='AWS region required for s3:// location'):
+            File('s3://fakebucket/fakekey', 'r', driver='ros3')
+
+        with pytest.raises(ValueError,
+                           match=r'^foo://wrong/scheme: S3 location must begin with'):
+            File('foo://wrong/scheme', 'r', driver='ros3')
+
+    def test_ros3_s3uri(self):
```
tacaswell, line 1037: Can we put this in its own file or make it out-in?  I'm super happy to pick up a network dependency in the test suite.

I think we also need to add a skip on both of these if h5py is not built with ros.

ajelenak: Don't know what "make it out-in" means.

I can certainly put all ROS3 tests in a single file, I just added these two to an already existing class.

The entire test class will now be skipped if the ros3 driver is not available:
https://github.com/h5py/h5py/blob/5006c597fce7054b516c93a91cc69ff8a5808cb1/h5py/tests/test_file.py#L1014-L1017

ajelenak: @tacaswell I moved the ros3 tests into a separate file.

tacaswell: meant "opt-in" sorry, typing failure 🐑 .

ajelenak: The ros3 tests are now in their own file: [test_ros3.py](https://github.com/h5py/h5py/blob/c70487c8d91ea23969a0b1b6d8171bcd53cc19d0/h5py/tests/test_ros3.py). The tests in this file will be skipped for HDF5 prior to 1.10.6 or if HDF5 was not built with the ros3 driver. Is this now acceptable?

tacaswell: I think we may still want to add a "--network" or "--no-network" flag.  I know that some of the linux distros run their build system in contexts where there is no network access so it is good if we could give them a way to disable the test that actually reads from s3 without having to patch the tests.

ajelenak: Added `--no-network` CLI option to pytest. Created a new `pytest.mark.nonetwork" marker and added it to the ros3 tests.

ajelenak: @tacaswell Are the current changes acceptable to resolve this conversation? When I run `pytest -rs --no-network h5py` the ros3 tests are all skipped:
```
SKIPPED [3] h5py/tests/test_ros3.py: No Internet
```

ajelenak:
Since there are no outstanding items to complete I will merge this PR on 2022-10-21 if nothing comes up until then.

ajelenak:
Closes #2133.

PR #2141, opened 2022-09-01
Fixed double whitespace in error message

Helveg:
Do what you do best. Nitpicking.


tacaswell (approved):

tacaswell:
Going to let CI spin before merging just to be safe, but 👍🏻 .

What is programming but arguing with computers and what are computers but the most literal nit-picky systems we have easy access to?

tacaswell:
The test failure is on install which is unrelated.

PR #2142, opened 2022-09-05
Fix incompatible function pointer types

ryanking13:
This PR fixs some imcompatible function pointer types. Compiling h5py with recent version of LLVM with `-Wincompatible-function-pointer-types` compile option is not happy about this.

`H5E_walk_t` [in HDF5 is defined as](https://github.com/HDFGroup/hdf5/blob/ae414872f50187e64cbd6cc8f076c22cf5df2d53/src/H5Epublic.h#L175):

```c
typedef herr_t (*H5E_walk2_t)(unsigned n, const H5E_error2_t *err_desc, void *client_data);
```

### DIFF
--- a/h5py/_errors.pyx
+++ b/h5py/_errors.pyx
@@ -94,7 +94,7 @@ cdef struct err_data_t:
     H5E_error_t err
     int n
 
-cdef herr_t walk_cb(int n, H5E_error_t *desc, void *e) nogil:
+cdef herr_t walk_cb(unsigned int n, const H5E_error_t *desc, void *e) nogil:
 
     cdef err_data_t *ee = <err_data_t*>e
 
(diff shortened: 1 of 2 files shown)


h5py/_errors.pyx
```diff
@@ -120,7 +120,7 @@ cdef int set_exception() except -1:
 
     err.n = -1
 
-    if H5Ewalk(<hid_t>H5E_DEFAULT, H5E_WALK_UPWARD, walk_cb, &err) < 0:
+    if H5Ewalk(<hid_t>H5E_DEFAULT, H5E_WALK_UPWARD, &walk_cb, &err) < 0:
```
tacaswell, line 123: Has this just been super broken?

ryanking13: No, it is not broken. I just realized that an operator `&` applied to a function pointer has [no effect](https://stackoverflow.com/questions/6893285/why-do-function-pointer-definitions-work-with-any-number-of-ampersands-or-as). So that part would not be necessary.

aragilar:
This looks good to me, any objections to me going ahead and merging it?

tacaswell (approved):

PR #2147, opened 2022-09-14
mpi: support using MS-MPI

mathstuf:
MS-MPI does not use an `mpicc` wrapper and instead provides its information through environment variables. When requested, use this information. Also requires `HDF5_MPI=ON` to avoid requesting MS-MPI in non-MPI contexts.

### DIFF
--- a/docs/build.rst
+++ b/docs/build.rst
@@ -238,3 +238,7 @@ by setting the ``HDF5_MPI`` environment variable::
 
 You will need a shared-library build of Parallel HDF5 as well, i.e. built with
 ``./configure --enable-shared --enable-parallel``.
+
+On Windows, MS-MPI is usually used which does not have an ``mpicc`` wrapper.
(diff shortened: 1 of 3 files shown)


mathstuf:
CI failure seems very unrelated (Linux 3.11 aarch64 wheelbuild fails).

tacaswell (approved):
This seems reasonable, but I have no expertise or way to test this.

mathstuf:
For links to my testing of it:

- Without this, [`mpi.h` cannot be found when using a parallel-enabled HDF5 on Windows](https://open.cdash.org/build/8154727/notes#note15).
- With it, it [does build](https://gitlab.kitware.com/paraview/paraview-superbuild/-/jobs/7293870) (ignore the packaging failure; the paths are too long for WIX to handle).

takluyver:
I know nothing of MS-MPI, but it looks like this won't change any behaviour unless you set `H5PY_MSMPI=ON`, so let's merge it. If any more work is needed, it will be up to the people using MS-MPI.

PR #2148, opened 2022-09-14
Use macOS 11 on CI

takluyver:
The macos-10.15 build environment is deprecated, with ~weekly brownouts to highlight that it will be removed in December.

https://github.com/actions/runner-images/issues/5583


takluyver:
Merging this as it's a simple change and I want to have CI working again.

Issue #2144, opened 2022-09-09
Long loading times of compound datasets when accessing fields with `np.asarray()`


tobiasjj:
I work in a team for a company and help to develop a Python package to analyze scientific data produced from devices. Recently, we stumbled over a major performance issue, when loading large compound datasets when accessing them via `np.asarray(dset.fields("FieldA"))`. We need to access it with `np.asarray()` for reasons. We figured out, that this is caused by `np.asarray()` iterating over all datapoints of the data via the `__getitem__()` method, as the `FieldsWrapper` object (returned by `fields()`) does not provide an `__array__()` method.

A simple solution would be to add the `__array__()` method to the `FieldsWrapper` class. I'm wondering if there is a specific reason behind the `FieldsWrapper` class not implementing the `__array__()` method or if it would be possible to add it? I would be happy to contribute. Should I open a PR?

What do you think?

takluyver:
My expectation when I added FieldsWrapper was that you'd do `dset.fields("FieldA")[:]` to get a NumPy array. But I don't see any issue with adding a `__array__` method as well.

PR #2151, opened 2022-09-15
Add an `__array__()` method to the `FieldsWrapper` class

tobiasjj:
**Why this PR**
When trying to access fields of compound datasets with numpy functions (like `np.asarray(dset.fields("FieldA"))`), loading might take very long, as the numpy function tries to retrieve the whole array by iterating over every single element via the `__getitem__()` method. To speed up the loading, I added the  `__array__()`[1] method to the `FieldsWrapper` class, which returns the whole array.

This solves issue https://github.com/h5py/h5py/issues/2144

[1] More details about [the `__array__()` method ](https://numpy.org/devdocs/user/basics.interoperability.html#the-array-method) in the numpy docs.

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -253,6 +253,12 @@ class FieldsWrapper:
             names = [names]
         self.read_dtype = readtime_dtype(prior_dtype, names)
 
+    def __array__(self, dtype=None):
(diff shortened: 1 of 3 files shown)


takluyver:
Thanks, this looks good to me.

PR #2157, opened 2022-10-02
Fix complex32

### LINKED ISSUE #2156: Complex float16 problems
I've got an HDF5 file that uses complex-valued data using the usual h5py {r, i} names for the members of a structured type (since HDF5 doesn't have a built in complex type). The problem is that the members are float16, and numpy doesn't have a corresponding type (see open issue [here](https://github.com/numpy/numpy/issues/14753)), which causes all kinds of grief. For example,
```Python
import h5py
import numpy as np
f = h5py.File("/tmp/foo.h5", "w")
c4 = np.dtype([('r', 'f2'), ('i', 'f2')])
ds = f.create_dataset("z", dtype=c4, shape=(10,))
print(ds.dtype)
```
Crashes with the error
```
  File "/private/tmp/test_c4.py", line 6, in <module>
    print(ds.dtype)
  File "/Users/bhawkins/Library/Python/3.10/lib/python/site-packages/h5py-3.7.0-py3.10-macosx-12-x86_64.egg/h5py/_hl/dataset.py", line 518, in dtype
    return self.id.dtype
  File "h5py/h5d.pyx", line 131, in h5py.h5d.DatasetID.dtype.__get__
  File "h5py/h5d.pyx", line 134, in h5py.h5d.DatasetID.dtype.__get__
  File "h5py/h5t.pyx", line 434, in h5py.h5t.TypeID.dtype.__get__
  File "h5py/h5t.pyx", line 435, in h5py.h5t.TypeID.dtype.__get__
  File "h5py/h5t.pyx", line 1231, in h5py.h5t.TypeCompoundID.py_dtype
TypeError: data type '<c4' not understood
```

Lots of methods like slicing need access to the `dtype` attribute and fail with the same error.

Until numpy supports complex float16, I'd prefer if `h5py.Dataset.dtype` just returned the structured type. I could attempt this change if such a PR would be appreciated.

bhawkins:
This is my attempt to fix #2156. It's basically a one-line change to restrict complex dtypes to the ones currently supported by numpy (float, double, and long double). Otherwise a compound dtype is returned.

### DIFF
--- a/h5py/h5t.pyx
+++ b/h5py/h5t.pyx
@@ -1223,7 +1223,8 @@ cdef class TypeCompoundID(TypeCompositeID):
         if len(field_names) == 2                                and \
             tuple(field_names) == (cfg._r_name, cfg._i_name)    and \
             field_types[0] == field_types[1]                    and \
-            field_types[0].kind == 'f':
+            field_types[0].kind == 'f'                          and \
+            field_types[0].itemsize in (4, 8, 16):
 
             bstring = field_types[0].str
             blen = int(bstring[2:])
--- a/h5py/tests/test_dataset.py
+++ b/h5py/tests/test_dataset.py
@@ -1037,6 +1037,13 @@ class TestDtype(BaseDataset):
         dset = self.f.create_dataset('foo', (5,), '|S10')
         self.assertEqual(dset.dtype, np.dtype('|S10'))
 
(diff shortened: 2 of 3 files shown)


takluyver:
Thanks, this looks good to me. :slightly_smiling_face:

PR #2165, opened 2022-10-25
Add CI/tox support for Python 3.11

aragilar:
This only does the minimal amount to get wheels running, further cleanup of the CI would be a good idea (e.g. dropping old Python versions, moving static CI tests to newer Python version).

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -27,7 +27,7 @@ jobs:
   pool:
     vmImage: ubuntu-20.04
   variables:
-    CIBW_BUILD: cp3{7,8,9,10}-manylinux_x86_64
+    CIBW_BUILD: cp3{7,8,9,10,11}-manylinux_x86_64
     CIBW_MANYLINUX_X86_64_IMAGE: ghcr.io/h5py/manylinux2010_x86_64-hdf5
     # Include less debugging info for smaller wheels (default is -g2)
     CIBW_ENVIRONMENT: "CFLAGS=-g1"
@@ -43,7 +43,7 @@ jobs:
     HDF5_VERSION: 1.12.2
     HDF5_DIR: $(HDF5_CACHE_DIR)/$(HDF5_VERSION)
     HDF5_VSVERSION: "16-64"
-    CIBW_BUILD: cp3{7,8,9,10}-win_amd64
+    CIBW_BUILD: cp3{7,8,9,10,11}-win_amd64
   steps:
     - template: ci/azure-pipelines-wheels.yml
       parameters:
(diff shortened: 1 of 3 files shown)


tacaswell:
"powercycled" to restart CI.

takluyver:
It looks like at least part of the issue is our custom manylinux+HDF5 docker images not having Python 3.11 yet. I've just launched a new build of those images - this is something we can hopefully all do at https://github.com/h5py/hdf5-manylinux/actions/workflows/build-image.yml .

kanderso-nrel:
@takluyver crossref https://github.com/h5py/hdf5-manylinux/issues/6 if you haven't already seen that.  Seems like `Dockerfile_x86_64` will need to update from `manylinux2010` before this will work?

takluyver:
Aha, thanks, I hadn't seen that. :slightly_smiling_face:

tacaswell:
Tried again now that https://github.com/h5py/hdf5-manylinux/actions/runs/3371047039 finished.

kanderso-nrel:
A hint for the failure: looks like `azure-pipelines.yml` is still referencing the `manylinux2010` image

azure-pipelines.yml
```diff
@@ -27,7 +27,7 @@ jobs:
   pool:
     vmImage: ubuntu-20.04
   variables:
-    CIBW_BUILD: cp3{7,8,9,10}-manylinux_x86_64
+    CIBW_BUILD: cp3{7,8,9,10,11}-manylinux_x86_64
     CIBW_MANYLINUX_X86_64_IMAGE: ghcr.io/h5py/manylinux2010_x86_64-hdf5
```
takluyver, line 31:
```suggestion
    CIBW_MANYLINUX_X86_64_IMAGE: ghcr.io/h5py/manylinux2014_x86_64-hdf5
```

takluyver:
It looks like this is working - no idea what's wrong with the Travis ppc64le build, but it's the same on master.

There's a warning on Azure saying:

> You should provide GitHub token if you want to download a python release. Otherwise you may hit the GitHub anonymous download limit.

Let's leave it as is for now, and hope that Azure will soon have Python 3.11 without needing to fetch it from Github.

Cadair:
Just so I know for SunPy's release planning is this likely to make it to a release in the short term or are there other things to do before hand.

takluyver:
I'm roughly thinking some time in the next few weeks. We don't specifically aim to release h5py after new Python/HDF5 releases, but it has also been a few months, and quite a few PRs merged, since 3.7, so it would be good to put a new release out.

Cadair:
Awesome thanks!

paulkorir:
Any update on this?

neutrinoceros:
@takluyver I'm sorry to bother you, I know that these requests can be painful. I maintain a couple packages for which this is the bottleneck, and we could use some visibility on when wheels for Python 3.11 may be available. It doesn't have to be soon (although the sooner, the better), but it'd be really helpful if we could just plan the transition downstream.

Issue #2125, opened 2022-07-28
Add Dataset.is_scale() method?


ajelenak:
The Dataset class currently has `make_scale()` method but lacks a check if an HDF5 dataset is a dimension scale. Hence, I propose to add an `is_scale()` method. It would be a shortcut for:
```python
h5ds.is_scale(self._id)
```
Any objections?

mullenkamp:
This would definitely be great.
What is currently the best way to check if a dataset is a scale?

ajelenak:
@mullenkamp Thanks! This issue has been quiet for some time, I agree with you obviously.

The current way to check is what I outlined in the first comment, something like this:
```python
import h5py.h5ds

# If ds is an h5py.Dataset object...
h5py.h5ds.is_scale(ds._id)
```

mullenkamp:
Oh! Thanks for that! I was originally a little confused by that one line.
That seems much better than my attempt which queries the attrs.

PR #2168, opened 2022-11-04
Add Dataset.is_scale() method

ajelenak:
Implements what was proposed in #2125.

### DIFF
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -574,11 +574,6 @@ Reference
 
        You can optionally pass a name to associate with this scale.
 
-    .. method:: is_scale()
-
-        Return ``True`` if the dataset is also a :ref:`dimension scale <dimension_scales>`,
-        ``False`` otherwise.
-
     .. method:: virtual_sources
 
        If this dataset is a :doc:`virtual dataset </vds>`, return a list of
@@ -666,6 +661,11 @@ Reference
     .. attribute:: dims
 
         Access to :ref:`dimension_scales`.
+
+    .. attribute:: is_scale
+
+        Return ``True`` if the dataset is also a :ref:`dimension scale <dimension_scales>`,
+        ``False`` otherwise.
 
     .. attribute:: attrs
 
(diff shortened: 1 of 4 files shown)


takluyver:
Should we make this a property (so you'd use e.g. `if dset.is_scale:` without `()`)? We already have `.is_virtual` as a property, so it would seem natural to follow that pattern, and I doubt we'd ever need parameters for it.

ajelenak:
I agree. Made the required changes.

epourmal:
Just FYI in case you want to consider... 

HDF5 has a configure option to use new object references for dimensions scales `--enable-dimension-scales-with-new-ref`. It was introduced in 1.12. 2( or 1?) for enabling DS interface with VOL connectors. See also [H5DSwith_new_ref](https://docs.hdfgroup.org/hdf5/develop/group___h5_d_s.html#gaed2b97139202dfe69f3f2a7364c10fbc). New object references are default for dimension scales for non-native VOL connectors.

ajelenak:
This PR is about using [`H5DSis_scale()`](https://docs.hdfgroup.org/hdf5/develop/group___h5_d_s.html#title9). Does this function behaves differently for new HDF5 object references?

epourmal:
> This PR is about using [`H5DSis_scale()`](https://docs.hdfgroup.org/hdf5/develop/group___h5_d_s.html#title9). Does this function behaves differently for new HDF5 object references?

I don't think this is a problem for h5py based on 1.12.2 and later since the function will be returning true or false as expected. The problem is in possible file incompatibility if the file was created with a pass-through VOL connector: a non-native VOL connectors always trigger usage of  new references in spite of the terminal VOL connector being a native one! plus the library may be configured to use new references in the first place. A user will be even not aware of incompatibility until he/she uses HDF5 software based on 1.10.* or earlier to read the file back. I hope the library will handle the case gracefully, but I don't think it was tested.

This is an HDF5 problem and has to be addressed there. I just wanted to raise awareness of possible HDF5 file backward compatibility issue. Maybe h5py as a high-level API has a better way of dealing with such issues until HDF5 fixes the problem and uses old references when a terminal connector is a native one. 

As you can see this is a slightly different problem from `is_virtual`.

h5py community was warned :-)

ajelenak:
Closes #2125.

PR #2178, opened 2022-11-24
Update how architectures are defined on Travis CI

takluyver:
It appears that how you specify ppc64le has changed on Travis, and this was leading to the cryptic "An error occurred while generating the build script"  error messages we've been getting.

https://docs.travis-ci.com/user/multi-cpu-architectures

This should get ppc64le builds working again, and I've cleaned up a couple of other things in the `.travis.yml` file while I was looking at it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: python
+os: linux
 dist: xenial
 
 notifications:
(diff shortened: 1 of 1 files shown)


takluyver:
Merging since this fixes CI. :slightly_smiling_face:

PR #2179, opened 2022-11-25
Move static metadata to pyproject.toml

mwtoews:
This PR moves static project metadata from `setup.py` to `pyproject.toml`, as is described by [PEP 621](https://peps.python.org/pep-0621/) and supported by [setuptools >=61](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html).

Build system changes:

- wheel does not need to be included in `requires` (most recent docs have removed this)
- Specific versions of Cython also don't need to be listed, replaced with `Cython ~=0.29`
- Alphasort `requires` list

Project metadata changes:

- Removed Andrew Collette from maintainers (kept in authors); replaced with @takluyver as this is observed from the [contributors](https://github.com/h5py/h5py/graphs/contributors)
- Change licence from "BSD" to SPDX identifier "BSD-3-Clause"
- Packages don't need to be manually listed, the are found using [custom discovery](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html)
- URLs table has been expanded slightly
- Readme (long_description) content-type is now specified, so that `twine check --strict dist/*` pass (`tox -e checkreadme`)

Other notes:

- The installed wheel now includes *.pxd and *.pyx files. Not sure why this is (perhaps due to custom discovery?), and I'm not sure if its a good or bad thing to include for a binary wheel.
- While there are [dynamic methods](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata) to extract version from (e.g.) `h5py.version.version`, this did not work, so this is best left to `setup.py` to evaluate for now.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,13 +1,80 @@
 [build-system]
-# Further build requirements - like numpy & cython - come from setup.py via
-# the PEP 517 interface.
 requires = [
-    "setuptools",
-    "wheel",
+    "Cython ~=0.29",
     "oldest-supported-numpy",
     "pkgconfig",
-    "Cython >=0.29; python_version<'3.8'",
-    "Cython >=0.29.14; python_version=='3.8'",
-    "Cython >=0.29.15; python_version>='3.9'",
+    "setuptools >=61",
 ]
 build-backend = "setuptools.build_meta"
+
+[project]
+name = "h5py"
+description = "Read and write HDF5 files from Python"
+authors = [
+    {name = "Andrew Collette", email = "andrew.collette@gmail.com"},
+]
+maintainers = [
+    {name = "Thomas Kluyver", email = "thomas@kluyver.me.uk"},
+]
+license = {text = "BSD-3-Clause"}
+classifiers = [
+    "Development Status :: 5 - Production/Stable",
+    "Intended Audience :: Developers",
+    "Intended Audience :: Information Technology",
+    "Intended Audience :: Science/Research",
+    "License :: OSI Approved :: BSD License",
+    "Operating System :: Unix",
+    "Operating System :: POSIX :: Linux",
+    "Operating System :: MacOS :: MacOS X",
+    "Operating System :: Microsoft :: Windows",
+    "Programming Language :: Cython",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: Implementation :: CPython",
+    "Topic :: Scientific/Engineering",
+    "Topic :: Database",
+    "Topic :: Software Development :: Libraries :: Python Modules",
+]
+requires-python = ">=3.7"
+dynamic = ["dependencies", "version"]
+
+[project.readme]
+text = """\
+The h5py package provides both a high- and low-level interface to the HDF5
+library from Python. The low-level interface is intended to be a complete
+wrapping of the HDF5 API, while the high-level component supports  access to
+HDF5 files, datasets and groups using established Python and NumPy concepts.
+
+A strong emphasis on automatic conversion between Python (Numpy) datatypes and
+data structures and their HDF5 equivalents vastly simplifies the process of
+reading and writing data from Python.
+
+Wheels are provided for several popular platforms, with an included copy of
+the HDF5 library (usually the latest version when h5py is released).
+
+You can also `build h5py from source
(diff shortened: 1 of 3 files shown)


aragilar:
I don't see there being too much of an issue doing the switch. I'd keep the manual listing of packages though: I highly doubt we're going to add a new package, and I'd rather avoid changing too many things at once (especially given the somewhat bespoke setup we have).

mwtoews:
Manual list of packages has been updated.

Also, finally figured out why  *.pxd and *.pyx files were included with wheels "By default, include-package-data is true in pyproject.toml" ([ref](https://setuptools.pypa.io/en/latest/userguide/datafiles.html#include-package-data)). This is set to false, so the wheel build contents are now comparable.

aragilar:
@mwtoews Are you able to rebase this to resolve the tox conflicts?

@takluyver This looks fine to me, but you're more familiar with the newer packaging standards than I am, can you have a quick look in case some is wrong?

mwtoews:
Conflicts are resolved, and a slightly awkward readme table section has been re-worked to `[project.readme]`. Ready for further review.

takluyver:
> I think in will both now allow invalid versions on cython on some versions of cpython 

In practice, I don't think this is too big a problem, because the expectation is now for isolated build environments, so it will normally be built with the newest allowed version of Cython, regardless of what's already installed in the destination environment.

It does mean providing slightly less information to downstream packagers about the required Cython version, but I would assume that if you get an error when building for a new Python version, upgrading Cython is a fairly obvious thing to try anyway.

> and puts an artificial cap on version of cython it claims to support (I regularly build h5py with the cython 3 pre-releases and the cython default branch)

We could relax this to `>= 0.29, <4`. But until Cython 3.0 is released, we don't know whether it will build a given version of h5py - e.g. I think the `cimport class` syntax I removed recently (#2195) would have broken with the next Cython pre-release. So I think it probably makes sense to be cautious and exclude 3.0 until we can check it.

If you do a non-isolated build, e.g. with the `dev-install.sh` script, these requirements aren't checked, so you can still try building with a newer Cython regardless of this. But it won't use Cython 3.0 automatically when it comes out.

pyproject.toml
```diff
@@ -1,13 +1,80 @@
...
+
+[project]
+name = "h5py"
+description = "Read and write HDF5 files from Python"
+authors = [
+    {name = "Andrew Collette", email = "andrew.collette@gmail.com"},
+]
+maintainers = [
+    {name = "Thomas Kluyver", email = "thomas@kluyver.me.uk"},
+]
```
takluyver, line 18: It certainly shouldn't be just me here. We should either get rid of the maintainers list altogether or add the other maintainers - at a minimum, @tacaswell, @aragilar and @ajelenak, though there are several others who've been given maintainer access at some point but aren't so active at present.

tacaswell: I propose we keep this list to "people willing to deal with emails about h5py".

takluyver: I'm happy enough for my email to be there - we shouldn't just point everyone to Andrew. I imagine that dealing with emails will mostly mean telling people to use the issue tracker.

pyproject.toml
```diff
@@ -1,13 +1,80 @@
...
 build-backend = "setuptools.build_meta"
+
+[project]
+name = "h5py"
+description = "Read and write HDF5 files from Python"
+authors = [
+    {name = "Andrew Collette", email = "andrew.collette@gmail.com"},
+]
+maintainers = [
+    {name = "Thomas Kluyver", email = "thomas@kluyver.me.uk"},
```
tacaswell, line 17:
```suggestion
    {name = "Thomas Kluyver", email = "thomas@kluyver.me.uk"},
    {name = "Thomas A Caswell", email = "tcaswell@bnl.gov"},
```

in the address I would like to use for this.

pyproject.toml
takluyver: Other maintainers (@aragilar @ajelenak @kif @aparamon ), please do say if you're happy for your name (and optionally email) to be in this list of maintainers. If I don't hear in the next few days, I'll merge this, though you're always welcome to add yourself later.

I'd rather be inclusive than exclusive in any kind of list of credits, so anyone who's made more than trivial contributions is welcome to be named as a maintainer if they'd like to be. But we won't share your real name or email on PyPI without your confirmation.

PR #2180, opened 2022-11-25
Better access to compression/filter information

### LINKED ISSUE #2161: Dataset.compression is None when using hdf5plugin compressors
## Description

When a hdf5 dataset is written using one of the compression filters from hdf5plugin, the dataset has a `compression` attribute of `None`.

This is a feature request to change that.

## Example

```python
import h5py, numpy as np
import hdf5plugin

f = h5py.File("tmp.h5", "w")
dset = f.create_dataset("X", data=np.random.randn(50, 100), **hdf5plugin.LZ4())

assert dset.compression is not None
```

```pytb
AssertionError                            Traceback (most recent call last)
Input In [4], in <cell line: 1>()
----> 1 assert dset.compression is not None
```

But, h5ls is able to see the name of the compression filter, so ideally h5py should too:

```python
f.close()
!h5ls -v tmp.h5
```

```
Opened "tmp.h5" with sec2 driver.
X                        Dataset {50/50, 100/100}
    Location:  1:800
    Links:     1
    Chunks:    {25, 50} 10000 bytes
    Storage:   40000 logical bytes, 40064 allocated bytes, 99.84% utilization
    Filter-0:  HDF5 lz4 filter; see http://www.hdfgroup.org/services/contributions.html-32004 OPT {0}
    Type:      native double
```

## Details

This is due to the possible values for Dataset.compression being hardcoded here: https://github.com/h5py/h5py/blob/1487a54fb5149603dcc32604df4db418ea4f5236/h5py/_hl/dataset.py#L556-L563

It looks like there is an API for getting the names of a filter: [`H5Zget_filter_info `](https://docs.hdfgroup.org/hdf5/develop/group___h5_z.html#ga9ef800ceec249c8819492545def9adba).

But I'm not sure how you would be able to tell whether a filter was a "compressor". 

The complicated solution would be to let hdf5plugin register it's compressors with `h5py`. Hopefully there is a more elegant solution.

## Version info

h5py    3.7.0
HDF5    1.12.2
Python  3.9.12 (main, Mar 26 2022, 15:52:10) 
[Clang 13.0.0 (clang-1300.0.29.30)]
sys.platform    darwin
numpy   1.22.4
cython (built with) 0.29.30
numpy (built against) 1.19.3
HDF5 (built against) 1.12.2

takluyver:
This follows from the discussion in #2161. The changes are:

- `dset.compression` will return 'unknown' rather than `None` if a plugin filter is used and none of the three built-in compression filters (gzip, lzf, szip) are included.
  - Technically the unknown filter doesn't have to be a compression filter, but it probably will be.
  - What should we do if there are plugin filters used *and* one that we recognise (like 'gzip')? In master, and in this PR at present, we still name the filter we recognise.
- New properties `dset.filter_ids` and `dset.filter_names` give you more information on all the filters in use, compression or otherwise.
  - If you need to get the filter options or flags, you still need to go to the low-level API: `dset.id.get_create_plist().get_filter(i)`
- `group.create_dataset_like()` can now copy a filter pipeline even if it doesn't recognise the filters in it.
  - I'm still figuring out parts of how this interacts with keyword arguments.

Closes #2161

### DIFF
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -641,6 +641,21 @@ Reference
 
         Whether Fletcher32 checksumming is enabled (T/F).  See :ref:`dataset_fletcher32`.
 
+    .. attribute:: filter_ids
+                   filter_names
+
+        The numeric filter IDs and the string names (as stored in the file) of
+        the filters in use. Each attribute is a tuple.
+
+        Filters are mostly used to compress data, but can also do things like
+        checksumming (see :ref:`dataset_compression`). Other attributes listed
+        above provide convenient shortcuts to check on common filters.
+        IDs for filters built into h5py can be found in the :mod:`h5py.h5z`
+        module, while filter IDs from plugins are listed in `HDF Group's registry
+        <https://portal.hdfgroup.org/display/support/Registered+Filter+Plugins>`_.
+
+        .. versionadded:: 3.8
+
     .. attribute:: fillvalue
 
         Value used when reading uninitialized portions of the dataset, or None
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -623,6 +623,11 @@ Reference
         String with the currently applied compression filter, or None if
         compression is not enabled for this dataset.  See :ref:`dataset_compression`.
 
+        This only recognises the built-in compression options ``'gzip'``,
+        ``'lzf'`` and ``'szip'``. Other compression mechanisms will show as
+        ``'unknown'`` from h5py 3.8. Use :attr:`filter_ids` and
+        :attr:`filter_names` to get more complete information.
+
     .. attribute:: compression_opts
 
         Options for the compression filter.  See :ref:`dataset_compression`.
--- a/h5py/_hl/filters.py
+++ b/h5py/_hl/filters.py
@@ -244,10 +244,15 @@ def fill_dcpl(plist, shape, dtype, chunks, compression, compression_opts,
     external = _normalize_external(external)
     # End argument validation
 
-    if (chunks is True) or \
-    (chunks is None and any((shuffle, fletcher32, compression, maxshape,
-                             scaleoffset is not None))):
-        chunks = guess_chunk(shape, maxshape, dtype.itemsize)
+    if chunks is None and any((
+        shuffle, fletcher32, compression, maxshape, scaleoffset is not None
+    )):
+        chunks = True
+
+    pre_chunked = plist.get_layout() == h5d.CHUNKED and plist.get_chunk() != ()
+    if chunks is True:
+        # Guess chunk shape unless a passed-in property list already has chunks
+        chunks = None if pre_chunked else guess_chunk(shape, maxshape, dtype.itemsize)
 
     if maxshape is True:
         maxshape = (None,)*len(shape)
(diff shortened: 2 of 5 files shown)


takluyver:
Well, it now looks like I've reimplemented part of HDF5's property list API in Python, to manage filters. That feels kind of silly, but I couldn't see a better way to allow `create_dataset_like()` to do things like 'replace gzip compression with Blosc' or 'leave the compression alone, but add a shuffle filter before the compression step', while also allowing for arbitrary filters.

ajelenak:
@takluyver thanks for this PR. Why should `Dataset.compression` only return something for those three compression filters? Also, are `lzf` and `szip` really built-in because libhdf5 only comes with `gzip`? Although, even this one is a build configuration option. How about expanding for all the registered compression filters? Another issue is that I am aware of some HDF5 datasets out there with more than one compression filter. Don't think that was intended but those data exist now.

Another suggestion: Would something like `Dataset.filters` instead of `filter_ids` and `filter_names` work? I feel like having more than one filter property is not necessary. The `filters` property could return the complete information for each filter: id, name, settings. Is there a performance hit for immediately collecting the full dataset filter information?

takluyver:
> Why should Dataset.compression only return something for those three compression filters?

As you've noted, the notion of reporting compression as a single string doesn't really work, because there can be multiple filters used - either by mistake, or filters like shuffle that are part of the compression even though they don't compress data themselves.

So the compromise I picked here is to add `dset.filter_names` to get a more complete view of the filter pipeline, but leave `dset.compression` working as before for the three names that you can pass in to `grp.create_dataset(..., compression=x)`. I think returning `'unknown'` for other filters makes it clearer that you can't pass it back in. We could make it guess when there's just 1 unknown filter, though.

> Would something like Dataset.filters instead of filter_ids and filter_names work?

It can certainly work technically, but my guess is that for the high-level API, having ways to get just the names or just the IDs makes more sense. So long as the stored names are meaningful, the two should give you the same information, but the names are convenient for human understanding, whereas the IDs are more reliable for writing code (e.g. `if BLOSC_ID in dset.filter_ids`).

I think the filter settings can be left to the low-level API. It's not much extra code to access them:

```python
dset.id.get_create_plist().get_filter(0)
```

docs/high/dataset.rst
```diff
@@ -667,6 +667,11 @@ Reference
         String with the currently applied compression filter, or None if
         compression is not enabled for this dataset.  See :ref:`dataset_compression`.
 
+        This only recognises the built-in compression options ``'gzip'``,
+        ``'lzf'`` and ``'szip'``. Other compression mechanisms will show as
+        ``'unknown'`` from h5py 3.8. Use :attr:`filter_ids` and
```
tacaswell, line 672:
```suggestion
        ``'unknown'`` from h5py 3.16. Use :attr:`filter_ids` and
```

h5py/_hl/filters.py
```diff
@@ -148,6 +148,113 @@ class Gzip(FilterRefBase):
     def __init__(self, level=DEFAULT_GZIP):
         self.filter_options = (level,)
 
+
+class FilterPipeline:
```
tacaswell, line 152: Can this be private?

takluyver:
I've opened #2792 with a simplified version of this, without the changes to `.create_dataset_like()`.

PR #2183, opened 2022-12-03
Micro optimizations to save a few ms here and there (for larger datasets)

hmaarrfk:
Admittedly, this is a tiny optimization, but it avoids a few checks that the dataset will do to ensure the `id` is valid.


hmaarrfk:
I don't think that these "hurt" readability. Though, they may be using more advanced concepts like `rank` instead of `len(shape)`

tacaswell (approved):

hmaarrfk:
Thank you!

PR #2154, opened 2022-09-20
fix wrong file driver version

hongxu-jia:
Due to commit [1] applied in hdf5 (1.13.2), import hdf5 failed

|>>> import h5py
|Traceback (most recent call last):
|  File "<stdin>", line 1, in <module>
|  File "/usr/lib/python3.10/site-packages/h5py/__init__.py", line 56, in <module>
|    from . import h5a, h5d, h5ds, h5f, h5fd, h5g, h5r, h5s, h5t, h5p, h5z, h5pl
|  File "h5py/h5fd.pyx", line 220, in init h5py.h5fd
|RuntimeError: Wrong file driver version # (wrong file driver version #)

Initial driver version to fix the error

[1] https://github.com/HDFGroup/hdf5/commit/42b767fc67ad1e13735e3cee2077f2e108f9463e

Signed-off-by: Hongxu Jia <hongxu.jia@windriver.com>

PR #2187, opened 2022-12-18
Add new H5FD_class_value_t typedef and fix wrong file driver version import error

ajelenak:
Enable `import h5py` to work for libhdf5-1.14.0. Does what  #2154 did but in the proper way.

### DIFF
--- a/h5py/api_types_hdf5.pxd
+++ b/h5py/api_types_hdf5.pxd
@@ -235,40 +235,100 @@ cdef extern from "hdf5.h":
     H5FD_MPIO_INDEPENDENT = 0,
     H5FD_MPIO_COLLECTIVE
 
+  # File driver identifier type and values
+  IF HDF5_VERSION >= (1, 14, 0):
+    ctypedef int H5FD_class_value_t
+
+    H5FD_class_value_t H5_VFD_INVALID      # -1
+    H5FD_class_value_t H5_VFD_SEC2         # 0
+    H5FD_class_value_t H5_VFD_CORE         # 1
+    H5FD_class_value_t H5_VFD_LOG          # 2
+    H5FD_class_value_t H5_VFD_FAMILY       # 3
+    H5FD_class_value_t H5_VFD_MULTI        # 4
+    H5FD_class_value_t H5_VFD_STDIO        # 5
+    H5FD_class_value_t H5_VFD_SPLITTER     # 6
+    H5FD_class_value_t H5_VFD_MPIO         # 7
+    H5FD_class_value_t H5_VFD_DIRECT       # 8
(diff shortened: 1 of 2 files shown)


ajelenak:
All current h5py tests pass on my macOS laptop with libhdf5-1.14.0.

tacaswell:
Why don't we want these changes on main?  So long as the are back-compatible I do not see the benefit of maintaining a second branch as long term I think we will need to keep supporting all versions of libhdf5 in the same Python code base.

Shouldn't these version checks be based on 1.13.x so that it works with the development versions?

ajelenak:
I thought to keep that branch until the 1.14.0 is released which should be this month I think. Also there were some CI job failures with my commits so I don't want to break master.

The 1.13 releases are becoming obsolete when 1.14.0 is out so I thought why bother.

tacaswell:
I am a bit confused by libhdf5's versioning, but if 1.14.0 is not out yet, does the dev version claim to be that version already?

My thought was to use the accurate 1.13.x version gating so that it is at least possible to test some of the changes via released versions of libhdf5.

If CI is failing on main with these patches, then the patches need to be fixed so that CI does not fail, hiding the failures in an extra branch just kicks the can down the road.

ajelenak:
Indeed, libhdf5 versioning was formulated before something like SemVer so is a bit unconventional.

The 1.13.x releases were not simple development ones. They were categorized as experimental on the road to 1.14.0, with no promise of backward compatibility between any of them.  It happened that h5py worked until 1.13.2. I see no good reason for h5py to support them now that 1.14.0 release is 7-10 days away.

ajelenak:
I am not sure if the CI is failing from my changes. I briefly looked them that day and it seemed they were caused by some kind of tox errors. h5py tests did not even run.

takluyver:
I reproduced the failure locally, but it seems to be already gone with the latest version of Tox - it was probably [this issue](https://github.com/tox-dev/tox/issues/2747). Tox 4.0 came out recently, so I guess they've been ironing out a few issues.

nadvornikjiri:
Hi,

since 1.14 is officially out now how do we want to proceed with this one? Will we keep it as a separate branch or merge it to the master and continue with new features there?

nadvornikjiri:
Btw also adding a link to the [forum](https://forum.hdfgroup.org/t/subfiling-support/10594)- basically, Dana confirmed what Aleksandar said here before.

ajelenak:
I will merge next Monday US time unless someone objects. All the current comments are resolved and all h5py tests pass with libhdf5-1.14.0 on my laptop.

takluyver:
Yes, merging it to master is the plan - see PR #2194.

PR #2190, opened 2022-12-21
Some fixes for testing with tox 4.0

takluyver:
I was overly optimistic in my assessment that a recent bugfix release of Tox had fixed the issue we were seeing - it had fixed the one issue I had looked at, but not a couple of others.

One config option has simply been renamed, and there's currently an issue with defining a mixture of environments that install an already-built wheel and environments that don't install the package at all. I copied a workaround described at https://github.com/tox-dev/tox/issues/2442#issuecomment-1347555684 .

### DIFF
--- a/ci/azure-pipelines-test-wheels.yml
+++ b/ci/azure-pipelines-test-wheels.yml
@@ -13,6 +13,9 @@ steps:
   displayName: Install tox
 - bash: |
     ls $PIPELINE_WORKSPACE/${{ parameters.platform }}_wheels/
+    echo "PIPELINE_WORKSPACE=$PIPELINE_WORKSPACE"
+    echo "parameters.platform=${{ parameters.platform }}"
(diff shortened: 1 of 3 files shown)


takluyver:
Well, I fixed the issues from the previous build, but some new ones have shown up for whatever reason. :confused:

takluyver:
I think it's significant that the failing jobs have this message:

```
ROOT: No tox.ini or setup.cfg or pyproject.toml found, assuming empty tox.ini at D:\a\1\s
```

It wasn't there when they were passing. :thinking:

takluyver:
OK, the last remaining failure is with the Windows wheel tests - it can't install the wheel package.

My best guess is that Tox 3.x accepted a glob pattern for `--installpkg`, maybe accidentally, whereas Tox 4 expects a concrete file path. We're using bash for the relevant step, which should expand the glob before calling Tox, but the mixture of Windows style `D:\` and Unix style `/d/` paths stops it working.

takluyver:
Finally everything is passing again. I'll merge this later today (UK time) if no-one objects.

PR #2193, opened 2023-01-03
Don't handle boolean selection on 1D dataset as fancy indexing

### LINKED ISSUE #2189: Large performance regression on boolean indexing in 3.7.0
We are seeing a very severe performance regression when using boolean indexing between h5py 3.6.0 and 3.7.0.

Bisecting suggests
first bad commit: [b556856819de6fcff7248479c6f42b18af6ced21] Convert boolian arrays into index arrays for fancy slicing

%%timeit of a test script on our data which makes heavy use of boolean indexing with 1-D arrays of length 1e5-1e6 
 
h5py @ 3051c76a:
21.9 s ± 191 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

h5py @ b5568568:
39min 23s ± 26.8 s per loop (mean ± std. dev. of 7 runs, 1 loop each)

RedHat Enterprise Linux 8 Conda environment. Same hdf5 version used for both tests.

h5py    3.6.0
HDF5    1.10.6
Python  3.10.8 (main, Nov 24 2022, 14:13:03) [GCC 11.2.0]
sys.platform    linux
numpy   1.23.4
cython (built with) 0.29.32
numpy (built against) 1.23.4
HDF5 (built against) 1.10.6

takluyver:
Hopefully closes #2189.

@tguruswamy can you test if this fixes your use case?

### DIFF
--- a/h5py/_selector.pyx
+++ b/h5py/_selector.pyx
@@ -180,6 +180,11 @@ cdef class Selector:
                 if a.ndim != 1:
                     raise TypeError("Only 1D arrays allowed for fancy indexing")
                 if a.dtype.kind == 'b':
+                    if self.rank == 1:
+                        # The dataset machinery should fall back to a faster
+                        # alternative (using PointSelection) in this case.
+                        # https://github.com/h5py/h5py/issues/2189
+                        raise TypeError("Use other code for boolean selection on 1D dataset")
                     if a.size != l:
                         raise TypeError("boolean index did not match indexed array")
                     a = a.nonzero()[0]
--- a/tox.ini
+++ b/tox.ini
@@ -90,7 +90,9 @@ skip_install=True
 # Work around https://github.com/tox-dev/tox/issues/2442
 package_env = DUMMY NON-EXISTENT ENV NAME
 deps=pre-commit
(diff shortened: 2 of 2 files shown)


takluyver:
@tacaswell @aragilar I'd like to get this in before doing a new release. It's not the most elegant solution, but it hopefully solves a ~100x performance regressions for a particular use case. @tguruswamy I'm still hoping you can confirm that this does actually solve the issue you see.

It's somewhat tricky to write a good test for. The same thing we care about works either way, it's just a matter of how quickly it works. We could obviously test that the code path I've modified rejects this specific case, but that's an implementation detail, not the outcome we care about.

This might be a good case for an asv benchmark, but we've never got round to setting up anything to run that regularly and spot regressions.

tguruswamy:
My use case does show a return to previous performance:

h5py @ 70cfcb88534492583bf199f64544034407e8bcef:
21.9 s ± 700 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

Thanks for the quick response, it's really appreciated.

tacaswell (approved):

takluyver:
Thanks for testing it :slightly_smiling_face:

PR #2194, opened 2023-01-04
Changes necessary to support HDF5 1.14

takluyver:
Now that 1.14 has been released, we should bring these changes into master, and also set up testing with 1.14.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -76,7 +76,7 @@ jobs:
       parameters:
         platform: macos_$(arch)
 
-- job: 'ubuntu1604'
+- job: 'ubuntu'
   pool:
     vmImage: ubuntu-20.04
   strategy:
@@ -103,12 +103,6 @@ jobs:
         HDF5_VERSION: 1.10.3
         HDF5_DIR: $(HDF5_CACHE_DIR)/$(HDF5_VERSION)
         H5PY_ENFORCE_COVERAGE: yes
-      py37-deps-hdf51106:
-        python.version: '3.7'
-        TOXENV: py37-test-deps
-        HDF5_VERSION: 1.10.6
-        HDF5_DIR: $(HDF5_CACHE_DIR)/$(HDF5_VERSION)
-        H5PY_ENFORCE_COVERAGE: yes
       py38-deps-hdf51107:
         python.version: '3.8'
         TOXENV: py38-test-deps
@@ -125,6 +119,12 @@ jobs:
         python.version: '3.9'
         TOXENV: py39-test-deps
         HDF5_VERSION: 1.12.2
+        HDF5_DIR: $(HDF5_CACHE_DIR)/$(HDF5_VERSION)
+        H5PY_ENFORCE_COVERAGE: yes
+      py310-deps-hdf51140:
+        python.version: '3.10'
+        TOXENV: py310-test-deps
+        HDF5_VERSION: 1.14.0
         HDF5_DIR: $(HDF5_CACHE_DIR)/$(HDF5_VERSION)
         H5PY_ENFORCE_COVERAGE: yes
     # do mpi tests
@@ -200,13 +200,6 @@ jobs:
         HDF5_VSVERSION: "16-64"
         H5PY_ENFORCE_COVERAGE: yes
     # 64 bit - HDF5 1.10
-      py38-hdf5110:
-        python.version: '3.8'
-        TOXENV: py38-test-deps,py38-test-mindeps,py38-test-deps-pre
-        HDF5_VERSION: 1.10.6
-        HDF5_DIR: $(HDF5_CACHE_DIR)/$(HDF5_VERSION)
-        HDF5_VSVERSION: "16-64"
-        H5PY_ENFORCE_COVERAGE: yes
       py39-hdf5110:
         python.version: '3.9'
         TOXENV: py39-test-deps,py39-test-mindeps,py39-test-deps-pre
@@ -220,6 +213,13 @@ jobs:
         python.version: '3.8'
         TOXENV: py38-test-deps
         HDF5_VERSION: 1.12.0
+        HDF5_DIR: $(HDF5_CACHE_DIR)/$(HDF5_VERSION)
+        HDF5_VSVERSION: "16-64"
+        H5PY_ENFORCE_COVERAGE: yes
+      py310-hdf5114:
+        python.version: '3.10'
+        TOXENV: py310-test-deps
+        HDF5_VERSION: 1.14.0
         HDF5_DIR: $(HDF5_CACHE_DIR)/$(HDF5_VERSION)
         HDF5_VSVERSION: "16-64"
         H5PY_ENFORCE_COVERAGE: yes
--- a/docs/high/file.rst
+++ b/docs/high/file.rst
@@ -452,10 +452,12 @@ Reference
                     ``h5.get_config().track_order``.
     :param fs_strategy: The file space handling strategy to be used.
             Only allowed when creating a new file. One of "fsm", "page",
-            "aggregate", "none", or None (to use the HDF5 default).
+            "aggregate", "none", or ``None`` (to use the HDF5 default).
     :param fs_persist: A boolean to indicate whether free space should be
             persistent or not. Only allowed when creating a new file. The
             default is False.
+    :param fs_page_size: File space page size in bytes. Only use when
+            fs_strategy="page". If ``None`` use the HDF5 default (4096 bytes).
     :param fs_threshold: The smallest free-space section size that the free
             space manager will track. Only allowed when creating a new file.
             The default is 1.


takluyver:
The Travis ppc64le builds seem to be persistently failing for no clear reason - they just get stuck at some point, and eventually time out. It's happening consistently on master as well, so it's not an issue with this branch. Although weirdly some other pull requests seem to have built successfully. :confused:

aragilar:
Should we merge this and ask @drew-parsons (being the Debian maintainer) or anyone else if they can test this on non x86 systems they have access to?

drew-parsons:
> Should we merge this and ask @drew-parsons (being the Debian maintainer) or anyone else if they can test this on non x86 systems they have access to?

I can test the updated h5py, but note that Debian has been tardy updating libhdf5. 1.10.8 is still being used in Debian unstable.

nadvornikjiri:
Hi everybody, I am testing it now with hdf5 version 1.14.0 (building from the git tag) with parallel support. However when I rebuild h5py (latest pull) with it it behaves strange.

import h5py indeed works, but it complains for example on h5py.File that:
```
import h5py
>>> h5py.File
AttributeError: module 'h5py' has no attribute 'File'
```

When I check the h5py module is indeed empty:
```
>>> for i in dir(h5py):
...     print(i)
... 
__doc__
__file__
__loader__
__name__
__package__
__path__
__spec__
```

I am building it the same way as with any earlier hdf5 version. 

Thanks for pointing me in the right direction.

takluyver:
I've no idea how a build error could cause that - the `File` class is defined in pure Python code.

Can you check `>>> h5py` in your interpreter, to see where it's loaded from? I would guess it's finding the wrong folder.

nadvornikjiri:
Well, this I see for the first time:
```
>>> import inspect
>>> inspect.getsourcefile(h5py)
  File "/usr/lib/python3.8/inspect.py", line 696, in getsourcefile
    filename = getfile(object)
  File "/usr/lib/python3.8/inspect.py", line 659, in getfile
    raise TypeError('{!r} is a built-in module'.format(object))
TypeError: <module 'h5py' (namespace)> is a built-in module
```

If I go to the venv folder then it is in there:
```
find . -name "*h5py*"
./lib/python3.8/site-packages/h5py-3.8.0-py3.8-linux-x86_64.egg
./lib/python3.8/site-packages/h5py-3.8.0-py3.8-linux-x86_64.egg/h5py
```

But indeed there are no __init__.py files in these folders.

Just to be sure, this is the procedure I'm using for building h5py for almost 2 years already:
```
export CC=mpicc
export HDF5_MPI="ON"
export HDF5_DIR=<hdf5_path>/hdf5
export LD_LIBRARY_PATH=<hdf5_path>/hdf5/lib:$LD_LIBRARY_PATH
python setup_configure.py --mpi
pip uninstall h5py
python setup.py clean --all
python setup.py install
```

Were there some changes lately with the h5py setup.py script?

takluyver:
Not lately, although I think your `python setup_configure.py --mpi` line has been a no-op for some time now - the environment variables are now used directly during the build, rather than having a separate configure step.

Using `setup.py install` is no longer recommended - I'd suggest using `pip install .` instead (see [the docs on building frmo source](https://docs.h5py.org/en/stable/build.html#source-installation)). I don't know if this is the cause of what you see, but it's worth a try.

nadvornikjiri:
Ah yes, that did it. My tests are now passing fine with 1.14.0. Interesting that I didn't have this problem with 1.13.x versions, only with the 1.14. I will be building with pip from now on.

nadvornikjiri:
I have now started to add some modifications to the low level API to support the Subfiling VFD ([H5Pset_fapl_subfiling()](https://docs.hdfgroup.org/hdf5/develop/group___f_a_p_l.html#ga65dbddbba216fdd0bdf99b4feaa74db2)). But in the middle of it I already realized somebody is working on that? 

```
(api_types_hdf5.pxd):
  # File driver identifier type and values
  IF HDF5_VERSION >= (1, 14, 0):
...
    H5FD_class_value_t H5_VFD_SUBFILING    # 12
...

```
But I didn't find the set_fapl function in the code or the docs.. Is somebody working on that? Maybe I should file another issue for it..

Thanks for pointing me.

takluyver:
I'm not aware of anyone working on it. I think all of those enum values were added when h5py was made compatible with HDF5 1.14 (#2187), but it wasn't specfically focused on subfiling.

nadvornikjiri:
Aha, and what's the usual process for adding these new features from new releases? Sorry for a basic question, I would be happy to make it myself as my premiere pull request, but not sure how to approach it.

I see that adding this VFD won't be that trivial - there are some configuration typedefs the subfiling fapl requires and to be honest I am not entirely sure where I should put them. Also, I guess it would make sense to make this another string option for the "driver" attribute of the File.open high-level API with some logic about the configuration setting/consistency checks.

takluyver:
No problem! There's no particular process - it's generally just done by whoever is interested. I might suggest approaching the low-level and high-level API as two separate steps - exposing new functionality in the low level is usually pretty mechanical, whereas the high level API may bring more design questions.

In terms of what you'll actually need to do:

- `api_types_hdf5.pxd` contains the non-function symbols we use from HFDF5. This is where you'll add things like typedefs. You can see it's divided into blocks by comments; try to add things in the relevant block. You can use `IF HDF5_VERSION >= (1, 14, 0):` to make them conditional on the HDF5 version (the capital `IF` is checked at build time).
- `api_functions.txt` lists the functions we use  from HDF5. This is where you'll add `H5Pset_fapl_subfiling`, again finding the right section. There's a custom syntax to give a minimum HDF5 version. This is used to generate a bunch of simple wrapper functions, which turn HDF5 errors into Python exceptions.
- Then `h5p.pyx` (and similar files) is where we expose the new functionality to be used from Python code. E.g. the function taking a File Access Property List as the first argument is wrapped in a method on the `PropFAID` class.

nadvornikjiri:
Hi there again, I'm now trying to put it together and I am a little lost on how I should handle the configuration structures. It's clear where to define the structures, but how to use them in the h5p.pyx function?

`def set_fapl_subfiling(self, threadpool_size, ioc_selection, stripe_count, stripe_size): ...`


There are 2 configuration structures - [H5FD_ioc_config_t](https://docs.hdfgroup.org/hdf5/develop/struct_h5_f_d__ioc__config__t.html) which is needed for the [H5Pset_fapl_ioc](https://docs.hdfgroup.org/hdf5/develop/group___f_a_p_l.html#ga714340ec9cbb65cb0a819e1d775825f0) method that is required for setting the parameter [ioc_fapl_id](https://docs.hdfgroup.org/hdf5/develop/struct_h5_f_d__subfiling__config__t.html#a1ab8be7dd1e8b8c50eb8bd1b87513bd0) of the [H5FD_subfiling_config_t ](https://docs.hdfgroup.org/hdf5/develop/struct_h5_f_d__subfiling__config__t.html) configuration needed by the [H5Pset_fapl_subfiling](https://docs.hdfgroup.org/hdf5/develop/group___f_a_p_l.html#ga65dbddbba216fdd0bdf99b4feaa74db2) method.

I have dug out that there is a similar class with parameters accessible as properties - the CacheConfig class in h5ac.pyx though it's not entirely clear to me where the self.cache_config property of that class is initialized.. Should I use something similar for these two additional config structures and add them as 2 separate classes and pyx +pxd files?

I wouldn't care too much about having these classes, but I certainly need to set parameters of the H5FD_subfiling_params_t structure such as the ioc_selection, stripe_size, and stripe_count.

Thanks for showing me the way, I can figure out the rest! :)

tacaswell:
Yes, that sounds like the right approach.  

At the c-level we need pointers to the right kind of structs to pass into libhdf5 so we need an exact mirror/exposure of all of the configuration structs in the low level (cython) API.

In the high-level API is where we make it "Pythonic" and (agreeing with Thomas) is where actual design questions come in.

nadvornikjiri:
Maybe I hid the question on what's not clear to me now - where is defined cache_config property of the CacheConfig class? Is there some guide on where I need to put it so the "cythonizer" picks it up correctly?

Thanks for help.

nadvornikjiri:
Actually, I already chose to do it already by the `H5AC_cache_config_t` example and already have it cythonizing correctly. However, when I reach the mpicc (building h5py.defs) then it complains:

```
/home/caucau/SDSSCube/ext_lib/h5py/h5py/defs.c:29207:80: error: unknown type name ‘H5FD_subfiling_config_t’; did you mean ‘H5FDsubfiling_H’?
 _t __pyx_f_4h5py_4defs_H5Pset_fapl_subfiling(hid_t __pyx_v_fapl_id, H5FD_subfiling_config_t const *__pyx_v_vfd_config) {
                                                                     ^~~~~~~~~~~~~~~~~~~~~~~
                                                                                H5FDsubfiling_H

```

The api_gen.py does not throw any errors, but I'm suspecting to not generating the pyx objects correctly.. I've added it by the example of the cache config to the api_types_hdf5.pxd:

```
# === H5FD - Subfiling ========================================================


  IF HDF5_VERSION >= (1, 14, 0):
    unsigned int H5FD_IOC_CURR_FAPL_VERSION
    unsigned int H5FD_SUBFILING_CURR_FAPL_VERSION
    unsigned int H5FD_IOC_FAPL_MAGIC
    unsigned int H5FD_SUBFILING_FAPL_MAGIC

    cdef extern from "H5FDioc.h":
      # === H5FD - IO Concentrator configuration API ================================
      cdef enum H5FD_subfiling_ioc_select_t:
        SELECT_IOC_ONE_PER_NODE,
        SELECT_IOC_EVERY_NTH_RANK,
        SELECT_IOC_WITH_CONFIG,
        SELECT_IOC_TOTAL

      ctypedef struct H5FD_ioc_config_t:
        uint32_t magic
        uint32_t version
        int32_t thread_pool_size

    cdef extern from "H5FDsubfiling.h":
      # === H5FD - Subfiling configuration API ================================
      ctypedef struct H5FD_subfiling_params_t:
        H5FD_subfiling_ioc_select_t ioc_selection
        int64_t stripe_size
        int32_t stripe_count

      ctypedef struct H5FD_subfiling_config_t:
        uint32_t magic
        uint32_t version
        hid_t ioc_fapl_id
        hbool_t require_ioc
        H5FD_subfiling_params_t shared_cfg

```

Thanks for investing the time into me, next time I'll be contributing less painfully :)..

takluyver:
Can you push your changes somewhere so I can have a look?

Normally all the structs and enums you need are available through `hdf5.h` (it includes all the other public `.h` files), so it shouldn't be necessary to `cdef extern` more files.

nadvornikjiri:
Where can I push it? I have a new branch that I could push to the h5py repo, but I don't have access (ERROR: Permission to h5py/h5py.git denied to nadvornikjiri).

takluyver:
The norm on Github is that you fork the repo (click the fork button on the top right of this page) and push to your fork. Then you can make a pull request back to the original repo.

nadvornikjiri:
Here you go https://github.com/nadvornikjiri/h5py - it's on branch subfiling.

takluyver:
I think subfiling is a feature specific to parallel HDF5, right? If so, the wrapper function definitions will need an MPI marker, like these:

https://github.com/h5py/h5py/blob/3507819de54b35af05b2ca8ca55ec7e7b60cb919/h5py/api_functions.txt#L381-L382

And then the low-level methods need to be conditional as well:

https://github.com/h5py/h5py/blob/3507819de54b35af05b2ca8ca55ec7e7b60cb919/h5py/h5f.pyx#L467

takluyver:
(Feel free to open a pull request, by the way - even if it's not ready to merge, it's a convenient way to review code)

nadvornikjiri:
OK, let's continue within the pull request above.

PR #2195, opened 2023-01-04
Remove unnecessary Cython 'cimport class' syntax

takluyver:
This appears to be an old leftover that hasn't been needed for a long time, and is being removed for Cython 3.0 (https://github.com/cython/cython/pull/4904). PyCharm's static analysis also doesn't recognise it, leading to spurious warnings.

### DIFF
--- a/h5py/h5s.pxd
+++ b/h5py/h5s.pxd
@@ -10,7 +10,7 @@
 
 from .defs cimport *
 
-from ._objects cimport class ObjectID
+from ._objects cimport ObjectID
 
 cdef class SpaceID(ObjectID):
     pass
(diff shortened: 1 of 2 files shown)


tacaswell (approved):

tacaswell:
This appears to be hanging on 

```
$ ls -lRa $HDF5_CACHE_DIR
```

on one of the travis jobs and has been for a while...

takluyver:
It seems to be inconsistent where the Travis ppc64le job gets stuck - I don't think it's always on the `ls` command, sometimes it's while compiling, and sometimes while it's still setting up (before running anything specific to h5py). We could try getting rid of the recursive `ls`, though. :shrug:

PR #2196, opened 2023-01-05
Remove recursive ls on Travis

takluyver:
In the vague hope that this might stop ppc64le builds getting stuck. I'm not convinced that it will, but it's worth a try.


takluyver:
It worked on the PR, though builds seem to fail more often for master than for PRs even where they don't touch this. I'll merge this change and we can see if it works in master.

takluyver:
Hasn't fixed it, unfortunately.

PR #2198, opened 2023-01-05
Build low-level API docs on ubuntu-latest image

takluyver:
Ubuntu 18.04 images are deprecated and will be removed in April. There's a brownout today which I noticed.

https://github.com/actions/runner-images/issues/6002


takluyver:
I also now updated the versions of some of the actions we're using, which got rid of a few deprecation warnings.

Issue #2199, opened 2023-01-06
ppc64le builds on Travis timing out


takluyver:
Creating an issue as a central place to record the issue and potential ideas. All ppc64le builds on Travis from master appear to be hanging, and then failing with the "No output has been received in the last 10m0s" message. Where it appears to get stuck varies - sometimes it seems to hang during the general setup before it runs anything from our repository.

Weirdly, quite a few pull request builds succeed, but the failure is very consistent on master. :confused: 

https://app.travis-ci.com/github/h5py/h5py/builds

Things we've already tried:

- Removing a recursive `ls` call where it sometimes seemed to be stuck (#2196)
- Using a newer Ubuntu version (#2197) 
- Deleting the cache for the master branch on Travis (via https://app.travis-ci.com/github/h5py/h5py/caches ) and re-running the build

The Travis forum for [multi CPU architectures](https://travis-ci.community/c/environments/multi-cpu-arch/96) contains questions from quite a few people seeing similar things, but I've not yet found any good advice on how to resolve it. There are [some pointers](https://travis-ci.community/t/builds-hang-with-output-truncated-mid-line/7611/14) that not all the output is showing up, so it might actually be getting stuck later than it appears to. Other threads (e.g. [this one](https://travis-ci.community/t/multiarch-builds-lxd-always-time-out-on-failure-instead-of-exit/7659)) say that you should avoid the `set -e` bash option in the main commands that Travis runs - but I don't think we're using that.

Unfortunately, I'm not aware of any other CI platform providing native ppc64le builds. You can use emulated builds on Github actions, but that's likely to be very slow.

vasole:
I'm experiencing the same issue and I have tried newer distributions too.

I have contacted Travis-CI support and they suggested to go back to focal what I have done. However, the problem seems to be still there:

https://app.travis-ci.com/github/vasole/pymca/builds/259446667

vasole:
Just in case, the trouble ticket at Travis is:

https://support.travis-ci.com/hc/en-us/requests/42395

vasole:
It seems to have started now (using focal).

takluyver:
Thanks @vasole ! I guess we can try with focal as well.

aragilar:
It looks like #2200 didn't fix this, should we drop travis for this, and look at https://osuosl.org/? From https://osuosl.org/services/powerdev/current-projects/ it looks like numpy is supported, I'm happy to fill out the form and ask them what's required on our side.

[...]

PR #2200, opened 2023-01-06
Try using Ubuntu 20.04 on Travis

takluyver:
See discussion on #2199

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: python
 os: linux
-dist: bionic
+dist: focal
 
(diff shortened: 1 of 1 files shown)


takluyver:
Same old failure :disappointed:

vasole:
I would suggest to contact travis support: `supportATtravis-ci.com`

Perhaps they did something behind the scenes...

vasole:
I do not have a support contract either, but they were reactive.

takluyver:
Not needed for now.

PR #2203, opened 2023-01-11
Allow arbitrary array-ish objects for fancy indexing

### LINKED ISSUE #1757: Using tuple as dataset index raises TypeError
### How to reproduce?

```python
import h5py

with h5py.File("test.h5", mode="w") as h5f:
    dset = h5f.create_dataset("matrix", (2, 2), dtype="=f8")

with h5py.File("test.h5", mode="r") as h5f:
    print(h5f["matrix"][0, (0, 1)])
```

### Actual result

```
  File "bug.py", line 7, in <module>
    print(h5f["matrix"][0, (0, 1)])
  File ".../h5py/_hl/dataset.py", line 777, in __getitem__
  File ".../h5py/_hl/selections.py", line 82, in select
  File "h5py/_selector.pyx", line 269, in h5py._selector.Selector.make_selection
  File "h5py/_selector.pyx", line 210, in h5py._selector.Selector.apply_args
TypeError: not all arguments converted during string formatting
```

### Expected result

```
[0. 0.]
```

### System info

* Operating System: Ubuntu 18.04.5 LTS
* Python version: 3.8.6
* Where Python was acquired: apt

Bisection result:
```text
474ef53d2698f6fa8988df2a2ec99525863bc8c4 is the first bad commit
commit 474ef53d2698f6fa8988df2a2ec99525863bc8c4
Author: Thomas Kluyver <thomas@kluyver.me.uk>
Date:   Thu Apr 2 15:37:32 2020 +0100

    Use selector in creating selections
```

### LINKED ISSUE #2118: regression: python range not accepted as FancySelection

takluyver:
Including tuples and range objects (though slices should generally be preferred over range). This should match numpy's behaviour more - anything that `asarray()` converts to a 1D array of integers should work.

Still under discussion with @tacaswell, but I think we're leaning this way.

Closes #1757
Closes #1758 
Closes #2118 
Closes #2119

### DIFF
--- a/h5py/_selector.pyx
+++ b/h5py/_selector.pyx
@@ -15,6 +15,7 @@ from numpy cimport (
 )
 from cpython cimport PyNumber_Index
 
+import collections.abc
 import numpy as np
 from .defs cimport *
 from .h5d cimport DatasetID
@@ -172,8 +173,8 @@ cdef class Selector:
                 continue
 
             # [[0, 2, 10]] - list/array of indices ('fancy indexing')
-            if isinstance(a, (list, np.ndarray)):
-                if isinstance(a, list) and len(a) == 0:
+            if isinstance(a, (np.ndarray, collections.abc.Sequence)):
+                if not isinstance(a, np.ndarray) and len(a) == 0:
                     a = np.asarray(a, dtype=np.intp)
                 else:
                     a = np.asarray(a)
--- a/h5py/_selector.pyx
+++ b/h5py/_selector.pyx
@@ -173,54 +173,55 @@ cdef class Selector:
                 continue
 
             # [[0, 2, 10]] - list/array of indices ('fancy indexing')
-            if isinstance(a, (np.ndarray, collections.abc.Sequence)):
-                if not isinstance(a, np.ndarray) and len(a) == 0:
-                    a = np.asarray(a, dtype=np.intp)
-                else:
-                    a = np.asarray(a)
+            if isinstance(a, np.ndarray):
                 if a.ndim != 1:
                     raise TypeError("Only 1D arrays allowed for fancy indexing")
-                if a.dtype.kind == 'b':
-                    if self.rank == 1:
-                        # The dataset machinery should fall back to a faster
-                        # alternative (using PointSelection) in this case.
-                        # https://github.com/h5py/h5py/issues/2189
-                        raise TypeError("Use other code for boolean selection on 1D dataset")
-                    if a.size != l:
-                        raise TypeError("boolean index did not match indexed array")
-                    a = a.nonzero()[0]
-                if not np.issubdtype(a.dtype, np.integer):
-                    raise TypeError("Indexing arrays must have integer dtypes")
-                if array_ix != -1:
-                    raise TypeError("Only one indexing vector or array is currently allowed for fancy indexing")
+            else:
+                arr = np.asarray(a)
+                if arr.ndim != 1:
+                    raise TypeError("Selection can't process %r" % a)
+                a = arr
+                if a.size == 0:
+                    # asarray([]) gives float dtype by default
+                    a = a.astype(np.intp)
 
-                # Convert negative indices to positive
-                if np.any(a < 0):
-                    a = a.copy()
-                    a[a < 0] += l
+            if a.dtype.kind == 'b':
+                if self.rank == 1:
+                    # The dataset machinery should fall back to a faster
+                    # alternative (using PointSelection) in this case.
+                    # https://github.com/h5py/h5py/issues/2189
+                    raise TypeError("Use other code for boolean selection on 1D dataset")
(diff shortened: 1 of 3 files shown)


takluyver:
For reference, I found the equivalent indexing code in NumPy here (written in C):

https://github.com/numpy/numpy/blob/df1ee34c7faea1cfc2aef0719444b6fa43887b42/numpy/core/src/multiarray/mapping.c#L408-L607

h5py/_selector.pyx
```diff
@@ -15,6 +15,7 @@ from numpy cimport (
 )
 from cpython cimport PyNumber_Index
 
+import collections.abc
```
tacaswell, line 18:
```suggestion
```

tacaswell (approved):
I think the other option is documenting how we differ from numpy and forever closing issues that the difference is intentional.

takluyver:
Yup, that sounds viable too. Fancy indexing in h5py is already more restricted than NumPy - e.g. it's only allowed on a single axis, and only with indices in increasing order - so we could reasonably document that it accepts limited types as well.

Other maintainers (@aragilar @ajelenak @kif etc.) feel free to weigh in. NumPy accepts all sorts of things for fancy indexing, and this PR would make h5py similar. Obviously there's an advantage to matching NumPy's API, but this can also be surprising in some cases - tuples require an odd special case to index a single axis, and range() objects would work but much slower than an equivalent slice().

One way or the other, I'd like to decide on this for 3.8 and close the various issues and PRs concerning it.

takluyver:
I don't know of any agreement along those lines.

I went and looked into Dask's code, from what I can see it appears to follow NumPy in allowing anything that can be converted into a NumPy array (on [this line](https://github.com/dask/dask/blob/dfcefd8891310c638922ee9d2546b2d24e647743/dask/array/slicing.py#L73)). It's even more relaxed than NumPy in that it will accept an array of floats for indexing if they're all close to integers.

tacaswell:
Given that we are all on the fence, I'm going to be the jerk who makes a decision and merges this :)

PR #2209, opened 2023-01-18
Python 3.11 build

johnnynunez:
Python 3.11 is out in anaconda https://github.com/ContinuumIO/anaconda-issues/issues/13082


takluyver:
Thanks, but we've already got some 3.11 testing - it works by building wheels in one job and then installing and testing these in another. Our CI is in a kind of constant tension between covering more cases and not getting too slow, so I don't want to add more jobs at the moment.

PR #2216, opened 2023-01-23
Try to fix ARM wheel upload error

aragilar:
It looks like the ARM wheels have been failing for ~7 months now (based on the uploaded wheels). This does the following to address that:
 * Make the script set -e, so that it errors if any command errors.
 * Add the missing conda-package-handling to hopefully let the script pass.

It's possible that this setup no longer works, and we need to use an alternative to anaconda-client to upload wheels (but it looks like other projects have no issues uploading, so we'll need to see what they do).

### DIFF
--- a/ci/upload_wheels.sh
+++ b/ci/upload_wheels.sh
@@ -1,5 +1,12 @@
-ANACONDA_ORG="scipy-wheels-nightly";
-pip install git+https://github.com/Anaconda-Server/anaconda-client;
+set -e
+
+ANACONDA_ORG="scipy-wheels-nightly"
+
+# unclear why conda-package-handling is not on PyPI, but anaconda-client needs
+# it, and anaconda-client on PyPI is *really* old and the devs seem to no longer
+# upload it?
+pip install git+https://github.com/conda/conda-package-handling/
+pip install git+https://github.com/Anaconda-Server/anaconda-client
 
 if [[ "$TRAVIS_EVENT_TYPE" != "cron" && -z "$TRAVIS_TAG" ]] ; then
   echo "Not uploading wheels (build not for cron or git tag)"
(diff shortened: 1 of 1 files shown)


takluyver:
Ah, thanks, I had opened #2217 before seeing this. If this works, it's probably better than my attempt (using an older version of anaconda-client).

takluyver:
And indeed it does seem to work - thanks!

I think we should consider this a short-term fix, though. It seems like the anaconda tools don't really support installing with pip, even though they've still got `setup.py` packaging in the repositories. Longer term, I guess we should either:

1. Set up a whole conda/mamba installation just to upload some packages (straightforward but unsatisfying)
2. Do the upload without Anaconda tools - presumably it's just some fairly straightforward HTTP requests we can do without special packages.
3. Upload somewhere else.

aragilar:
I figured I'd make the smallest possible change. It look like matplotlib uses mamba as part of github actions: https://github.com/matplotlib/matplotlib/blob/8ed4c02633dd7ee1e940e3b15d4eafd738ff1d6c/.github/workflows/nightlies.yml.

I'm unclear what's the advantage of https://anaconda.org/scipy-wheels-nightly/ v.s. some other hosting (Does it provide a PyPI-like index?), and is it used for anything else other storage (e.g. is anyone doing tests against h5py dev versions)?

takluyver:
> I'm unclear what's the advantage of https://anaconda.org/scipy-wheels-nightly/ v.s. some other hosting (Does it provide a PyPI-like index?)

I think there just aren't that many places that we can easily upload all those files, from the command line, and keep them organised together in a list. We need either a pretty generous amount of storage space, or something that will automatically delete them after a set time (our other wheels come as Azure build artifacts, which are only kept for a couple of weeks or so). And naturally we want it to be free, so we don't have to figure out who's going to pay.

> and is it used for anything else other storage (e.g. is anyone doing tests against h5py dev versions)?

I don't think so - we've only ever uploaded the ARM wheels there, and I assume if people were using it we'd get pressure to upload the other packages there too.

takluyver:
Unfortunately, this doesn't fix it - it looks like anaconda-client really doesn't expect to be used outside a conda environment.

New plan: I'm going to try to use Travis' support for [uploading to Github releases](https://docs.travis-ci.com/user/uploading-artifacts/).

Issue #2215, opened 2023-01-23
3.8.0  manylinux_2_17_aarch64 build is missing in pypi


Chizkiyahu:
To assist reproducing bugs, please include the following:
 * Operating System - aarch64 (docker on mac m1)
 * Python version  3.7-3.10
 * h5py version 3.8.0
  * The full traceback/stack trace shown (if it appears)

### Dockerfile that works 
```
FROM python:3.9-slim
RUN pip install h5py==3.7.0
 ```
### Dockerfile that fails
```
FROM python:3.9-slim
RUN pip install h5py==3.8.0
 ```
 log 
 ```
#5 [2/2] RUN pip install h5py==3.8.0
#5 sha256:b52b13691e8b2670dfaf38bfd1594024d9105a451ae19f70aee567da098ebcd4
#5 1.317 Collecting h5py==3.8.0
#5 1.753   Downloading h5py-3.8.0.tar.gz (400 kB)
#5 2.069      ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 400.8/400.8 KB 1.3 MB/s eta 0:00:00
#5 2.200   Installing build dependencies: started
#5 10.63   Installing build dependencies: finished with status 'done'
#5 10.63   Getting requirements to build wheel: started
#5 10.84   Getting requirements to build wheel: finished with status 'done'
#5 10.91   Installing backend dependencies: started
#5 12.45   Installing backend dependencies: finished with status 'done'
#5 12.46   Preparing metadata (pyproject.toml): started
#5 12.65   Preparing metadata (pyproject.toml): finished with status 'done'
#5 12.85 Collecting numpy>=1.14.5
#5 12.94   Downloading numpy-1.24.1-cp39-cp39-manylinux_2_17_aarch64.manylinux2014_aarch64.whl (14.0 MB)
#5 15.76      ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 14.0/14.0 MB 5.0 MB/s eta 0:00:00
#5 15.78 Building wheels for collected packages: h5py
#5 15.78   Building wheel for h5py (pyproject.toml): started
#5 16.03   Building wheel for h5py (pyproject.toml): finished with status 'error'
#5 16.04   error: subprocess-exited-with-error
#5 16.04   
#5 16.04   × Building wheel for h5py (pyproject.toml) did not run successfully.
#5 16.04   │ exit code: 1
[...]

PR #2218, opened 2023-01-24
Try uploading wheels from Travis to Github releases instead of anaconda.org

takluyver:
Another attempt at solving #2215.

I'll probably merge this later today and then create a tag so we can get the 3.8 wheels for aarch64.


takluyver:
I briefly created a 'machine user' to own the token, but then realised that Github's new 'fine grained' tokens let me do the same thing without another account, so I've deleted `h5py-uploader` again. The 'fine grained' tokens can only be valid for a maximum of 1 year, though, which is annoying given that that's one or two releases for us.

tacaswell:
I think we want to keep the upload to anaconda for nightlies though?

takluyver:
Oh, sorry, I merged before seeing your comment. If the nightlies have been broken for 7 months without anyone noticing, I think that suggests it's a waste of effort to fix the upload. Hopefully the Travis deployment thing will work with less hassle.

takluyver:
Of course we can always revisit that if we do want nightly aarch64 wheels. For now I'd just like to get them up for the release.

takluyver:
This worked, but I had to switch back from the new 'fine grained' access token to a 'classic' token with much broader scope. I might create a machine user again to limit the scope to this one repo, or maybe Travis will add support for the fine-grained tokens.

tacaswell:
> Of course we can always revisit that if we do want nightly aarch64 wheels. For now I'd just like to get them up for the release.

Oh, I missed that this was _only_ disabling aarch64 wheels.

takluyver:
Yup, the x64 and ARM Mac wheels are produced on Azure, and available for download from there as 'artifacts'. They're automatically cleaned up a couple of weeks after the build.

PR #2223, opened 2023-01-30
Document that wheels don't include ROS3 support

### LINKED ISSUE #2208: h5py was built without ROS3 support, can't use ros3 driver
Operating System : Windows 10
Python version : 3.10
h5py version : 
D:\Program Files\Python310>pip uninstall h5py
Found existing installation: h5py 3.7.0

发生异常: ValueError
h5py was built without ROS3 support, can't use ros3 driver
  File "D:\hdf5Test.py", line 71, in read_hdf5
    f = h5py.File(h5_path, 'r', driver='ros3', aws_region=b'xxxxx', secret_id=b'xxxx', secret_key=b'xxxx')

takluyver:
Closes #2208.

### DIFF
--- a/docs/high/file.rst
+++ b/docs/high/file.rst
@@ -115,6 +115,13 @@ of supported drivers and their options:
         The argument values must be ``bytes`` objects. All three arguments are
         required to activate AWS authentication.
 
+        .. note::
+           Pre-built h5py packages on PyPI do not include this S3 support. If
+           you want this feature, you could use packages from conda-forge, or
+           :ref:`build h5py from source <source_install>` against an HDF5 build
+           with S3 support. Alternatively, use the :ref:`file-like object
+           <file_fileobj>` support with a package like s3fs.
+
 
 .. _file_fileobj:
 


docs/high/file.rst
```diff
@@ -115,6 +115,13 @@ of supported drivers and their options:
         The argument values must be ``bytes`` objects. All three arguments are
         required to activate AWS authentication.
 
+        .. note::
+           Pre-built h5py packages on PyPI do not include this S3 support. If
```
tacaswell, line 119:
```suggestion

           Pre-built h5py packages on PyPI do not include this S3 support. If
```

Doesn't rst require a blank line here?

tacaswell: nvm, https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html in pretty clear I am wrong.

tacaswell (approved):
I'm not stoked that we are including details about one of the many packaging systems that h5py binaries get built for in the library docstring (we would not add "debian makes such-and-such choice") but agree this is the pragmatic place to put this information to make it discover-able.

PR #2224, opened 2023-01-30
Add collective metadata functions to the low level API

jchelly:
I'd like to add the following functions to the low level API:

  * H5Pset_all_coll_metadata_ops()
  * H5Pget_all_coll_metadata_ops
  * H5Pset_coll_metadata_write()
  * H5Pset_coll_metadata_write()

For context: I'm using h5py on a  HPC cluster to process simulation outputs stored as HDF5. The code is distributed over multiple compute nodes which have 128 cores each.  In order to make use of all of the cpu cores I run python using mpi4py with one process per core. I'm using collective I/O to read the input simulation data and write out the results.

This puts quite a load on the Lustre parallel file system, and I think it's probably because every process accesses the files independently for metadata operations. I'm hoping that can be alleviated by having HDF5 do all file access in collective mode so that only a few processes per node need to access the file system.

For my use case I just need to put the whole file in collective metadata mode. To do that I've added get/set_all_coll_metadata_ops() and get/set_coll_metadata_write() methods to h5p.PropFAID. The HDF5 documentation says that H5Pset_all_coll_metadata_ops() can also be called on group, dataset, datatype, link, or attribute access property lists. Of those I think h5py only exposes link and dataset access property lists so I also added get/set_all_coll_metadata_ops() to h5p.PropLAID and h5p.PropDAID.

### DIFF
--- a/h5py/api_functions.txt
+++ b/h5py/api_functions.txt
@@ -328,7 +328,10 @@ hdf5:
   1.12.1 herr_t H5Pget_file_locking(hid_t fapl_id, hbool_t *use_file_locking, hbool_t *ignore_when_disabled)
   1.10.7-1.10.99 herr_t H5Pset_file_locking(hid_t fapl_id, hbool_t use_file_locking, hbool_t ignore_when_disabled)
   1.12.1 herr_t H5Pset_file_locking(hid_t fapl_id, hbool_t use_file_locking, hbool_t ignore_when_disabled)
-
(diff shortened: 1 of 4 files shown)


roblatham00:
Just driving by to show my support.  These optimizations are critical for any non-trivial level of scaling and I hope these can be added to h5py soon

ptim0626:
Hi, thanks @jchelly for implementing this! May I ask what the status of this PR? I was looking for the exact low-level API functions in h5py and found this PR! I am more than happy to contribute to the code for any remaining work so this will be available in h5py in the future.

PR #2229, opened 2023-02-09
Adding `h5z.register_filter` wrapper of `H5zregister`

### LINKED ISSUE #2228: Adding H5Zregister function wrapper
The [H5Zregister](https://docs.hdfgroup.org/hdf5/develop/group___h5_z.html#title6) function is not available in `h5py`.

It can be useful to have it available for registering filter from Python.

This would be useful in `hdf5plugin` where we now use `ctypes` to wrap this function, see:
https://github.com/silx-kit/hdf5plugin/blob/b53d70b702154a149a60f653c49b20045228aa16/src/hdf5plugin/_utils.py#L63-L76

t20100:
This PR proposes to provide access to `H5Zregister` through `h5z.register_filter` function which takes as single parameter the memory address of the structure defining the filter.

It also adds a basic test of registering/unregistering a dummy filter defined from Python.

closes #2228

### DIFF
--- a/h5py/h5z.pyx
+++ b/h5py/h5z.pyx
@@ -102,8 +102,19 @@ def get_filter_info(int filter_code):
 def register_filter(Py_ssize_t cls_pointer_address):
     '''(INT cls_pointer_address) => BOOL
 
-    Register a new filter from a memory address pointing a buffer containing a
-    `H5Z_class1_t` or `H5Z_class2_t` data structure describing the filter.
+    Register a new filter from the memory address of a buffer containing a
+    ``H5Z_class1_t`` or ``H5Z_class2_t`` data structure describing the filter.
+
+    `cls_pointer_address` can be retrieved from a HDF5 filter plugin dynamic
+    library::
+
+        import ctypes
+
+        filter_clib = ctypes.CDLL("/path/to/my_hdf5_filter_plugin.so")
+        filter_clib.H5PLget_plugin_info.restype = ctypes.c_void_p
+
+        h5py.h5z.register_filter(filter_clib.H5PLget_plugin_info())
+
     '''
     return <int>H5Zregister(<const void *>cls_pointer_address) >= 0
 
--- a/h5py/h5z.pyx
+++ b/h5py/h5z.pyx
@@ -16,6 +16,8 @@ from ._objects import phil, with_phil
 
 # === Public constants and data structures ====================================
 
+CLASS_T_VERS = H5Z_CLASS_T_VERS
+
 FILTER_LZF = H5PY_FILTER_LZF
 
 FILTER_ERROR    = H5Z_FILTER_ERROR
--- a/h5py/h5z.pyx
+++ b/h5py/h5z.pyx
@@ -97,6 +97,16 @@ def get_filter_info(int filter_code):
 
 
 @with_phil
+def register_filter(Py_ssize_t cls_pointer_address):
+    '''(INT cls_pointer_address) => BOOL
+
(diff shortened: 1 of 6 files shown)


h5py/h5z.pyx
```diff
@@ -16,6 +16,8 @@ from ._objects import phil, with_phil
 
 # === Public constants and data structures ====================================
 
+CLASS_T_VERS = H5Z_CLASS_T_VERS
```
t20100, line 19: Used in the test

h5py/h5z.pyx
```diff
@@ -96,6 +98,16 @@ def get_filter_info(int filter_code):
     return flags
 
 
+@with_phil
+def register_filter(Py_ssize_t cls_pointer_address):
+    '''(INT cls_pointer_address) => BOOL
```
t20100, line 103: I couldn't find a better way than passing the memory address as a `int`.

It works both for passing a ctypes struct from Python and with a `ctypes.c_void_p` retrieved from a C function call (use case for `hdf5plugin` package).

t20100:
I just realized there's already a `test_unregister_filter` in `test_filters.py`...
Maybe best to put h5z tests together.

Should I move the new `test_register_filter` (and all the ctypes definitions that goes with it) in `test_filters.py`?

h5py/h5z.pyx
takluyver: Hmm, that's kind of ugly, but I also don't have a better idea. Do you want to add those two ways to get a pointer-as-python-int to the docstring?

h5py/tests/test_h5z.py
```diff
@@ -0,0 +1,73 @@
...
+    h5z.register_filter(addressof(dummy_filter_class))
+
+    assert h5z.filter_avail(filter_id)
+    filter_flags = h5z.get_filter_info(filter_id)
+    assert (
+        filter_flags
+        == h5z.FILTER_CONFIG_ENCODE_ENABLED | h5z.FILTER_CONFIG_DECODE_ENABLED
+    )
+
+    h5z.unregister_filter(filter_id)
```
takluyver, line 71:
```suggestion
    try:
        assert h5z.filter_avail(filter_id)
        filter_flags = h5z.get_filter_info(filter_id)
        assert (
            filter_flags
            == h5z.FILTER_CONFIG_ENCODE_ENABLED | h5z.FILTER_CONFIG_DECODE_ENABLED
        )
    finally:
        h5z.unregister_filter(filter_id)
```

Just to make sure it's unregistered again even if the test fails.

takluyver:
Now that we've got a `test_h5z` module, I think it's OK for `test_unregister_filter` to live there. I generally like having one test module per API module. Presumably it just didn't seem worth creating a whole new module for a single test of a few lines. Practicality beats purity. :slightly_smiling_face:

h5py/h5z.pyx
t20100: `ctypes` actually converts `c_void_p` to `int` when used as function return type:

ctypes type | C type | Python type
-- | -- | --
c_void_p | void* | int or None

From [ctypes's fundamental data types](https://docs.python.org/3/library/ctypes.html#fundamental-data-types).

I've added a code snippet showing how to register a filter from a hdf5 filter plugin dynamic library.

I would not document the 2nd way about how to register a pure Python filter as done in the test, even if it's possible, it's not really something I would promote.

t20100:
I've updated the `register_filter` docstring and moved `test_unregister_filter` to `test_h5z.py`.

takluyver:
Thanks, this looks good. :slightly_smiling_face:

PR #2232, opened 2023-02-10
Dataset direct chunk read to user provided buffer

### LINKED ISSUE #2231: Allowing to pass an output buffer for dataset direct chunk read operations.
The way `DatasetID.read_direct_chunk` is currently implemented requires for each chunk a memory allocation (`emalloc`) and a memory copy to return a `bytes` object (`ret = data[:read_chunk_nbytes]`):
https://github.com/h5py/h5py/blob/3507819de54b35af05b2ca8ca55ec7e7b60cb919/h5py/h5d.pyx#L563-L569

This allocation+copy would not be need if the user could provide its own buffer to read the chunk in as does the HDF5 C API.

t20100:
This PR proposes to add a new method `DatasetID.read_direct_chunk_tobuffer` which allows to read a chunk  directly in a user provided object supporting Python [buffer protocol](https://docs.python.org/3/c-api/buffer.html). As opposed to `DatasetID.read_direct_chunk_tobuffer`, it spares a memory copy and enables the caller to provide its own buffer.

An alternative could be to extend `DatasetID.read_direct_chunk` with an extra `out=None` argument which can take a buffer as input.
We tried a few things:

- Returning the number of bytes read if `out` is not None: [here](https://github.com/jonwright/h5py/blob/3d067e8d911517db1d3cc5e141a457a119a729c1/h5py/h5d.pyx#L520-L594).
  The down side is that the returned type changes depending on the inputs.
- Returning a ndarray "view" on the out buffer: [here](https://github.com/jonwright/h5py/blob/db004d834f415cc4f7b5622c6904cbda7701ab46/h5py/h5d.pyx#L527-L602).
  The down side is that it is slower because it allocates more Python objects and anyway does not returns the same types.

Here is a quick&dirty performance comparison reading chunks from the [kevlar.h5](http://www.silx.org/pub/pyFAI/pyFAI_UM_2020/data_ID13/kevlar.h5) file with the script below:

```
read_direct_chunk_tobuffer: 0.17198193189688027s
read_direct_chunk: 0.25807333388365805s
```

<details>
<summary>Test script</summary>

```python
import h5py
import timeit

with h5py.File("kevlar.h5", "r") as h5f:
    dataset = h5f["entry/data/data"]
    out = bytearray(dataset.nbytes // len(dataset))
    t0 = timeit.default_timer()
    for index in range(len(dataset)):
        dataset.id.read_direct_chunk_tobuffer((index, 0, 0), out)
    dt = timeit.default_timer() - t0
    print(f"read_direct_chunk_tobuffer: {dt}s")

with h5py.File("kevlar.h5", "r") as h5f:
    dataset = h5f["entry/data/data"]
    t0 = timeit.default_timer()
    for index in range(len(dataset)):
        dataset.id.read_direct_chunk((index, 0, 0))
    dt = timeit.default_timer() - t0
    print(f"read_direct_chunk: {dt}s")
```

</details>

This was initiated by @jonwright and we worked together on this proposition.

closes #2231

### DIFF
--- a/h5py/h5d.pyx
+++ b/h5py/h5d.pyx
@@ -28,7 +28,9 @@ from ._proxy cimport dset_rw
 from ._objects import phil, with_phil
 from cpython cimport PyObject_GetBuffer, \
                      PyBUF_ANY_CONTIGUOUS, \
+                     PyBUF_WRITABLE, \
                      PyBuffer_Release
+from cpython.mem cimport PyMem_Malloc, PyMem_Free
 
 
 # Initialization
@@ -515,6 +517,78 @@ cdef class DatasetID(ObjectID):
                     H5Sclose(space_id)
 
     IF HDF5_VERSION >= (1, 10, 2):
+
+        def read_direct_chunk_tobuffer(self, offsets, out, PropID dxpl=None):
+            """ (offsets, out, PropID dxpl=None)
+
+            Reads data directly to an output buffer from a chunk at position
+            specified by the `offsets` argument and bypasses any filters HDF5
+            would normally apply to the written data. However, the written data
+            may be compressed or not.
+
+            The `out` argument must support the Py_buffer interface (e.g.,
+            `bytearray` or `numpy.ndarray`) and be large enough to contain the
+            whole chunk.
+
+            Returns a tuple containing the `filter_mask` and the number of bytes
+            read into the `out` buffer.
+
+            `filter_mask` is a bit field of up to 32 values. It records which
+            filters have been applied to this chunk, of the filter pipeline
+            defined for that dataset. Each bit set to `1` means that the filter
+            in the corresponding position in the pipeline was not applied to
+            compute the raw data. So the default value of `0` means that all
+            defined filters have been applied to the raw data.
+
+            Feature requires: 1.10.2 HDF5
+            """
+
+            cdef hid_t dset_id
+            cdef hid_t dxpl_id
+            cdef hid_t space_id
+            cdef hsize_t *offset = NULL
+            cdef int rank
+            cdef uint32_t filters
+            cdef hsize_t chunk_bytes
+            cdef Py_buffer view
+            cdef int nb_offsets = len(offsets)
+
+            dset_id = self.id
+            dxpl_id = pdefault(dxpl)
+            space_id = H5Dget_space(dset_id)
+            if space_id == -1:  # H5I_INVALID_HID
+                raise RuntimeError("Cannot retrieve dataset space")
+            rank = H5Sget_simple_extent_ndims(space_id)
+            H5Sclose(space_id)
+
+            if nb_offsets != rank:
+                raise ValueError(
+                    f"offsets length ({nb_offsets}) must match dataset rank ({rank})"
+                )
+
+            PyObject_GetBuffer(out, &view, PyBUF_ANY_CONTIGUOUS | PyBUF_WRITABLE)
+            try:
+                offset = <hsize_t*> PyMem_Malloc(sizeof(hsize_t)*rank)
+                if not offset:
+                    raise MemoryError()
+                convert_tuple(offsets, offset, rank)
+                H5Dget_chunk_storage_size(dset_id, offset, &chunk_bytes)
+
+                if <hsize_t>view.len < chunk_bytes:
+                    raise ValueError(
+                        f"out buffer is not large enough: {view.len} bytes, requires {chunk_bytes} bytes"
+                    )
+
+                IF HDF5_VERSION >= (1, 10, 3):
+                    H5Dread_chunk(dset_id, dxpl_id, offset, &filters, <char *>view.buf)
+                ELSE:
+                    H5DOread_chunk(dset_id, dxpl_id, offset, &filters, <char *>view.buf)
+            finally:
+                if offset:
+                    PyMem_Free(offset)
+                PyBuffer_Release(&view)
+
+            return filters, chunk_bytes
 
         def read_direct_chunk(self, offsets, PropID dxpl=None):
             """ (offsets, PropID dxpl=None)
--- a/h5py/h5d.pyx
+++ b/h5py/h5d.pyx
@@ -16,6 +16,7 @@ include "config.pxi"
 # Compile-time imports
 
 from collections import namedtuple
+cimport cython
 from ._objects cimport pdefault
 from numpy cimport ndarray, import_array, PyArray_DATA
 from .utils cimport  check_numpy_read, check_numpy_write, \
@@ -26,10 +27,12 @@ from .h5p cimport PropID, propwrap
 from ._proxy cimport dset_rw
 
 from ._objects import phil, with_phil
-from cpython cimport PyObject_GetBuffer, \
-                     PyBUF_ANY_CONTIGUOUS, \
-                     PyBUF_WRITABLE, \
-                     PyBuffer_Release
+from cpython cimport PyBUF_ANY_CONTIGUOUS, \
+                     PyBuffer_Release, \
+                     PyBytes_AsString, \
+                     PyBytes_FromStringAndSize, \
+                     PyObject_GetBuffer
+
 from cpython.mem cimport PyMem_Malloc, PyMem_Free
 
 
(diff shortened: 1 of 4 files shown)


takluyver:
Thanks for this, I think it will be a useful addition.

My initial reaction is that I'd prefer it to be an option for the existing method `read_direct_chunk` method, as you discuss already. I think Numpy arrays are relatively expensive to create, but I'd hope there's some built-in Python type we could create relatively cheaply to represent the data read without copying it. Perhaps a `bytearray` or a `memoryview`? This would hopefully be similar to the `bytes` returned by the existing code.

I see in one of your alternative commits a surprised comment "this copies !?". I think it does, and I think you're right that that's not ideal. While we're here, I wonder if we can fix that. It looks like the C-API functions `PyBytes_FromStringAndSize` and `PyBytes_AsString` can be used to create a bytes object with the specified size and then write into it from C. Or perhaps we can add an option which returns a slightly different type but avoids the copy.

t20100:
Yes, using `PyBytes_FromStringAndSize` and `PyBytes_AsString` works fine and is faster than current state!
I really didn't expected there was a way to set the `bytes` content after creation.
I'll rework this PR and look for a return type in case the `out` buffer is provided.

t20100:
As discussed, I updated this PR to:
- Provide a single `read_direct_chunk` method with an optional `out` argument.
- Use `PyBytes_FromStringAndSize` and `PyBytes_AsString` when `out` is `None` to spare a memory allocation and copy.

With this implementation, when `out` is provided, it must be an object supporting the buffer protocol and providing a writable contiguous 1D buffer of bytes.
In this case it returns a Python `memoryview` which is this simplest implementation.

I tested a couple of approaches (see [in this branch](https://github.com/t20100/h5py/blob/read_direct_chunk_test/h5py/h5d.pyx)):
- Returning a wrapper object rather than a memory view ([see here](https://github.com/t20100/h5py/blob/19aa77750ada858e99faafa943657f41955748b4/h5py/h5d.pyx#L161)) to allow providing an API closer to `bytes`. This seems slightly faster than the `memoryview` implementation.
- Having `read_direct_chunk` return an object rather than a 2-tuple (yet keeping compatible) to allow extending returned information ([see here](https://github.com/t20100/h5py/blob/19aa77750ada858e99faafa943657f41955748b4/h5py/h5d.pyx#L203)).

h5py/h5d.pyx
```diff
@@ -534,47 +541,64 @@ cdef class DatasetID(ObjectID):
...
-            if len(offsets) != rank:
-                raise TypeError("offset length (%d) must match dataset rank (%d)" % (len(offsets), rank))
+            if nb_offsets != rank:
+                raise ValueError(
+                    f"offsets length ({nb_offsets}) must match dataset rank ({rank})"
+                )
+
+            offset = <hsize_t*> PyMem_Malloc(sizeof(hsize_t)*rank)
+            if not offset:
+                raise MemoryError()
```
takluyver, line 575: We mostly use our internal `emalloc` function, which deals with raising MemoryError. But I don't really understand when we need to use the Python malloc wrapper (as you've done) or the HDF5 one (`H5allocate_memory`). :confused:

takluyver: A bit of reading suggests that what's important is that it's the same memory management system allocating and freeing the memory. If that's how you understand it too, then I'd suggest we stick to our internal `emalloc` utility function for consistency, since we're in full control of where this is allocated and freed.

h5py/tests/test_h5d_direct_chunk.py
```diff
@@ -113,3 +114,81 @@ def test_read_write_chunk(self):
...
+    def test_fail_buffer_too_small(self, writable_file):
+        ref_data = numpy.arange(16).reshape(4, 4)
+        dataset = writable_file.create_dataset(
+            "uncompressed", data=ref_data, chunks=ref_data.shape)
+
+        out = bytearray(ref_data.nbytes // 2)
+        with pytest.raises(ValueError):
+            dataset.id.read_direct_chunk((0, 0), out=out)
+
+    def test_fail_buffer_readonly(self, writable_file):
```
takluyver, line 177: I'm curious, which bit actually errors on a read-only `out` object?

takluyver:
Thanks, this is looking pretty good to me.

> Returning a wrapper object rather than a memory view ([see here](https://github.com/t20100/h5py/blob/19aa77750ada858e99faafa943657f41955748b4/h5py/h5d.pyx#L161)) to allow providing an API closer to bytes. This seems slightly faster than the memoryview implementation.

My leaning is to keep it simple, i.e. use the native Python memoryview rather than a custom object.

> Having read_direct_chunk return an object rather than a 2-tuple (yet keeping compatible) to allow extending returned information ([see here](https://github.com/t20100/h5py/blob/19aa77750ada858e99faafa943657f41955748b4/h5py/h5d.pyx#L203)).

Again, I like the simple option. I don't think we need any extra returned objects at the moment, and I don't think it's that likely that we will in the future.

h5py/tests/test_h5d_direct_chunk.py
takluyver: nvm, I see it in the Cython docs - the lack of a `const` declaration for the parameter rejects read-only buffers.

h5py/h5d.pyx
```diff
@@ -534,47 +541,64 @@ cdef class DatasetID(ObjectID):
...
+            space_id = H5Dget_space(dset_id)
+            if space_id == -1:  # H5I_INVALID_HID
+                raise RuntimeError("Cannot retrieve dataset space")
             rank = H5Sget_simple_extent_ndims(space_id)
+            H5Sclose(space_id)
 
-            if len(offsets) != rank:
-                raise TypeError("offset length (%d) must match dataset rank (%d)" % (len(offsets), rank))
+            if nb_offsets != rank:
+                raise ValueError(
```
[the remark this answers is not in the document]
t20100: You're right, it's best to keep it as it was and keep consistency with other `DatasetID` methods. I'll remove this change.

That's a matter of interpreting "wrong type" and "wrong value" in the case of tuple with different sizes in [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError) doc:

> Passing arguments of the wrong type (e.g. passing a [list](https://docs.python.org/3/library/stdtypes.html#list) when an [int](https://docs.python.org/3/library/functions.html#int) is expected) should result in a [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError), but passing arguments with the wrong value (e.g. a number outside expected boundaries) should result in a [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError).

h5py/h5d.pyx
t20100: I updated it to use `emalloc` as the other methods.

It was to use Python memory heap rather than system `malloc` as suggested in cython's [memory_allocation tutorial](https://cython.readthedocs.io/en/stable/src/tutorial/memory_allocation.html):

> Note that the C-API functions for allocating memory on the Python heap are generally preferred over the low-level C functions above as the memory they provide is actually accounted for in Python’s internal memory management system. They also have special optimisations for smaller memory blocks, which speeds up their allocation by avoiding costly operating system calls.

h5py/h5d.pyx
```diff
@@ -534,47 +539,59 @@ cdef class DatasetID(ObjectID):
...
+                    retval = PyBytes_FromStringAndSize(NULL, chunk_bytes)
+                    chunk_buffer = PyBytes_AsString(retval)
+                else:
+                    out_bytes = out.shape[0]  # Fast way to get out length
+                    if out_bytes < chunk_bytes:
+                        raise ValueError(
+                            f"out buffer is only {out_bytes} bytes, {chunk_bytes} bytes required"
+                        )
+                    retval = memoryview(out[:chunk_bytes])
+                    chunk_buffer = &out[0]
```
takluyver, line 584: Just to check, what happens if you try to read an unallocated chunk (which has storage size 0) with a zero-sized buffer? I saw that you turned off bounds checking, so I just want to be sure it won't segfault.

h5py/h5d.pyx
takluyver: Thanks. Maybe we do want to use that, but if so, let's do it in a separate PR. :+1:

h5py/h5d.pyx
t20100: It fails in HDF5's read chunk function, the same way as without the `out` argument:

```python
f = h5py.File("test.h5", "w")
ds = f.create_dataset("data", shape=(100,), dtype=numpy.uint16, chunks=(10,), compression='gzip')
f.flush()

out = bytearray(ds.dtype.itemsize * numpy.prod(ds.chunks))
ds.id.read_direct_chunk((0,), out=out)
```
```
OSError                                   Traceback (most recent call last)
Input In [42], in <cell line: 1>()
----> 1 ds.id.read_direct_chunk((0,), out=out)

File h5py/h5d.pyx:587, in h5py.h5d.DatasetID.read_direct_chunk()
    585 
    586                 IF HDF5_VERSION >= (1, 10, 3):
--> 587                     H5Dread_chunk(dset_id, dxpl_id, offset, &filters, chunk_buffer)
    588                 ELSE:
    589                     H5DOread_chunk(dset_id, dxpl_id, offset, &filters, chunk_buffer)

OSError: Can't read unprocessed chunk data (storage is not initialized)
```

Worth a test?

takluyver: No, I think that's OK without a test. I don't want to check that it raises OSError because I don't think OSError is quite right, but I don't know what is right, so that's good enough for now. I think there are various places where we misuse OSError (and possibly other exception types).

takluyver:
Thanks!

PR #2234, opened 2023-02-17
Remove support for using dset.astype() as a context manager

takluyver:
Finally removing a wart that I've been planning to get rid of for years, the use of a thread-local variable in the dataset class.

We added and recommended the new way of using this (`dset.astype('f4')[:]`) in 3.0, released October 2020. We added a deprecation warning for the context manager (`with dset.astype('f4'):`) in 3.6, released November 2021 (PR #1842).

### DIFF
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -503,12 +503,8 @@ Reference
             >>> out.dtype
             dtype('int16')
 
-        .. versionchanged:: 3.0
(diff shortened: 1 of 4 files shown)


takluyver:
@tacaswell @aragilar - are you happy with removing this for 3.9, or do you think we should give people more time to see the deprecation warning?

aragilar:
I think go ahead and remove it.

tacaswell:
Belatedly :+1:

PR #2235, opened 2023-02-17
Try to fix codecov for pull requests

takluyver:
I realised while poking around codecov that it was seeing a mixture of reports uploaded for the PR, some of which were associated with the *merge* commit and some with the *head* commit. In particular, the Azure builds send reports for the merge commit, and the Appveyor builds (which test on 32-bit Windows) send them for the merge commit, even though they're testing the merge commit.

I think I've found a workaround to send Appveyor coverage reports for the merge commit. Hopefully this will make the coverage change report on PRs more accurate.

While I was looking at this, I also discovered that the codecov config file was invalid, and I updated the Appveyor config to use Python 3.8 as the 'outer' Python that installs and runs tox.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -56,7 +56,7 @@ test_script:
   # Note that you must use the environment variable %PYTHON% to refer to
   # the interpreter you're using - Appveyor does not do anything special
   # to put the Python evrsion you want to use on PATH.
-  - "py -3.8 -m tox -v"
+  - "py -3.8 -m tox run --discover %PYTHON%\\python.exe"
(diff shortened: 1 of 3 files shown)


takluyver:
> No change to project coverage :+1: 

This is what I was going for. :slightly_smiling_face:

takluyver:
That seems to have worked too. All the uploaded coverage reports are now associated with the head commit of the PR.

takluyver:
Merging this myself because it only touches CI, not any part of h5py itself.

PR #2242, opened 2023-03-03
BUG: handle deprecation warning from numpy 1.25 (np.product is deprecated in favour of np.prod)

neutrinoceros:
In repsonse to https://github.com/numpy/numpy/pull/23314

### DIFF
--- a/h5py/_hl/filters.py
+++ b/h5py/_hl/filters.py
@@ -41,6 +41,7 @@ from collections.abc import Mapping
 import operator
 
 import numpy as np
+from .base import product
 from .compat import filename_encode
 from .. import h5z, h5p, h5d, h5f
 
@@ -358,7 +359,7 @@ def guess_chunk(shape, maxshape, typesize):
 
     # Determine the optimal chunk size in bytes using a PyTables expression.
     # This is kept as a float.
-    dset_size = np.prod(chunks)*typesize
+    dset_size = product(chunks)*typesize
     target_size = CHUNK_BASE * (2**np.log10(dset_size/(1024.*1024)))
 
     if target_size > CHUNK_MAX:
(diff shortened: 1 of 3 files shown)


takluyver:
Thanks! While we're looking at this, though, I think all or most of these occurrences can be replaced by our internal `product` function. We found at some point that for a handful of integers, as in a shape tuple, it's quicker to multiply them in Python than to get Numpy involved.

https://github.com/h5py/h5py/blob/7705f672cc6e091b069900ea260c3a27f329445a/h5py/_hl/base.py#L504-L513

neutrinoceros:
@takluyver cool ! I've replaced every occurrences where `np.product` was used on an array shape.

h5py/_hl/filters.py
```diff
@@ -358,7 +358,7 @@ def guess_chunk(shape, maxshape, typesize):
 
     # Determine the optimal chunk size in bytes using a PyTables expression.
     # This is kept as a float.
-    dset_size = np.product(chunks)*typesize
+    dset_size = np.prod(chunks)*typesize
```
tacaswell, line 361: I would move these over too?  I think chunks are still shape-like.

Not going to block merging over.

tacaswell (approved):

h5py/_hl/filters.py
neutrinoceros: Your call, I just didn't want to optimize code that I didn't immediately understand :)

neutrinoceros: done ! FWIW, I measured that `h5py._hl.base.product` was indeed faster than `np.prod` on small arrays such as `chunks`

takluyver:
Thanks @neutrinoceros !

pllim:
Hello. When will this patch be in a release? I see this warning now with numpy 1.25rc1. Thank you!

PR #2244, opened 2023-03-19
Expose low level getter/setter for elink access flags and fixes missing lapl arguments

cderemble:
With this PR, one could set the default open mode for external links to readonly so target linked files would not be modified, regardless of the open mode of the main file.

### DIFF
--- a/h5py/__init__.py
+++ b/h5py/__init__.py
@@ -56,7 +56,7 @@ _register_lzf()
 from . import h5a, h5d, h5ds, h5f, h5fd, h5g, h5r, h5s, h5t, h5p, h5z, h5pl
 
 from ._hl import filters
-from ._hl.base import is_hdf5, HLObject, Empty, dlapl, dlcpl
+from ._hl.base import is_hdf5, HLObject, Empty
 from ._hl.files import (
     File,
     register_driver,
--- a/h5py/__init__.py
+++ b/h5py/__init__.py
@@ -56,7 +56,7 @@ _register_lzf()
 from . import h5a, h5d, h5ds, h5f, h5fd, h5g, h5r, h5s, h5t, h5p, h5z, h5pl
 
 from ._hl import filters
-from ._hl.base import is_hdf5, HLObject, Empty
+from ._hl.base import is_hdf5, HLObject, Empty, dlapl, dlcpl
 from ._hl.files import (
     File,
     register_driver,
--- a/h5py/_hl/group.py
+++ b/h5py/_hl/group.py
@@ -508,7 +508,11 @@ class Group(HLObject, MutableMappingHDF5):
     @with_phil
     def __contains__(self, name):
         """ Test if a member name exists """
-        return self.id.__contains__(self._e(name), lapl=self._lapl)
+        if hasattr(h5g, "_path_valid"):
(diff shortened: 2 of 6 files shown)


cderemble:
related issue:
https://forum.hdfgroup.org/t/can-we-use-external-links-to-read-only-data-in-writable-files/9390

h5py/__init__.py
```diff
@@ -56,7 +56,7 @@
 from . import h5a, h5d, h5ds, h5f, h5fd, h5g, h5r, h5s, h5t, h5p, h5z, h5pl
 
 from ._hl import filters
-from ._hl.base import is_hdf5, HLObject, Empty
+from ._hl.base import is_hdf5, HLObject, Empty, dlapl, dlcpl
```
takluyver, line 59:
```suggestion
from ._hl.base import is_hdf5, HLObject, Empty
```

Modifying global default property lists is definitely not how we should encourage people to use the API.

We should instead come up with some way to get custom options to the operations that need them, whether by passing in a property list or passing in the relevant options and constructing a property list internally. This may be somewhat more complex, so we might want to expose the low-level functionality in this PR and then figure out the high-level API as a later step.

news/h5p_elink_acc_flags.rst
```diff
@@ -0,0 +1,30 @@
+New features
+------------
+
+* Low-level ``h5p.set_elink_acc_flags`` function that allows to set the external link traversal file access flag in a link access property list.
+* Low-level ``h5p.get_elink_acc_flags`` function that retrieves the external link traversal file access flag from the specified link access property list.
```
takluyver, line 5: This is the sort of addition that would generally go under 'Exposing HDF5 functions' - we're exposing additional functions from the HDF5 C API to be called from Python. :slightly_smiling_face:

h5py/__init__.py
cderemble: reverted

h5py/_hl/group.py
```diff
@@ -508,6 +508,8 @@ def __reversed__(self):
     @with_phil
     def __contains__(self, name):
         """ Test if a member name exists """
+        if hasattr(h5g, "_path_valid"):
+            return h5g._path_valid(self.id, self._e(name), self._lapl)
```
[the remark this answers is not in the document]
cderemble: this is tricky: I am not sure how to set ``lapl`` from grp.id...
but I made the function more similar to the h5g one

news/h5p_elink_acc_flags.rst
cderemble: changed news accordingly

h5py/_hl/group.py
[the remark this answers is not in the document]
cderemble: This is possible in python, but it seems it does not work in cython:
https://cython.readthedocs.io/en/latest/src/userguide/special_methods.html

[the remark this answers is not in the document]
takluyver: OK, fair enough.

takluyver:
Thanks, this looks good. So we'll need a second part where we make it possible to control this setting somehow when following links. Do you want to open an issue to discuss that?

PR #2247, opened 2023-04-03
BLD: fix for cython3 compatibility

tacaswell:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [build-system]
 requires = [
-    "Cython ~=0.29",
+    "Cython >=0.29.31,<1",
     "oldest-supported-numpy",
     "pkgconfig",
     "setuptools >=61",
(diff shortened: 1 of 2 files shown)


tacaswell:
xref https://github.com/scipy/scipy/issues/17234#issuecomment-1493423202

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [build-system]
 requires = [
-    "Cython ~=0.29",
+    "Cython >=0.29.31",
```
takluyver, line 3: I'm a bit wary of declaring Cython 3 compatibility before the final version is out (I think the `~=0.29` specifier is equivalent to `>=0.29, <1`). Though I see they're onto betas now, so hopefully it's not far off.

tacaswell: maybe `>=0.29.31, <1` ?  We need to put a floor on the patch version so that the keyword exists.

takluyver: Yup, if we do that, I think it's fine.

tacaswell: done.

Issue #2152, opened 2022-09-15
Add __array__() method to AstypeWrapper and AsStrWrapper


takluyver:
To support things like `np.asarray(dset.astype(np.int32))` and `np.asarray(dset.asstr())`.

See #2151 (doing the same for `dset.fields()`) for an example of what this will involve.

ethanotran:
Can I work on this? Looking to make my first contribution to open source.

takluyver:
Yes, please do! Have a look at PR #2151 for an example, and the [contributing docs](https://docs.h5py.org/en/stable/contributing.html#how-to-get-your-code-into-h5py), and feel free to ask if you have any questions.

aragilar:
Closed via #2269.

PR #2269, opened 2023-05-27
Add __array__ method to Astype and AsStr wrappers

cydanil:
This PR adds `__array__` to `AstypeWrapper` and `AsStrWrapper`, as suggested in #2152 .  

Cheers,  
Cyril


aragilar:
Looks good to me, thanks @cydanil.

PR #2272, opened 2023-06-02
Remove `import base` inside `make_new_dset`

levonwaveye:
Import within a function that is called for every `create_dataset` call has performance impact. In some scenarios we create many datasets with a 'core' driver and this import has overhead in configurations with slower imports. `.base` is already imported in the header so there seems to be no value in having it be imported locally, unless I am missing something.

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -17,7 +17,7 @@ import sys
 import numpy
 
 from .. import h5, h5s, h5t, h5r, h5d, h5p, h5fd, h5ds, _selector
-from .base import HLObject, phil, with_phil, Empty, cached_property, find_item_type
+from .base import HLObject, phil, with_phil, Empty, cached_property, find_item_type, array_for_new_object
 from . import filters
 from . import selections as sel
 from . import selections2 as sel2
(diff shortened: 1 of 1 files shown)


tacaswell:
This seems reasonable from a code style point of view (and as `.base` is already imported at the top I do not see what we would be gaining from re-importing locally.

I am curious what you can do to the import system to make re-importing an already imported module (which should more-or-less be a dictionary lookup) slow!

levonwaveye:
Seems imports within a function stay local to that function and are not cached between function calls. See some comparisons e.g. in https://wiki.python.org/moin/PythonSpeed/PerformanceTips#Import_Statement_Overhead. For us this was a non-issues on some platforms (Mac, Linux), but made `make_new_dset` a very heavy call on Windows. `AttributeManager.create` makes the same `array_for_new_object` call without an import, and it performed well.

tacaswell:
That suggests the import machinery is about the same speed as doing the `string.lower` function call, not that the import machinery is particularly expensive.

Do you have any benchmarking on this?

```
In [1]: %time import numpy as np
CPU times: user 15.6 ms, sys: 6.51 ms, total: 22.2 ms
Wall time: 21.8 ms

In [2]: %time import numpy as np
CPU times: user 2 µs, sys: 0 ns, total: 2 µs
Wall time: 2.86 µs

In [3]: def foo():
   ...:     import numpy
   ...:

In [4]: %time foo()
CPU times: user 3 µs, sys: 0 ns, total: 3 µs
Wall time: 3.81 µs

In [5]:
```

Note the first time is slow (the actual import) and the next two are 4 orders of magnitude faster (even inside a funcion).

To be clear, I am in favor of taking this change, I just do not believe it will have any meaningful performance impact.

levonwaveye:
That's fair. We only have 3% of time spent in builtin.import. I could not re-compile on Windows to compare, so likely misattributed where the performance slowdown is coming from there. Seems `dset_id.write` is much heavier than the import, even for small arrays.

tacaswell (approved):

PR #2273, opened 2023-06-07
Replace some more uses of np.product() with Python function

takluyver:
We've found in the past that calling `np.product()` for small amounts of data like shape tuples is actually much slower than doing the same calculation in Python - the overhead of numpy's flexible machinery outweighs the core calculation. This converts some more cases which we had missed previously; based on searching the code, I think these are the last ones.

In particular, my colleague @philsmt was doing something which involved creating thousands of small attributes, and found that calls to `np.product()` took something like 1/4 of the time in attribute creation.

### DIFF
--- a/h5py/_hl/attrs.py
+++ b/h5py/_hl/attrs.py
@@ -166,7 +166,7 @@ class AttributeManager(base.MutableMappingHDF5, base.CommonStateObject):
             # is compatible, and reshape if needed.
             else:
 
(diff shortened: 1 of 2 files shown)


takluyver:
This is a fairly simple change, and it's passing all tests, so I'll merge it in a couple of days if there are no comments in the meantime.

tacaswell (approved):

tacaswell:
nice that it drops a bunch of `dtype` specification as well.

PR #2274, opened 2023-06-07
Replace attributes by simple delete & recreate

takluyver:
Following the discussion on #2206, this makes `.attrs.create()` and `.attrs[name] =` replace existing attributes by simply deleting them before creating a new one, rather than creating a new attribute with a temporary name and then renaming it. This works around some long-standing issues triggered by renaming attributes, as discussed in #1385.

The 'temporary name' scheme was introduced in #501, but as far as I can tell, it's orthogonal to the main goal of that PR, which was about array dtypes. It does mean that if there's an error creating or writing the new attribute, the previous value is preserved, which is worth something. But we prepare the HDF5 datatype and dataspace before removing an existing attribute, which will hopefully catch most error cases.

### DIFF
--- a/h5py/_hl/attrs.py
+++ b/h5py/_hl/attrs.py
@@ -122,6 +122,7 @@ class AttributeManager(base.MutableMappingHDF5, base.CommonStateObject):
             Data type of the attribute.  Overrides data.dtype if both
             are given.
         """
+        name = self._e(name)
 
         with phil:
             # First, make sure we have a NumPy array.  We leave the data type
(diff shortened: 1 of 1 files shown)


h5py/_hl/attrs.py
```diff
@@ -189,35 +190,24 @@ def create(self, name, data, shape=None, dtype=None):
...
-                    h5a.delete(self._id, self._e(name))
-                    h5a.rename(self._id, self._e(tempname), self._e(name))
             except:
                 attr.close()
-                h5a.delete(self._id, self._e(tempname))
+                h5a.delete(self._id, name)
                 raise
-            finally:
-                attr.close()
+            attr.close()
```
tacaswell, line 210: 🚲 🏠   : put the `finally` back and remove the `attr.close` in it?

tacaswell (approved):
Modulo a very minor comment

tacaswell:
I am not sure what race condition we are trying to protect from here.  We are already in `with phil` so anything in this Python process will not conflict. 

I am not sure off the top of my head what SWMR and attributes do though.

h5py/_hl/attrs.py
tacaswell: oh, fair.  I had not considered the need to close before deleting.

PR #2280, opened 2023-06-20
Error on repeated slicing/indexing of a VirtualSource object

### LINKED ISSUE #2276: Repeated/Multiple slicing of VirtualSource gives unexpected results.
I'm using the virtual dataset features of h5py and I'm encountering an issue. When slicing multiple times a VirtualSource I get wrong results. The sliced virtual source do not correspond to the sliced data. Below a reproducible example.

```python
import h5py
import numpy as np

name = "data"
shape = (100,)
dtype = "i"
data = np.arange(shape[0])

# create a file that will be used as a virtual source
with h5py.File("source.h5", "w") as file:
    dset = file.create_dataset(name, shape, dtype, data)
vsource = h5py.VirtualSource("source.h5", name, shape)

# slice the source and the data twice in the same manner
data = data[80:90][::2]
vsource = vsource[80:90][::2]

# to get the content of the virtual source we write it to disk an read it
layout = h5py.VirtualLayout(vsource.shape, vsource.dtype)
layout[...] = vsource
with h5py.File("vds.h5", "w") as file:
    file.create_virtual_dataset(name, layout)
with h5py.File("vds.h5", "r") as file:
    values = file[name][...]

assert np.all(values == data)  # Raise error
```

Looking at the values it looks like the second slice resets the offset to zero:

```python
>>> values
array([0., 2., 4., 6., 8.])
```

Which is confirmed there :

```python
>>> print(h5py.VirtualSource("source.h5", name, shape)[80:90].sel._sel)
>>> print(h5py.VirtualSource("source.h5", name, shape)[80:90][:].sel._sel)
((80,), (10,), (1,), (False,))
((0,), (10,), (1,), (False,))
```

Indeed in the source code we can see in the vds.py file that the __getitem__ method of the VirtualSource object does not use previous slicing information to update the sel attribute. It only uses the shape of the actual sliced object.

```python
def __getitem__(self, key):
    tmp = copy(self)
    tmp.sel = select(self.shape, key, dataset=None)
    _convert_space_for_key(tmp.sel.id, key)
    return tmp
```

It would be nice if we could change this or at least throw a warning to prevent users from getting unexpected slicing results.

I'm using h5py version 3.8.0.

takluyver:
E.g. `vsource[10:20][::2]` currently appears to work, but does the wrong thing. It's somewhat complex to make it do the right thing, and the need doesn't come up very often, so this just makes the failure explicit.

Closes #2276.

### DIFF
--- a/h5py/_hl/vds.py
+++ b/h5py/_hl/vds.py
@@ -132,15 +132,19 @@ class VirtualSource:
             self.maxshape = tuple([h5s.UNLIMITED if ix is None else ix
                                    for ix in maxshape])
         self.sel = SimpleSelection(shape)
+        self._all_selected = True
 
     @property
     def shape(self):
         return self.sel.array_shape
 
     def __getitem__(self, key):
+        if not self._all_selected:
+            raise RuntimeError("VirtualSource objects can only be sliced once.")
         tmp = copy(self)
         tmp.sel = select(self.shape, key, dataset=None)
         _convert_space_for_key(tmp.sel.id, key)
+        tmp._all_selected = False
         return tmp
 
 class VirtualLayout:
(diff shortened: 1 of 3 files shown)


tacaswell:
The failure is from upload to codedov failing.

tacaswell (approved):

PR #2301, opened 2023-08-27
Add support for AWS session token

ajelenak:
This token can now be used by the ros3 driver as part of AWS temporary security credentials.


ajelenak:
Just pushed a commit that resolves an issue related to the 1.10 API upgrade. Since this PR is already a month old, I will merge it tomorrow. It should be really good to have it in the next release.

PR #2306, opened 2023-09-06
h5fd: add `noexcept` to function casts

mathstuf:
Required with either Cython 3.0.0 it seems.

---
I don't have `tox` handy, but can fix if needed; these all seem to be under the line length limit already used within the file at least.


tacaswell:
I do not think that this is sufficient.  per https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html you should only use this if you know the functions will never raise and these function call back into Python and _may_ raise exceptions.

See the discussion in https://github.com/h5py/h5py/issues/2268.

tacaswell (changes requested):
I am not sure that this is actually doing what we want (as at least some of those methods _can_ raise exceptions)

mathstuf:
Closing in preference for a better fix. See #2268 for progress.

PR #2308, opened 2023-09-12
Build wheels & test with HDF5 1.14.2

takluyver:
Which was just released.

The version for the Linux wheels will need to be updated separately in the repo that creates the docker image, but I'll let the tests run before changing that.


takluyver:
@matham I've held off updating the Mac OS wheel builds to HDF5 1.14 because the patches for cross compiling break. If you've got time at some point, could you see if the patches are still needed, and update them if so? Thanks. :slightly_smiling_face: .

This PR doesn't need to wait for that change, I think.

Issue #2312, opened 2023-09-27
Raise minimum supported HDF5 version?


takluyver:
We haven't bumped the minimum supported HDF5 version in several years, and we still have a couple of tests that run with HDF5 1.8 (on Windows, due to some quirks of history). Meanwhile, HDF group are moving to a model where they only release updates for the 1.14 series from later this year. This diagram is [from their Github repo](https://github.com/HDFGroup/hdf5/blob/develop/doc/img/release-schedule.png):

![release-schedule](https://github.com/h5py/h5py/assets/327925/c38cd7d9-6e20-4ac3-92a4-c88984763136)

So I think it's reasonable for h5py to drop support for old versions of HDF5. This can simplify the code, because various things that are currently conditional on the HDF5 version, such as virtual dataset support, can be assumed. It can also reduce the CI requirements, at least in theory.

For the first step, I propose that we support **HDF5 1.10.4** and above. This is still quite conservative - 1.10.4 is from October 2018, and it's the version in the oldest supported Ubuntu & Debian releases today (20.04 & 10), although RHEL 7, which is still supported, has 1.8.12 (I used [this site](https://pkgs.org/search/?q=hdf5&on=summary) to check versions). I think we could probably require a much more recent version than that, and I'd like to raise the minimum further, but I suggest we don't do it all at once.

takluyver:
Thanks! We're already using a relatively recent version of HDF5 when we bundle it (I've opened #2308 to update that as well), this is about how old an HDF5 version you can use if you build h5py yourself.

aragilar:
I think moving to 1.10.4 is reasonable, given the 1.8 series has ended, and I don't think we get to assume new features even if we were to require (say) 1.8.23. I think if anyone wants any h5py bug fixes to run on 1.8, they can perform said backporting themselves.

tacaswell:
I think going forward "h5py supports the versions of libhdf5 that hdf5group supports" is the correct policy.

We are not going to pull down any old artifacts and if people really (really) need to support and old hdf5 they can (worst case) go back to a tag.

takluyver:
Great - I've already prepared #2313 raising the minimum to 1.10.4, so it sounds like we can merge that as a starting point anyway.

[...]

PR #2313, opened 2023-09-27
Require HDF5 1.10.4 or above

takluyver:
Drop support for very old versions of HDF5. For the background to this, see #2312.


takluyver:
The discussion in #2312 is all about how much newer an HDF5 we can require; we all seem to be on board with dropping the 1.8 series like this PR does, so I'm going to merge this now.

PR #2319, opened 2023-10-04
Build & test wheels on Python 3.12

### LINKED ISSUE #2310: cp312 wheels
Now that numpy 1.26 has been released, is there a timeline for releasing cp312 wheels for h5py?

takluyver:
Closes #2310

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -27,7 +27,7 @@ jobs:
   pool:
     vmImage: ubuntu-20.04
   variables:
-    CIBW_BUILD: cp3{8,9,10,11}-manylinux_x86_64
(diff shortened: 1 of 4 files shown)


takluyver:
Good, that seems to work. I was a bit worried that we might have to fix Cython 3 compatibility (#2268) before we could support 3.12.

aragilar:
We probably want to add 3.12 to tox as well. We don't include classifiers for Python versions, so we don't need to update them.

takluyver:
Thanks, good point. I've added the tox config now.

tacaswell:
Should we backport this to a branch one commit past the last tag to avoid having to do a release to publish 312 wheels?

Insert my normal complaints about the issues of bundling the binary artifact generation logic into the project source code.

tacaswell (approved):

PR #2331, opened 2023-10-18
`IOError` → `OSError`

DimitriPapadopoulos:
The following exceptions are kept for compatibility with previous versions; starting from Python 3.3, they are aliases of [`OSError`](https://docs.python.org/3/library/exceptions.html#OSError).
* _exception_ **`EnvironmentError`**
* _exception_ **`IOError`**
* _exception_ **`WindowsError`**

See:
https://docs.python.org/3/library/exceptions.html#IOError


takluyver:
That seems reasonable, thanks.

PR #2333, opened 2023-10-18
Use `yield from` where possible

DimitriPapadopoulos:

### DIFF
--- a/h5py/_hl/dims.py
+++ b/h5py/_hl/dims.py
@@ -53,8 +53,7 @@ class DimensionProxy(base.CommonStateObject):
 
     @with_phil
     def __iter__(self):
-        for k in self.keys():
-            yield k
+        yield from self.keys()
 
     @with_phil
     def __len__(self):


tacaswell (approved):

other/unterm-generator.py
```diff
@@ -7,8 +7,7 @@
 
 def yield_groups(filename):
     with h5py.File(filename, 'r') as fh:
-        for group in fh:
-            yield group
```
takluyver, line 11: Please leave this one alone - it's a script that checked for a very specific bug, and even a small change could affect how it behaves. It's not part of the h5py package, the code style really doesn't matter.

(Alternatively, find a version of h5py from before the bug was fixed, and verify that the modified script still shows the same bug. But that's not really a good use of time. :wink: )

DimitriPapadopoulos: I will follow your adivce. Just give me some time to revert one of the files.

DimitriPapadopoulos: Done.

takluyver:
Thank you!

PR #2335, opened 2023-10-18
Directly initialize tuple using generator

DimitriPapadopoulos:

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -334,10 +334,10 @@ class ChunkIterator:
         self._layout = dset.chunks
         if source_sel is None:
             # select over entire dataset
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks, but this seems merely 'different' to me, not obviously better.

DimitriPapadopoulos:
Well, it's not just different:
* The existing code:
  1. creates a list using list comprehension,
  2. creates a set from this list.
* The proposed code:
   1. creates a set using set comprhension.

I am not saying these details won't be optimized away by Python. However, in terms of readability, why use **list comprehension** to create a **set**, when Python provides **set comprehension**?

takluyver:
All good, thanks.

PR #2343, opened 2023-11-17
Support for Blosc2 ND arrays with direct chunking (optimized)

ivilata:
This PR adds optional support for the optimized reading of n-dimensional slices of Blosc2-compressed datasets, leveraging direct chunk access (skipping the slow HDF5 filter pipeline) and 2-level partitioning into chunks and then smaller blocks (so that less data is actually decompressed).  A similar technique was adopted not long ago by PyTables in PyTables/PyTables#1056 (hence the same PR title), with remarkable speed-ups, see this blog post: <https://www.blosc.org/posts/pytables-b2nd-slicing/>.

The feature is optional in the sense that it depends on the presence of both hdf5plugin and Python-Blosc2, and it degrades gracefully to normal filter pipeline-based behaviour if Python-Blosc2 is missing. An extra feature `blosc2` has been added to `pyproject.toml` so that the extra dependencies are only installed if `pip install h5py[blosc2]` (or similar) is explicitly requested. Even if the required dependencies are installed, the user may still force-disable the optimized reading by setting `BLOSC2_FILTER=1` in the environment. All this is documented in the Installation and Datasets sections of the Manual.

We tried to make the changes as minimal and contained as possible: besides the (very limited) changes to doc files, `pyproject.toml` and `tox.ini` (for extra optional deps), and the `Dataset` class (a dozen of lines with a similar behaviour to fast read attempts), actual support code is just a single Python file `h5py/_hl/blosc2.py` (most of it docstrings/comments), plus a simple example file and a reasonably exhaustive unit test file.

With the code as is, we already obtained some 2x-3x speedups in h5py in a benchmark very similar to the one used in the mentioned blog post, between normal filter-based slicing and optimized direct 2-level partitioned slicing.

We'd be very happy to see this PR accepted, however we're open to redo it following a different approach if you think that should fit h5py better.

Thank you very much for your consideration!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,9 @@ jobs:
     - arch: ppc64le
       python: 3.9
       env:
-      - TOXENV=py39-test-deps-blosc2-deps
+      # TODO: This should be py39-test-deps-blosc2-deps (as below),
+      # but C-Blosc2 (as of 2.11.2) doesn't build on ppc64le.
+      - TOXENV=py39-test-deps
       - TOX_OPTS="--force-dep numpy<1.23"
       - HDF5_VERSION=1.10.5
       - HDF5_DIR=$HDF5_CACHE_DIR/$HDF5_VERSION
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ jobs:
     - arch: ppc64le
       python: 3.9
       env:
-      - TOXENV=py39-test-deps
+      - TOXENV=py39-test-deps-blosc2-deps
       - TOX_OPTS="--force-dep numpy<1.23"
       - HDF5_VERSION=1.10.5
       - HDF5_DIR=$HDF5_CACHE_DIR/$HDF5_VERSION
@@ -39,7 +39,7 @@ jobs:
       python: 3.9
       virt: vm
       env:
-        - TOXENV=py39-test-deps
+        - TOXENV=py39-test-deps-blosc2-deps
         - TOX_OPTS=""
         - HDF5_VERSION=1.10.5
         - HDF5_DIR=$HDF5_CACHE_DIR/$HDF5_VERSION
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,7 +9,7 @@ environment:
     ### USING HDF5 1.10 ###
 
     - PYTHON: "C:\\Python38"
-      TOXENV: "py38-test-deps-blosc2-deps"
+      TOXENV: "py38-test-deps"
       HDF5_VSVERSION: "14"
       HDF5_DIR: "C:\\hdf5\\%HDF5_VERSION%\\%HDF5_VSVERSION%"
       HDF5_VERSION: "1.10.6"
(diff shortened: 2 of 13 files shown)


ivilata:
BTW, here are the results of the preliminary benchmarks that I mention in the PR description, where the speedups in h5py from filter (green) to optimized (purple) can be seen (along with the equivalent optimization in PyTables). It's also true that the last update to `hdf5-blosc2` in hdf5plugin to support b2nd did apparently slow down filter-based access. We're still profiling everything…

b2nd_compare_getslice

ivilata:
I've tried to enable testing Bloc2 optimized slicing in CI runs (except for ppc64le, where C-Blosc2 fails to build), but I'm finding issues with the AppVeyor Windows builds. All three Python 3.8, 3.9 & 3.11 runs fail with this funny build error:

```
  Getting requirements to build wheel: finished with status 'error'
py311-test-deps-blosc2-deps: exit 1 (42.53 seconds) C:\projects\h5py> python -I -m pip install blosc2>=2.3.0 hdf5plugin>=4.3.0 numpy>=1.23.2 pytest pytest-cov pytest-mpi>=0.2 pid=5812
  py311-test-deps-blosc2-deps: FAIL code 1 (43.98 seconds)
  evaluation failed :( (44.42 seconds)
  error: subprocess-exited-with-error
[…]
  reading manifest file 'h5py.egg-info\SOURCES.txt'
 […]
    File "C:\Users\appveyor\AppData\Local\Temp\1\pip-build-env-_g20yr09\overlay\Lib\site-packages\setuptools\_distutils\util.py", line 139, in convert_path
      raise ValueError("path '%s' cannot be absolute" % pathname)
  ValueError: path '/home/takluyver/Code/h5py/h5py/_conv.c' cannot be absolute
  [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error
Getting requirements to build wheel did not run successfully.
exit code: 1
```

Example output [here](https://ci.appveyor.com/project/h5py/h5py/builds/48565722/job/2b9i50weexu3dpf6). I'm a little bit at a loss on what to try now, any help is much appreciated. :slightly_smiling_face:

ivilata:
In the end I disabled the Blosc2 tests for Win32 in ed99de9 (Blosc2 in not very well supported on 32-bit architectures anyway).

ivilata:
> It's also true that the last update to `hdf5-blosc2` in hdf5plugin to support b2nd did apparently slow down filter-based access.

I found some bug in the benchmark and it may be the case that the new version of `hdf5-blosc2` is *not* slower than the old one. However, the fixed results are quite similar.

h5py-pytables-big-chunk

takluyver:
Thanks for this work, but I'd prefer h5py to remain as a more-or-less Pythonic wrapper for HDF5, and not grow extra code paths to bypass HDF5 in specific circumstances, even if big speedups are possible. It's already enough work to deal with HDF5 versions, different drivers and compile options, Python versions and Numpy versions, without involving extra libraries.

Can you get everything you need for this from the public API of h5py? If so, perhaps you can maintain this as a separate package, with an API like `BloscReader(dataset)[:10]`.

@tacaswell, @aragilar, feel free to disagree with me :slightly_smiling_face:

ivilata:
Thanks @takluyver anyway for having a look at this! I understand your position, and fortunately everything more or less relies on the public API indeed, so I'll start a different project as you suggested. Hopefully most of the code that I developed should be reusable nearly as-is. :slightly_smiling_face:

I'm closing the PR. Cheers!

takluyver:
Thanks Ivan! When you're ready, we could add it to the list of related
projects on this page: https://docs.h5py.org/en/stable/related_projects.html

PR #2345, opened 2023-11-28
Fix Cython 3 compatibility

djhoese:
Closes #2268.

This is a first step in Cython 3 compatibility. This is based on a branch started by @takluyver. The basic idea so far is that functions needed to be defined with `except X` or `noexcept` to define how errors are handled. The complication was in h5py/h5fd.pyx which casts the h5py version of the file functions (which use a wrapper struct) to match the HDF5 version of the function. Due to what I think is a bug in Cython 3, you can't include an `except X` statement in your casting so as suggested on the Cython mailing list I created ctypedefs for every function pointer that needed one.

TODO:

- [ ] Review all performance warnings during compilation. Cython points out that many functions are not handled yet and it is now assuming that exceptions will be raised and the GIL is acquired to check for an exception.
- [ ] Not needed for initial compatibility, but removal of all IF statements since it is being removed from the Cython language. There are some suggestions on how to not use them, but I'm not sure they will work for all cases used in h5py.

Note the alternative suggested by da-woods (not tagging here to not be annoying) in the above mentioned issue of just specifying the Cython option to force the old behavior.

https://github.com/h5py/h5py/issues/2268#issuecomment-1829300815

#### Example of performance warning

```
performance hint: h5py/_locks.pxi:114:5: Exception check on 'unlock_lock' will always require the GIL to be acquired.
Possible solutions:
        1. Declare the function as 'noexcept' if you control the definition and you're sure you don't want the function to raise exceptions.
        2. Use an 'int' return type on the function to allow an error code to be returned.
```

PR #2349, opened 2023-12-04
Require NumPy < 2 (for now)

takluyver:
From discussion in #2345: NumPy 2.0 is coming, and will break ABI compatibility.

For the future, it looks like we'll be able to [build with newer Numpy specifying the desired compatibility](https://numpy.org/doc/stable/release/1.25.0-notes.html#compiling-against-the-numpy-c-api-is-now-backwards-compatible-by-default) rather than building with the oldest supported Numpy. We should investigate that once Numpy 2.0 is close to ready (we should get at least 6 weeks with an RC). But for now, let's just make it clear that the package expects Numpy 1.x.


aragilar:
@takluyver Should we merge this?

takluyver:
Closing in favour of #2401

Issue #2298, opened 2023-08-22
Support for H5D multi functions?


jreadey:
It seems that the multi dataset functions (H5Dread_multi/H5Dwrite_multi) that were added in HDF5 1.12 are not supported in h5py.  Would there be interest among the h5py community in having these?

Some background: H5Dread_multi allows reading multiple selections from multiple datasets in one call.  H5Dwrite_multi is the same thing for writing.  In parallel HDF these can provide a significant speed up compared to do the same selections sequentially.  For regular HDF5 - not so much unfortunately.

Here's the RFC that provides more detail into the implementation: https://docs.hdfgroup.org/hdf5/rfc/H5HPC_MultiDset_RW_IO_RFC.pdf.

mattjala:
I'm working on implementing h5py wrappers for `H5Dread_multi`/`H5Dwrite_multi`, hopefully I'll have a PR ready within the next couple days.

PR #2351, opened 2023-12-05
Implement low level H5Dread_multi call

mattjala:
This implements the function `read_multi` in `h5d.pyx`, in order to call HDF5's `H5Dread_multi`.

To avoid duplicating type conversion handling, I modified `dset_rw` to handle both the single and multi dataset cases. 

After review, I plan to implement a similar function for `H5Dwrite_multi`, and then high-level calls for both. 

Addresses #2298


h5py/h5d.pyx
```diff
@@ -151,6 +151,55 @@ def open(ObjectID loc not None, char* name, PropID dapl=None):
     """
     return DatasetID(H5Dopen(loc.id, name, pdefault(dapl)))
 
+IF HDF5_VERSION >= (1, 14, 0):
+    @with_phil
+    def read_multi(count, list dataset_ids, list mspace_ids, list fspace_ids,
+        list type_ids, list bufs not None, PropID dxpl=None):
```
mattjala, line 157: These parameters are lists instead of numpy arrays to avoid an issue with the destination buffers. 

Specifically, the individual destination buffer for each dset read should be a numpy array. If `bufs` were itself also a numpy array containing these arrays, then it would copy those arrays instead of pointing to the same memory, and the original destination arrays the user created wouldn't get populated.

takluyver:
As `dset_rw` is kind of critical to h5py, and this introduces extra mallocs in there - i.e. extra places that might fail or slow things down - I wonder if it would make sense to write a separate `dset_rw_multi`. Even if that means some code duplication, it should mean the addition doesn't make the existing APIs any worse.

h5py/h5d.pyx
takluyver: Thanks, that makes sense.

The Python function probably shouldn't take `count` as the first parameter - it's necessary in C to tell you how many items you can read after a pointer, but lists have a length.

But if that's the case, I'd probably do a check up front that the lists are all the same length, and fail with a clear error if not.

h5py/tests/test_h5d_direct_chunk.py
```diff
@@ -186,3 +186,75 @@ def test_fail_buffer_not_contiguous(self, writable_file):
...
+        with h5py.File(filename, "w") as filehandle:
+            shape = (10, 10, 10)
+            dt = numpy.int32
+
+            # Write data
+            data_in = numpy.reshape(numpy.arange(numpy.prod(shape)), shape)
+            dataset = filehandle.create_dataset("data", shape,
+                                                dtype=dt, data=data_in)
+
+            self.assertTrue(numpy.array_equal(data_in, dataset[...]))
```
takluyver, line 205:
```suggestion
            numpy.testing.assert_array_equal(data_in, dataset[...])
```

And throughout. Specific assert methods like this can give you more useful information if the thing they're checking fails, whereas `assertTrue` can only tell you it was not true.  E.g.

```
AssertionError: 
Arrays are not equal

(shapes (5,), (6,) mismatch)
 x: array([0, 1, 2, 3, 4])
 y: array([0, 1, 2, 3, 4, 5])
```

h5py/tests/test_h5d_direct_chunk.py
```diff
@@ -186,3 +186,75 @@ def test_fail_buffer_not_contiguous(self, writable_file):
         out = array[:, :, ::2]  # Array is not contiguous
         with pytest.raises(ValueError):
             dataset.id.read_direct_chunk((0, 0), out=out)
+
+
+@ut.skipIf(h5py.version.hdf5_version_tuple < (1, 14, 0),
+           "read_multi requires HDF5 >= 1.14.0")
+class TestReadMulti(TestCase):
```
takluyver, line 193: This is in the file `test_h5d_direct_chunk.py`, the rest of which is all about the direct chunk read/write methods. I think this file is small enough that we may as well just rename it to `test_h5d.py`.

h5py/h5d.pyx
```diff
@@ -151,6 +151,63 @@ def open(ObjectID loc not None, char* name, PropID dapl=None):
     """
     return DatasetID(H5Dopen(loc.id, name, pdefault(dapl)))
 
+IF HDF5_VERSION >= (1, 14, 0):
+    @with_phil
+    def rw_multi(list dataset_ids, list mspace_ids, list fspace_ids,
```
mattjala, line 156: This could be broken up into `read_multi` and `write_multi` for readability, but they would be identical except for one parameter passed to `dset_rw_multi`.

mattjala:
I have an implementation of a high-level `MultiManager` class that's ready for review. Would it be best to bring that into this PR, or make a separate request after this is merged?

ajelenak:
@takluyver @tacaswell Is there anything else that needs to be done in this PR? Would be nice to merge it soon.

aragilar:
It's not clear to me if the issues @takluyver had were resolved (otherwise I'd be happy to merge this if it got rebased)?

I'm not clear if this is going to be the only new way to access datasets, or if there are going to be further ones, but in the high level code, I wonder if it makes sense to pull out the conversion/parsing of parameters into a single function so that changes there are consistent for both code paths?

mattjala:
@aragilar I've rebased my changes and factored out some parameter processing that was shared between `Dataset` and `MultiManager`.

PR #2358, opened 2023-12-22
Creates access to remaining H5O_info_t struct fields, Closes #2347

### LINKED ISSUE #2347: Btree and index info through H5o
I'm trying to optimize dataset layouts and chunk sizes. Part of this optimization requires looking at how much metadata overhead we are incurring vs a particular layout and chunk size.

Looking through `h5py.h5o` I see there is no definition for this struct which provides access btree and index size. ..which comes from `metasize.obj` field.

https://docs.hdfgroup.org/hdf5/v1_12/struct_h5_o__info1__t.html

What I am asking is, is my understanding true? Can I submit a PR for it?

Delengowski:
These are just exposing additional fields of `H5O_info_t` struct and has no direct interface on high level classes so I don't think tests are necessary. Let me know if otherwise.

### DIFF
--- a/h5py/h5o.pyx
+++ b/h5py/h5o.pyx
@@ -84,13 +84,55 @@ cdef class _OHdr(_ObjInfoBase):
     property nmesgs:
         def __get__(self):
             return self.istr[0].hdr.nmesgs
+    property nchunks:
+        def __get__(self):
+            return self.istr[0].hdr.nchunks
+    property flags:
+        def __get__(self):
+            return self.istr[0].hdr.flags
 
     def __init__(self):
         self.space = _OHdrSpace()
         self.mesg = _OHdrMesg()
 
     def _hash(self):
-        return hash((self.version, self.nmesgs, self.space, self.mesg))
+        return hash((self.version, self.nmesgs, self.nchunks, self.flags, self.space, self.mesg))
+
+cdef class _MetaSizeObj(_ObjInfoBase):
+
+    property index_size:
+        def __get__(self):
+            return self.istr[0].meta_size.obj.index_size
+    property heap_size:
+        def __get__(self):
+            return self.istr[0].meta_size.obj.heap_size
+
+    def _hash(self):
+        return hash((self.index_size, self.heap_size))
+
+cdef class _MetaSizeAttr(_ObjInfoBase):
+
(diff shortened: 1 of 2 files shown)


h5py/h5o.pyx
```diff
@@ -84,13 +84,55 @@ cdef class _OHdr(_ObjInfoBase):
...
+
+    property index_size:
+        def __get__(self):
+            return self.istr[0].meta_size.attr.index_size
+    property heap_size:
+        def __get__(self):
+            return self.istr[0].meta_size.attr.heap_size
+
+    def _hash(self):
+        return hash((self.index_size, self.heap_size))
```
takluyver, line 123: I think these two classes can be condensed into one - we'll need two instances, but the corresponding C structs are of the same type (`H5_ih_info_t`).

Delengowski: Yes, I tried to make them one but was struggling with the Cython code to access `attr` vs `obj` field from `self.istr[0].meta_size`.

Its all very new to me. I'll take another crack.

Delengowski: I believe I got it working. It built and I was able to access the structs in ipython.

```
In [1]: import h5py

In [2]: h5py.__file__
Out[2]: '/home/delen/miniconda3/envs/h5py_dev/lib/python3.10/site-packages/h5py/__init__.py'

In [3]: from h5py.h5o import get_info

In [4]: h5 = h5py.File("./h5py/h5py/tests/data_files/vlen_string_dset.h5")

In [5]: get_info(h5['DS1'].id).meta_size.obj.index_size
Out[5]: 0

In [6]: get_info(h5['DS1'].id).meta_size.attr.index_size
Out[6]: 0

In [7]: get_info(h5['DS1'].id).meta_size.attr.heap_size
Out[7]: 0

In [8]: get_info(h5['DS1'].id).meta_size.obj.heap_size
Out[8]: 0
```

takluyver:
Thanks, this looks good to me. Could I ask you to add a news entry for it?

https://docs.h5py.org/en/stable/contributing.html#write-a-release-note

Delengowski:
Closes #2347

takluyver:
Thanks @Delengowski !

PR #2360, opened 2024-01-09
Add methods to Group to visit links

### LINKED ISSUE #671: f.visit and f.visititems don't visit h5py.ExternalLink type of items
Hello,

Running

``` python
import numpy as np
import h5py

with h5py.File('file2', 'w', libversion='latest') as f:
    g = f.create_group('grp')
    g.create_dataset('X', data=np.random.randn(5,5))

with h5py.File('file1', 'w', libversion='latest') as f:
    g = f.create_group('grp')
    g['X'] = h5py.ExternalLink('file1', 'grp/X')
    g.create_dataset('Y', data=np.random.randn(5,5))

with h5py.File('file1', 'r', libversion='latest') as f:
    def func(name):
        print name
    print 'Visited:'
    f.visit(func)
    print 'Manually:'
    print 'f.keys()', f.keys()
    print "f['grp'].keys()", f['grp'].keys()
```

prints

> Visited:
> grp
> grp/Y
> Manually:
> f.keys() [u'grp']
> f['grp'].keys() [u'X', u'Y']

Obs.:

> ~> python -c "import h5py; print h5py.__version__"
> 2.5.0

Danilo

loichuder:
Fix #671

This PR adds the methods `visit_links` and `visititems_links` as proposed by @takluyver in https://github.com/h5py/h5py/issues/671#issuecomment-1866088248.

I also added tests and the news file (I hope it is fine?)

### DIFF
--- a/h5py/_hl/group.py
+++ b/h5py/_hl/group.py
@@ -613,6 +613,9 @@ class Group(HLObject, MutableMappingHDF5):
     def visit(self, func):
         """ Recursively visit all names in this group and subgroups.
 
+        Note: visit ignores soft and external links. To visit those, use
+        visit_links.
+
         You supply a callable (function, method or callable object); it
         will be called exactly once for each link in this group and every
         group below it. Your callable must conform to the signature:
@@ -638,6 +641,9 @@ class Group(HLObject, MutableMappingHDF5):
 
     def visititems(self, func):
         """ Recursively visit names and objects in this group.
+
+        Note: visititems ignores soft and external links. To visit those, use
+        visititems_links.
 
         You supply a callable (function, method or callable object); it
         will be called exactly once for each link in this group and every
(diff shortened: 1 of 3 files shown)


h5py/_hl/group.py
```diff
@@ -667,6 +673,63 @@ def proxy(name):
...
+        Example:
+
+        # Get a list of all softlinks in the file
+        >>> mylist = []
+        >>> def func(name, link):
+        ...     if isinstance(link, SoftLink):
+        ...         mylist.append(name)
+        ...
+        >>> f = File('foo.hdf5')
+        >>> f.visititems_links(func)
```
loichuder, line 724: Docstring copy-pasted from `visititems` and adapted to links (including the example)

h5py/tests/test_group.py
```diff
@@ -697,6 +697,48 @@ def test_bailout(self):
...
+        """ All subgroups and links are visited """
+        l = []
+        self.f.visit_links(l.append)
+        self.assertSameElements(l, self.groups + self.links)
+
+    def test_visititems(self):
+        """ All links are visited """
+        l = []
+        comp = [(x, type(self.f.get(x, getlink=True))) for x in self.groups + self.links]
+        self.f.visititems_links(lambda x, y: l.append((x, type(y))))
```
loichuder, line 731: I resorted to test for the type of links because apparently it is not possible to test for link equality??

```python
>>> l = f.get('grp1', getlink=True)
>>> l2 = f.get('grp1', getlink=True)
>>> l == l2
False
```

:thinking:

takluyver: That might be easy to fix, but it raises the question: if two different files contain links with the same path, which would point to different objects because of starting at different places, do we expect them to be equal?

At present, the Python `SoftLink` & `ExternalLink` classes just store the path & (for external links) the filename. They don't keep any reference to the file they came from.

My leaning is that that's OK, two links with the same path can be equal, even if within their files they would resolve differently. :shrug:

h5py/_hl/group.py
```diff
@@ -667,6 +673,63 @@ def proxy(name):
                 return func(name, self[name])
             return h5o.visit(self.id, proxy)
 
+    def visit_links(self, func):
+        """ Recursively visit all names in this group and subgroups, including links.
```
takluyver, line 677: Technically, all names in a group are links - but normally they're hard links, pointing to an object rather than a name.

IDK how important it is to make this clear, but it does lead to another significant difference between these methods & the existing (object-based) ones. If you have two hardlinks to the same object, the link-based functions will see both of them, but the object-based functions will only see one - whichever one they reach first.

h5py/tests/test_group.py
loichuder: Well, I guess it depends on your usecase.

Personally, I can't think of a usecase where you would want to check for link equality, except when testing. :shrug:

h5py/_hl/group.py
loichuder: Understood. I pushed a commit to mention this in the docstring.

takluyver:
Thanks!

PR #2368, opened 2024-01-16
DO NOT MERGE: pytest potential fix

woutdenolf:
Tests pytest fix https://github.com/pytest-dev/pytest/pull/11821 which tries to fix https://github.com/pytest-dev/pytest/issues/11816.


woutdenolf:
Should be fixed by https://github.com/pytest-dev/pytest/pull/11825

PR #2381, opened 2024-02-16
Fix dset.iter_chunks() with selection

Delengowski:
Found a solution that works with other test cases.

The way I saw the problem, we would calculate a chunk_slice that was not a subset of the requested slice. So the I added a check that checks that the calculated chunk slice is a subset of the requested slice.

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -353,7 +353,6 @@ class ChunkIterator:
             index = s.start // self._layout[dim]
             self._chunk_index.append(index)
 
-
     def __iter__(self):
         return self
 
@@ -430,7 +429,7 @@ class Dataset(HLObject):
     def fields(self, names, *, _prior_dtype=None):
         """Get a wrapper to read a subset of fields from a compound data type:
 
-                            >>> 2d_coords = dataset.fields(['x', 'y'])[:]
+        >>> 2d_coords = dataset.fields(['x', 'y'])[:]
 
         If names is a string, a single field is extracted, and the resulting
         arrays will have that dtype. Otherwise, it should be an iterable,
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -325,9 +325,7 @@ class ChunkIterator:
     """
     def __init__(self, dset, source_sel=None):
         self._shape = dset.shape
-        self._rank = len(dset.shape)
-        self._slices = [slice(None)]*self._rank
-
+        rank = len(dset.shape)
 
         if not dset.chunks:
             # can only use with chunked datasets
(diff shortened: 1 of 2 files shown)


aragilar:
Looks like codecov had an issue uploading, rerunning the failing jobs.

Delengowski:
Adds a small runtime hit when there is no selection.  I'm wondering that if `souce_sel` is passed as `None` we should keep that information and not perform that check at all. This is all assuming my check is the correct way to go about it and there's not a simpler solution.

```python
import h5py
import numpy
a = numpy.arange(25).reshape(5, 5)
h5f = h5py.File('iterchunks.h5', 'w')
d = h5f.create_dataset('test1', data=a, chunks=(2, 2))
```


This branch

```python
In [22]: %timeit -n 1000000 -r 7 [x for x in d.iter_chunks()]
37.3 µs ± 1.25 µs per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

Master

```python
In [10]: %timeit -n 1000000 -r 7 [x for x in d.iter_chunks()]
32 µs ± 265 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

h5py/_hl/dataset.py
```diff
@@ -336,61 +338,72 @@ def __init__(self, dset, source_sel=None):
...
+                        self._slices[dim] = self._calculate_next_chunk(dim)
+                        self._chunk_index[dim] += 1
+                        break
                 # we still have room to extend along this dimensions
-                return tuple(slices)
+                break
 
             if dim > 0:
                 # reset to the start and continue iterating with higher dimension
                 self._chunk_index[dim] = 0
```
takluyver, line 392: Looking at the code in master at present, this is the bit I'd be looking to change. When it hits the end of the selection in one dimension, this is telling it to go back to the start in that dimension - but the overall start of the dimension rather than the start of the selection. So I would try something like this:

```suggestion
                self._chunk_index[dim] = s.start // self._layout[dim]
```

If that works, I suspect it might fix it without the added complexity elsewhere in the class. In particular, I'm suspicious of the line `self._chunk_index[dim] = 1` you've added a few lines above. I suspect this hardcoded 1 works for the particular case you've added as a test, but will still be wrong in other cases.

Delengowski: I will get this tomorrow. I hope it's simpler bc I felt what I was doing was complicated but it was only solution I was conceptualizing.

Delengowski: `git checkout master -- h5py/_hl/dataset.py` and swapping that does indeed make it pass for the additional test case.

This is good because I was uncomfortable with the slowness I introduced in the common case.

Delengowski:
Current test failures seem unrelated


https://dev.azure.com/h5pyappveyor/h5py/_build/results?buildId=1917&view=logs&j=fb5a0786-7759-5eb0-8063-f3845f8c552c&t=efe2eb3b-410d-520c-c622-b15b2e00631e&l=334

takluyver:
Thanks! Yes, I think the failures are where it's trying to test with numpy 2, which is in beta now, so nothing to do with this. #2386 is working on that.

takluyver:
Close/reopen to re-run tests with that PR merged.

Issue #2379, opened 2024-02-15
h5py build errors when building with gcc 14


befeleme:
During the build of h5py with gcc 14 as RPM in Fedora Linux there appear errors  that lead to a failed build. 

To assist reproducing bugs, please include the following:
 * Fedora Linux 40
 * system Python 3.12
 * h5py version: 3.10.0
 * HDF5 version: 1.12.1
 * The full traceback/stack trace shown (if it appears): full log: https://kojipkgs.fedoraproject.org/work/tasks/1570/113281570/build.log

```
gcc -fno-strict-overflow -Wsign-compare -DDYNAMIC_ANNOTATIONS_ENABLED=1 -DNDEBUG -fexceptions -fexceptions -fexceptions -O2 -flto=auto -ffat-lto-objects -fexceptions -g -grecord-gcc-switches -pipe -Wall -Wno-complain-wrong-lang -Werror=format-security -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=3 -Wp,-D_GLIBCXX_ASSERTIONS -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -fstack-protector-strong -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -mbranch-protection=standard -fasynchronous-unwind-tables -fstack-clash-protection -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -fopenmp -fPIC -DH5_USE_110_API -DH5Rdereference_vers=2 -DNPY_NO_DEPRECATED_API=0 -Ih5py -I/builddir/build/BUILD/h5py-3.10.0/serial/lzf -I/usr/lib64/python3.12/site-packages/numpy/core/include -I/usr/include/python3.12 -c /builddir/build/BUILD/h5py-3.10.0/serial/h5py/h5l.c -o build/temp.linux-aarch64-cpython-312/builddir/build/BUILD/h5py-3.10.0/serial/h5py/h5l.o
/builddir/build/BUILD/h5py-3.10.0/serial/h5py/h5l.c: In function ‘__pyx_pf_4h5py_3h5l_9LinkProxy_12get_val’:
/builddir/build/BUILD/h5py-3.10.0/serial/h5py/h5l.c:8145:107: error: passing argument 4 of ‘__pyx_f_4h5py_4defs_H5Lunpack_elink_val’ from incompatible pointer type [-Wincompatible-pointer-types]
[...]

PR #2382, opened 2024-02-22
Use libc.stdint instead of numpy

opoplawski:
Additional fix for #2379 

See also https://github.com/numpy/numpy/issues/25869


aragilar:
This looks fine to me, thanks @opoplawski for fixing this.

PR #2386, opened 2024-02-23
BUG: Fix nightly wheel build for NumPy 2.0

larsoner:
I tried using the shiny new nightly wheels but [they don't work with NumPy 2.0.0.dev0](https://github.com/mne-tools/mne-python/actions/runs/8021028446/job/21912196809?pr=12457#step:17:18296) because they're not built against NumPy 2.0. It seems like they should be so I'm trying it by setting some pyproject.toml options.

No idea if this is a good idea in terms of how releases are done (?), but it should fix the nightly wheels. If it's bad for releases, maybe a second job that builds them against NumPy 2.0 would be good, in which case env vars could be used instead of pyproject.toml.


larsoner:
I think if `h5py` does not plan to release again before NumPy 2.0 comes out, then this could probably be merged as-is (once I fix the Python 3.8 issue). Then once NumPy 2.0 comes out some of the `--pre` stuff can go away (as can `oldest-supported-numpy`) but otherwise this should work. If `h5py` *does* plan to release *and* the wheel builds from Azure are used as part of that... then we need some sort of conditionals in the scripts or redundant runs. One option would be to set up the nightly version to use GHA rather than `azure`, too. I think it's simpler there in terms of configuration.

takluyver:
Do we have a rough sense of when Numpy 2.0 is expected? The wheels we build on Azure are the ones we publish for a release.

I don't have any specific plan for the next release of h5py, but it's been a few months since the last release, and I'd rather not be stuck in a state where we can't release until Numpy 2.0 drops.

tacaswell:
On Matplotlib https://github.com/matplotlib/matplotlib/blob/5e347777192b37ac77decc4f975e53ed9d87973b/.github/workflows/cibuildwheel.yml#L103-L108

```
      CIBW_BUILD_FRONTEND: >-
        ${{ ((github.event_name == 'push' && github.ref == 'refs/heads/main') ||
              (github.event_name == 'pull_request' && github.base_ref == 'main')) &&
            'pip; args: --pre
             --extra-index-url "https://pypi.anaconda.org/scientific-python-nightly-wheels/simple"' ||
            'build' }}
```

we condition this on if the build is triggered from main or a tag.  I assume that is something we can also do with azure (but do not know how off the top of my head).

takluyver:
I don't love this solution: the point of building & testing wheels in CI is that we don't get nasty surprises when we try to make a release (or when it's already made :scream: ). Doing something different at release time, either with conditions in the pipeline or by reverting a change just before release, leaves the door open to unforeseen problems.

Is it viable just to hold off on this until Numpy 2.0 is looking closer (on the order of a couple of weeks rather than 3 months)? Or set up a separate repo with CI to build & test against numpy 2.0? I appreciate that it's a good idea for there to be tests somewhere with a development version of Numpy, but I'd rather it didn't replace our normal CI based on the Numpy that people are using today.

larsoner:
> Is it viable just to hold off on this until Numpy 2.0 is looking closer (on the order of a couple of weeks rather than 3 months)? Or set up a separate repo with CI to build & test against numpy 2.0? I appreciate that it's a good idea for there to be tests somewhere with a development version of Numpy, but I'd rather it didn't replace our normal CI based on the Numpy that people are using today.


Okay how about instead an entirely separate job that builds and tests wheels against the nightly NumPy from scientific-python-nightly-wheels? That way you get the benefit of testing against numpy-dev -- which is useful in itself -- but also as a side benefit can then upload those wheels to the same repo. I've pushed a version of that in case it seems viable.

larsoner:
Okay:

- Enabled for PRs only Linux wheel building for now -- I think it's the most critical case and also the easiest one to add. Later refactoring of `CIBW_` params could add macOS and Windows without too much effort I think.
- It runs on PRs to make sure things here are NumPy-dev compatible, building all Linux wheels but testing just on 3.12 for simplicity/speed
- The action takes ~10 min, which seems like a reasonable amount of time considering the `linux_wheels` build job plus one test job on Azure alone takes ~13 min. It also runs concurrently with the Azure stuff so hopefully won't cause a backlog during PRs :crossed_fingers: 
- Has a scheduled nightly run to upload wheels to Anaconda cloud (in theory!). At least all the expected wheels are in the right place in the `dist/` directory in the run:

https://github.com/h5py/h5py/actions/runs/8053630312/job/21996858474?pr=2386#step:4:11  

Pushing a commit to make the "Upload wheels" job conditional on being part of a scheduled run (and also hopefully fix coverage upload, which silently failed on the last commit).

larsoner:
Friendly ping on this one -- would help projects that depend on scientific-python-nightly-wheels I think! I'll update to latest `main` to make sure things are still green :+1:

h5py/tests/test_dataset.py
```diff
@@ -912,7 +912,7 @@ def test_vlen_unicode(self):
     def test_string_fixed(self):
         """ Assignment of fixed-length byte string produces a fixed-length
         ascii dataset """
-        self.f['x'] = np.string_("Hello there")
+        self.f['x'] = np.bytes_("Hello there")
```
larsoner, line 915: Without this on 2.0:
```
E           AttributeError: `np.string_` was removed in the NumPy 2.0 release. Use `np.bytes_` instead.. Did you mean: 'strings'?
```

h5py/tests/test_dtype.py
```diff
@@ -132,7 +132,7 @@ def test_write_empty_vlen(self):
         fname = self.mktemp()
         with h5py.File(fname, 'w') as f:
             d = np.core.records.fromarrays([[], []], names='a,b', formats='|V16,O')
-            dset = f.create_dataset('test', data=d, dtype=[('a', '|V16'), ('b', h5py.special_dtype(vlen=np.float_))])
+            dset = f.create_dataset('test', data=d, dtype=[('a', '|V16'), ('b', h5py.special_dtype(vlen=np.float64))])
```
larsoner, line 135:
```
E           AttributeError: `np.float_` was removed in the NumPy 2.0 release. Use `np.float64` instead.. Did you mean: 'float16'?
```

tox.ini
```diff
@@ -11,11 +11,12 @@ deps =
...
+    # For --pre, 2.0.0b1 has a header size difference that has been reverted, so avoid it
     py38-deps: numpy>=1.17.5
-    py39-deps: numpy>=1.19.3
-    py310-deps: numpy>=1.21.3
-    py311-deps: numpy>=1.23.2
-    py312-deps: numpy>=1.26.0
+    py39-deps: numpy>=1.19.3,!=2.0.0b1
+    py310-deps: numpy>=1.21.3,!=2.0.0b1
+    py311-deps: numpy>=1.23.2,!=2.0.0b1
+    py312-deps: numpy>=1.26.0,!=2.0.0b1
```
larsoner, line 19: Fixes failures in this PR and in `master` like

https://dev.azure.com/h5pyappveyor/h5py/_build/results?buildId=1914&view=logs&jobId=a5aa7219-b716-5eaa-2dae-b4098305f378

larsoner:
Took a few rounds of commits but got things green again -- looks like this PR now also fixes some failures on `master` at the moment (can split those off if you want!).

takluyver:
Thanks, this looks good to me now - we can check that things work with pre-release NumPy, but still also know that it builds with a released version at the same time.

I'll merge this now to fix the CI on master as well.

Issue #2394, opened 2024-03-19
Dataset to numpy conversion replaces part of the data with zeros


cbassa:
An occasional problem I run into is that when reading an HDF5 dataset with h5py and converting it to a numpy array, a large fraction of the output array is replaced with zeros. This appears to only be the case when reading the entire dataset, as slices of a subset of the dataset results in the correct data being returned. What may be going wrong here?

The example code below yields these results:
```
Fraction of zeros with np.fromfile:  0.0007078439597315436
Fraction of zeros with cast to np.array:  0.633608081655481
Fraction of zeros with read_direct:  0.633608081655481
Fraction of zeros with horizontal slice:  0.633608081655481
Fraction of zeros with horizontal and vertical slice:  0.633608081655481
Fraction of zeros with almost complete vertical slice:  0.0007078439597315436
Fraction of zeros with complete vertical slice:  0.633608081655481
```
As the data and header are split into separate files, reading the data with `np.fromfile` returns the expected input, which only has 0.07% consisting of zeros. All other approaches, except the partial slicing, return arrays in which 63% has been replaced with zeros. 

The input array has a shape of 228864x6400 values as float32.

```
#!/usr/bin/env python3
import h5py
import numpy as np

if __name__ == "__main__":
    # File names
    h5fname = "L2037874_SAP000_B000_S0_P000_bf.h5"
    rawfname = "L2037874_SAP000_B000_S0_P000_bf.raw"
    group_name = "SUB_ARRAY_POINTING_000/BEAM_000/STOKES_0"

    # Read with numpy directly
    data = np.fromfile(rawfname, dtype="float32").reshape(-1, 6400)
    print("Fraction of zeros with np.fromfile: ", np.sum(data == 0) / np.prod(data.shape))

    # Cast to numpy array
    h5 = h5py.File(h5fname, "r")
    data = np.array(h5[group_name])
    h5.close()
    print("Fraction of zeros with cast to np.array: ", np.sum(data == 0) / np.prod(data.shape))

    # Use read_direct
    h5 = h5py.File(h5fname, "r")
    data = np.zeros((228864, 6400), dtype="float32")
    h5[group_name].read_direct(data)
    h5.close()
    print("Fraction of zeros with read_direct: ", np.sum(data == 0) / np.prod(data.shape))    

    # Using horizontal slice
    h5 = h5py.File(h5fname, "r")
    data = h5[group_name][:]
    h5.close()
[...]

PR #2398, opened 2024-03-21
Allow creation of expandable external datasets

### LINKED ISSUE #2396: Fix `maxshape` and `external` keywords interaction when creating an external dataset
External datasets have data in a raw binary file external to their HDF5 file and can be extended along the first dimension. Presence of the `maxshape` argument when creating an extendable external dataset currently always triggers the chunked layout which causes a libhdf5 error. Setting of chunked layout should take into account also the presence of the `external` keyword which declares external raw binary files with dataset data.

See #2394 issue for details.

 * Operating System: macOS 14.4 (Sonoma)
 * h5py version: latest master branch

h5py    3.10.0
HDF5    1.14.3
Python  3.12.0 | packaged by conda-forge | (main, Oct  3 2023, 08:36:57) [Clang 15.0.7 ]
sys.platform    darwin
numpy   1.26.0
cython (built with) 3.0.9
numpy (built against) 1.26.2
HDF5 (built against) 1.14.3

ajelenak:
External HDF5 datasets can be expanded along their first dimension but cannot have chunked layout. This PR avoids setting chunked layout when creating a dataset if both `maxshape` and `external` keyword arguments are present.

Fixes #2396.

### DIFF
--- a/h5py/_hl/filters.py
+++ b/h5py/_hl/filters.py
@@ -246,8 +246,13 @@ def fill_dcpl(plist, shape, dtype, chunks, compression, compression_opts,
     # End argument validation
 
     if (chunks is True) or \
-    (chunks is None and any((shuffle, fletcher32, compression, maxshape and not len(external),
-                             scaleoffset is not None))):
+    if (chunks is True) or (chunks is None and any((
+            shuffle,
+            fletcher32,
+            compression,
+            (maxshape and not len(external)),
+            scaleoffset is not None,
+    ))):
         chunks = guess_chunk(shape, maxshape, dtype.itemsize)
 
     if maxshape is True:
--- a/h5py/_hl/filters.py
+++ b/h5py/_hl/filters.py
@@ -246,7 +246,7 @@ def fill_dcpl(plist, shape, dtype, chunks, compression, compression_opts,
     # End argument validation
 
     if (chunks is True) or \
-    (chunks is None and any((shuffle, fletcher32, compression, maxshape,
+    (chunks is None and any((shuffle, fletcher32, compression, maxshape and not len(external),
                              scaleoffset is not None))):
         chunks = guess_chunk(shape, maxshape, dtype.itemsize)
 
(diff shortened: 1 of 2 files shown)


h5py/_hl/filters.py
```diff
@@ -246,7 +246,7 @@ def rq_tuple(tpl, name):
     # End argument validation
 
     if (chunks is True) or \
-    (chunks is None and any((shuffle, fletcher32, compression, maxshape,
+    (chunks is None and any((shuffle, fletcher32, compression, maxshape and not len(external),
                              scaleoffset is not None))):
```
takluyver, line 250: While we're looking at this, I think it could be formatted more clearly.

```suggestion
    if (chunks is True) or (chunks is None and any((
            shuffle,
            fletcher32,
            compression,
            (maxshape and not len(external)),
            scaleoffset is not None,
    ))):
```

Other than this, the change LGTM.

tacaswell:
Should there also be work upstream in libhdf5 to reject creating these broken data sets?

With this change can the user still explicitly ask for a broken dataset?

takluyver:
I think HDF5 already rejects the combination of chunked & external storage (from #2394: `ValueError: Unable to synchronously create dataset (external storage not supported with chunked layout)`). But external & expandable (along axis 0) should work.

tacaswell:
> I think HDF5 already rejects the combination of chunked & external storage

But then how did the user end up with a hdf5 header which is invalid?  Or am I confused by what is going on and we using this code on access?

ajelenak:
Indeed, in other cases I tried libhdf5 raised errors.

ajelenak:
That HDF5 file is not invalid. I got educated afterward about expanding external dataset along first dimension without the need for chunking. That's why I made this PR, to allow doing so from h5py.

tacaswell:
> That HDF5 file is not invalid. 

Will this fix our reading as well then?

takluyver:
The code involved here is only called when creating datasets, so I don't believe it will fix reading. @ajelenak spotted a separate issue while investigating #2394, which remains mysterious AFAIK.

tacaswell (approved):

Issue #2353, opened 2023-12-11
Numpy 2.0 compat plan


larsoner:
I read through https://github.com/h5py/h5py/pull/2345#issuecomment-1839031905, https://github.com/numpy/numpy/issues/24300, and related issues and based on my (limited) understanding of the issue, I think that compiling with NumPy 2.0 will generate binaries that work with both NumPy >= 2.0 and NumPy < 2.0. As a downstream user of h5py I've been using this repo to use h5py nightlies (which I assume is managed by h5py maintainers?):

https://7933911d6844c6c53a7d-47bd50c35cd79bd838daf386af554a83.ssl.cf2.rackcdn.com

It has been possible for some months (and thus seemingly reliable by now) to get a 2.0.0.dev NumPy from https://pypi.anaconda.org/scientific-python-nightly-wheels/simple, and many other libraries (scikit-learn, scipy, etc.) now build their nightly wheels against itn. Would it be possible to migrate the `rackcdn.com` h5py binaries to build against NumPy 2.0.0.dev? Then downstream libraries (like mine) could resume testing against NumPy 2.0.0.dev *plus* h5py (and anything that depends on it). Currently we have to choose between testing NumPy 2.0.0.dev *or* h5py because libraries that build against NumPy < 2.0 (like h5py on `rackcdn`) can't be imported when using the NumPy 2.0.0.dev wheels.

tacaswell:
That url is a 404 for me.

I don't think we build and upload nightly anyplace.  As part of CI we build and upload the wheels to azure...but I don't think they automatically get moved to any index.

larsoner:
Oof you're right, that URL is not useful! I thought it was being used but wasn't.

Would it be possible to:

1. Build the nightlies against NumPy 2.0.0.dev if they aren't already, and
2. Connect with the scientific-python-nightly-wheels folks to upload the wheels to https://pypi.anaconda.org/scientific-python-nightly-wheels/simple

? It seems like the leg work for (1) needs to be done at some point in the next two months anyway, and starting it now will give at least a couple of extra weeks given the timeline from the NumPy issue:

> the first RC will be end of Dec with the release expected in early Feb 2024

For (2) I'm happy to look into connecting with the right people to get the wheels uploaded (permissions or whatever). But @tacaswell I assume you might already be familiar given involvement with matplotlib :)

tacaswell:
[...]

PR #2401, opened 2024-04-02
Try building with numpy 2.0 rc1

takluyver:
See discussion on #2353.

### DIFF
--- a/h5py/tests/common.py
+++ b/h5py/tests/common.py
@@ -125,7 +125,7 @@ class TestCase(ut.TestCase):
                 normalized_arr_dtype = repack_fields(arr.dtype)
 
             assert normalized_dset_dtype == normalized_arr_dtype, \
-                "Dtype mismatch (%s vs %s)%s" % (normalized_dset_dtype, normalized_dset_dtype, message)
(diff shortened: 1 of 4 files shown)


takluyver:
Numpy 2.0 requires Python >= 3.9, whereas we're still testing against 3.8. So for now I've made it conditional: it continues using `oldest-supported-numpy` to build on Python 3.8.

We can discuss dropping Python 3.8 as a separate question. The [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) recommendation is to drop it later this month, though I tend to err on the cautious side with old Python versions, not least as we're still using 3.8 in some places at XFEL.

takluyver:
Looks like we've got some issues on Windows, mostly similar to:

```
>           assert normalized_dset_dtype == normalized_arr_dtype, \
                "Dtype mismatch (%s vs %s)%s" % (normalized_dset_dtype, normalized_dset_dtype, message)
E           AssertionError: Dtype mismatch (int64 vs int64)
```

takluyver:
Ah, I see there's a bug in the helper function that makes the message less than helpful.

takluyver:
That seems to be working; I'll merge in another day or so if there are no objections.

aragilar:
Looks good to me, and agree on keeping python 3.8 supported for now.

Issue #2403, opened 2024-04-03
`test_reopen` fails for forthcoming libhdf5-1.14.4


ajelenak:
Just fyi... These lines in the `test_reopen` of the test_file.py cause `OSError` with the code of the future libhdf5 1.14.4 release:
https://github.com/h5py/h5py/blob/078a481371dd6a46e260eeb196a4ef80e32c6a8a/h5py/tests/test_file.py#L926-L927
The change in behavior is likely caused by this (text from [RELEASE.txt](https://github.com/HDFGroup/hdf5/blob/8245d3ab75a3f897e92a8d3a885c3b3898ec1887/release_docs/RELEASE.txt#L687-L712)):
```
    - Fixed H5Fget_access_plist so that it returns the file locking
      settings for a file

      When H5Fget_access_plist (and the internal H5F_get_access_plist)
      is called on a file, the returned File Access Property List has
      the library's default file locking settings rather than any
      settings set for the file. This causes two problems:

        - Opening an HDF5 file through an external link using H5Gopen,
          H5Dopen, etc. with H5P_DEFAULT for the Dataset/Group/etc.
          Access Property List will cause the external file to be opened
          with the library's default file locking settings rather than
          inheriting them from the parent file. This can be surprising
          when a file is opened with file locking disabled, but its
          external files are opened with file locking enabled.

        - An application cannot make use of the H5Pset_elink_fapl
          function to match file locking settings between an external
          file and its parent file without knowing the correct setting
          ahead of time, as calling H5Fget_access_plist on the parent
          file will not return the correct settings.

      This has been fixed by copying a file's file locking settings
      into the newly-created File Access Property List in H5F_get_access_plist.

      This fix partially addresses GitHub issue #4011
```

aragilar:
https://github.com/h5py/h5py/blob/066cad17312a592b26f59f91ef383dbde29bd142/h5py/_hl/files.py#L155-L156 and hence https://github.com/h5py/h5py/blob/066cad17312a592b26f59f91ef383dbde29bd142/h5py/h5p.pyx#L1518-L1525 are where this logic is handled. I'm not sure what `H5Pset_file_locking` is doing with `ignore_when_disabled=False`, as that's what's different to the other cases.

ajelenak:
Should there be a note in the documentation about the change for the `best-effort` option? Someone may have used it and now it will throw an error.

takluyver:
[...]

PR #2405, opened 2024-04-05
Prepare to release 3.11.0

takluyver:
rever steps `version_bump` and `changelog`, followed by manual cleanup.


ajelenak:
When is the release date? Would it be possible to add support for libhdf5-1.14.4, due out later next week?

takluyver:
There isn't a set date, but Numpy 2.0 is coming quite soon, and I imagine there are downstream packages that would like to test with Numpy 2.0rc1 and h5py before Numpy 2 is final. So there is some time pressure.

Do you know how much effort supporting HDF5 1.14.4 is likely to be? Is #2403 likely to be the only issue?

ajelenak:
There's native (libhdf5) support for float16 datatype. I have branch and will make a PR so the changes can be discussed.

takluyver:
Please do! My instinct, though, is that cutting over from our added float16 type might be thorny enough that we want to get 3.11 out the door before tackling that. Let's see what the PR looks like, though.

ajelenak:
See #2406.

takluyver:
3.11 is out on PyPI now.

Unfortunately we don't have Linux aarch64 wheels at the moment. These were built on Travis CI using their 'partner queues', where ARM & IBM provide their respective CPU architectures for open source projects to use. Travis builds haven't been launching for a few months; I asked Travis support about this a few days ago, but so far there's no movement. Maybe this will come back for a future release, maybe we'll move it to another CI provider, or maybe it's just more than we have room to support. :shrug:

lpsinger:
cibuildwheel provides instructions for how to do aarch64 builds under qemu emulation on free tier GitHub Actions.

PR #2416, opened 2024-04-24
TST: avoid a numpy deprecation warning in test_write_empty_vlen

neutrinoceros:
Fixes a couple deprecation warnings caught while running the test suite with `pytest -Werror` against numpy dev.

Namely
```
DeprecationWarning: numpy.core is deprecated and has been renamed to numpy._core. The numpy._core namespace contains private NumPy internals and its use is discouraged, as NumPy internals can change without warning in any release. In practice, most real-world usage of numpy.core is to access functionality in the public NumPy API. If that is the case, use the public NumPy API. If not, you are using NumPy internals. If you would still like to access an internal attribute, use numpy._core.records.
```
xref: https://github.com/numpy/numpy/pull/24634

And
```python-traceback
DeprecationWarning: __array__ implementation doesn't accept a copy keyword, so passing copy=False failed. __array__ must implement 'dtype' and 'copy' keyword arguments.
```
xref: https://github.com/numpy/numpy/pull/25168


h5py/_hl/dataset.py
```diff
@@ -251,10 +255,12 @@ def __len__(self):
         """
         return len(self._dset)
 
-    def __array__(self):
+    def __array__(self, dtype=None, copy=_NUMPY_COPY_IF_NEEDED):
+        if dtype not in (None, object):
+            raise ValueError
```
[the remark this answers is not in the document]
tacaswell: I think a fully fleshed out error as this is a wrapper to get Python strings out of the array!

h5py/tests/test_dtype.py
```diff
@@ -129,9 +129,14 @@ def test_vlen_enum(self):
 
 class TestEmptyVlen(TestCase):
     def test_write_empty_vlen(self):
+        try:
+            from numpy._core import records
```
tacaswell, line 133: Is there a more public place to get this from?

neutrinoceros: thanks. Done !

h5py/_hl/dataset.py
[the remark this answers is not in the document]
neutrinoceros: Thanks, done !

h5py/tests/test_dtype.py
neutrinoceros: thanks for asking, indeed there is ! `np.rec` is the public namespace for this, and is available in numpy 1 and 2 alike !

neutrinoceros:
@tacaswell should I add tests for any/all new (uncovered) paths ?

tacaswell:
Yes please if it is not excessively complicated to make the tests work with both np1 and np2.

neutrinoceros:
Took me longer than I anticipated but I added tests to cover *most* new paths (I couldn't figure out how to use `Dataset.fields` properly so `FieldsWrapper.__array__` still has an uncovered branch).
In the process I learned more about what these various classes do and I changed my approach. I think supporting `copy=False` in numpy 2.0 (which means: "raise an exception if you cannot avoid a copy/allocation) doesn't make sense for these classes since they always return newly allocated memory (as far as I understand), so I'm raising an exception where due.

tacaswell (approved):

h5py/_hl/dataset.py
```diff
@@ -213,10 +214,16 @@ def __len__(self):
         """
         return len(self._dset)
 
-    def __array__(self, dtype=None):
+    def __array__(self, dtype=None, copy=True):
+        if copy not in (True, _NUMPY_COPY_IF_NEEDED):
```
takluyver, line 218: The docs seem to suggest that NumPy 1.x never passes the `copy` parameter to `__array__` (e.g. [here](https://numpy.org/doc/stable/reference/arrays.classes.html#numpy.class.__array__) & [here](https://numpy.org/doc/stable/reference/c-api/array.html#c.PyArray_FromArrayAttr) only document it taking a dtype argument). This fits with a brief test I did on NumPy 1.26.2. If that's right, then I think we can simplify this:

```suggestion
    def __array__(self, dtype=None, copy=True):
        if copy is False:
```

h5py/_hl/dataset.py
```diff
@@ -213,10 +214,16 @@ def __len__(self):
...
+            raise ValueError(
+                f"AstypeWrapper.__array__ received {copy=} "
+                f"but memory allocation cannot be avoided on read"
+            )
+
         data = self[:]
         if dtype is not None:
-            data = data.astype(dtype)
+            return data.astype(dtype, copy=_NUMPY_COPY_IF_NEEDED)
```
takluyver, line 226:
```suggestion
            return data.astype(dtype, copy=False)
```

I'm pretty sure this will work on both NumPy versions. `data` at this point is a newly allocated array from reading the data.

[array.astype() docs for Numpy 2.x](https://numpy.org/devdocs/reference/generated/numpy.ndarray.astype.html#numpy.ndarray.astype) seem to confirm this should still work.

h5py/_hl/dataset.py
```diff
@@ -251,10 +258,19 @@ def __len__(self):
         """
         return len(self._dset)
 
-    def __array__(self):
+    def __array__(self, dtype=None, copy=True):
+        if dtype not in (None, object):
+            raise TypeError(
+                "AsStrWrapper.__array__ doesn't support the dtype argument"
+            )
+        if copy not in (True, _NUMPY_COPY_IF_NEEDED):
```
takluyver, line 266:
```suggestion
        if copy is False:
```

Same as above.

h5py/_hl/dataset.py
```diff
@@ -251,10 +258,19 @@ def __len__(self):
...
+            raise ValueError(
+                f"AsStrWrapper.__array__ received {copy=} "
+                f"but memory allocation cannot be avoided on read"
+            )
         return numpy.array([
             b.decode(self.encoding, self.errors) for b in self._dset
-        ], dtype=object).reshape(self._dset.shape)
+        ], dtype=object, copy=_NUMPY_COPY_IF_NEEDED).reshape(self._dset.shape)
```
takluyver, line 273:
```suggestion
        ], dtype=object).reshape(self._dset.shape)
```

We're passing in a list, so Numpy is always creating a new (object) array here in any case. So we may as well not clutter the code up.

h5py/_hl/dataset.py
```diff
@@ -268,11 +284,17 @@ def __init__(self, dset, prior_dtype, names):
             names = [names]
         self.read_dtype = readtime_dtype(prior_dtype, names)
 
-    def __array__(self, dtype=None):
+    def __array__(self, dtype=None, copy=True):
+        if copy not in (True, _NUMPY_COPY_IF_NEEDED):
```
takluyver, line 288:
```suggestion
        if copy is False:
```

As above.

h5py/_hl/dataset.py
```diff
@@ -268,11 +284,17 @@ def __init__(self, dset, prior_dtype, names):
...
+            raise ValueError(
+                f"FieldsWrapper.__array__ received {copy=} "
+                f"but memory allocation cannot be avoided on read"
+            )
         data = self[:]
         if dtype is not None:
-            data = data.astype(dtype)
-        return data
+            return data.astype(dtype, copy=_NUMPY_COPY_IF_NEEDED)
```
takluyver, line 295:
```suggestion
            return data.astype(dtype, copy=False)
```

h5py/_hl/dataset.py
```diff
@@ -1049,11 +1071,16 @@ def write_direct(self, source, source_sel=None, dest_sel=None):
                 self.id.write(mspace, fspace, source, dxpl=self._dxpl)
 
     @with_phil
-    def __array__(self, dtype=None):
+    def __array__(self, dtype=None, copy=True):
         """ Create a Numpy array containing the whole dataset.  DON'T THINK
         THIS MEANS DATASETS ARE INTERCHANGEABLE WITH ARRAYS.  For one thing,
         you have to read the whole dataset every time this method is called.
         """
+        if copy not in (True, _NUMPY_COPY_IF_NEEDED):
```
takluyver, line 1079:
```suggestion
        if copy is False:
```

takluyver (commented):
Thanks! I think the NumPy stuff can be simplified a bit, I've put some comments about it inline.

h5py/_hl/dataset.py
neutrinoceros: good call. Indeed the `copy` argument in `__array__` is new in numpy 2.0 !

h5py/_hl/dataset.py
```diff
@@ -31,6 +31,7 @@
 _LEGACY_GZIP_COMPRESSION_VALS = frozenset(range(10))
 MPI = h5.get_config().mpi
 
+_NUMPY_COPY_IF_NEEDED = False if numpy.__version__.startswith("1.") else None
```
takluyver, line 34: I think this line is no longer needed.

```suggestion
```

takluyver:
Thanks!

neutrinoceros:
@takluyver these deprecation warnings are starting to affect downstream CI (I've heard of astropy, but there may be others). My assumption is that we'll release h5py 3.12 sometimes in the next 2 months for Python 3.13 wheels anyway, and I don't think that's a problem if not, but could you confirm ? Thank you !

takluyver:
Yup, I think we can do a new release fairly soon. I'd be tempted to wait for Python 3.13, but we can also go before that if we need to.

neutrinoceros:
with the rc1 out we should be able to ship wheels for it soon. This is happening in #2470

takluyver:
Ah, yes, that's right. But possibly we should wait for numpy to make wheels first?

neutrinoceros:
We don't *have* to. In fact matplotlib 3.9.2 already has cp313 wheels on PyPI, even though neither numpy or contourpy does, at the moment.
That said it's absolutely reasonable to wait for numpy if you'd like.

takluyver:
I doubt there's much pressure until numpy has wheels ready, and it gives us a chance to catch any random errors from numpy's wheels being built differently from how numpy gets built in our CI, so I probably would wait for that.

Python 3.13 also includes the option to disable the GIL, doesn't it? Do you know if we need to do anything around that? Using h5py you don't get much benefit from disabling the GIL, because HDF5 has it's own big lock, but I imagine someone will want to try.

neutrinoceros:
Yes there's a couple small things we can do to at least not get in the way of downstream testers. I've been learning about this topic lately and I should be able to expedite such changes in the coming days.

takluyver:
Thanks!

Issue #2417, opened 2024-05-01
Unable to build from source with parallel support on Python 3.12


nima3333:
Hello,

When trying to build on Python 3.12 with Parallel support, I got the error following error:

```
$ export HDF5_DIR="/usr/local"
$ export HDF5_MPI="ON"
$ export CC=/usr/bin/mpicc
$ python -m pip install --no-binary=h5py h5py
Collecting h5py
  Using cached h5py-3.11.0.tar.gz (406 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Installing backend dependencies ... error
  error: subprocess-exited-with-error

  × pip subprocess to install backend dependencies did not run successfully.
  │ exit code: 1
  ╰─> [1045 lines of output]
      Collecting wheel
        Using cached wheel-0.43.0-py3-none-any.whl.metadata (2.2 kB)
      Collecting mpi4py==3.1.4
        Using cached mpi4py-3.1.4.tar.gz (2.5 MB)
        Installing build dependencies: started
        Installing build dependencies: finished with status 'done'
        Getting requirements to build wheel: started
        Getting requirements to build wheel: finished with status 'done'
        Preparing metadata (pyproject.toml): started
        Preparing metadata (pyproject.toml): finished with status 'done'
      Using cached wheel-0.43.0-py3-none-any.whl (65 kB)
      Building wheels for collected packages: mpi4py
        Building wheel for mpi4py (pyproject.toml): started
        Building wheel for mpi4py (pyproject.toml): finished with status 'error'
        error: subprocess-exited-with-error

        × Building wheel for mpi4py (pyproject.toml) did not run successfully.
        │ exit code: 1
        ╰─> [1020 lines of output]
            running bdist_wheel
            running build
            running build_src
            running build_py
            creating build
            creating build/lib.linux-x86_64-cpython-312
            creating build/lib.linux-x86_64-cpython-312/mpi4py
            copying src/mpi4py/__init__.py -> build/lib.linux-x86_64-cpython-312/mpi4py
            copying src/mpi4py/__main__.py -> build/lib.linux-x86_64-cpython-312/mpi4py
            copying src/mpi4py/run.py -> build/lib.linux-x86_64-cpython-312/mpi4py
            copying src/mpi4py/bench.py -> build/lib.linux-x86_64-cpython-312/mpi4py
            creating build/lib.linux-x86_64-cpython-312/mpi4py/futures
[...]

PR #2418, opened 2024-05-01
Python 3.12 requires mpi4py>=3.1.5 in setup.py

nima3333:
See issue #2417.
As for Python 3.11, I added the last version of mpi4py as a requirement.

By the way, I had to bump the `check-manifest` rev in pre-commit (`.pre-commit-config.yaml`) to be able to run `tox -e pre-commit` with Python 3.12. (0.39 -> 0.49)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,7 @@ if setup_configure.mpi_enabled():
     # incompatible with newer setuptools.
     RUN_REQUIRES.append('mpi4py >=3.1.1')
     SETUP_REQUIRES.append("mpi4py ==3.1.1; python_version<'3.11'")
-    SETUP_REQUIRES.append("mpi4py ==3.1.4; python_version=='3.11'")
+    SETUP_REQUIRES.append("mpi4py ==3.1.4; python_version=='3.11.*'")
     SETUP_REQUIRES.append("mpi4py ==3.1.6; python_version>='3.12'")
 
 # Set the environment variable H5PY_SETUP_REQUIRES=0 if we need to skip
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,8 @@ if setup_configure.mpi_enabled():
     # incompatible with newer setuptools.
     RUN_REQUIRES.append('mpi4py >=3.1.1')
     SETUP_REQUIRES.append("mpi4py ==3.1.1; python_version<'3.11'")
-    SETUP_REQUIRES.append("mpi4py ==3.1.4; python_version>='3.11'")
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -47,7 +47,8 @@
     # incompatible with newer setuptools.
     RUN_REQUIRES.append('mpi4py >=3.1.1')
     SETUP_REQUIRES.append("mpi4py ==3.1.1; python_version<'3.11'")
-    SETUP_REQUIRES.append("mpi4py ==3.1.4; python_version>='3.11'")
+    SETUP_REQUIRES.append("mpi4py ==3.1.4; python_version=='3.11'")
```
tacaswell, line 50: Should this be
```suggestion
    SETUP_REQUIRES.append("mpi4py ==3.1.4; python_version=='3.11.*'")
```

to catch all versions of py3.11?  I am always a bit fuzzy on the grammar of these version limits.

nima3333: Yes, you are right

tacaswell (approved):

tacaswell:
@takluyver Would like a second set of eyes on the pinning.

setup.py
```diff
@@ -47,7 +47,8 @@
     # incompatible with newer setuptools.
     RUN_REQUIRES.append('mpi4py >=3.1.1')
     SETUP_REQUIRES.append("mpi4py ==3.1.1; python_version<'3.11'")
-    SETUP_REQUIRES.append("mpi4py ==3.1.4; python_version>='3.11'")
+    SETUP_REQUIRES.append("mpi4py ==3.1.4; python_version=='3.11.*'")
+    SETUP_REQUIRES.append("mpi4py ==3.1.6; python_version>='3.12'")
```
takluyver, line 51: This listing is based on the idea that when building (with `SETUP_REQUIRES`), we want the oldest compatible version of packages, and the buiilt package will then be compatible with newer ones. I.e. if building with 3.1.5 works for Python 3.12, go for that version.

IDK if doing this is actually useful for mpi4py, though. @aragilar might have a better idea.

aragilar: I'm not familiar enough with the changes made to mpi4py to be sure (from what I can see in the changelog, the changes were mostly minor), but my feeling is locking in the setup.py wouldn't be that useful, given you probably want to install mpi4py separately to check which MPI version you're using (rather than pulling in a wheel), and use `--no-build-isolation` so you're matching the build environment to the run environment. I think leaving this as is fine for now, given mpi4py 4 looks like it's going to come this year (based on its changelog), and it's not clear if or how that will affect us.

takluyver: Thanks @aragilar ! I'll merge this as is, then.

Issue #2419, opened 2024-05-02
Segfault after running test suite with hdf5 1.14.4.2


antonio-rojas:
After running the test suite against hdf5 1.14.4.2, we get the attached segfault. All tests pass correctly, the segfault happens *after* the test suite has finished running.

 * Operating System: Arch Linux x86_64
 * Python version: 3.12.3
 * Where Python was acquired: system
 * h5py version: 3.11.0
 * HDF5 version: 1.14.4.2
 * The full traceback/stack trace shown (if it appears)
 ```
#0  0x00007613d3646da8 in H5T__unlock_cb (_dt=0x5b1cc4fe5920, id=216172782113784908, _udata=0x7ffee4b314ec)
    at /usr/src/debug/hdf5/hdf5-1.14.4-2/src/H5T.c:1648
#1  0x00007613d357db2a in H5I__iterate_cb (_key=0x0, _udata=<synthetic pointer>, _item=0x5b1cc51c8270)
    at /usr/src/debug/hdf5/hdf5-1.14.4-2/src/H5Iint.c:276
#2  H5I_iterate (type=H5I_DATATYPE, func=0x7613d3646da0 <H5T__unlock_cb>, udata=0x7ffee4b314ec, app_ref=false)
    at /usr/src/debug/hdf5/hdf5-1.14.4-2/src/H5Iint.c:1584
#3  0x00007613d3647195 in H5T_top_term_package () at /usr/src/debug/hdf5/hdf5-1.14.4-2/src/H5T.c:1703
#4  0x00007613d3446ca5 in H5_term_library () at /usr/src/debug/hdf5/hdf5-1.14.4-2/src/H5.c:456
#5  0x00007613d5a5cb36 in __run_exit_handlers (status=status@entry=0, listp=0x7613d5bf6680 <__exit_funcs>, run_list_atexit=run_list_atexit@entry=true, 
    run_dtors=run_dtors@entry=true) at exit.c:108
#6  0x00007613d5a5cc80 in __GI_exit (status=status@entry=0) at exit.c:138
#7  0x00007613d5e7f49e in Py_Exit (sts=0) at Python/pylifecycle.c:3060
#8  0x00007613d5e74dfb in handle_system_exit () at Python/pythonrun.c:756
#9  0x00007613d5e74747 in _PyErr_PrintEx (tstate=0x7613d6222ae8 <_PyRuntime+459656>, set_sys_last_vars=1) at Python/pythonrun.c:765
[...]

PR #2422, opened 2024-05-05
Lock native float16 dtype only if available

ajelenak:
I've created this PR to start discussion how to solve the #2419 issue. Not sure if the current fix is the right approach but locking the `H5T_NATIVE_FLOAT16` datatype now happens only if it's not equal to `H5I_INVALID_HID`.


jhendersonHDF:
Thanks for this proposed fix @ajelenak! I have a branch I'm working on that fixes the crashing issues in #2419, but I'm still hitting a corrupted heap/use after free issue that I suspect comes from some interaction between h5py's custom datatype conversion methods and the changes to type conversion in 1.14.4 that avoids unnecessary copying of datatypes. It appears the library is ending up with multiple IDs pointing to the same datatype pointer, leading to issues when the library is terminating and closing IDs. Still tracking this down.

takluyver:
Is only `H5T_NATIVE_FLOAT16` affected by this, or do `H5T_IEEE_F16BE` & `...LE` need similar checks?

jhendersonHDF:
> Is only `H5T_NATIVE_FLOAT16` affected by this, or do `H5T_IEEE_F16BE` & `...LE` need similar checks?

The `H5T_IEEE_F16BE` and `H5T_IEEE_F16LE` macros will always map to a valid HDF5 datatype. Only `H5T_NATIVE_FLOAT16` could be invalid, depending on the build.

ajelenak:
> I haven't checked whether `H5Tget_class(H5I_INVALID_HID)` would throw an error or not.

Update:
```
import h5py
a = h5py.h5t.TypeID(-1)
a.get_class()
```
Produces an exception:

    ValueError: Not a datatype (not a datatype)

aragilar:
This seems like a reasonable fix to me, and if there are any other types that need this conditional check, then they can be added in future PRs. Thanks @ajelenak.

PR #2423, opened 2024-05-06
Document in strings.rst usage of special_dtype(vlen=str)

doronbehar:
I have been using the example I documented here for a while now. Felt it was missing. I had to learn this from the question [here](https://stackoverflow.com/questions/68500454/can-i-use-h5py-to-write-strings-to-an-hdf5-file-in-one-line-rather-than-looping).

### DIFF
--- a/docs/strings.rst
+++ b/docs/strings.rst
@@ -24,9 +24,19 @@ Storing strings
 ---------------
 
 When creating a new dataset or attribute, Python ``str`` or ``bytes`` objects
-will be treated as variable-length strings, marked as UTF-8 and ASCII respectively.
-Numpy bytes arrays (``'S'`` dtypes) make fixed-length strings.
-You can use :func:`.string_dtype` to explicitly specify any HDF5 string datatype.
+should be treated as variable-length strings, marked as UTF-8 and ASCII
+respectively.
+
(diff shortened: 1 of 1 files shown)


aragilar:
Rather than `h5py.special_dtype(vlen=str)` (which is oldish and somewhat implicit about what it means, though I don't think we'll drop support for it), you should use `h5py.string_dtype(encoding='utf-8', length=None)` (or, since those are the default values `h5py.string_dtype()`).

doronbehar:
Hmm I wonder how I missed that back then... The docs indeed mention `[string_dtype](https://docs.h5py.org/en/stable/special.html#h5py.string_dtype)`, but they don't tell you to use that for datasets, and only gives examples for attributes. I pushed something.

takluyver:
Thanks! I think this could benefit from an example. But it's a fairly confusing topic, because of all the different possible string types, and I think it needs careful presentation to avoid muddying the waters. So I hope you won't mind if I open a new PR reworking this example to show the different possibilities.

takluyver:
I've tried to do that in #2424.

takluyver:
I'm going to close this as I've merged #2424. Thanks!

doronbehar:
Thank you for your help.

PR #2424, opened 2024-05-07
Add examples of different ways of creating string datasets

takluyver:
Building on #2423, trying to illustrate the different possible ways to create string datasets.

### DIFF
--- a/docs/strings.rst
+++ b/docs/strings.rst
@@ -28,6 +28,24 @@ will be treated as variable-length strings, marked as UTF-8 and ASCII respective
 Numpy bytes arrays (``'S'`` dtypes) make fixed-length strings.
 You can use :func:`.string_dtype` to explicitly specify any HDF5 string datatype.
 
+::
+
+    string_data = ["varying", "sizes", "of", "strings"]
+
+    # Variable length strings (implicit)
+    f['vlen_strings1'] = string_data
+
+    # Variable length strings (explicit)
+    ds = f.create_dataset('vlen_strings2', shape=4, dtype=h5py.string_dtype())
+    ds[:] = string_data
+
+    # Fixed length strings (implicit) - longer strings are truncated
+    f['fixed_strings1'] = np.array(string_data, dtype='S6')
+
+    # Fixed length strings (explicit) - longer strings are truncated
+    ds = f.create_dataset('fixed_strings2', shape=4, dtype=h5py.string_dtype(length=6))
+    ds[:] = string_data
+
 When writing data to an existing dataset or attribute, data passed as bytes is
 written without checking the encoding. Data passed as Python ``str`` objects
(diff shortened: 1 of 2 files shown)


docs/strings.rst
```diff
@@ -28,6 +28,24 @@ will be treated as variable-length strings, marked as UTF-8 and ASCII respective
 Numpy bytes arrays (``'S'`` dtypes) make fixed-length strings.
 You can use :func:`.string_dtype` to explicitly specify any HDF5 string datatype.
 
+::
+
+    string_data = ["varying", "sizes", "of", "strings"]
+
+    # Variable length strings (implicit)
+    f['vlen_strings1'] = string_data
```
doronbehar, line 36: This `f` is supposed to be the database? What is supposed to happen when you are doing it implicitly?

docs/strings.rst
```diff
@@ -28,6 +28,24 @@ will be treated as variable-length strings, marked as UTF-8 and ASCII respective
...
+
+    # Variable length strings (explicit)
+    ds = f.create_dataset('vlen_strings2', shape=4, dtype=h5py.string_dtype())
```
doronbehar, line 39: It'd be nice if there were some examples with `data=` keyword argument - those are the ones that got me confused.

docs/strings.rst
```diff
@@ -28,6 +28,24 @@ will be treated as variable-length strings, marked as UTF-8 and ASCII respective
 Numpy bytes arrays (``'S'`` dtypes) make fixed-length strings.
 You can use :func:`.string_dtype` to explicitly specify any HDF5 string datatype.
 
+::
```
doronbehar, line 31:
```suggestion
You can use :func:`.string_dtype` to explicitly specify any HDF5 string datatype,
as shown in the examples below::
```

docs/strings.rst
takluyver: `f` for file - this is a convention across the h5py docs, e.g. look further down this page, or at the [dataset page](https://docs.h5py.org/en/stable/high/dataset.html). We don't really use the word database at all; HDF5 also talks about files rather than databases.

'Implicit' here just means you're not telling h5py a dtype, so it's guessing based on the object you give it.

docs/strings.rst
takluyver: Passing `data=` should be equivalent to the implicit case (`f[name] =`) if you don't pass `dtype=`, or to the explicit case if you do. I'm showing the explicit dtype case as two lines to illustrate that it lets you create the dataset before you have all the data.

I'd rather not make this example longer by showing more possible ways to do the same thing. The explanation at [creating datasets](https://docs.h5py.org/en/stable/high/dataset.html#creating-datasets) could probably be improved as well.

PR #2427, opened 2024-05-16
TST: make sure file descriptors are closed before tests return

neutrinoceros:
Continuing to scratch my own itch for running the test suite with warnings promoted to errors (`pytest -Werror`), these are the only two tests I see failing locally (but there could be more, some tests are simply skipped on my system), and they were simple enough to fix.

I opened a couple similar PRs in the past, fixing mostly deprecation warnings from numpy that I discovered in CI of packages that depend on h5py and promote warnings as errors. Would you consider doing the same thing directly in h5py's CI so you don't need reports from downstream in the future ?


tacaswell:
Thank you for this PR.

I am in favor of promoting warnings to errors in our CI and would merge that PR!

PR #2428, opened 2024-05-16
TST: promote most warnings to errors in CI

neutrinoceros:
A continuation of #2427
Let's try promoting all warnings to errors and see what happens.

### DIFF
--- a/h5py/tests/test_file2.py
+++ b/h5py/tests/test_file2.py
@@ -172,6 +172,11 @@ class TestFileObj(TestCase):
         finally:
             os.remove(fname)
 
+    @pytest.mark.filterwarnings(
+        # on Windows, a resource warning may be emitted
+        # when this test returns
+        "ignore:unclosed file:ResourceWarning"
+    )
     def test_TemporaryFile(self):
         # in this test, we check explicitly that temp file gets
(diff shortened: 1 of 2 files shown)


h5py/tests/test_file2.py
```diff
@@ -172,6 +172,11 @@ def test_file(self):
         finally:
             os.remove(fname)
 
+    @pytest.mark.filterwarnings(
+        # on Windows, a resource warning may be emitted
+        # when this test returns
+        "ignore:unclosed file:ResourceWarning"
+    )
```
neutrinoceros, line 179: this isn't fully satisfactory but I couldn't find a way to rewrite this particular test any differently, and the warning only seems to happen on Windows + Python <= 3.11, so just being conservative and continuing to ignore it seemed like the simplest option. Happy to revisit if desired.

takluyver: I think you're right that there's no easy way to avoid this - the test is specifically looking for behaviour that happens when the temporary file object is cleaned up because its reference count has gone to 0. This is also what triggers the warning - the file is closed by refcounting rather than explicitly with `.close()` or a `with` block.

takluyver:
Thanks @neutrinoceros !

PR #2433, opened 2024-05-27
Fix H5F_close_degree_t and update default link access properties

### LINKED ISSUE #2431: External link files are opened with file locking enabled even if main file is opened with locking disabled
<details><summary>Summary of the h5py configuration</summary>

```
h5py    3.11.0  # current master
HDF5    1.14.4
Python  3.11.2 (main, Mar 13 2023, 12:18:29) [GCC 12.2.0]
sys.platform    linux
numpy   2.0.0rc2
cython (built with) 3.0.10
numpy (built against) 2.0.0rc2
HDF5 (built against) 1.14.4
```

</details>

We ran into an issue when accessing external links with HDF5's file locking disabled.
The target file of the external link is always opened with file locking enabled even if the main file was opened with file locking disabled, while one would expect to use consistent file access properties.

This has been first raised to the HDFGroup (see: https://forum.hdfgroup.org/t/external-link-and-file-locking-disabled-issue/12012) and fixed in libhdf5 v1.14.4.

Since `h5py` uses it's own default link access property list `h5py.base.default_lapl` which defines an external link file access property list (`set_elink_fapl`), libhdf5 default is ignored:

https://github.com/h5py/h5py/blob/902e32c3fdc06f1e0feaa29717df83e92953dc2a/h5py/_hl/base.py#L129-L135

<details><summary>Script to reproduce the issue</summary>

```python
import multiprocessing
import time
import h5py

def create_files():
    with h5py.File("linked.h5", "w") as h5f:
        h5f["data"] = (1, 2, 3)

    with h5py.File("main.h5", "w") as h5f:
        h5f["data"] = (3, 2, 1)
        h5f["link"] = h5py.ExternalLink("linked.h5", "/data")


def read():
    with h5py.File("main.h5", "r", locking=False) as h5f:
        print("data:", h5f["/data"][()])
        print("link:", h5f["/link"][()])


if __name__ == "__main__":
    print("* Create files")
    create_files()
    print("* Open files in append mode")
    main = h5py.File("main.h5", "a")
    linked = h5py.File("linked.h5", "a")
    print("* Read from subprocess")
    context = multiprocessing.get_context("spawn")
    process = context.Process(target=read)
    process.start()
    time.sleep(1)
```

</details>

Also, while looking at this issue, I found an issue in the values of `H5F_close_degree_t` enum which differs from libhdf5 ones:

- libhdf5's [H5Fpublic.h](https://github.com/HDFGroup/hdf5/blob/d3f5e07b1eea230c7ab340ac9fe24dffa8406355/src/H5Fpublic.h#L114-L119) and [version from 23 years ago](https://github.com/HDFGroup/hdf5/blob/d28fd08f4a9593bd8bb51ebc17b8b6555b3885cd/src/H5Fpublic.h#L72-L85):

```
typedef enum H5F_close_degree_t {
    H5F_CLOSE_DEFAULT = 0, /**< Use the degree pre-defined by underlying VFD */
    H5F_CLOSE_WEAK    = 1, /**< File closes only after all opened objects are closed */
    H5F_CLOSE_SEMI    = 2, /**< If no opened objects, file is closed; otherwise, file close fails */
    H5F_CLOSE_STRONG  = 3  /**< If there are opened objects, close them first, then close file */
} H5F_close_degree_t;
```

- h5py: https://github.com/h5py/h5py/blob/902e32c3fdc06f1e0feaa29717df83e92953dc2a/h5py/api_types_hdf5.pxd#L117-L121

I'll prepare a PR to fix the enum and to propose ways to solve this issue for discussion.

t20100:
This PR proposes 2 fixes (I can split it in 2 PRs if you prefer):
- Update the values of the `H5F_close_degree_t` enum to be the same as in [H5Fpublic.h](https://github.com/HDFGroup/hdf5/blob/d3f5e07b1eea230c7ab340ac9fe24dffa8406355/src/H5Fpublic.h#L114-L119)
- Update the default "link access property list" used by `h5py` to be the default one and add a test of external link access with file locking disabled which is impaired by the fact that `h5py` overrides the default "link access property list".

I'm not sure what the side effect of the default "link access property list" change can be!

The fact that the following code has been using the wrong value for `CLOSE_STRONG` and was actually using `CLOSE_SEMI` for over a decade without notice makes me think it should have little impact:

https://github.com/h5py/h5py/blob/902e32c3fdc06f1e0feaa29717df83e92953dc2a/h5py/_hl/base.py#L129-L135

Also, looking at the hdf5 documentation, the "link access property list" is actually not always used:

- [H5Lcreate_external documentation](https://docs.hdfgroup.org/hdf5/v1_14/v1_14_4/group___h5_l.html#ga15dfaeb9b1c0b3136533cb97ee45e683):

  > lapl_id is the link access property list used in traversing the new link. Note that an external file opened by the traversal of an external link is always opened with the weak file close degree property setting, H5F_CLOSE_WEAK (see H5Pset_fclose_degree()); any file close degree property setting in lapl_id is ignored.
  > ...
  > Attention
  >     A file close degree property setting (H5Pset_fclose_degree()) in the external link file access property list or in the external link callback function will be ignored. A file opened by means of traversing an external link is always opened with the weak file close degree property setting, H5F_CLOSE_WEAK .

- [H5Oget_info_by_idx*() H5Oget_info_by_name*()](https://docs.hdfgroup.org/hdf5/v1_14/v1_14_4/group___h5_o.html#title23)

  > The link access property list, lapl_id, is not currently used; it should be passed in as H5P_DEFAULT

So this PR proposes to let libhdf5 use its default (and so disable file locking when accessing external links with libhdf5 >=v1.14.4).

There is a more conservative way to fix the external link+file locking disabled issue (#2431).
It is to copy the file locking properties from the current file into the "link access property list":

```python
class Group(HLObject, MutableMappingHDF5):
   ...
    @property
    def _lapl(self):
        """ Fetch the link access property list appropriate for this object
        """
        lapl = h5p.create(h5p.LINK_ACCESS)
        fapl = h5p.create(h5p.FILE_ACCESS)
        fapl.set_fclose_degree(h5f.CLOSE_STRONG)
        group_fapl = self.file.id.get_access_plist()
        use_file_locking, ignore_when_disabled = group_fapl.get_file_locking()
        fapl.set_file_locking(use_file_locking, ignore_when_disabled)
        lapl.set_elink_fapl(fapl)
        return lapl
```

closes #2431


t20100:
I'm also worried about the side effects....

My guess is that within `h5py` there is no object to close when closing a file (is that right?) otherwise with the `CLOSE_SEMI`  mode that was used before this PR the file closing would have failed... so the "close_degree" doesn't seems to matter here.

> H5F_CLOSE_SEMI    = 2, /**< If no opened objects, file is closed; otherwise, file close fails */

Maybe advices from the hdgroup would be valuable here.

Issue #2408, opened 2024-04-10
Arm wheels missing from latest release


westonpace:
These files are present in the [3.10 release](https://pypi.org/project/h5py/3.10.0/#files) but the equivalent files are missing from the [3.11 release](https://pypi.org/project/h5py/3.11.0/#files):

```
h5py-3.10.0-cp311-cp311-manylinux_2_17_aarch64.manylinux2014_aarch64.whl
h5py-3.10.0-cp310-cp310-manylinux_2_17_aarch64.manylinux2014_aarch64.whl
h5py-3.10.0-cp39-cp39-manylinux_2_17_aarch64.manylinux2014_aarch64.whl
h5py-3.10.0-cp38-cp38-manylinux_2_17_aarch64.manylinux2014_aarch64.whl
```

westonpace:
Further context: https://github.com/h5py/h5py/pull/2405#issuecomment-2047248192

glenn-jocher:
We've seen the same problem at https://github.com/ultralytics/ultralytics. Missing aarch64 wheels have crashed the Arm64 Docker deployments that install tensorflow (with h5py as a required tensorflow dependency). 

We've temporarily resolved this by avoiding h5py==3.11.0 installs on aarch64 in https://github.com/ultralytics/ultralytics/pull/9956

lpsinger:
cibuildwheel provides [instructions for how to do aarch64 builds under qemu emulation on the free tier of GitHub Actions](https://cibuildwheel.pypa.io/en/stable/faq/#emulation).

tacaswell:
I would be happy to review a PR moving us to cibuildwheel on GHA.

This issue highlights some of the inherent problems in the way wheels get built (see https://pypackaging-native.github.io for more details).

lpsinger:
> I would be happy to review a PR moving us to cibuildwheel on GHA.

I would find that a little daunting for h5py because of the many different old CI workflows this would have to replace: Travis, Azure Pipelines, and the existing nightly GH Actions wheel job.

But all you really have to do is add this to your `pyproject.toml` file, and cibuildwheel will do the rest for you:

```toml
[tool.cibuildwheel.linux]
# Build wheels for aarch64 under emulation.
archs = ["auto", "aarch64"]
```

aragilar:
[...]

PR #2434, opened 2024-05-27
TST: Test aarch64 wheel building using GHA

larsoner:
Reading https://github.com/h5py/h5py/issues/2408, I *think* this might be a low-effort way to see if `aarch64` wheels could be built using GHA emulation. I guess we'll see if https://github.com/pypa/cibuildwheel/issues/1771 is a problem.

If it works I think in theory they could be *tested* using emulation as well, especially if the `cibuildwheel` action were used. Or maybe the build will already test... no sure 🤷

### DIFF
--- a/.github/workflows/nightlies.yml
+++ b/.github/workflows/nightlies.yml
@@ -34,6 +34,12 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: "3.11"
+      # For aarch64 support
+      # https://cibuildwheel.pypa.io/en/stable/faq/#emulation
+      - uses: docker/setup-qemu-action@v3
+        with:
+          platforms: all
+        if: runner.os == 'Linux'
       - name: Install cibuildwheel
         run: python -m pip install cibuildwheel==2.16.5
       - name: Build wheels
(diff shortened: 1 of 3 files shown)


larsoner:
Looks like it might actually work. However, the native builds only take ~100s, the emulated ones take ~1200s. So probably don't want to build aarch64 on every PR, and maybe not on the nightlies. Happy to close this or try to adapt it into a solution depending on what people want -- mostly I just wanted to see if it worked as a potential solution!

hoechenberger:
It could be a solution at least for releases, though!

tacaswell:
I agree too long for every PR, but in the nightlies and/or on merge to main would be fine at these times.

hoechenberger:
This would be a great addition indeed!

larsoner:
Okay for `nightlies.yml` I limited the set it produces on PRs to 3.9/3.12 on x86_64 only:
```
Building limited wheels for PR with commit message: TST: Ping
```
On PRs if there's a `[aarch64]` in the commit message, then it will try to build 3.9/3.10/3.11/3.12 x x86_64/aarch64. And on anything other than PRs (push, workflow_dispatch, or schedule) it will build that complete set of variants.

Note that this only touches the `nightlies.yml`. I went to make the equivalent change to Azure only to look at the config options and note that it's not supported on Azure in [the cibuildwheel support table](https://cibuildwheel.pypa.io/en/stable/#usage). So something more will need to be done to enable wheel support on releases.

cbrnr:
@larsoner I'm just curious, is there a particular reason why you are not using the cibuildwheel action (and instead "manually" run cibuildwheel)? Would the action not simplify a few things?

larsoner:
I don't totally remember why -- I do usually use the action in my projects 🤔. I think it was probably to more closely match the existing Azure-based wheel infrastructure in this repo. That may maintainers here can (in theory) more easily make changes both places since the infrastructure is more similar.

tacaswell:
I suspect our wheel building infrastructure has a very heavy path dependence (back to the old macpython projects on travis).  I think we were an early adopter of "build all the wheels in the cloud" and got to a local minimum of "working" before the current round of GHA work.

Speaking only for myself, I do not see any reason not to simplify the wheel building and move it all to GHA other than we have a (mostly) working solution already.

njzjz:
Note that arm-based Linux runner will be GA at the end of this year: https://github.blog/changelog/2024-06-03-actions-arm-based-linux-and-windows-runners-are-now-in-public-beta/

larsoner:
Okay it turned out to be fairly easy to fold this into #2444 so I'll close this. Can reopen if #2444 stalls!

Issue #2430, opened 2024-05-25
Difficulty in Determining Format for Datatypes Associated with Compound Dataset


madduxc:
Documentation - not a bug.

I am using Python version 3.9.7 and h5py version 3.7.0 on Windows 10 OS.  These were acquired through a company Artifactory account.  I have been working with HDF files as a primary source of data storage for the last two years, converting data from csv, numpy, pandas, and Nastran HDF outputs, primarily generating mapped compound datasets.   In building converters for numpy and pandas data, I had some difficulty in finding documentation on the required format of compound datatypes.

I believe that new users could benefit from more thorough documentation of the format required to generate compound datasets manually, especially for datatypes not native to HDF (like boolean) or from numpy or pandas conversion. Given the amount and qualify of the documentation present, I believe the community would benefit from a brief description in the datsets.rst and perhaps an example file or two as a way for the new user to get started.  

I would be willing to contribute to this effort and eager to receive and incorporate input on various methods, as my experience is still limited.

aragilar:
This seems reasonable to me. We use numpy's dtype to handle things at the high level, which also doesn't really explain compound dtypes (at least from my recollection, I haven't looked at that part of the numpy docs for a while). Feel free to create a PR with some suggestions (and/or link to appropriate spots in the numpy docs).

madduxc:
I should be ready for a PR tonight - enough to get things started at least.  Thanks.

PR #2437, opened 2024-06-01
Documentation for Compound Datasets

madduxc:
Addition of a section to Dataset.rst describing the steps taken to manually generate a compound dataset.

Related to Issue #2430

### DIFF
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -111,35 +111,60 @@ safe to use with very large target selections.  It is supported for the above
 Compound Datasets
 ~~~~~~~~~~~~~~~~~
 
-Compound datasets can be used to store different types of data in separate columns.  This practice can be valuable when used in conjunction with attributes to generate linked databases.  Keys can be assigned to string columns, with sets of mixed data (in columns) following. 
-As with simple datasets, new compound datasets are created using either Group.create_dataset() or Group.require_dataset().  However, in compound datasets, dtype assignment is not optional.  Initializing the compound dataset follows the form::
+In some cases it may desirable to stored mixed types of data in a single dataset.  Compound 
+datasets can be used for this purpose, provided that all of the data in a single column 
+is homogenous.  This practice can be valuable when used in conjunction with attributes 
+and mapping to generate linked datsets simulating a database structure.  
+
+As with simple datasets, new compound datasets are created using either Group.create_dataset() 
+or Group.require_dataset().  However, in compound datasets, dtype and shape assignment are not 
+optional, and as stated above it is necessary to separate field names from the slices.  
+
+Initializing the compound dataset follows the form::
 
     >>> ds = fn.require_dataset(ds_path, data=HDFData, dtype=DataType, shape=rShape)
 
-where dtype is a list of tuples containing the header, type, and array size (for lists) for each column in the dataset.  As an example, in order to create a compound dataset containing a string, integer, floating point, and list of integers. 
+where dtype is a list of tuples containing the field name, type, and array size (for lists) for 
+each column in the dataset.  As an example, in order to create a compound dataset containing a 
+string, integer, list of booleans, and floating point data::
 
 1)	Create the data from the your data source.  Data is entered row-by-row as a list of tuples::
 
     >>> HDFData = [
-    >>>		(‘loc1’, 1, [1, 2], 1.005),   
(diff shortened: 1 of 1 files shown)


Delengowski:
Is it worth noting the nesting of dtypes? It's possible for a field to be an array. I think you may even be able to nest compound types, I think.

miccoli:
If I got it right, compound types are mapped to numpy structured dtypes: maybe adding an explicit cross reference to the relevant docs, <https://numpy.org/doc/stable/user/basics.rec.html#structured-datatypes>, could be useful.

docs/high/dataset.rst
```diff
@@ -96,14 +96,70 @@ To retrieve the contents of a `scalar` dataset, you can use the same
...
+is homogenous.  This practice can be valuable when used in conjunction with attributes
+and mapping to generate linked datsets simulating a database structure.
+
+As with simple datasets, new compound datasets are created using either Group.create_dataset()
+or Group.require_dataset().  However, in compound datasets, dtype and shape assignment are not
+optional, and as stated above it is necessary to separate field names from the slices.
+
+Initializing the compound dataset follows the form::
+
+    >>> ds = fn.require_dataset(ds_path, data=HDFData, dtype=DataType, shape=rShape)
```
takluyver, line 121: For examples, let's stick to `.create_dataset()` - or the mapping style shorthand `f['name'] = x` - rather than `.require_dataset()`, which is an extra thing to understand.

docs/high/dataset.rst
```diff
@@ -96,14 +96,70 @@ To retrieve the contents of a `scalar` dataset, you can use the same
...
+2.	Declare the datatype for each column by reference to the field name.  Datatype is entered row-by-row as a list of tuples, with array dimensions (beginning with length) given as the third entry in the tuple::
+
+        >>> DataType = [
+        >>>		(‘Location’, ‘S10’),
+        >>>		(‘LocInt’, ‘int’),
+        >>>		(‘LocList’, ‘bool’, 2),
+        >>>		(‘LocFloat’, ‘float’),
+        >>>     ]
+
+3.	Define the shape as an integer, which is equal to the number of rows in the dataset::
```
takluyver, line 146: Like any other dataset, compound datasets can be multi-dimensional, with a shape tuple. You're not limited to a linear list with columns. I have a case at work where we're using 2D compound datasets, so this isn't just theoretical. :wink: 

(But also, as mentioned above, if you have the data already, you don't need to pass an explicit shape)

docs/high/dataset.rst
```diff
@@ -96,14 +96,70 @@ To retrieve the contents of a `scalar` dataset, you can use the same
...
+(as shown above) can be provided, or a variable-length datatype can be declared using
+string_dtype, as shown below::
+
+    >>> sdt = h5py.string_dtype(encoding='utf-8')
+    >>> DataType = [
+    >>>		(‘Location’, sdt),
+    >>>		... ,
+
+Note that fixed-length datatypes can truncate strings without notification.  Also note that this
+declaration can be used to override a given datatype (int to float, etc.).
```
takluyver, line 161: AFAIK, strings in compound datasets have the same behaviour & limitations as strings in h5py in general, so we should link to the strings page rather than repeating a subset of that information.

takluyver (commented):
Thanks! I think it makes sense to have a section here explaining what compound datasets are and how to use them. I think it could do with being a bit more general, and refer to other relevant docs where necessary.

docs/high/dataset.rst
```diff
@@ -69,7 +69,7 @@ following slicing arguments are recognized:
     * At most one ``Ellipsis`` (``...``) object
     * An empty tuple (``()``) to retrieve all data or `scalar` data
 
-Here are a few examples (output omitted).
+Here are a few examples (output omitted).::
```
takluyver, line 72:
```suggestion
Here are a few examples (output omitted)::
```

`.:` in the output looks ugly. :wink:

docs/high/dataset.rst
```diff
@@ -458,7 +514,7 @@ Reference
 
         Check that the dataset is accessible.
         A dataset could be inaccessible for several reasons. For instance, the
-        dataset, or the file it belongs to, may have been closed elsewhere.
+        dataset, or the file it belongs to, may have been closed elsewhere.:
```
takluyver, line 517:
```suggestion
        dataset, or the file it belongs to, may have been closed elsewhere::
```

PR #2447, opened 2024-06-22
MNT: enable cp313 nightly builds

neutrinoceros:
NumPy is already making nightlies for CPython 3.13, and after testing on my fork, it seems we're already compatible (at least on ubuntu) so might as well add nightlies for it too.


aragilar:
Catching issues early sounds like a great idea, thanks @neutrinoceros! We probably want to decide on a frequency to update the manylinux images (as mentioned in https://github.com/h5py/hdf5-manylinux/issues/11), but that shouldn't block merging this.

Issue #2019, opened 2021-12-16
Potential deadlock from some combination of `h5py`, `gc`, and `fsspec`


TomAugspurger:
Version info:
<details>

```
Summary of the h5py configuration
---------------------------------

h5py    3.6.0
HDF5    1.12.1
Python  3.8.12 | packaged by conda-forge | (default, Oct 12 2021, 21:59:51) 
[GCC 9.4.0]
sys.platform    linux
sys.maxsize     9223372036854775807
numpy   1.20.0
cython (built with) 0.29.24
numpy (built against) 1.18.5
HDF5 (built against) 1.12.1
```

</details>

This script creates a deadlock, typically by the 12th or so iteration:

```python
import fsspec
import h5netcdf
import h5py


def main():
    url = "https://pangeo.blob.core.windows.net/pangeo-public/hurs_day_ACCESS-CM2_historical_r1i1p1f1_gn_1950.nc"
    for i in range(100):
        with fsspec.open(url).open() as o:
            h5netcdf.core.File(o, mode="r", decode_vlen_strings=True)

        print("--------------------------------------", i)


if __name__ == "__main__":
    main()
```

The notable features are:

1. It uses fsspec / aiohttp to open the URL and provide a file-like object. fsspec does some things with threads and the asyncio event loop that I don't really understand (cc @martindurant)
2. It uses `h5netcdf` to open the file. Apologies for not having a completely minimal example, but I haven't been able to reproduce the hang with *just* h5py. https://github.com/dask/dask/issues/7547#issuecomment-872563337 indicates that this might be interacting with CPython's garbage collector, so it's *possibele* that h5netcdf is creating the circular references necessary to trigger the deadlock

Here's the traceback when I interrupt it:

```pytb
(notebook) jovyan@jupyter-taugspurger-40microsoft-2ecom:~/cmip6$ $ python hang.py 
-------------------------------------- 0
-------------------------------------- 1
-------------------------------------- 2
-------------------------------------- 3
-------------------------------------- 4
^CTraceback (most recent call last):
  File "hang.py", line 16, in <module>
    main()
  File "hang.py", line 10, in main
    h5netcdf.core.File(o, mode="r", decode_vlen_strings=True)
  File "/srv/conda/envs/notebook/lib/python3.8/site-packages/h5netcdf/core.py", line 780, in __init__
    super(File, self).__init__(self, self._h5path)
[...]

PR #2449, opened 2024-06-23
WHL: enable cp313t nightly builds

neutrinoceros:
Follow up to #2447, this time add cp313*t* wheels for CPython free-threading (a.k.a "no GIL") flavor.
Currently this requires building Cython from source: effort to support free-threading [is underway](https://github.com/cython/cython/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A%22nogil+CPython%22) but not released yet.
To my surprise, the trial run on my fork revealed no problem, so I guess we can just roll with it ?

### DIFF
--- a/.github/workflows/nightlies.yml
+++ b/.github/workflows/nightlies.yml
@@ -39,12 +39,13 @@ jobs:
       - name: Build wheels
         run: python -m cibuildwheel --output-dir wheelhouse
         env:
-          CIBW_BEFORE_BUILD: "pip install --pre --extra-index-url https://pypi.anaconda.org/scientific-python-nightly-wheels/simple numpy && pip install \"Cython>=0.29.31,<4\" pkgconfig \"setuptools>=61\""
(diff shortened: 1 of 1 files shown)


neutrinoceros:
sorry I didn't realize my PRs would conflict with yours.
#2444 is big enough that I don't fancy adding this patch there, but I'm okay with this one waiting for yours.

tacaswell:
We provide our own global lock so I'm not surprised the tests pass.

neutrinoceros:
Nice. So h5py is already and will remain thread safe under the new Python build ?

tacaswell:
That is my 80% expectation but threads always find a way to surprise you (see https://github.com/h5py/h5py/issues/2019) and we do some fancy stuff with gc around `nonlocal_close`).

neutrinoceros:
let's drop this for now: there's already a conflict now and by the time I circle back to it it'll be almost completely redundant.

PR #2463, opened 2024-07-29
Expose fill time property via the high-level dataset creation interface

### LINKED ISSUE #2459: Add option to never write fill value via the high-level dataset creation interface
This is not a bug report.

Since this is my first issue here, I just want to take this chance to thank all the developers in `h5py` for providing this convenient Python wrapper for hdf5! 

#### Background
For parallel writing to chunked storage, the chunks are allocated when the dataset is created and in `h5py` [it enforces the time of writing fill value on allocation](https://github.com/h5py/h5py/blob/master/h5py/_hl/filters.py#L262). From the comment there I guess it is to ensure all values have been properly initialised to avoid reading uninitialised memory. However, this is unnecessary when we know we are definitely going to overwrite all the fill values in our whole dataset by our actual data. 

#### Current limitation
I found out [you can pass your dataset creation property list using the keyword `dcpl`](https://github.com/h5py/h5py/blob/master/h5py/_hl/dataset.py#L39) from `create_dataset`, so potentially I can use the `set_fill_time` method to change it to `FILL_TIME_NEVER` for my own dcpl, however it will be [overwritten by `FILL_TIME_ALLOC`](https://github.com/h5py/h5py/blob/master/h5py/_hl/filters.py#L262) later. Another route is to bypass the high-level interface entirely by using the exposed C API, but I will miss all the validations provided to other settings.

#### Suggestion
I would like to suggest exposing the option of never writing fill values via the high-level `create_dataset`/`require_dataset` interface. One way of doing it is to provide a parameter in those methods, such as something like `write_fill` with options `"alloc"`/`"never"`/`"ifset"` ([the current options in hdf5](https://docs.hdfgroup.org/hdf5/v1_14/_h5_dpublic_8h.html#aa39293626c4e68dd28b06c0dc84bde4a)), default it to `"alloc"` and this will keep the original behaviour. By consciously passing `write_fill="never"`, the user is aware of and responsible for this.  

Another (less obvious) way is to check the setting of `dcpl` to see what fill time option is set, and if equates to the enum value of `H5D_FILL_TIME_NEVER`, just keep it.

#### Potential action
[Setting the fill time to `H5D_FILL_TIME_NEVER` can have significant performance improvement for parallel writing to chunked storage](https://www.hdfgroup.org/wp-content/uploads/2022/05/2022_ECP-Parallel-HDF5-Performance-Tuning.pdf). I hope this can benefit users who know they will definitely overwrite the fill values by their data. I am more than happy to contribute to this if `h5py` developers are happy about this idea and this won't cause significant safety and security issues.

Thank you.

Useful link:
[Storage allocation and fill summary](https://docs.hdfgroup.org/hdf5/v1_14/_h5_d__u_g.html#subsubsec_dataset_allocation_store)

ptim0626:
This PR exposes the fill time property via the high-level dataset creation interface. This is my first PR to `h5py` so feel free to let me know what I could have missed. Thank you.

For the motivation of this PR, please see #2459.

Ready for review.

Upon merging, this closes #2459.

### DIFF
--- a/docs/high/group.rst
+++ b/docs/high/group.rst
@@ -386,10 +386,10 @@ Reference
             following choices: `alloc`, write fill value before writing
             application data values or when the dataset is created; `never`,
             never write fill value; `ifset`, write fill value if it is defined.
-            Default to `alloc` for chunked storage, `ifset` for contiguous
-            storage. If the whole dataset is going to be written by the
-            application, setting this to `never` can avoid unnecessary writing
-            of fill value and potentially improve performance.
+            Default to `ifset`, which is the default of HDF5 library. If the
+            whole dataset is going to be written by the application, setting
+            this to `never` can avoid unnecessary writing of fill value and
+            potentially improve performance.
 
         :keyword track_times: Enable dataset creation timestamps (**T**/F).
 
--- a/docs/high/group.rst
+++ b/docs/high/group.rst
@@ -382,6 +382,15 @@ Reference
         :keyword fillvalue: This value will be used when reading
                             uninitialized parts of the dataset.
 
+        :keyword fill_time: Control when to write the fill value. One of the
+            following choices: `alloc`, write fill value before writing
+            application data values or when the dataset is created; `never`,
+            never write fill value; `ifset`, write fill value if it is defined.
+            Default to `alloc` for chunked storage, `ifset` for contiguous
+            storage. If the whole dataset is going to be written by the
+            application, setting this to `never` can avoid unnecessary writing
+            of fill value and potentially improve performance.
+
         :keyword track_times: Enable dataset creation timestamps (**T**/F).
 
         :keyword track_order: Track attribute creation order if
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -38,7 +38,8 @@ def make_new_dset(parent, shape=None, dtype=None, data=None, name=None,
                   fillvalue=None, scaleoffset=None, track_times=False,
                   external=None, track_order=None, dcpl=None, dapl=None,
                   efile_prefix=None, virtual_prefix=None, allow_unknown_filter=False,
-                  rdcc_nslots=None, rdcc_nbytes=None, rdcc_w0=None):
+                  rdcc_nslots=None, rdcc_nbytes=None, rdcc_w0=None, *,
+                  fill_time=None):
     """ Return a new low-level dataset identifier """
 
     # Convert data to a C-contiguous ndarray
@@ -104,7 +105,8 @@ def make_new_dset(parent, shape=None, dtype=None, data=None, name=None,
     dcpl = filters.fill_dcpl(
         dcpl or h5p.create(h5p.DATASET_CREATE), shape, dtype,
         chunks, compression, compression_opts, shuffle, fletcher32,
-        maxshape, scaleoffset, external, allow_unknown_filter)
+        maxshape, scaleoffset, external, allow_unknown_filter,
+        fill_time=fill_time)
 
     if fillvalue is not None:
         # prepare string-type dtypes for fillvalue
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -105,7 +105,7 @@ def make_new_dset(parent, shape=None, dtype=None, data=None, name=None,
     dcpl = filters.fill_dcpl(
         dcpl or h5p.create(h5p.DATASET_CREATE), shape, dtype,
         chunks, compression, compression_opts, shuffle, fletcher32,
-        maxshape, scaleoffset, external, allow_unknown_filter, fill_time)
+        maxshape, scaleoffset, external, allow_unknown_filter)
 
     if fillvalue is not None:
         # prepare string-type dtypes for fillvalue
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -35,11 +35,10 @@ MPI = h5.get_config().mpi
 def make_new_dset(parent, shape=None, dtype=None, data=None, name=None,
                   chunks=None, compression=None, shuffle=None,
                   fletcher32=None, maxshape=None, compression_opts=None,
-                  fillvalue=None, fill_time=None, scaleoffset=None,
-                  track_times=False, external=None, track_order=None,
-                  dcpl=None, dapl=None, efile_prefix=None, virtual_prefix=None,
-                  allow_unknown_filter=False, rdcc_nslots=None,
-                  rdcc_nbytes=None, rdcc_w0=None):
+                  fillvalue=None, scaleoffset=None, track_times=False,
+                  external=None, track_order=None, dcpl=None, dapl=None,
+                  efile_prefix=None, virtual_prefix=None, allow_unknown_filter=False,
+                  rdcc_nslots=None, rdcc_nbytes=None, rdcc_w0=None):
     """ Return a new low-level dataset identifier """
 
     # Convert data to a C-contiguous ndarray
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -105,7 +105,7 @@ def make_new_dset(parent, shape=None, dtype=None, data=None, name=None,
     dcpl = filters.fill_dcpl(
         dcpl or h5p.create(h5p.DATASET_CREATE), shape, dtype,
         chunks, compression, compression_opts, shuffle, fletcher32,
-        maxshape, scaleoffset, external, allow_unknown_filter)
+        maxshape, scaleoffset, external, allow_unknown_filter, write_fill)
 
     if fillvalue is not None:
         # prepare string-type dtypes for fillvalue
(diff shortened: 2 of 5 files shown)


ptim0626:
The pipeline failed because of hdf5 download link and deprecated macOS image, after merging #2464 this should be fixed.

h5py/_hl/dataset.py
```diff
@@ -35,10 +35,11 @@
 def make_new_dset(parent, shape=None, dtype=None, data=None, name=None,
                   chunks=None, compression=None, shuffle=None,
                   fletcher32=None, maxshape=None, compression_opts=None,
-                  fillvalue=None, scaleoffset=None, track_times=False,
-                  external=None, track_order=None, dcpl=None, dapl=None,
-                  efile_prefix=None, virtual_prefix=None, allow_unknown_filter=False,
-                  rdcc_nslots=None, rdcc_nbytes=None, rdcc_w0=None):
+                  fillvalue=None, fill_time=None, scaleoffset=None,
```
neutrinoceros, line 38: This change is API breaking: we cannot add new arguments at arbitrary positions between two existing positionally-allowed arguments.

Please move this new argument to the end of the signature and make it keyword-only.

h5py/_hl/filters.py
```diff
@@ -146,7 +150,7 @@ def __init__(self, level=DEFAULT_GZIP):
 
 def fill_dcpl(plist, shape, dtype, chunks, compression, compression_opts,
               shuffle, fletcher32, maxshape, scaleoffset, external,
-              allow_unknown_filter=False):
+              allow_unknown_filter=False, fill_time=None):
```
neutrinoceros, line 153: This is better, but I'd still prefer we make it keyword only:
```suggestion
              allow_unknown_filter=False, *, fill_time=None):
```

neutrinoceros (changes requested):
Hi ! Thanks for your contribution !
Here's a first shallow review, we can dig deeper once it's addressed.

h5py/_hl/dataset.py
ptim0626: Thanks for pointing this out, I haven't considered this.

h5py/_hl/filters.py
```diff
@@ -259,7 +263,23 @@ def rq_tuple(tpl, name):
 
     if chunks is not None:
         plist.set_chunk(chunks)
-        plist.set_fill_time(h5d.FILL_TIME_ALLOC)  # prevent resize glitch
+
+    if fill_time is None:
+        if chunks is None:
+            fill_time = 'ifset'
+        else:
+            fill_time = 'alloc'  # prevent resize glitch
```
takluyver, line 271: Are these the same defaults as HDF5 itself? If so, can we skip calling `set_fill_time` entirely in this case, and leave it up to HDF5, so we don't have to ensure we're doing the same thing for all corner cases & with future changes in HDF5.

h5py/_hl/filters.py
```diff
@@ -259,7 +263,23 @@ def rq_tuple(tpl, name):
...
+    else:
+        invalid_fill_time = (isinstance(fill_time, str) and
+                             fill_time.lower() not in
```
takluyver, line 274: Our existing one-of-these-strings parameters, like `compression`, `driver` and `mode`, are case-sensitive, and I think this is a common pattern (e.g. the `byteorder=` parameter for `int.from_bytes`. So let's stick with that here.

h5py/_hl/filters.py
```diff
@@ -259,7 +263,23 @@ def rq_tuple(tpl, name):
...
+                             ('alloc', 'never', 'ifset'))
+        fill_time_not_str = not isinstance(fill_time, str)
```
takluyver, line 276: If we're going to the trouble of checking the type separately, we could raise a TypeError. Or we could keep the code simple and just check `fill_time in _FILL_TIME_ENUM`. :slightly_smiling_face:

h5py/_hl/filters.py
ptim0626: The default from hdf5 is 'ifset', H5D_FILL_TIME_IFSET ([1.10](https://docs.hdfgroup.org/hdf5/v1_10/group___d_c_p_l.html#ga92c5eb5ee19bfd4a9184cf0428d1b00c), [1.12](https://docs.hdfgroup.org/hdf5/v1.12/group___d_c_p_l.html#ga92c5eb5ee19bfd4a9184cf0428d1b00c) and [v1.14](https://docs.hdfgroup.org/hdf5/v1_14/group___d_c_p_l.html#ga92c5eb5ee19bfd4a9184cf0428d1b00c)), and I believe this won't be changed. The original behaviour in h5py is to set H5D_FILL_TIME_ALLOC when it is chunked to 'prevent resize glitch'. I am trying to preserve the original behaviour in h5py (that 'prevent resize glitch' is from 13 years ago!).

h5py/tests/test_dataset.py
```diff
@@ -476,6 +476,74 @@ def test_exc(self):
...
+        dset = self.f.create_dataset('foo', (10,), chunks=(2,), fillvalue=4.0)
+        plist = dset.id.get_create_plist()
+        self.assertEqual(plist.get_fill_time(), h5py.h5d.FILL_TIME_ALLOC)
+        self.assertEqual(dset[0], 4.0)
+        self.assertEqual(dset[7], 4.0)
+
+    def test_compressed_default(self):
+        """ Fill time is ALLOC for compressed dataset (chunked) """
+        dset = self.f.create_dataset('foo', (10,), compression='gzip',
+                                     fillvalue=4.0)
```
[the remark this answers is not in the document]
ptim0626: Thanks for the tip!

h5py/_hl/filters.py
takluyver: Ah, sorry, I'd missed this behaviour was already there.

I'm inclined to get rid of that and use HDF5's defaults. There's a good chance whatever it was working round was fixed years ago, and I can't see any further details at a quick look,  so I suspect the only way to find out if it's still needed is to change it and see if anything breaks. We should make sure it's in the release notes if we change the default, though, so it's easy to find the relevant change if something does go wrong.

h5py/_hl/filters.py
ptim0626: Yes I agree with the simple check `fill_time in _FILL_TIME_ENUM`. I think I was trying to prevent users passing the enum value directly and somehow guided by some type check...

h5py/_hl/filters.py
ptim0626: Adopted.

h5py/_hl/filters.py
ptim0626: From the comment it looks like the purpose is to prevent access to uninitialised memory. If `max_shape` is `(10,)` and the user just write 5 values, the result of reading the last 5 values later without fill value could be some garbage.  

I can try to remove this behaviour to see if it passes the tests/CI.

ptim0626: Tests passed locally if I remove the original h5py behaviour. Let see the CI.

takluyver:
Thanks!

I think allocating space in a file without writing it gives zeros rather than junk data (at least `man 2 truncate` seems to guarantee this). Maybe you can see uninitialised memory if using the 'core' file driver, which operates in-memory?

Other maintainers: this now changes a default to match HDF5's own default (H5D_FILL_TIME_ALLOC always), at my suggestion. Speak up if you see any issue with that.

takluyver:
You're right, @ajelenak, my mistake. I was writing the comment too quickly.

docs/high/group.rst
```diff
@@ -382,6 +382,15 @@ Reference
         :keyword fillvalue: This value will be used when reading
                             uninitialized parts of the dataset.
 
+        :keyword fill_time: Control when to write the fill value. One of the
+            following choices: `alloc`, write fill value before writing
+            application data values or when the dataset is created; `never`,
+            never write fill value; `ifset`, write fill value if it is defined.
+            Default to `alloc` for chunked storage, `ifset` for contiguous
+            storage. If the whole dataset is going to be written by the
```
takluyver, line 390: Don't forget to update this.

ptim0626:
> I think allocating space in a file without writing it gives zeros rather than junk data (at least `man 2 truncate` seems to guarantee this). Maybe you can see uninitialised memory if using the 'core' file driver, which operates in-memory?

I am not entirely sure about this. From [here in the API reference](https://docs.hdfgroup.org/hdf5/v1_14/_h5_d__u_g.html#title5), if you look at the table 'Storage allocation and fill summary', it specifically says '_A read of unwritten data returns undefined values_'. So perhaps from HDF5 side, it doesn't guarantee they will be zeros.

The default fill value is zero (from HDF5), however if the fill value is never written, they may not necessarily be zeros.

docs/high/group.rst
ptim0626: Thanks for reminding!

ptim0626:
I added two tests for resizing chunked dataset.
- when fill value and fill time are default (`0` and `ifset` respectively`), the resized dataset will be filled by zeros.
- when fill value is not the default, and fill time is `never`, the resized dataset won't be filled by the user-defined fill value

takluyver:
Thanks @ptim0626 !

One more thing: can you add in the news entry that the default changed for the chunked case? Put it under 'deprecations' for now, I'll tweak the header to something like 'Breaking changes & deprecations' when I compile the release notes. :slightly_smiling_face: 

Other than that, I think this is ready to go - thanks for your patience.

PR #2464, opened 2024-07-30
MNT: Fix hdf5 download link in CI

ptim0626:
This small PR fixes the download link of hdf5 libraries, the location `https://www.hdfgroup.org/ftp/...` is not available and the source files are located under `https://support.hdfgroup.org/ftp/...`.

Ready for review.

### DIFF
--- a/ci/get_hdf5_if_needed.sh
+++ b/ci/get_hdf5_if_needed.sh
@@ -35,7 +35,7 @@ else
         echo "building HDF5"
 
         IFS='.-' read MAJOR_V MINOR_V REL_V PATCH_V <<< "$HDF5_VERSION"
-        # the assets in GitHub (currently) has two naming conventions
+        # the assets in GitHub (currently) have two naming conventions
         if [[ -n "${PATCH_V}" ]]; then
             ASSET_FMT1="hdf5-${MAJOR_V}_${MINOR_V}_${REL_V}-${PATCH_V}"
             ASSET_FMT2="hdf5_${MAJOR_V}.${MINOR_V}.${REL_V}.${PATCH_V}"
--- a/ci/get_hdf5_if_needed.sh
+++ b/ci/get_hdf5_if_needed.sh
@@ -64,7 +64,7 @@ else
         url_base="https://github.com/HDFGroup/hdf5/archive/refs/tags/"
         curl -fsSL -o "hdf5-$HDF5_VERSION.tar.gz" "${url_base}${ASSET_FMT1}.tar.gz" || curl -fsSL -o "hdf5-$HDF5_VERSION.tar.gz" "${url_base}${ASSET_FMT2}.tar.gz"
 
-        tar -xzvf hdf5-$HDF5_VERSION.tar.gz --one-top-level --strip-components=1
+        mkdir -p hdf5-$HDF5_VERSION && tar -xzvf hdf5-$HDF5_VERSION.tar.gz --strip-components=1 -C hdf5-$HDF5_VERSION
 
         pushd hdf5-$HDF5_VERSION
 
--- a/ci/get_hdf5_if_needed.sh
+++ b/ci/get_hdf5_if_needed.sh
@@ -34,9 +34,16 @@ else
     else
         echo "building HDF5"
 
-        MINOR_V=${HDF5_VERSION#*.}
-        MINOR_V=${MINOR_V%.*}
-        MAJOR_V=${HDF5_VERSION/%.*.*}
+        IFS='.-' read MAJOR_V MINOR_V REL_V PATCH_V <<< "$HDF5_VERSION"
+        # the assets in GitHub (currently) has two naming conventions
+        if [[ -n "${PATCH_V}" ]]; then
+            ASSET_FMT1="hdf5-${MAJOR_V}_${MINOR_V}_${REL_V}-${PATCH_V}"
+            ASSET_FMT2="hdf5_${MAJOR_V}.${MINOR_V}.${REL_V}.${PATCH_V}"
+        else
+            ASSET_FMT1="hdf5-${MAJOR_V}_${MINOR_V}_${REL_V}"
+            ASSET_FMT2="hdf5_${MAJOR_V}.${MINOR_V}.${REL_V}"
+        fi
+
         if [[ $MAJOR_V -gt 1 || $MINOR_V -ge 12 ]]; then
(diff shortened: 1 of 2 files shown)


neutrinoceros:
Hi @ptim0626, thanks for your contribution !
Both changes look fine but as they're not related to each other, it'd be preferable to separate them into two distinct PRs. Can you ?

ptim0626:
@neutrinoceros Yes I agree, I just found this out while working on this PR. I will split it.

ajelenak:
Just want to mention that the libhdf5 source is also available from https://github.com/HDFGroup/hdf5/releases. For example: https://github.com/HDFGroup/hdf5/archive/refs/tags/hdf5_1.14.4.3.tar.gz. This should be more stable link than the HDFG ones.

ptim0626:
Thanks for pointing this out. I found those in HDFG are more consistent in their naming convention, e.g. for 1.12 the github one is [https://github.com/HDFGroup/hdf5/archive/refs/tags/hdf5-1_12_3.tar.gz](https://github.com/HDFGroup/hdf5/archive/refs/tags/hdf5-1_12_3.tar.gz) (note the underscores), but for 1.14 it seems it switches to decimal-delimited. I guess pointing to a location with more a consistent naming convention is more convenient for the CI.

tacaswell:
We should use the github URLs as @ajelenak suggests please.

ptim0626:
I have no preference on the sources, whether from HDFG or their GitHub are both fine. I modified the code to get the source from their GitHub releases in Linux, with some logics to deal with their naming conventions (no idea why the naming is not consistent, in contrary to those in HDFG).

I will work on the corresponding part when building in Windows.

ptim0626:
Same modification to Windows building script done.  Most of the changes handle breaking change in the naming, e.g. `1.14.4.3` vs `1_14_1-2`.

takluyver:
Thanks! I'm going to close & reopen this to check it with #2444 merged.

ci/get_hdf5_if_needed.sh
```diff
@@ -54,9 +61,15 @@ else
...
-        curl -fsSLO "https://www.hdfgroup.org/ftp/HDF5/releases/hdf5-$MAJOR_V.$MINOR_V/hdf5-$HDF5_VERSION/src/hdf5-$HDF5_VERSION.tar.gz"
-        tar -xzvf hdf5-$HDF5_VERSION.tar.gz
+        # temporarily turn off -e option as curl will emit non-zero exit code
+        # as only one assert format is valid
+        # (curl exits with 22 with http code >400)
+        set +e
+        curl -fsSL -o "hdf5-$HDF5_VERSION.tar.gz" "https://github.com/HDFGroup/hdf5/archive/refs/tags/{${ASSET_FMT1},${ASSET_FMT2}}.tar.gz"
+        set -e
+
+        tar -xzvf hdf5-$HDF5_VERSION.tar.gz --one-top-level --strip-components=1
```
takluyver, line 71: It looks like the Mac version of the `tar` command doesn't have the `--one-top-level` flag, unfortunately.

ci/get_hdf5_if_needed.sh
```diff
@@ -54,9 +61,15 @@ else
...
         pushd /tmp
-        #                                   Remove trailing .*, to get e.g. '1.12' ↓
-        curl -fsSLO "https://www.hdfgroup.org/ftp/HDF5/releases/hdf5-$MAJOR_V.$MINOR_V/hdf5-$HDF5_VERSION/src/hdf5-$HDF5_VERSION.tar.gz"
-        tar -xzvf hdf5-$HDF5_VERSION.tar.gz
+        # temporarily turn off -e option as curl will emit non-zero exit code
+        # as only one assert format is valid
+        # (curl exits with 22 with http code >400)
+        set +e
+        curl -fsSL -o "hdf5-$HDF5_VERSION.tar.gz" "https://github.com/HDFGroup/hdf5/archive/refs/tags/{${ASSET_FMT1},${ASSET_FMT2}}.tar.gz"
+        set -e
```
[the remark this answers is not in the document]
ptim0626: @takluyver Thanks for the suggestion. Initially I used curl globbing to reuse the connection, but two separate instances of curl (at least for now) shouldn't really matter, and there won't be the weird flipping of `set +e` and `set -e`, which I also don't like!

ci/get_hdf5_if_needed.sh
ptim0626: @takluyver Thanks for pointing this out, amended.

larsoner (approved):
All green except Travis which seems to have timed out after 5065m (!), should be good to go I think?

ci/get_hdf5_if_needed.sh
```diff
@@ -34,9 +34,16 @@ else
     else
         echo "building HDF5"
 
-        MINOR_V=${HDF5_VERSION#*.}
-        MINOR_V=${MINOR_V%.*}
-        MAJOR_V=${HDF5_VERSION/%.*.*}
+        IFS='.-' read MAJOR_V MINOR_V REL_V PATCH_V <<< "$HDF5_VERSION"
+        # the assets in GitHub (currently) has two naming conventions
```
neutrinoceros, line 38:
```suggestion
        # the assets in GitHub (currently) have two naming conventions
```

ci/get_hdf5_win.py
```diff
@@ -114,8 +123,9 @@ def build_hdf5(version, hdf5_file, install_path, cmake_generator, use_prefix):
         copy(f, pjoin(install_path, 'lib'))
 
 
-def get_cmake_config_path(version, extract_point):
-    return pjoin(extract_point, REL_PATH_TO_CMAKE_CFG.format(version=version))
+def get_cmake_config_path(extract_point, zip_file):
+    dir_suffix = basename(zip_file).replace(".zip", "")
```
neutrinoceros, line 127: Marginally frustrating that this requires Python 3.9 and I *just* proposed to drop 3.8
```suggestion
    dir_suffix = basename(zip_file).removesuffix(".zip")
```

neutrinoceros: (please do not apply this)

takluyver:
Yeah, I've restarted the Travis job a few times, but it always seems to time out. We can merge this as is, since the CI is broken anyway. If Travis doesn't sort itself out in a few days, maybe we just give up on it. The only reason we still use it is for testing on pcc64le; I don't know of any other CI service that offers that.

PR #2465, opened 2024-07-30
MNT: Retire macOS-11 in CI

ptim0626:
This PR updates the mac image to `macOS-12` as [`macOS-11` is retired from  2024-06-28](https://learn.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#recent-updates).

Ready for review.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -51,7 +51,7 @@ jobs:
 
 - job: 'macos_wheels'
   pool:
-    vmImage: macOS-11
+    vmImage: macOS-12
   variables:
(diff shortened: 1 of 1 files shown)


neutrinoceros:
Thanks for taking the time to split up your PR ! I'm actually not sure we should merge this given that #2444 is expected to be merged very shortly and this generates a merge conflict we can avoid.

ptim0626:
ah sorry for my ignorance about this, not aware this is something work in progress. If a merge conflict can be avoided, I better close this.

PR #2469, opened 2024-08-04
WHL: enable cp313 wheels for aarch64 Linux

neutrinoceros:
🤞🏻

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -56,6 +56,7 @@ jobs:
           - "3.10"
           - "3.11"
           - "3.12"
+          - "3.13.0rc1"
         include:
           - os: ubuntu-latest
             arch: x86_64


neutrinoceros:
I don't think errors are related. What seems more likely is that some archive file at https://www.hdfgroup.org/ftp/HDF5/releases/hdf5-1.12/hdf5-1.12.2/src/hdf5-1.12.2.tar.gz got corrupted. Who should this be reported to ?

aragilar:
@jreadey might know?

I'm not clear which version of HDF5 we're supposed to be using, I think the 1.12 series is no longer supported (with 1.10 and 1.14 being currently, though I think 1.10 was ending soon)? It may be easier to change the HDF5 release used?

ajelenak:
@aragilar https://github.com/HDFGroup/hdf5#release-schedule

Both HDF5 1.10 and 1.12 are deprecated now.

neutrinoceros:
The error also affects `master` [example logs](https://github.com/h5py/h5py/actions/runs/10234841667/job/28314849918), so it's clearly not related to this PR.
For my own sake, xref https://github.com/h5py/h5py/issues/2312

cbrnr:
How are you planning to deal with missing wheels for 3.13? For example, NumPy doesn't seem to have wheels for 3.13 yet, so do you build from source (meaning that you probably need additional build time depenencies like Meson etc.)?

neutrinoceros:
I opened this early just to get a baseline, but you're right that trying to go faster than numpy will only make things more complicated for us; that is not the plan.
I'll push this forward more seriously after numpy commits some `cp313` wheels to PyPI.
Also note that we're already uploading nightly `cp313` wheels for manylinux, built against numpy's own nightlies.

neutrinoceros:
This was superseded by #2470, except for the part on aarch64 Linux, so I narrowed the scope of this PR

neutrinoceros:
For posterity: I reverted this in #2491 because it never seemed to work.

PR #2473, opened 2024-08-15
Allow for a .N patch version in HDF5_VERSION environment variable

takluyver:
I accidentally broke CI by setting HDF5_VERSION in the hdf5-manylinux images to 1.14.4.3, forgetting that this is also used in h5py's build machinery. This should ignore the `.3` patch version.


takluyver:
Everything's passed except for Travis, which is stuck again, so I'm going to merge this to fix CI. Review is still welcome post-merge.

Issue #2475, opened 2024-08-16
Support for free-threading CPython


neutrinoceros:
As hinted to in #2416, I think there are a couple small things we can do to help the community of downstream testers to run h5py on free-threaded CPython. I just have a couple minutes on my hands so instead of *doing* them, I'll just write a list for now.
- review unsafe C-API usage (most importantly use of borrowed refs), see https://docs.python.org/3.13/howto/free-threading-extensions.html
- declare support in all C and Cython extensions (see https://py-free-threading.github.io/porting/ and https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#compiler-directives)
- setup some minimal validation in CI (like running tests in a single thread on Python 3.13t on ubuntu and check that everything passes and the GIL stays disabled)

I am hoping this is a relatively lightweight job, so right now I think that doing all this in time for h5py 3.12 is feasible, so let me put it in the milestone.

ajelenak:
What implications of running h5py with free-threaded CPython will be on libhdf5 which is not thread-safe by default? I don't think h5py currently uses thread-safe build of libhdf5.

tacaswell:
We already have our own global lock on any access to libhdf5 https://github.com/h5py/h5py/blob/7d4c3a15fc53f7d0a5b6c8e9895a9b3fb9632f12/h5py/_objects.pyx#L20-L46.  The only place I am currently worried is in the code around the non-local close registry https://github.com/h5py/h5py/blob/7d4c3a15fc53f7d0a5b6c8e9895a9b3fb9632f12/h5py/_objects.pyx#L95-L98 but it looks like everything is already behind `phil`

neutrinoceros:
Given the discussion in #2478 I think I'd like to wait on Cython 3.1 before making any more progress on this. In the mean time, it shouldn't be blocking the 3.12 release.

tacaswell:
I _think_ it is also reasonable to build cp313t wheels, but leave the "not freethread safe" state on it.  That way downstream can still get pre-built wheels and then turn the safety checks off to test.

neutrinoceros:
How about we do this but just for nightlies for now ?

neutrinoceros:
So that's basically what I tried in #2449, but the reason I abandoned it is that I think Cython might soon have a stable release that supports cp313t. If Cython 3.1 is not out in a month we can revive #2449, sounds good ?

SerGeRybakov:
[...]

PR #2478, opened 2024-08-19
BLD: add freethreading_compatible Cython markers

neutrinoceros:
partially address #2475
All this does is avoid reenabling the GIL on import for extension modules, and on Python 3.13t: it is not a claim for complete thread-safety, rather a declaration of intention that we want to do it and make it easier for downstream testers at least try it out so we can prioritize real-world applications when we start actually implementing thread safety (if needed).

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -98,6 +98,7 @@ jobs:
           CIBW_BUILD_FRONTEND: ${{ env.CIBW_BUILD_FRONTEND }}
           CIBW_PRERELEASE_PYTHONS: ${{ env.CIBW_PRERELEASE_PYTHONS }}
           CIBW_BEFORE_TEST: ${{ env.CIBW_BEFORE_TEST }}
+          CIBW_ENABLE: cpython-freethreading
 
       - run: python ./ci/bundle_hdf5_whl.py wheelhouse
         if: runner.os == 'Windows'
--- a/api_gen.py
+++ b/api_gen.py
@@ -144,6 +144,7 @@ class Line:
 
 raw_preamble = """\
 # cython: language_level=3
+# cython: freethreading_compatible=True
 #
 # Warning: this file is auto-generated from api_gen.py. DO NOT EDIT!
 #
@@ -156,6 +157,7 @@ from .api_types_ext cimport *
 
 def_preamble = """\
 # cython: language_level=3
+# cython: freethreading_compatible=True
 #
 # Warning: this file is auto-generated from api_gen.py. DO NOT EDIT!
 #
@@ -169,6 +171,7 @@ from .api_types_ext cimport *
 
 imp_preamble = """\
 # cython: language_level=3
+# cython: freethreading_compatible=True
 #
 # Warning: this file is auto-generated from api_gen.py. DO NOT EDIT!
 #
--- a/ci/cibw_test_command.sh
+++ b/ci/cibw_test_command.sh
@@ -11,7 +11,7 @@ echo "PROJECT_PATH=$PROJECT_PATH"
 WHEEL_PATH=$2
 echo "WHEEL_PATH=$WHEEL_PATH"
 
-export PYVER=$(python -c "import sys; print(''.join(map(str, sys.version_info[:2])))")
+export PYVER=$(python -c "import sys; print(''.join(map(str, sys.version_info[:2])) + ('t' if (sys.version_info>=(3,13) and not sys._is_gil_enabled()) else ''))")
 ENVLIST="py$PYVER-test-deps,py$PYVER-test-mindeps,py$PYVER-test-deps-pre"
 export H5PY_TEST_CHECK_FILTERS=1
 echo "ENVLIST=$ENVLIST"
--- a/h5py/_conv.pyx
+++ b/h5py/_conv.pyx
@@ -1,5 +1,6 @@
 # cython: profile=False
 # cython: language_level=3
+# cython: freethreading_compatible=True
 # This file is part of h5py, a Python interface to the HDF5 library.
 #
 # http://www.h5py.org
--- a/h5py/_errors.pyx
+++ b/h5py/_errors.pyx
@@ -1,3 +1,4 @@
+# cython: freethreading_compatible=True
 # This file is part of h5py, a Python interface to the HDF5 library.
 #
 # http://www.h5py.org
--- a/h5py/_objects.pyx
+++ b/h5py/_objects.pyx
@@ -1,3 +1,4 @@
+# cython: freethreading_compatible=True
 # This file is part of h5py, a Python interface to the HDF5 library.
 #
 # http://www.h5py.org
--- a/h5py/_proxy.pyx
+++ b/h5py/_proxy.pyx
@@ -1,4 +1,5 @@
 # cython: profile=False
+# cython: freethreading_compatible=True
 
 # This file is part of h5py, a Python interface to the HDF5 library.
 #
--- a/h5py/_selector.pyx
+++ b/h5py/_selector.pyx
@@ -1,4 +1,5 @@
 # cython: language_level=3
+# cython: freethreading_compatible=True
 """Class to efficiently select and read data from an HDF5 dataset
 
 This is written in Cython to reduce overhead when reading small amounts of
--- a/h5py/h5.pxd
+++ b/h5py/h5.pxd
@@ -1,4 +1,5 @@
 # cython: language_level=3
+# cython: freethreading_compatible=True
 # This file is part of h5py, a Python interface to the HDF5 library.
 #
 # http://www.h5py.org
--- a/h5py/h5.pyx
+++ b/h5py/h5.pyx
@@ -1,4 +1,5 @@
 # cython: language_level=3
+# cython: freethreading_compatible=True
 # This file is part of h5py, a Python interface to the HDF5 library.
 #
 # http://www.h5py.org
--- a/h5py/h5a.pyx
+++ b/h5py/h5a.pyx
@@ -1,4 +1,5 @@
 # cython: language_level=3
+# cython: freethreading_compatible=True
 # This file is part of h5py, a Python interface to the HDF5 library.
 #
 # http://www.h5py.org
--- a/h5py/h5ac.pyx
+++ b/h5py/h5ac.pyx
@@ -1,3 +1,4 @@
+# cython: freethreading_compatible=True
 # This file is part of h5py, a Python interface to the HDF5 library.
 #
 # http://www.h5py.org
--- a/h5py/h5d.pyx
+++ b/h5py/h5d.pyx
@@ -1,4 +1,5 @@
 # cython: language_level=3
+# cython: freethreading_compatible=True
 ## This file is part of h5py, a Python interface to the HDF5 library.
 #
 # http://www.h5py.org
--- a/h5py/h5ds.pyx
+++ b/h5py/h5ds.pyx
@@ -1,4 +1,5 @@
 # cython: language_level=3
+# cython: freethreading_compatible=True
 # This file is part of h5py, a Python interface to the HDF5 library.
 #
 # http://www.h5py.org
--- a/h5py/h5f.pyx
+++ b/h5py/h5f.pyx
@@ -1,3 +1,4 @@
+# cython: freethreading_compatible=True
 # This file is part of h5py, a Python interface to the HDF5 library.
 #
 # http://www.h5py.org
--- a/h5py/h5fd.pyx
+++ b/h5py/h5fd.pyx
@@ -1,3 +1,4 @@
+# cython: freethreading_compatible=True
 # This file is part of h5py, a Python interface to the HDF5 library.
 #
 # http://www.h5py.org
--- a/h5py/h5g.pyx
+++ b/h5py/h5g.pyx
@@ -1,4 +1,5 @@
 # cython: language_level=3
+# cython: freethreading_compatible=True
 # This file is part of h5py, a Python interface to the HDF5 library.
 #
 # http://www.h5py.org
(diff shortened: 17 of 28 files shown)


takluyver:
Thanks!

So am I understanding correctly that binaries built with this change - including the wheels we publish - would leave the GIL off in free-threaded Python, potentially exposing the user to bugs if our code is not completely thread safe?

Is there a way we could make this a convenient option to set at build time, but still have the wheels we publish on PyPI require the GIL for now? I.e. I want the first round of people testing h5py without the GIL to build h5py from source, as a strong signal that you're testing something that may not work.

I can see the messaging in the Python docs is that free-threading is experimental, but I expect people are going to be *really* excited about this. I'd much rather enthusiastic non-experts see h5py as incompatible for now, than have it claim compatibility when it might not work.

neutrinoceros:
freethreading python actually requires special wheels altogether, and there are still not built by default with cibuildwheel. This changes nothing for regular wheels, and only makes a difference for anyone we wants to build from source in freethreading Python. I very much agree that downstream testers should build from source (or maybe get a nightly if we can set this up for ubuntu), and I am not proposing we publish stable cp313t wheels for now !

takluyver:
Great, thanks!

So these markers allow it to build freethreading compatible binaries if you ask. How do you ask for that?

neutrinoceros:
> So these markers allow it to build freethreading compatible binaries if you ask.

Not *quite*: you can already build free-threading binaries without them, it's just that, by default the GIL is re-enabled when one imports *any* extension that doesn't explicitly declare that it's fine with having the GIL released, and you get the following warning:
```python-traceback
RuntimeWarning: The global interpreter lock (GIL) has been enabled to load module 'h5py._errors', which has not declared that it can run safely without the GIL. To override this behavior and keep the GIL disabled (at your own risk), run with PYTHON_GIL=0 or -Xgil=0.
```
So it is already possible to get the same effect (disable the warning) on the user side by invoking python with the appropriate flags (and maybe we should actually just let users do that for now, up for debate).

>  How do you ask for that?

Locally, one needs to [install Python 3.13t](https://py-free-threading.github.io/installing_cpython/). Cython 3.1.0 is also needed (but not released yet) so it should be built from source or installed from https://pypi.anaconda.org/scientific-python-nightly-wheels/simple/ . Build isolation should also be turned off for this reason.

In cibuildwheel, [free-threaded builts are controled by a flag](https://cibuildwheel.pypa.io/en/stable/options/#free-threaded-support) which is off by default.

Also note that NumPy 2.1.0 has cp313 wheels for mac and linux, but not for windows, so if anyone wants to try it there, they'll also need to build numpy from source (which I guess will likely not work out of the box).

takluyver:
Right, sorry, replace 'free-threading binaries' with 'extension modules that don't trigger re-enabling the GIL'.

Thanks, I hadn't realised that it was possible to override the automatic re-enabling of the GIL. That will presumably be the easiest way to test h5py until we're ready to publish wheels that claim to be compatible with free-threading.

neutrinoceros:
I think the first piece of the puzzle would be something like
```patch
diff --git a/.github/workflows/build_wheels.yml b/.github/workflows/build_wheels.yml
index 2e9a1b6a..6fff93df 100644
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -118,6 +118,7 @@ jobs:
           TOX_TEST_LIMITED: ${{ env.TOX_TEST_LIMITED }}
           CIBW_PRERELEASE_PYTHONS: ${{ env.CIBW_PRERELEASE_PYTHONS }}
           CIBW_BEFORE_TEST: ${{ env.CIBW_BEFORE_TEST }}
+          CIBW_FREE_THREADED_SUPPORT: ${{ (github.event_name == 'schedule' || github.event_name == 'workflow_dispatch') }}
         if: steps.triage.outputs.skip != '1'

       # And upload the results
```

However, as I mentionned earlier, building requires Cython 3.1 (unreleased) and it looks like it could land soon enough that it's not worth special casing it now.
Another (small) hurdle is that setting `PYTHON_GIL=0` unconditionally doesn't work since the regular build for Python 3.13 will err with
```
Fatal Python error: config_read_gil: PYTHON_GIL / -X gil are not supported by this build
```

tacaswell:
I do not think we should mark as "free-threading-safe" until we have done some testing forcing YOLO mode (e.g. `PYTHON_GIL=0`).  I think it reasonable to expect down-stream to test in the same way until we are sure.

To start with we should add a cp313t job to the test matrix that runs the current test suite with `PYTHON_GIL=0` and see what happens!

neutrinoceros:
I except this will become significantly easier to setup soon: `actions/setup-pyhon` just merged [support for free-threaded builds](https://github.com/actions/setup-python/pull/973) a couple days ago. We just need to wait for a new release (currently, the latest one is 5.4.0)

rgommers:
> We just need to wait for a new release (currently, the latest one is 5.4.0)

FYI, v5.5.0 is available now and has full free-threading support.

> `+          CIBW_FREE_THREADED_SUPPORT: ...
`

Note that this just changed to `CIBW_ENABLE: cpython-freethreading` (see https://cibuildwheel.pypa.io/en/stable/options/#enable)

neutrinoceros:
thanks for reminding me Ralf !

rgommers:
Are you planning to wait until the Cython 3.1.0 release before merging this? Even then, it's probably pretty aggressive to require that immediately as the minimum supported version. If not, you're probably a lot better off passing the `freethreading_compatible=true` dynamically rather than in directive comments. Something along the lines of:
```python
compiler_directives = {}
if Version(cython_version) >= Version("3.1.0b1"):
    compiler_directives["freethreading_compatible"] = True

setup(
    ext_modules=cythonize(
        extensions,
        compiler_directives=compiler_directives,
    )
)
```
(from https://py-free-threading.github.io/porting-extensions/#__tabbed_1_3)

That way this PR will mergeable now, and you don't unnecessarily force distros onto a brand new Cython release.

neutrinoceros:
Thanks for the suggestion. Indeed, I intend to polish this before undrafting; bumping our Cython requirement is a shortcut I'm taking to experiment quickly, not something I want to ship.

neutrinoceros:
Failing on all platforms with error messages similar to
```
ERROR: h5py-3.13.0-cp313-cp313t-manylinux_2_17_aarch64.manylinux2014_aarch64.whl is not a supported wheel on this platform.
```

I would normally suspect an outdated `pip` to raise this error, but it also looks like cibuildwheel correctly installs the latest version (currently `25.0.1`), so I don't quite get what's going on.

neutrinoceros:
Ah ! I agree that running tox in a containerized context is weird, I just failed to realize that it *might* be where the problem was, thanks for pointing it out !
The issue is that, in our current CI setup, wheel building is 100% coupled with tox; it'd be indeed desirable to separate them, but we would either loose part of our test matrix *or* end up duplicating a lot of logic from `tox.ini`... that is, *if* we choose to keep using `tox`. I'm not a fan of this tool so I would lean towards just dropping it entirely, but obviously this needs discussion first (and I don't *expect* my admittedly radical take to be consensual).

rgommers:
> (and I don't expect my admittedly radical take to be consensual).

no idea about preferences/concensus in this project, but I'd say it's far from radical at least. I quite dislike `tox` myself, as do quite a few people I collaborate with, and I consider the design fundamentally flawed. It makes it effectively impossible to use uv/conda/pixi/etc. (and people are really attached to their tool of choice here). Imho such tooling should be layered such that you can define your test tasks without strong coupling to environment setup. `spin` has the right idea there, you can use it within any environment manager - and absent a tool like `spin` you should be able to use `pytest` directly).

In terms of environment/dependency management, there's effectively very little of interest in `tox.ini` besides the `numpy` and Python versions by the way. The more interesting/challenging dependencies seem to be HDF5, MPI, and compilers - none of which can be controlled with `tox`.

rgommers:
FYI from colleagues I heard that in a couple of projects switching from `tox` to `nox` unblocked free-threading support. We don't know conclusively that there are no issues in `nox` (there's no issue, PR or docs that touches on this) but anecdotally installing packages and running tests works. `nox` is very similar to `tox`, so perhaps a less invasive change.

neutrinoceros:
https://github.com/tox-dev/tox/issues/3391 was just resolved in tox 4.26.0, and Cython 3.1 was shipped last week. Let's see where we're at.

neutrinoceros:
This is starting to look good. It'd be great to also set `PYTHON_GIL=0`, but this can only be done *conditionally* this the gil-enabled build of Python 3.13 errors if this variable is defined (even if set to `1` ...). I think I'll post-pone this bit for now and finally undraft this first PR.

takluyver:
If I've understood this correctly, the markers and the `CIBW_ENABLE` option would combine to create wheels which claim they are compatible with free-threading (i.e. importing h5py does not trigger re-enabling the GIL)? If that is right, please hold off on merging for now. I could of course exclude the `cp313t` wheel manually when doing a release, but I'd rather not have to remember that.

I think the first step is to get some realistic testing of h5py with the GIL disabled, before we claim it's compatible. Running our test suite is a good first step - I think this PR does this, though it would be good to ensure the GIL is not getting accidentally reenabled during the test run. But I don't think most of our existing tests really exercise multiple threads. We should either look for a realistic example using threads & h5py, or create our own stress-tests doing HDF5 operations from multiple threads. The callback-based iterators like `group.visit()` and `dset.id.chunk_iter()` could be useful for this, because we can pause in the middle of an HDF5 call. It would also be a good idea to test the file-like-object support, which involves HDF5 calling back into Python at a lower level.

This doesn't necessarily need to be part of our regular test suite for now; it could be a separate script that we run manually for starters. But I think testing with `PYTHON_GIL=0` (or YOLO-mode, thanks @tacaswell ) should come before claiming compatibility.

takluyver:
Correct me if I've misunderstood, but I believe there are two levels of
free-threading compatibility: building for the cp313t ABI, and the flag
that tells Python not to re-enable the GIL when the extension module is
loaded. I think this PR currently does both, and I don't want to do the
second step just yet. People who want to take their chances and test
without the GIL can do so anyway by setting PYTHON_GIL=0. I'd like to hear
from some people trying that before going to step 2. I think the flag
within the extension modules is more prominent than docs or trove
classifiers.

In terms of testing, I think the starting point is just to do a bunch of
HDF5 operations from concurrent threads: create and delete groups,
datasets, attributes, write into chunked datasets so chunks have to be
allocated, change any flexible metadata, whatever you can think of. The
places where HDF5 can call back into Python could be a particular
challenge. I hope that having our own big lock in h5py will mean it all
works, but it's a pretty big complex thing, and I don't know where we might
be unwittingly relying on the GIL.

rgommers:
> In terms of testing, I think the starting point is just to do a bunch of HDF5 operations from concurrent threads:

[`pytest-run-parallel`](https://github.com/Quansight-Labs/pytest-run-parallel) does a lot of this automatically, so should smoke out most the obvious issues without having to write dedicated concurrent tests. See https://py-free-threading.github.io/testing/ for some more context. I'd start there; install it in your test env locally, add `--parallel-threads=2` and see what fails. Once that is happy, push it a bit (say `--parallel-threads=12`).

neutrinoceros:
Sounds reasonable. In fact, running our test suite with `pytest-run-parallel` with 4 threads immediately reveals thread safety issues (seg faults all over the place), so I'm convinced we shouldn't allow downstream testing without `PYTHON_GIL=0` just yet.

neutrinoceros:
Actually, if I only use 2 threads, I get clearer error messages.
Most of the exceptions look like these
```
(...)
PARALLEL FAILED h5py/tests/test_group.py::TestAdditionalMappingFuncs::test_update_dict - OSError: Unable to create link (name already exists)
PARALLEL FAILED h5py/tests/test_group.py::TestAdditionalMappingFuncs::test_update_iter - OSError: Unable to create link (name already exists)
PARALLEL FAILED h5py/tests/test_group.py::TestAdditionalMappingFuncs::test_update_kwargs - OSError: Unable to create link (name already exists)
PARALLEL FAILED h5py/tests/test_group.py::TestGet::test_get_class - ValueError: Unable to synchronously create group (name already exists)
PARALLEL FAILED h5py/tests/test_group.py::TestGet::test_get_default - ValueError: Unable to synchronously create group (name already exists)
PARALLEL FAILED h5py/tests/test_group.py::TestGet::test_get_link - ValueError: Unable to synchronously create group (name already exists)
PARALLEL FAILED h5py/tests/test_group.py::TestGet::test_get_link_class - ValueError: Unable to synchronously create group (name already exists)
(...)
```

I think they are all legit exceptions from HDF5 itself ? I guess these tests should be marked as `@pytest.mark.thread_unsafe` (defined in `pytest-run-parallel`) for now, and we should really be focusing on everything *else* that's failing.
I'm going to run experiments in that direction in a different branch.

neutrinoceros:
It's clear now that most of this PR is premature, so, switching back to draft mode.

takluyver:
Yeah, if two threads try to create an object in the same file with the same name, it's expected that one of them will fail (with or without the GIL in play). That's not *unsafe*, just the tests are written with an assumption that doesn't hold when running multiple of the same in parallel.

neutrinoceros:
Of course, my mistake. I'm trying to get rid of this assumption in #2588

djhoese:
I'm bouncing around my dependencies checking on 3.14 and free-threading support and found this PR. One thing you could add to this PR is the free-threading classifier in pyproject.toml:

```
Programming Language :: Python :: Free Threading :: 2 - Beta
```

See https://py-free-threading.github.io/porting/#define-and-document-thread-safety-guarantees

Thanks for working on this!

neutrinoceros:
at this stage I would rather go with `Programming Language :: Python :: Free Threading :: 1 - Unstable`, and even that is premature in our case, but I'll circle back to it !

crusaderky:
I've branched off this PR and opened https://github.com/h5py/h5py/pull/2650 to pull this through the finishing line

neutrinoceros:
closing in favor of https://github.com/h5py/h5py/pull/2650

Issue #2445, opened 2024-06-12
Use of IF in Cython is deprecated


larsoner:
Building with latest Cython and HDF5 you'll see tens/hundreds of warnings of the form:
```
warning: h5py/_objects.pyx:215:12: The 'IF' statement is deprecated and will be removed in a
future Cython version. Consider using runtime conditions or C macros instead. See
https://github.com/cython/cython/issues/4310
```
These make the GH actions logs difficult to navigate.

I had a brief look at the `h5py` code and the solution wasn't immediately obvious, so I also opened https://github.com/cython/cython/pull/6243 to see if we can clean up the build logs in the interim.

takluyver:
There's a  fair bit of discussion (not specific to use in h5py) at https://github.com/cython/cython/issues/4310 . It seems like Cython depreecacted something that quite a lot of people are using without a simple way to replace it. There was even [a comment in 2022](https://github.com/cython/cython/issues/4310#issuecomment-1245862843) saying that it was OK to ignore it for now until the Cython developers had worked out a better way forwardss. I had hoped that the deprecation might be reconsidered, but unfortunately that doesn't seem to be happening.

There was also a proposal (https://github.com/cython/cython/issues/2596 ) to integrate a templating tool, which I think could address similar use cases, but it seems like it's not high on anyone's priority list. :confused:

neutrinoceros:
If I recall correctly, the original plan was for Cython to remove `IF/ELSE` in version 3.1, and maybe they changed plans, but it would seem appropriate to set our upper bound to `<3.1` for now. In my opinion that's the only thing we can do about it on our end and it would be sufficient to close this ticket until further notice.

neutrinoceros:
that's probably good enough until the plan is better defined. Should we close this then ?

larsoner:
If the latest `dev` version of Cython is used (like in #2449), now that https://github.com/cython/cython/pull/6243 has landed the warnings could at least be suppressed. But I'd suggest to leave this issue open in any case as a reminder to think about / address it at some point.

neutrinoceros:
I've studied this a bit more closely and here's what I found:
`api_gen.py` is generating Cython code and making use of `IF` statements here:
https://github.com/h5py/h5py/blob/7d4c3a15fc53f7d0a5b6c8e9895a9b3fb9632f12/api_gen.py#L213

Importantly, it'll wrap many blocks in `IF HDF5_VERSION >= (x, y, z):`
[...]

PR #2479, opened 2024-08-20
BLD: call `api_gen.run` unconditionally

neutrinoceros:
As discussed in #2445, and as suggested in https://github.com/h5py/h5py/issues/2445#issuecomment-2296653468
This is one small step towards solving #2445; the next logical step is to stop using Cython's `IF` in the generated `defs.pyx`.

### DIFF
--- a/api_gen.py
+++ b/api_gen.py
@@ -25,7 +25,22 @@
 """
 
 import re
-import os.path as op
+import os
+from hashlib import md5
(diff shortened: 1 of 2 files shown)


tacaswell (approved):

tacaswell:
It looks like travis is timing out on building the env to run the tests.

PR #2480, opened 2024-08-20
BLD: only generate Cython code that will compile against current configuration

neutrinoceros:
Build on top of #2479 and partially address #2445. This eliminates about 60% of Cython warnings related to the deprecation of `IF` in my case.


takluyver:
Thanks, this looks good to me. :+1:

PR #2485, opened 2024-08-23
DOC: unlimited maxshape for 1D Datasets

scimax:
I ran into a minor lack of documentation, which might actually be pretty obvious for advanced users, and I also figured it out quite fast, but I anyway wanted to ask if it is of interest, to extend the documentation on this level. Just to illustrate the problem, here a MWE:

```py
N_tsvd_iterations = 10
with h5py.File("foobar.h5", 'w-') as f:
    dset = f.create_dataset('test_data', (N_tsvd_iterations,), maxshape=None, 
                      chunks=True, dtype=np.float64)
    print(dset.shape)
    print(dset.maxshape)
    dset = f.create_dataset('test_data_2', (N_tsvd_iterations,), maxshape=(None,), 
                      chunks=True, dtype=np.float64)
    print(dset.shape)
    print(dset.maxshape)
```
The output reads as follows:
```
(10,)
(10,)
(10,)
(None,)
```

While it is somewhat obvious, since passing `maxshape=None` is identical to not passing it at all, one may anyways run into the issue when extending a finite `maxshape` to an unlimited one, i. e. naively replacing the given finite integer with `None`.

Therefore, I mentioned this case both in the docstring of `create_dataset` and in the main documentation of `Dataset`. Maybe there are better ways of handling this special case, but for the moment, I thought it would be good to just document it.

What do you think about this case?

PS: I did not run any tests as asked in the template as I didn't change any python code, only docs. Should I anyway do it? Probably it would make more sense to test build the documentation, wouldn't it?

### DIFF
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -214,6 +214,11 @@ axes using ``None``::
 
     >>> dset = f.create_dataset("unlimited", (10, 10), maxshape=(None, 10))
 
+While integers can be passed to ``maxshape`` for one-dimensional datasets, a tuple
+of length 1 must be passed for an unlimited maximum shape of the 1D-shaped dataset,
+i.e. ``(None,)``. Using only ``None`` instead of an integer will lead to the default 
+behaviour, having a ``maxshape`` identical to ``shape``.
+
 .. note:: Resizing an array with existing data works differently than in NumPy; if
     any axis shrinks, the data in the missing region is discarded.  Data does
     not "rearrange" itself as it does when resizing a NumPy array.
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -214,7 +214,7 @@ axes using ``None``::
 
     >>> dset = f.create_dataset("unlimited", (10, 10), maxshape=(None, 10))
 
-For a 1D dataset, ``maxshape`` can be an integer instead of a tuple. But to make an 
+For a 1D dataset, ``maxshape`` can be an integer instead of a tuple. But to make an
 unlimited 1D dataset, ``maxshape`` must be a tuple ``(None,)``. Passing ``None`` gives
 the default behaviour, where the initial size is also the maximum.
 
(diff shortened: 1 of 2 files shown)


docs/high/dataset.rst
```diff
@@ -214,6 +214,11 @@ axes using ``None``::
 
     >>> dset = f.create_dataset("unlimited", (10, 10), maxshape=(None, 10))
 
+While integers can be passed to ``maxshape`` for one-dimensional datasets, a tuple
+of length 1 must be passed for an unlimited maximum shape of the 1D-shaped dataset,
+i.e. ``(None,)``. Using only ``None`` instead of an integer will lead to the default
```
miccoli, line 219: instead of a **tuple**

scimax: > instead of a **tuple**


Makes sense, but I rather meant when you want to turn a limited maximum shape, given by an integer, into an unlimited maximum shape. In this case you have to replace the integer with a tuple, not with ``None``. Probably I have rephrase it to make it clear.

miccoli: (disclaimer: I'm not a maintainer, so mine is just a user-opinion ;-)
May I propose:

Please note that for unlimited one-dimensional datasets the `(None, )` tuple cannot be substituted by a bare `None` because this would be interpreted as the default value of `maxshape`, i.e. having a `maxshape` identical to `shape`.

Feel free to improve my English, which I know is pretty poor.

scimax: I like this one! I will edit the docs accordingly!

scimax: Sorry, it took me almost forever. It just fell back in priorities of my tasks. I'm happy that I could keep track of it. (sorry for the poor commit message. I was just too quick)

h5py/_hl/group.py
```diff
@@ -89,7 +89,10 @@ def create_dataset(self, name, shape=None, dtype=None, data=None, **kwds):
 
         maxshape
             (Tuple or int) Make the dataset resizable up to this shape. Use None for
-            axes you want to be unlimited. Integers can be used for 1D shape.
+            axes within the tuple you want to be unlimited. Integers can be used for 1D shape.
+            For 1D datsets with quasi-unlimited maxshape, a shape tuple of length 1 must be
```
takluyver, line 93:
```suggestion
            For 1D datsets with unlimited maxshape, a shape tuple of length 1 must be
```

I know 'quasi-unlimited' is technically more accurate, but I think it's a distraction from the point we're trying to make. No-one is assuming that passing `(None,)` somehow gives you infinite storage. :wink:

PR #2486, opened 2024-08-23
Increase minimum supported HDF5 to 1.10.6

takluyver:
Gets rid of a few more uses of Cython `IF`. The tag for 1.10.6 is from December 2019, so this is still pretty cautious.

:crossed_fingers: I haven't broken Travis again.


takluyver:
CI picked up that Ubuntu 20.04 still has HDF5 1.10.4, which I had missed. I think I'm OK with dropping support for that at this point.

h5py/h5p.pyx
```diff
@@ -1313,8 +1313,8 @@ cdef class PropFAID(PropInstanceID):
         Set the compatibility level for file format. Legal values are:
 
         - h5f.LIBVER_EARLIEST
-        - h5f.LIBVER_V18 (HDF5 1.10.2 or later)
-        - h5f.LIBVER_V110 (HDF5 1.10.2 or later)
+        - h5f.LIBVER_V18
+        - h5f.LIBVER_V110
         - h5f.LIBVER_V112 (HDF5 1.11.4 or later)
```
tacaswell, line 1318:
```suggestion
        - h5f.LIBVER_V112
```

h5py/h5p.pyx
```diff
@@ -1346,8 +1346,8 @@ cdef class PropFAID(PropInstanceID):
         Get the compatibility level for file format. Returned values are from:
 
         - h5f.LIBVER_EARLIEST
-        - h5f.LIBVER_V18 (HDF5 1.10.2 or later)
-        - h5f.LIBVER_V110 (HDF5 1.10.2 or later)
+        - h5f.LIBVER_V18
+        - h5f.LIBVER_V110
         - h5f.LIBVER_V112 (HDF5 1.11.4 or later)
```
tacaswell, line 1351:
```suggestion
        - h5f.LIBVER_V112
```

tacaswell (approved):
Two minor comments about stripping additional conditionals in docstrings.

h5py/h5p.pyx
takluyver: These notes are still relevant as long as we support any HDF5 1.10 version, no? I could simplify them to say 1.12 or later, since I assume no-one is now using the 1.11 development series.

takluyver: I guess we could also get rid of the comments altogether & just say that any specific version number won't be available if you're using an older HDF5.

tacaswell: oh, 🤦🏻 I missread these comments as 1.10.4.

PR #2490, opened 2024-08-30
WHL: don't skip cp39 nightly wheels

neutrinoceros:
Fix an issue in scheduled jobs where skipping cp39 wheels results in no identifier being selected on aarch64 ([example logs](https://github.com/h5py/h5py/actions/runs/10500649952/job/29089357892))
In fact the associated comment is out of date so I don't think we *need* to skip it.


neutrinoceros:
This is inherently impossible to test in normal CI since it only affects conditions on a scheduled run, so I'll merge now and check that everything's alright tomorrow.

PR #2491, opened 2024-08-31
WHL: remove broken job for cp313 wheels on aarch64 Linux

neutrinoceros:
Follow up to #2490, revert #2469 (this never worked)

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -56,7 +56,6 @@ jobs:
           - "3.10"
           - "3.11"
           - "3.12"
-          - "3.13.0rc1"
         include:
           - os: ubuntu-latest
             arch: x86_64


neutrinoceros:
This wheel can be re-enabled after Python 3.13 final is out. This means that if we release h5py 3.12 before that, this wheel will be missing (at first). We should still be able to upload it manually after the fact if we want to.
FYI @takluyver

takluyver:
Fine by me - do we understand what's not working? (I haven't looked at the logs yet)

neutrinoceros:
Kinda. Here's the error:
```
No build identifiers selected: BuildSelector(build_config='cp3130rc1-*_aarch64', skip_config='pp* *musllinux*', requires_python=<SpecifierSet('>=3.9')>, prerelease_pythons=False, free_threaded_support=False)
```

there's a obvious conflict between `build_config='cp3130rc1-*_aarch64'` and `prerelease_pythons=False`, but I've tried changing the latter to `True`, to no avail. I think the real problem is that the identifier `cp3130rc1` is invalid (it should be `cp313`), which I believe is produced by the following line https://github.com/h5py/h5py/blob/e558a348fecb95aa951ea81ce5a29bab12b27e68/ci/triage_build.sh#L50
Maybe this is easy to fix in bash but I don't feel competent here. Just in case, let me ping @larsoner too

takluyver:
Thanks, that makes sense. I always have to look up how to do these manipulations in bash ([Shell Parameter Expansion](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html)), but I think something like this could strip off the part we don't need:

```shell
PYTHON="${PYTHON%.0rc*}"
```

PR #2499, opened 2024-09-18
BUG: hotfix portability for ppc64(le) archs

neutrinoceros:
This is meant to addressed a regression caused by #2482
Hopefully I got it right this time; my main mistake was that I misread the documentation and thought `os.uname()` returned a string when it really returns a 5 elements tuple (the last of which being the string I was after) 

I switched from `os.uname` to `platform.uname` for two reasons:
- it closer mimics how `UNAME_MACHINE` is defined in Cython: https://github.com/cython/cython/blob/master/Cython/Compiler/Scanning.py#L88-L89
- it is documented as "fairly portable" (while `os.uname` is only available on UNIX)

I'll see if CI pass with the minimal patch first, but I expect I should be able to simplify the logic here by avoiding special-casing Windows.

### DIFF
--- a/h5py/h5t.pyx
+++ b/h5py/h5t.pyx
@@ -28,7 +28,7 @@ from .utils cimport  emalloc, efree, require_tuple, convert_dims,\
 
 # Python imports
 import codecs
-import os
+import platform
 import sys
 from collections import namedtuple
 import numpy as np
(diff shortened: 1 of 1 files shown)


neutrinoceros:
First check passed ! now let's do the more aggressive simplification

neutrinoceros:
@takluyver would you do me the honor ?

takluyver:
Gladly :smile:

PR #2502, opened 2024-09-22
RFC: migrate away from deprecated old-style Cython properties

### LINKED ISSUE #2501: RFC: migrate out of deprecated Cython proprety syntax
We have about 100 use of old-style `propertry` in Cython files. This style is documented as deprecated
https://cython.readthedocs.io/en/latest/src/userguide/extension_types.html#properties
(in fact, the reason I learned about it is that my code editor highlights these, apparently)

The [changelog](https://cython.readthedocs.io/en/latest/src/changes.html#id262) specifies that this style is deprecated since Cython 0.24, and although there's no clear expiration date, we could probably get started on this before it becomes an emergency.

neutrinoceros:
Close #2501

### DIFF
--- a/h5py/_objects.pyx
+++ b/h5py/_objects.pyx
@@ -170,17 +170,17 @@ cdef class ObjectID:
     cdef size_t _pyid
     """
 
-    property fileno:
-        def __get__(self):
-            cdef H5G_stat_t stat
-            with _phil:
-                H5Gget_objinfo(self.id, '.', 0, &stat)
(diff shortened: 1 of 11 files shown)


neutrinoceros:
To reviewers: I recommend turning off whitespace diff in GitHub for this one (most of the diff is code being de-dented)

aragilar:
Nice, this makes the code so much more readable! Thanks @neutrinoceros.

Issue #2505, opened 2024-09-25
h5py 3.12: import error on Windows


robomics:
Attempting to import a freshly installed version of `h5py` on Windows results in the following import error:

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "C:\hostedtoolcache\windows\Python\3.[12](https://github.com/robomics/2024-09-25-h5py-bugreport/actions/runs/11038497694/job/30661854085#step:5:13).6\x64\Lib\site-packages\h5py\__init__.py", line 25, in <module>
    from . import _errors
ImportError: DLL load failed while importing _errors: The specified module could not be found.
```

I managed to reproduce this bug on `windows-latest` images on GitHub Actions using Python 3.9-3.12.

See the most recent workflow run [here](https://github.com/robomics/2024-09-25-h5py-bugreport/actions/workflows/ci.yml) for more details.

takluyver:
Based on the number of reactions this has got in a few hours, I've yanked the 3.12.0 release for now until we have a better idea what's going on. I may still un-yank it again if it turns out that only a small fraction of Windows users are affected, although I don't know how I'd find that out.

I'm about to go to bed, and it's not easy for me to debug Windows issues even when I'm awake, since I don't have a convenient Windows machine to test on. So it would be good if some of the affected Windows users can investigate further.

Some clues:

-  Yanked releases aren't deleted, you can still download them [from PyPI](https://pypi.org/project/h5py/3.12.0/#files), or by specifying `pip install h5py==3.12.0`.
- We changed the CI platform on which we build wheels since 3.11, so it's likely that something to do with the new Github actions CI is the culprit. #2444 was the main PR for this.
- We do test the wheels after building them ([e.g. this CI run](https://github.com/h5py/h5py/actions/runs/11036731225/job/30655879090)), and they're passing the tests in our CI.
- I know of an ancient but useful tool for investigating these DLL load errors: https://dependencywalker.com/

takluyver:
Actually, it might be quite simple. When building Windows wheels on the old CI, we called a script `bundle_hdf5_whl.py`:

https://github.com/h5py/h5py/blob/3.11.0/ci/azure-pipelines-wheels.yml#L59

I don't think this is called anywhere in the new CI. That's meant to copy the HDF5 DLLs into the Windows wheels (like a simplified version of `auditwheel` on Linux or `delocate` on Mac).

[...]

PR #2507, opened 2024-09-26
WHL: hotfix missing dll in Windows wheels

neutrinoceros:
This does two things to address #2505:
- tentatively fix windows wheels by running the `ci/bundle_hdf5_whl.py` script as hinted to by @takluyver (https://github.com/h5py/h5py/issues/2505#issuecomment-2375396971). This is kind of a stab in the dark as it doesn't change how the wheel is *tested*, but this aligns with how it was done for h5py 3.11 so maybe it's good enough as a hotfix ?
- adjust a broken requirement on `numpy>=2.1.0dev` for nightly wheels (which cannot work on CPython 3.9 and prevented building and uploading most of our recent nightlies)

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -121,6 +121,9 @@ jobs:
           CIBW_BEFORE_TEST: ${{ env.CIBW_BEFORE_TEST }}
         if: steps.triage.outputs.skip != '1'
 
+      - run: python ./ci/bundle_hdf5_whl.py wheelhouse
+        if: steps.triage.outputs.skip != '1' && runner.os == 'Windows'
+
       # And upload the results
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks! It looks like the bundle script ran successfully in CI. I'd like at least one Windows user to try a wheel built from this PR and confirm that it works, then I'll prepare a 3.12.1 release.

neutrinoceros:
Just a note that *if* nobody steps up to test it from gh artifacts we could also merge the PR and wait for people to test next nightlies.

takluyver:
If no-one tests it from the PR, I'm not sure anyone will test it from the nightlies either. I might just make the release, that will get people testing it. :smiling_imp:

Issue #2446, opened 2024-06-13
Broken Links on the documentation


tabedzki:
There are broken links under the Compression part of the [Dataset page](https://docs.h5py.org/en/stable/high/dataset.html#custom-compression-filters).

ajelenak:
HDF Group's documentation links are undergoing changes. h5py documentation will be updated once the links are stable again.

One of the broken links is currently at https://github.com/HDFGroup/hdf5_plugins/blob/master/docs/RegisteredFilterPlugins.md.

neutrinoceros:
@ajelenak I wasn't aware these changes were on-going. Do you happen to know whether #2440 and #2443 are stable already ?

ajelenak:
The forum links are ok, but the portal link will go away.

ajelenak:
This issue should have been closed when the above PR was merged so doing it now manually.

PR #2512, opened 2024-10-03
Update HDF Group’s links

ajelenak:
This updates the HDF Group's links in the h5py documentation. I failed to find replacements for two old links but the rest should hopefully last for a long time. 🤞

### DIFF
--- a/docs/high/file.rst
+++ b/docs/high/file.rst
@@ -381,8 +381,8 @@ chunk cache*. They apply to all datasets unless specifically changed for each on
   ``rdcc_nbytes`` bytes. For maximum performance, this value should be set
   approximately 100 times that number of chunks. The default value is 521.
 
-Chunks and caching are described in greater detail in the `HDF5 documentation (TODO link)
-<.>`_.
+Chunks and caching are described in greater detail in the `HDF5 documentation
+<https://support.hdfgroup.org/documentation/hdf5-docs/advanced_topics/chunking_in_hdf5.html>`_.
 
 .. _file_alignment:
 
--- a/docs/high/file.rst
+++ b/docs/high/file.rst
@@ -401,7 +401,7 @@ takes effect and the alignment in bytes within the file. The alignment is
 measured from the end of the user block.
 
 For more information, see the official HDF5 documentation `H5P_SET_ALIGNMENT
-<https://portal.hdfgroup.org/display/HDF5/H5P_SET_ALIGNMENT>`_.
+<https://support.hdfgroup.org/documentation/hdf5/latest/group___f_a_p_l.html#gab99d5af749aeb3896fd9e3ceb273677a>`_.
 
 .. _file_meta_block_size:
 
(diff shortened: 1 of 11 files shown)


ajelenak:
Fixes #2446.

docs/high/file.rst
```diff
@@ -381,8 +381,8 @@ chunk cache*. They apply to all datasets unless specifically changed for each on
   ``rdcc_nbytes`` bytes. For maximum performance, this value should be set
   approximately 100 times that number of chunks. The default value is 521.
 
-Chunks and caching are described in greater detail in the `HDF5 documentation
-<https://portal.hdfgroup.org/display/HDF5/Chunking+in+HDF5>`_.
+Chunks and caching are described in greater detail in the `HDF5 documentation (TODO link)
```
takluyver, line 384: https://portal.hdfgroup.org/documentation/hdf5-docs/advanced_topics/chunking_in_hdf5.html ?

ajelenak: portal -> support: https://support.hdfgroup.org/documentation/hdf5-docs/advanced_topics/chunking_in_hdf5.html. This link is also going to go away soon but let's use while it is still working. I will make this change.

docs/related_projects.rst
```diff
@@ -16,8 +16,8 @@ the two projects (see :ref:`h5py_pytable_cmp`), as does the
 
 .. seealso::
 
-   `HDF Group's list of HDF5 tools
-   <https://portal.hdfgroup.org/display/HDF5/HDF5+Tools+by+Category>`_
+   `HDF Group's list of HDF5 tools (TODO link)
```
takluyver, line 19: I can still see the old version of this page, with a deprecation notice, at https://docs.hdfgroup.org/archive/support/products/hdf5_tools/toolsbycat.html . The notice says it 'will remain online', though presumably at some point it will break.

https://portal.hdfgroup.org/documentation/hdf5/latest/_view_tools_command.html looks roughly similar on the new site, though it only covers command-line tools, not GUIs. :shrug:

docs/high/file.rst
takluyver: :face_exhaling: The portal links seem to be the ones that search engines find. If you want to change that, I believe that either HTTP 301 redirects or adding a `rel="canonical"` link should tell search engines which location is preferred.

Thanks!

docs/related_projects.rst
ajelenak: > I can still see the old version of this page, with a deprecation notice, at https://docs.hdfgroup.org/archive/support/products/hdf5_tools/toolsbycat.html . The notice says it 'will remain online', though presumably at some point it will break.

https://support.hdfgroup.org/archive/support/products/hdf5_tools/toolsbycat.html This is the old support website, not updated in 7 years, and never will. Definitely not to be used for anything.

> https://portal.hdfgroup.org/documentation/hdf5/latest/_view_tools_command.html looks roughly similar on the new site, though it only covers command-line tools, not GUIs. 🤷

https://support.hdfgroup.org/documentation/hdf5/latest/_view_tools_command.html This one is part of the libhdf5 documentation and I don't know if it will ever include third-part tools. That's why I did not want to use it.

takluyver: Fair enough, in that case let's just remove this seealso entirely.

takluyver:
Thanks!

Issue #2514, opened 2024-10-07
h5py 3.12: `KeyError` with in-memory files using `io.BytesIO` and `ExternalLink`


paulmueller:
With h5py 3.12, I am encountering KeyErrors in an edge case where I am using `h5py.File` on top of `io.BytesIO` and writing `ExternalLink` to that in-memory file.

Since this code worked with h5py 3.11 and earlier, I believe this is a bug. Please let me know if I am doing something fundamentally wrong here and I will adapt my code instead :+1: .

Here is a script to reproduce the errors (works with h5py 3.11 and fails with h5py 3.12) including the error messages I get as comments. The difference between the two methods is that in the first case, the link is created in the root of the in-memory file, and in the second case the link is created in a subgroup.

```python

import io
import pathlib

import h5py
import numpy as np
import tempfile


def test_unable_to_open_object_component_not_found():
    path = pathlib.Path(tempfile.mkdtemp(
        prefix="h5py_test_unable_to_open_object_component_not_found_")) / "test.rtdc"
    fd = io.BytesIO()

    with h5py.File(path, "w", libver="latest") as h5:
        h5["subdir/data"] = np.arange(100)

    with h5py.File(fd, "w", libver="latest") as h5:
        h5["data"] = h5py.ExternalLink(str(path), f"/subdir/data")

    with h5py.File(fd, "r") as h5:
        # The following raises:
        #   File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
        #   File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
        #   File "/home/paul/repos-work/DC/dclab/.env/lib/python3.11/site-packages/h5py/_hl/group.py", line 357, in __getitem__
        #     oid = h5o.open(self.id, self._e(name), lapl=self._lapl)
        #           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        #   File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
        #   File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
        #   File "h5py/h5o.pyx", line 257, in h5py.h5o.open
        # KeyError: 'Unable to synchronously open object (component not found)'
        print(h5["data"])


def test_cant_find_object_too_many_links():
    path = pathlib.Path(tempfile.mkdtemp(
[...]

PR #2517, opened 2024-10-09
Add File.in_memory() alternative constructor

takluyver:
From discussion on #2514, this adds a high-level way of working with 'files' in memory.

```python
# Start a new HDF5 file in memory
f = h5py.File.in_memory()
f['a'] = [1, 2, 3]

# Get the file data as bytes, e.g. to send over the network
f.flush()
hdf_data = f.id.get_file_image()

# Turn the bytes back into an h5py File object
f2 = h5py.File.in_memory(hdf_data)
```

Questions:

1. Looking at [the code in the core driver](https://github.com/HDFGroup/hdf5/blob/bcc795fcf66f5ce72b76dab4b4df5136f2558303/src/H5FDcore.c#L752-L754), there's an annoying check that the given filename doesn't exist on the filesystem. We bypass this when creating a new 'file', but I can't see a way to avoid it when we're passing in a file image.

```c
    if ((file_image_info.buffer != NULL) && !(H5F_ACC_CREAT & flags)) {
        if ((fd = HDopen(name, o_flags, H5_POSIX_CREATE_MODE_RW)) >= 0) {
            HDclose(fd);
            HGOTO_ERROR(H5E_FILE, H5E_FILEEXISTS, NULL, "file already exists");
        }
```

For now I just use a filename that's unlikely to exist - `h5py_in_memory_nonfile_<N>`. We could create a temporary directory to almost guarantee that we pass a filename that doesn't exist. But the aim is to avoid touching the filesystem, so that doesn't seeem ideal.

2. I found after writing this that there's an `H5LTopen_file_image()` function in the high-level library, exposed in h5py's low-level API as `h5py.h5f.open_file_image()`. This has options to avoid unnecessary copying of the file data, but these options are not yet exposed, and I didn't fancy playing with reference counts to ensure the image data is freed at the right time. Without that option, it looks like it does essentially the same as this code.
3. I haven't included a read-only option; in-memory files are always writable. It wouldn't be hard to add read-only mode if we want to.

### DIFF
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -11,8 +11,9 @@
     Implements high-level support for HDF5 file objects.
 """
 
-import sys
+import inspect
 import os
+import sys
 from warnings import warn
 
 from .compat import filename_decode, filename_encode
@@ -108,10 +109,12 @@ def registered_drivers():
     return frozenset(_drivers)
 
 
-def make_fapl(driver, libver, rdcc_nslots, rdcc_nbytes, rdcc_w0, locking,
-              page_buf_size, min_meta_keep, min_raw_keep,
-              alignment_threshold, alignment_interval, meta_block_size,
-              **kwds):
+def make_fapl(
+    driver, libver=None, rdcc_nslots=None, rdcc_nbytes=None, rdcc_w0=None,
+    locking=None, page_buf_size=None, min_meta_keep=0, min_raw_keep=0,
+    alignment_threshold=1, alignment_interval=1, meta_block_size=None,
+    **kwds
+):
     """ Set up a file access property list """
     plist = h5p.create(h5p.FILE_ACCESS)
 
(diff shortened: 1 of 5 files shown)


h5py/_hl/files.py
```diff
@@ -567,6 +570,53 @@ def __init__(self, name, mode='r', driver=None, libver=None, userblock_size=None
...
+        """
+        for k in ('driver', 'locking', 'backing_store'):
+            if k in kwargs:
+                raise TypeError(
+                    f"File.in_memory() got an unexpected keyword argument {k!r}"
+                )
+        fcpl_kwargs = {}
+        for k in inspect.signature(make_fcpl).parameters:
+            if k in kwargs:
+                fcpl_kwargs = kwargs.pop(k)
```
tacaswell, line 597:
```suggestion
                fcpl_kwargs[k] = kwargs.pop(k)
```

takluyver: Thanks, good catch. I'll ensure this is covered in tests at the same time.

tacaswell (approved):

tacaswell:
The failures were upload to codecov, but restarted them be careful.

takluyver:
I've opened #2558 for the libver issue I spotted.

Issue #2451, opened 2024-07-01
Memory leak when updating attrs in newly created group


MatthieuDartiailh:
# Description

I have a system that open-close multiple times the same file and each time create a new group and update its attributes. While this pattern may not be ideal, it works BUT it does leak memory. On my machine running the attached MRE lead to a total memory increase of about 2.2 Mo. If I comment out the attrs update the memory increase is below 100 ko (I guess Python is shuffling around some stuff which explain this). 

I tried to (but it does not change anything):
- create_group instead of require_group and to pass track_order=True
- use attrs.create
- directly set the attribute `attrs["a"] = "b"`

# MRE

```python
import gc
import psutil
import h5py

print(h5py.version.info)

with h5py.File("test.h5", "w") as f:
    pass

ref = psutil.Process().memory_full_info().uss
for i in range(1000):
    with h5py.File("test.h5", "r+") as f:
        g = f.require_group(f"i{i}")
        g.attrs.update({"a": "b"})
    gc.collect()
    if i % 10 == 0:
        print("Memory usage: ", psutil.Process().memory_full_info().uss - ref)

```

# Version info

h5py    3.11.0
HDF5    1.14.2
Python  3.11.3 | packaged by Anaconda, Inc. | (main, Apr 19 2023, 23:46:34) [MSC v.1916 64 bit (AMD64)]
sys.platform    win32
sys.maxsize     9223372036854775807
numpy   1.26.4
cython (built with) 3.0.10
numpy (built against) 2.0.0rc1
HDF5 (built against) 1.14.2

MatthieuDartiailh:
Trying more variations I noticed the following:
- storing an integer or an np.void() reduce the leakage to about 100 ko
- using tracemalloc, it appears when storing strings or bytes we leak objects from encodings/ascii, when storing integer I still see large counts for Group and Attrs objects but I am not sure I fully understand how to read tracemalloc outputs.

tacaswell:
This may be a leak in libhdf5 as I can not reproduce in with 1.14.4 (h5py + hdf5 from Arch) but I do see it with 1.14.2 (from our wheels) and with conda-forge packages (1.14.3).

I think the best fix here is to do a new release and build against 1.14.4.

@ajelenak Is this a known bug/fix on the libhdf5 side?

jhendersonHDF:
[...]

PR #2520, opened 2024-10-29
Bundle hdf5 1.14.5

takluyver:
From discussion on #2451, I noticed that we'd got different versions on different platforms.

The corresponding update for Linux is at https://github.com/h5py/hdf5-manylinux/pull/14


takluyver:
Hmm, it fails to build HDF5 on Windows:

```text
CMake Error at CMakeFilters.cmake:109 (get_filename_component):
  get_filename_component unknown component
  D:/a/h5py/h5py/zlib-msvc-x64/build/native/lib_release/zlib.lib
```

takluyver:
https://github.com/HDFGroup/hdf5/issues/5026

takluyver:
```
   _______________ TestUnicode.test_unicode_hdf5_python_consistent _______________
  
  self = <h5py.tests.test_file.TestUnicode testMethod=test_unicode_hdf5_python_consistent>
  
      def test_unicode_hdf5_python_consistent(self):
          """ Unicode filenames can be used, and seen correctly from python
          """
          fname = self.mktemp(prefix=chr(0x201a))
          with File(fname, 'w') as f:
              print(os.listdir(self.tempdir))
  >           assert os.path.exists(fname)
  E           AssertionError: assert False
  E            +  where False = <function exists at 0x000001D0D315F670>('C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\h5py-test_vfneg1eg\\‚a_0fq1ds.hdf5')
  E            +    where <function exists at 0x000001D0D315F670> = <module 'ntpath' from 'C:\\Users\\runneradmin\\AppData\\Local\\pypa\\cibuildwheel\\Cache\\nuget-cpython\\python.3.9.13\\tools\\lib\\ntpath.py'>.exists
  E            +      where <module 'ntpath' from 'C:\\Users\\runneradmin\\AppData\\Local\\pypa\\cibuildwheel\\Cache\\nuget-cpython\\python.3.9.13\\tools\\lib\\ntpath.py'> = os.path
  
  py39-test-deps\lib\site-packages\h5py\tests\test_file.py:677: AssertionError
  ---------------------------- Captured stdout call -----------------------------
  ['9zab0xea.hdf5', 'jt640bo2.hdf5', 'lr6fxx8b.hdf5', 'â€š18x70rit.hdf5', 'â€ša_0fq1ds.hdf5']
```

h5py/tests/test_file.py
```diff
@@ -673,7 +673,8 @@ def test_unicode_hdf5_python_consistent(self):
         """
         fname = self.mktemp(prefix=chr(0x201a))
         with File(fname, 'w') as f:
-            self.assertTrue(os.path.exists(fname))
+            print(os.listdir(self.tempdir))
+            assert os.path.exists(fname)
```
tacaswell, line 677:
```suggestion
        assert os.path.exists(fname)
```

maybe check after we have closed the file?

tacaswell:
This feel very familiar and an issue with filesystems on windows having slightly different semantics about when things are visible than *nix....

takluyver:
Maybe, but two of the files `os.listdir()` sees in the directory start with `â€š`, which is what the character we're using turns into if you mess up encodings like this:

```
>>> chr(0x201a).encode('utf-8').decode('cp1252')
'â€š'
```

takluyver:
A change in HDF5 1.14.4 has essentially broken passing UTF-8 filenames to HDF5 on Windows. :disappointed: 

https://github.com/HDFGroup/hdf5/issues/5037

jhendersonHDF:
@takluyver I wonder if this happens to be the same issue in https://github.com/HDFGroup/hdf5/issues/4599?

takluyver:
Yup, I think that looks like the same thing.

takluyver:
The latest commit ([b72c52b](https://github.com/h5py/h5py/pull/2520/commits/b72c52b26e46f3a06ac64b0f4f1da07bf38c0133)) goes back to using the Windows code-page encoding on HDF5 1.14.4 and above (hopefully we can add an upper bound to that in the future). I think this is the right thing for h5py to do in isolation.

The bit I'm not sure about is whether we go ahead with bundling 1.14.5 with a known regression (but presumably other fixes), or wait and hope that a new HDF5 release soon will fix this.

tacaswell:
I am in favor of shipping the new version of libhdf5 as, if I understand this correctly, the issue is between us, hdf5 and the windows file system such that the resulting files (other than the name of the file) is not changed.

takluyver:
Shipping the new version means Windows users will no longer be able to create or open HDF5 files with a path containing characters outside their current codepage. E.g. users in English speaking & Western European countries can use [a selection of accented letters and common symbols](https://en.wikipedia.org/wiki/Windows-1252#Codepage_layout), but not other alphabets like Greek, Cyrillic or Japanese. This was the status before h5py 3.0.

I'm happy enough with that if you are. Obviously I'd document it in the release notes.

tacaswell:
Ah, I was misunderstanding what b72c52b26e46f3a06ac64b0f4f1da07bf38c0133 did.  To say it back:

 - in linux/osx -> no change
 - on windows:
    - prior to 1.14.4 all characters are allowed in file names
    - 1.14.4 and 1.14.5 only characters is current code page are allowed in filenames

This raises the question to me of what things like the windows file browser and shell do with file names that have characters outside of the current code page?  If that is already hard / has sharp edges I'm OK with publishing with this regression.


-----

As a side note conda-forge is building h5py double with both 1.14.3 and 1.14.4 to give users the option of going back, but we don't really have that as an option due to the limitations of wheels.

takluyver:
Yup, your description of the change looks right.

> This raises the question to me of what things like the windows file browser and shell do with file names that have characters outside of the current code page?

Windows-native software like the file browser will - I assume - use the unicode APIs for the filesystem, so it shouldn't be limited by the code page. I'd guess Powershell also does this; I'm not so sure about cmd.exe, the classic 'Command prompt' shell. Python has used the unicode APIs since 3.6 ([PEP 529](https://peps.python.org/pep-0529/)). So it will stick out somewhat.

Users who do need to use characters beyond the current code page could of course stick to h5py 3.12. Or in theory they could build a newer h5py with an older HDF5 - but on Windows in particular this is probably not easy.

Alternatively, we could:

1. Hold back the bundled libhdf5 for Windows to the last version which expects UTF-8 paths (1.14.3, I think). This is implicitly saying that this issue is more important than anything fixed since then, such as the memory leak discussed in #2451.
2. Patch our bundled libhdf5 to keep using UTF-8 paths. But then we either do the wrong thing when built with vanilla HDF5, or we need some extra mechanism to tell if we're using a patched version (in addition to the version check in this PR).

Neither of those options sound good to me, so I'm inclined to update it and document the issue.

tacaswell:
> Patch our bundled libhdf5 to keep using UTF-8 paths. But then we either do the wrong thing when built with vanilla HDF5, or we need some extra mechanism to tell if we're using a patched version (in addition to the version check in this PR).

This is a non-starter for me.  I could see making a case that down stream packagers (which for windows is just CF) should ship a patched version of 1.14.5 but that is because they have the machinery to both apply patches reliably and track with h5py builds rely on them.  Leaving that aside, I also don't think we want to get into the business of risking injecting bugs into libhdf5!

I agree the least-bad option is to update all of the wheels and document this as an issue.

jhendersonHDF:
Hi all,

this is probably a bit late now (I don't know the current state of the next release of h5py), but after a lot of discussion we decided we're just going to do the sensible thing: revert the offending PR and have a 1.14.6 release soon after testing. Even though this will break the fix we made for https://forum.hdfgroup.org/t/open-create-hdf5-files-with-non-utf8-chars-such-as-shift-jis/11785, we want the h5py, conda-forge and other communities to have a release with the previous behavior, as well as the fix for the memory leak mentioned above.

We have to do a bit of work in creating the branch and testing, but once everything is settled I'll likely be asking for help with testing, if possible, to make sure the release is in a good state for h5py and that we haven't broken anything else.

takluyver:
Thanks! We haven't released yet, and there aren't all that many changes since h5py 3.12, so we will most likely wait until we can bundle HDF5 1.14.6 before doing a new release.

jhendersonHDF:
Hi again all,

things have mostly settled down for the 1.14.6 release at this point and I believe what's remaining is just documentation tasks. We have a recent snapshot of the branch at https://github.com/HDFGroup/hdf5/releases/tag/snapshot-1.14 (ignore the text there for now as it needs updated), with current development at https://github.com/HDFGroup/hdf5/tree/hdf5_1_14_6. There were some changes around the handling of ZLIB and SZIP and compression filters in general that we brought back from the main development branch to fix some important issues. Hopefully that didn't make matters worse, considering https://github.com/HDFGroup/hdf5/issues/5026, but otherwise changes in the branch were mostly documentation and CI-related. Please let me know if this release looks like it's back on track with the previous behavior and if we happened to break other things for h5py. Thanks!

takluyver:
Thanks! I've tested that out in #2547 and it seems to work OK (there's one CI failure that's unrelated).

We still also have a patched version of `CMakeFilters.cmake` because of https://github.com/HDFGroup/hdf5/issues/5026 . I haven't really tried to understand this, but if it's possible to fix it in 1.14.6 so we don't need a modified version of that file to build, that would be good.

jhendersonHDF:
> Thanks! I've tested that out in #2547 and it seems to work OK (there's one CI failure that's unrelated).
> 
> We still also have a patched version of `CMakeFilters.cmake` because of [HDFGroup/hdf5#5026](https://github.com/HDFGroup/hdf5/issues/5026) . I haven't really tried to understand this, but if it's possible to fix it in 1.14.6 so we don't need a modified version of that file to build, that would be good.

Thanks for testing! I hope that we'll get a chance to look at the issue in `CMakeFilters.cmake` for the release, but in the spirit of trying not to introduce too many functional changes for what should be a boring release, this may get delayed until the next release. If all goes well, I believe we're looking at trying to have the release out by the end of this week or soon after.

takluyver:
That sounds good, thanks! :+1:

jhendersonHDF:
Hi all,

with https://github.com/HDFGroup/hdf5/pull/5284 recently merged into the hdf5_1_14_6 branch you shouldn't need to patch `CMakeFilters.cmake` anymore (though it would be good to test to make sure). After resolving some small things we decided not to try and finish the release today, but it should be out early next week.

takluyver:
Thanks, that seemed to work OK (I tested in #2547 again).

PR #2524, opened 2024-11-15
Fix isolated build for Python 3.13 by using mpi4py==4.0.1

### LINKED ISSUE #2523: MPI build broken with Python 3.13
In https://github.com/h5py/h5py/blob/5c4871e37a08d7b030dc6c18df21f45cd714e6c6/setup.py#L51C1-L51C68, there is

```
    SETUP_REQUIRES.append("mpi4py ==3.1.6; python_version>='3.12'")
```

However, I think mpi4py 3.1.6 is not compatible with Python 3.13 (compile error).

I was able to compile mpi4py 4.0.1 on Python 3.13.

It would be interesting to know if h5py is compatible with mpi4py 4...

paugier:
Fixes #2523


takluyver:
Thanks!

PR #2529, opened 2024-11-26
Expose H5Sselect_shape_same function

takluyver:
I wanted to use this, and conveniently it's available from HDF5 1.10.6, which is the minimum version we support.

https://support.hdfgroup.org/documentation/hdf5/latest/group___h5_s.html#title45


takluyver:
It seems the docs were wrong, it was added in 1.10.7.

https://github.com/HDFGroup/hdf5/pull/5148

takluyver:
I'll merge this after the weekend if there are no comments.

PR #2534, opened 2024-12-02
BLD: fix running api_gen.py directly

### LINKED ISSUE #2533: Wrong number of arguments in call to `run()` in api_gen.py.
 * Operating System : debian testing derivative
 * Python version (e.g. 3.10): 3.11
 * Where Python was acquired - custom
 * h5py version (e.g. 3.9): head
 * HDF5 version (e.g. 1.12.2): 1.12.2

The error trace:

```
h5py/api_gen.py", line 328, in <module>
    run()
TypeError: run() missing 1 required positional argument: 'config'
```

Nothing above really matters. The issue is that the call doesn't have the right number of args from here:

https://github.com/h5py/h5py/blob/35e68c16760bbd28f9909e6af93c2a476c15a6b8/api_gen.py#L321-L327

```python
def run(config):  # <-- config option without a 
    lp = LineProcessor(config)
    lp.run()


if __name__ == '__main__':
    run()  # <-- No config arg passed.
```

I don't see anything about passing a config in the source:

```
find . -type f | grep -v git/ | xargs grep api_gen.py
./api_gen.py:# Warning: this file is auto-generated from api_gen.py. DO NOT EDIT!
./api_gen.py:# Warning: this file is auto-generated from api_gen.py. DO NOT EDIT!
./api_gen.py:# Warning: this file is auto-generated from api_gen.py. DO NOT EDIT!
./docs/whatsnew/3.12.rst:  ``api_gen.py`` should be run automatically to recreate this (:pr:`2479`,
./docs/contributing.rst:2. Auto-generated Cython wrappers for the C API (``api_gen.py``)
./docs/contributing.rst:run the file ``api_gen.py`` to re-generate the Cython wrappers from
./docs/contributing.rst:    $ python api_gen.py
./MANIFEST.in:include api_gen.py
./h5py/api_functions.txt:# Defines the HDF5 C API functions wrapped by h5py. See also api_gen.py.
```

neutrinoceros:
Close #2533

### DIFF
--- a/api_gen.py
+++ b/api_gen.py
@@ -28,6 +28,7 @@ import re
 import os
 from hashlib import md5
 from pathlib import Path
+from setup_configure import BuildConfig
 
 
 def replace_or_remove(new: Path) -> None:
@@ -318,7 +319,9 @@ cdef {0.code} {0.fname}({0.sig}) except {0.err_value}:
         self.cython_imp.write(imp)
 
 
-def run(config):
+def run():
+    # Get configuration from environment variables
+    config = BuildConfig.from_env()
     lp = LineProcessor(config)
     lp.run()
 
--- a/setup_build.py
+++ b/setup_build.py
@@ -152,7 +152,7 @@ class h5py_build_ext(build_ext):
 
         # Refresh low-level defs if missing or stale
         print("Executing api_gen rebuild of defs")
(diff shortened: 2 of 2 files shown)


schwehr:
Thanks for the fix!

PR #2544, opened 2025-01-17
Remove Travis CI configuration

### LINKED ISSUE #2543: Travis builds stopped again
Travis CI last built on 31st December. I can see a message on the Travis web interface:

> We are unable to start your build at this time. You exceeded the number of users allowed for your plan. Please review your plan details and follow the steps to resolution. 

I can email Travis support, they normally get things resolved when I get in touch with them. But then after a while it breaks again in a different way. I get the feeling that while they do still support open source projects, it's a bit of a corner case for them now.

So I'm inclined to stop trying to use Travis. That would mean abandoning testing on ppc64le CPUs, though - as far as I know, Travis remains the only (free) CI service to offer that. @kif I think it was ESRF that was interested in PowerPC most recently - is that still important for you? Do you know of any better way we could routinely test it?

takluyver:
Closes #2543.

Travis has become unreliable in ways that we can't fix. We'll stop testing on ppc64le for now, and leave it to downstreams that actually package for that architecture.


takluyver:
Looks like the Mac failures are not a random thing, something must have changed about the Mac build environment on Github. It's failing to find libhdf5 to build h5py with.

```
      Loading library to get build settings and version: libhdf5.dylib
      on sys.platform='darwin' with platform.machine()='x86_64'
      Library dirs checked: []
      error: dlopen(libhdf5.dylib, 0x0006): tried: 'libhdf5.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OSlibhdf5.dylib' (no such file), '/usr/lib/libhdf5.dylib' (no such file, not in dyld cache), 'libhdf5.dylib' (no such file)
      [end of output]
```

This is coming from Homebrew. Is anyone familiar with Mac and willing to dig in a bit?

takluyver:
I think the primary point is to check that building from source with an obvious way of getting HDF5 works, but given that the wheel builds generally test the newest HDF5, it doesn't hurt to have some tests with older versions. :shrug:

takluyver:
Trying with HDF5 1.12 in #2545.

tacaswell (approved):

tacaswell:
I merged #2545 and approved this one.  Merge as-is or power-cycle to re-run the OSX jobs at your discretion.

takluyver:
Thanks, I'll merge this as is. There's hopefully no way the presence of the Travis file can affect Azure.

PR #2545, opened 2025-01-17
Try HDF5 1.14 in Homebrew on Azure CI

takluyver:
See if this fixes CI


takluyver:
Nope:

> No available formula with the name "hdf5@1.12". Did you mean hdf5@1.10 or hdf5@1.8?

takluyver:
OK, it seems to work with `hdf5`, which is current version 1.14.5, installed from Homebrew. We still test with HDF5 1.10 on Linux & Windows, so this is probably OK.

tacaswell (approved):

PR #2547, opened 2025-01-28
Test with Hdf5 1.14.6

takluyver:
https://github.com/h5py/h5py/pull/2520#issuecomment-2616881325

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -79,7 +79,7 @@ jobs:
       - name: Cache built HDF5
         uses: actions/cache@v4
         with:
-          key: ${{ matrix.os }}-${{ matrix.arch }}-6
+          key: ${{ matrix.os }}-${{ matrix.arch }}-7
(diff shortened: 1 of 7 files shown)


takluyver:
I think the segfaults on Ubuntu aarch64 are unrelated - we haven't updated HDF5 for that build anyway.

neutrinoceros:
Seconded. This is a symptom of a much deeper issue that seemingly affects a very large number of ecosystems at the moment; see https://github.com/tonistiigi/binfmt/issues/215

takluyver:
I accidentally pushed the commit above (to try building without the patched cmake file) to master - I thought I was on this branch, but I wasn't.

I've reverted that, redone it here, and I've enabled Github's branch protection for master, so I can't push directly to it again. That means changes need to be made in pull requests and merged into master in Github, which I think we always do anyway. If this turns out to be a problem, we can disable it again.

takluyver:
Closing in favour of #2554.

PR #2548, opened 2025-01-28
MAINT: Simplify aarch64 building

larsoner:
Switch to using the public beta `ubuntu-24.04-arm` runner, which should speed up the aarch64 wheel build enough so that all the triaging hoops can be removed :crossed_fingers:

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -9,17 +9,12 @@ concurrency:
 
 # 1. Move as much as possible to pyproject.toml instead of using env vars
 # 2. Use native compiling for macOS via -13 (Intel) and -14 (arm) runners
-# 3. For aarch64, only ever test py312 and skip testing other python versions
-#    (for speed).
-# 4. Split aarch64 python version builds over jobs, for all other systems/archs build
-#    wheels for all Python versions in a single job (for speed).
-# 5. On pushes, build and test all wheels and push as release artifacts (in
+# 3. Use native compiling for aarch64 using public beta runner
+# 4. On pushes, build and test all wheels and push as release artifacts (in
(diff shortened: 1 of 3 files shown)


.github/workflows/build_wheels.yml
```diff
@@ -9,17 +9,12 @@ concurrency:
 
 # 1. Move as much as possible to pyproject.toml instead of using env vars
 # 2. Use native compiling for macOS via -13 (Intel) and -14 (arm) runners
-# 3. For aarch64, only ever test py312 and skip testing other python versions
-#    (for speed).
-# 4. Split aarch64 python version builds over jobs, for all other systems/archs build
-#    wheels for all Python versions in a single job (for speed).
-# 5. On pushes, build and test all wheels and push as release artifacts (in
+# 3. Use native compiling for aarch64 using public beta runner
```
neutrinoceros, line 12: this comment will probably not age well. Maybe if we add a link to the public announcement on the same line it'll be easier to spot when it goes out of date ?

neutrinoceros (commented):
nice clean up ! Other than my single trivial comment, LGTM
Thanks !

.github/workflows/build_wheels.yml
larsoner:
```suggestion
# 3. Use native compiling for aarch64 using public beta runner
#    https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/
```

neutrinoceros (approved):

neutrinoceros:
:shipit:

takluyver:
I just went to try this and found that you'd already done it - thanks! :tada:

PR #2550, opened 2025-02-04
Dataset views overhaul

crusaderky:
- Add support for `dtype` parameter in `np.array` applied to a `asstr` view
- Fix crash in `np.array` applied to a `astype`, `asstr`, or `fields` view of a 0-dimensional dataset
- Fix crash in `np.array(ds.asstr())` where ds.ndim > 1
- Skip double conversion in `np.array(ds.astype(...), dtype=...)`
- `np.array(ds.asstr(), dtype=int)` now raises ValueError instead of TypeError, in line with other numpy dtype conversions
- Added array-like properties to `astype`, `asstr`, and `fields` views
- Improve test coverage for `ds.fields`


takluyver:
Thanks, this looks like a nice improvement.

I think `AsStrView` should override dtype as well: for fixed length strings the Dataset object has a dtype like `dtype('S3')`, and `.asstr()` converts that to an object array.

h5py/_hl/dataset.py
```diff
@@ -198,47 +199,77 @@ def open_dset(parent, name, dapl=None, efile_prefix=None, virtual_prefix=None,
...
 
-        data = self[:]
+        data = self[()]
+        if not self.ndim:
+            return numpy.asarray(data, dtype=dtype or self.dtype)
         if dtype is not None:
             return data.astype(dtype, copy=False)
         return data
 
+class AstypeView(AbstractView):
```
takluyver, line 246: While we're renaming this anyway, we can make the naming pattern a bit more consistent (matching AsStrView).

```suggestion
class AsTypeView(AbstractView):
```

takluyver:
Thanks, this looks good. Could I ask you to summarise the main user visible changes in a [news snippet file](https://docs.h5py.org/en/stable/contributing.html#write-a-release-note)? Even if this is just something generic like 'various improvements to `.astype()`, `.asstr()` and `.fields()` views.

PR #2552, opened 2025-02-07
Lazily load dcpl: it's not needed for reading

ArvidJB:
Datasets have a `_dcpl` (data creation property list) field. For virtual Datasets it can be very expensive to load this list since it contains the entire virtual to underlying source mapping. Since this field is only actually needed for *writing* to the Dataset this commit defers loading the `_dcpl`.

This improves performance for opening a virtual Dataset like one created in `versioned_hdf5` quite substantially. Previously:
```
In [3]: with tempfile.TemporaryDirectory() as d:
   ...:     with h5py.File(f'{d}/data.h5', 'w') as f:
   ...:         vf = VersionedHDF5File(f)
   ...:         with vf.stage_version('r0') as sv:
   ...:             sv.create_dataset('values', data=np.random.rand(365, 12345), chunks=(10, 100))
   ...:     with h5py.File(f'{d}/data.h5', 'r') as f:
   ...:         vf = VersionedHDF5File(f)
   ...:         cv = vf[vf.current_version]
   ...:         def open_dataset():
   ...:             values = cv['values']
   ...:         print(timeit.timeit(open_dataset, number=100))
   ...:
3.910119968932122
```
With the changes in this commit:
```
In [5]: with tempfile.TemporaryDirectory() as d:
   ...:     with h5py.File(f'{d}/data.h5', 'w') as f:
   ...:         vf = VersionedHDF5File(f)
   ...:         with vf.stage_version('r0') as sv:
   ...:             sv.create_dataset('values', data=np.random.rand(365, 12345), chunks=(10, 100))
   ...:     with h5py.File(f'{d}/data.h5', 'r') as f:
   ...:         vf = VersionedHDF5File(f)
   ...:         cv = vf[vf.current_version]
   ...:         def open_dataset():
   ...:             values = cv['values']
   ...:         print(timeit.timeit(open_dataset, number=100))
   ...:
1.0834132973104715
```

Note that the performance here is still pretty poor since there are some performance issues in the underlying HDF5 library implementation as well which also copy the entire virtual "layout" when a Dataset is accessed. I also hope to fix these issues in HDF5 itself.

### DIFF
--- a/h5py/_hl/filters.py
+++ b/h5py/_hl/filters.py
@@ -308,7 +308,7 @@ def fill_dcpl(plist, shape, dtype, chunks, compression, compression_opts,
 
 def get_filter_name(code):
     """
-    Return the name of the compression filter for a given filter code.
+    Return the name of the compression filter for a given filter identifier.
 
     Undocumented and subject to change without warning.
     """
--- a/h5py/_hl/filters.py
+++ b/h5py/_hl/filters.py
@@ -306,16 +306,23 @@ def fill_dcpl(plist, shape, dtype, chunks, compression, compression_opts,
 
     return plist
 
+def get_filter_name(code):
+    """
+    Return the name of the compression filter for a given filter code.
+
+    Undocumented and subject to change without warning.
+    """
+    filters = {h5z.FILTER_DEFLATE: 'gzip', h5z.FILTER_SZIP: 'szip',
+               h5z.FILTER_SHUFFLE: 'shuffle', h5z.FILTER_FLETCHER32: 'fletcher32',
+               h5z.FILTER_LZF: 'lzf', h5z.FILTER_SCALEOFFSET: 'scaleoffset'}
+    return filters.get(code, str(code))
+
 def get_filters(plist):
     """ Extract a dictionary of active filters from a DCPL, along with
     their settings.
 
     Undocumented and subject to change without warning.
     """
-
-    filters = {h5z.FILTER_DEFLATE: 'gzip', h5z.FILTER_SZIP: 'szip',
-               h5z.FILTER_SHUFFLE: 'shuffle', h5z.FILTER_FLETCHER32: 'fletcher32',
-               h5z.FILTER_LZF: 'lzf', h5z.FILTER_SCALEOFFSET: 'scaleoffset'}
 
     pipeline = {}
 
@@ -343,7 +350,7 @@ def get_filters(plist):
             if len(vals) == 0:
                 vals = None
 
-        pipeline[filters.get(code, str(code))] = vals
+        pipeline[get_filter_name(code)] = vals
 
     return pipeline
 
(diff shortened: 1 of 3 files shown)


ArvidJB:
Note that `test_compression_number_invalid` now fails because `IndexError` is no longer raised on `Dataset` object construction. Instead it's now raised when `_filters` or `compression`. etc., is accessed. Is that acceptable?

takluyver:
Thanks, this looks like a good idea. :+1: 

> Note that test_compression_number_invalid now fails because IndexError is no longer raised on Dataset object construction. Instead it's now raised when _filters or compression. etc., is accessed. Is that acceptable?

IMO, it would be acceptable if this was about constructing a new Python `Dataset` object wrapping an existing HDF5 dataset.  But the test failure looks like it allows actually creating a new HDF5 dataset object to succeed when previously it failed; presumably it creates a dataset which can't be written to.

```text
               # Using gzip compression requires a compression level specified in compression_opts
              with self.assertRaises(IndexError):
  >               self.f.create_dataset('foo', (20, 30), compression=h5py.h5z.FILTER_DEFLATE)
  E               AssertionError: IndexError not raised
```

I don't think this is acceptable. Ideally we could catch this in some way before the file is modified, but if that's not practical we should make sure to trigger the error within `.create_dataset()`

ArvidJB:
> I don't think this is acceptable. Ideally we could catch this in some way before the file is modified, but if that's not practical we should make sure to trigger the error within `.create_dataset()`

I agree, that's probably best. I will try to implement that.

ArvidJB:
I ended up implementing a "compression" roundtrip check in `make_new_dset`. I think this works correctly for all the ways "compression" can be specified (name, int, `FilterRefBase`) and I also tested it on "lz4hc" and "blosc" compression.

What's a good name for the `get_filter_name` helper function I added?

takluyver:
Thanks! I think this will only really check the filters that `get_filters` knows about the options for, i.e. 'gzip' (really zlib) and szip, but that's better than nothing,  and I don't think HDF5 gives us an easy way to do this for arbitrary features.

`get_filter_name` seems like an OK name to me. It's not public API in any case.

h5py/_hl/filters.py
```diff
@@ -306,16 +306,23 @@ def rq_tuple(tpl, name):
 
     return plist
 
-def get_filters(plist):
-    """ Extract a dictionary of active filters from a DCPL, along with
-    their settings.
+def get_filter_name(code):
+    """
+    Return the name of the compression filter for a given filter code.
```
takluyver, line 311:
```suggestion
    Return the name of the compression filter for a given filter identifier.
```

Just to align with [what HDF group call this](https://github.com/HDFGroup/hdf5_plugins/blob/master/docs/RegisteredFilterPlugins.md). I don't think we need to change it in the code, though.

takluyver:
Thanks! Could I ask you to add a [news fragment](https://docs.h5py.org/en/stable/contributing.html#write-a-release-note) summarising this change?

ArvidJB:
> Thanks! Could I ask you to add a [news fragment](https://docs.h5py.org/en/stable/contributing.html#write-a-release-note) summarising this change?

I added an entry, although I am not sure which category it should belong to? For now I added it to "New Features", but it feels like not a big enough change to belong there...

takluyver:
Thanks! That's great, so long as it's in there I'll reshuffle everything when I'm preparing the next release.

PR #2554, opened 2025-02-08
Build with HDF5 1.14.6

takluyver:
This is released now. Supersedes #2547.

A separate change is still needed for the Docker images we build Linux wheels with.


takluyver:
I've updated HDF5 in the Docker images, and re-run the two Linux wheel jobs here so that they used 1.14.6. I think this is ready. :+1:

tacaswell (approved):

PR #2556, opened 2025-02-10
BLD: fix build-time compatibility with numpy 2.3

### LINKED ISSUE #2555: BLD: incoming incompatibility with numpy 2.3
numpy dev just expired deprecations that I just learned we relied on: [example logs](https://github.com/h5py/h5py/actions/runs/13232352872)

Basically, this line won't compile now that all 3 `NPY_*` constants it uses were removed:
https://github.com/h5py/h5py/blob/c02a7e70c55733746948d73b1ccc66a093ef9e01/h5py/_conv.pyx#L698

This affects cp311+ wheels only because numpy 2.3 doesn't support Python 3.10 and earlier.
xref: https://github.com/numpy/numpy/pull/28254

neutrinoceros:
Close #2555

### DIFF
--- a/h5py/utils.pyx
+++ b/h5py/utils.pyx
@@ -13,7 +13,7 @@
 from numpy cimport ndarray, import_array,\
                    NPY_UINT16, NPY_UINT32, NPY_UINT64,  npy_intp,\
                    PyArray_SimpleNew, PyArray_FROM_OTF,\
-                   NPY_CONTIGUOUS, NPY_NOTSWAPPED, NPY_FORCECAST
+                   NPY_ARRAY_C_CONTIGUOUS, NPY_ARRAY_NOTSWAPPED, NPY_ARRAY_FORCECAST
 
 # Initialization
 import_array()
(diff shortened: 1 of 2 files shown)


h5py/utils.pyx
```diff
@@ -164,7 +164,10 @@ cdef object create_hsize_array(object arr):
     else:
         raise RuntimeError("Can't map hsize_t %d to Numpy typecode" % sizeof(hsize_t))
 
-    return PyArray_FROM_OTF(arr, typecode, NPY_CONTIGUOUS | NPY_NOTSWAPPED | NPY_FORCECAST)
+    return PyArray_FROM_OTF(arr, typecode,
+        NPY_ARRAY_C_CONTIGUOUS
```
neutrinoceros, line 168: This replacement wasn't immediately clear to me: there's no `NPY_ARRAY_CONTIGUOUS`, but inspecting numpy's source makes it clear why: `NPY_CONTIGUOUS` was always an alias for `NPY_C_CONTIGUOUS`
see https://github.com/numpy/numpy/blob/fd8a68e4978defd094cffa23c71d7de7fb213e3f/numpy/_core/include/numpy/npy_1_7_deprecated_api.h#L31

takluyver:
Thanks! Do you know if there's a minimum version of Numpy required for the new names? It looks from the numpy issue like the deprecation might have been there since Numpy 1.7 (!), but I don't know the structure of Numpy well enough to know if that applies to all these names.

neutrinoceros:
The "new" APIs were introduced in numpy 1.7 alongside the deprecations, so it should be very safe to move now
xref https://github.com/numpy/numpy/pull/16986

takluyver:
Great, thank-you!

PR #2557, opened 2025-02-11
Native support for Numpy 2's variable-width strings

### LINKED ISSUE #2551: Numpy 2 variable-width strings
I'm working on a PR to support reading and writing [numpy 2's variable-width strings](https://numpy.org/devdocs/user/basics.strings.html#variable-width-strings) without passing through object dtype.

How should the UI look like?

For `__getitem__`:

1. if numpy 2 is installed and `H5Tget_class(tid) == H5T_STRING and H5Tis_variable_str(tid)` and the meta says it's unicode, the dataset dtype will automatically change to `np.dtypes.StringDType()`. `ds.asstr` becomes a no-op which also returns StringDType.
This is the most natural interface IMHO, but it's also a breaking change.

2. `ds[:]` still returns a numpy array of `byte` objects. `ds.asstr` detects if numpy 2 is installed and automatically returns a StringDType if the encoding is UTF-8. This is also a breaking change.

3. Nothing changes by default. `ds.asstr` gains a new parameter, `dtype=np.dtype(object)`. If the user explicitly sets `np.dtypes.StringDType()` or an alias for it (`dtype="T"`), then the returned array is of the desired type.

4. as (3), but the default is `dtype=None`, which issues a FutureWarning that the returned type is object but in the future will change to StringDType. At a later date, remove the warning and switch the default to StringDType.

5. As (1), but with some sort of opt-in global flag that is later flipped to opt-out.

For `create_dataset`:
1. `h5py.string_dtype()` continues returning `np.dtype('O', metadata={'vlen': str})`. When someone calls `__setitem__` on such an array, there is magic happening that bypasses the conversion through object type if the parameter's type is StringDType.

2. `h5py.string_dtype()` returns StringDType. This would be cleaner but a breaking change.

3. As (1), but with some sort of opt-in global flag that is later flipped to opt-out.

crusaderky:
Closes #2551

- [x] main implementation
- [x] performance benchmarks
- [x] no breaking changes (as discussed in #2551)
- [x] don't break compilation on numpy <2.3
- [x] don't break execution on numpy 1
- [x] compile on numpy 2.0
- [x] unit tests
- [x] documentation

# Smoke test
[npystrings_demo.py](https://github.com/crusaderky/h5py/blob/npystrings_demo/npystrings_demo.py) 

# Performance test
[npystrings_bench.py](https://github.com/crusaderky/h5py/blob/npystrings_demo/npystrings_bench.py) 

1_000_000 strings of random width 0~100 characters
(82 MiB dataset) on local NVMe

object dtype (master)
    write: 0.463s
    read : 0.529s

StringDType()
    write: 0.252s
    read : 0.390s

On-disk format is identical between the two.

### DIFF
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -518,9 +518,8 @@ Reference
             >>> out = dset[:500].astype('int16')  # Typically slower
 
         In case of variable-width strings, calling ``.astype('T')``
-        (NumPy's native variable-width strings) is zero cost and
-        faster than reading the data into an object-type array;
-        read more at :ref:`npystrings`.
+        (NumPy's native variable-width strings) is more efficient than reading
+        the data into an object-type array; read more at :ref:`npystrings`.
 
         .. versionchanged:: 3.14
            Added support for NumPy variable-width strings (``dtype='T'``).
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -504,7 +504,7 @@ Reference
 
     .. method:: astype(dtype)
 
-        Return a wrapper allowing you to read data as a particular
+        Return a read-only view allowing you to read data as a particular
         type.  Conversion is handled by HDF5 directly, on the fly::
 
             >>> dset = f.create_dataset("bigint", (1000,), dtype='int64')
@@ -517,13 +517,13 @@ Reference
 
             >>> out = dset[:500].astype('int16')  # slower
 
-        Calling ``.astype('T')`` (NumPy's native variable-width strings)
-        on a dataset returns a writable dataset with the specified dtype;
+        In case of variable-width strings, calling ``.astype('T')``
+        (NumPy's native variable-width strings) is zero cost and
+        faster than reading the data directly into an object-type array;
         read more at :ref:`npystrings`.
-        For all other dtypes, this method returns a read-only view.
 
         .. versionchanged:: 3.14
-           ``astype('T')`` returns a writeable dataset.
+           Added support for NumPy variable-width strings (``dtype='T'``).
 
         .. versionchanged:: 3.9
            :meth:`astype` can no longer be used as a context manager.
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -508,9 +508,22 @@ Reference
         type.  Conversion is handled by HDF5 directly, on the fly::
 
             >>> dset = f.create_dataset("bigint", (1000,), dtype='int64')
-            >>> out = dset.astype('int16')[:]
+            >>> out = dset.astype('int16')[:500]
             >>> out.dtype
             dtype('int16')
+
+        This is typically faster than reading the data and then
+        converting it with NumPy:
+
+            >>> out = dset[:500].astype('int16')  # slower
+
+        Calling ``.astype('T')`` (NumPy's native variable-width strings)
+        on a dataset returns a writable dataset with the specified dtype;
+        read more at :ref:`npystrings`.
+        For all other dtypes, this method returns a read-only view.
+
+        .. versionchanged:: 3.14
+           ``astype('T')`` returns a writeable dataset.
 
         .. versionchanged:: 3.9
            :meth:`astype` can no longer be used as a context manager.
@@ -525,6 +538,12 @@ Reference
        encoding and errors work like ``bytes.decode()``, but the default
        encoding is defined by the datatype - ASCII or UTF-8.
        This is not guaranteed to be correct.
+
+       .. note::
+          If you don't require backwards compatibility with NumPy 1.x or
+          h5py <3.14, you should consider reading into NumPy native strings
+          insteads, which can be much faster, with ``Dataset.astype('T')``.
+          Read more at :ref:`npystrings`.
 
        .. versionadded:: 3.0
 
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -512,7 +512,7 @@ Reference
             >>> out.dtype
             dtype('int16')
 
-        This is typically faster than reading the data and then
+        This can be faster than reading the data and then
         converting it with NumPy:
 
             >>> out = dset[:500].astype('int16')  # slower
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -515,11 +515,11 @@ Reference
         This can be faster than reading the data and then
         converting it with NumPy:
 
-            >>> out = dset[:500].astype('int16')  # slower
+            >>> out = dset[:500].astype('int16')  # Typically slower
 
         In case of variable-width strings, calling ``.astype('T')``
         (NumPy's native variable-width strings) is zero cost and
-        faster than reading the data directly into an object-type array;
+        faster than reading the data into an object-type array;
         read more at :ref:`npystrings`.
 
         .. versionchanged:: 3.14
@@ -542,7 +542,8 @@ Reference
        .. note::
           If you don't require backwards compatibility with NumPy 1.x or
           h5py <3.14, you should consider reading into NumPy native strings
-          insteads, which can be much faster, with ``Dataset.astype('T')``.
+          instead, which can be much faster, with
+          :meth:`.astype('T')<astype>`.
           Read more at :ref:`npystrings`.
 
        .. versionadded:: 3.0
--- a/docs/strings.rst
+++ b/docs/strings.rst
@@ -68,21 +68,24 @@ Data passed as NpyStrings is always encoded as UTF-8.
 NumPy variable-width strings
 ----------------------------
 
-Starting with NumPy 2.0 and h5py 3.14, you can also use ``dtype('T')`` to specify
-native NumPy variable-width strings. However, note that on a round-trip you will
-read object arrays of ``bytes`` again. This is to ensure that NumPy 1.x and 2.x
-behave in the same way unless the user explicitly requests native strings::
+Starting with NumPy 2.0 and h5py 3.14, you can also use ``np.dtypes.StringDType()``,
+or ``dtype('T')`` for short, to specify native NumPy variable-width string dtypes,
+a.k.a. NpyStrings.
+However, note that when you open a dataset that you created as a StringDType, its dtype
+will be object type with ``bytes`` elements again. This is because the two data types
+are identical in the HDF5 file, and to ensure that NumPy 1.x and 2.x behave in the same
+way unless the user explicitly requests native strings::
 
-    # NpyStrings (implicit)
+    # StringDType (from input data)
     npdata = np.asarray(string_data, dtype='T')
     ds = f.create_dataset('npystrings1', data=npdata)
 
-    # NpyStrings (explicit)
-    ds = f.create_dataset('npystrings2', shape=4, dtype='T')
+    # StringDType (explicit dtype parameter)
+    ds = f.create_dataset('npystrings2', shape=(4, ), dtype='T')
     ds[:] = string_data
 
-    # On a round-trip, you're back to bytes by default.
-    # Calling .astype('T') creates a writeable view with NpyStrings dtype.
+    # When you reopen the dataset, you're back to bytes by default.
+    # Calling .astype('T') creates a Dataset object with StringDType dtype.
     ds = f['npystrings1'].astype('T')
 
 What about NumPy's ``U`` type?
--- a/docs/strings.rst
+++ b/docs/strings.rst
@@ -85,7 +85,7 @@ native strings.
 In order to efficiently write NpyStrings to a dataset, simply assign a NumPy array with
 StringDType to it, either through ``__setitem__`` or :meth:`.Dataset.create_dataset`.
 To read NpyStrings out of any string-type dataset, use :meth:`.Dataset.astype('T')`.
-In both cases, there will be no intermediate conversion to object-type array, even if 
+In both cases, there will be no intermediate conversion to object-type array, even if
 the dataset's dtype is object type::
 
     # These three syntaxes are equivalent:
--- a/docs/strings.rst
+++ b/docs/strings.rst
@@ -14,7 +14,9 @@ Reading strings
 
 String data in HDF5 datasets is read as bytes by default: ``bytes`` objects
 for variable-length strings, or numpy bytes arrays (``'S'`` dtypes) for
-fixed-length strings. Use :meth:`.Dataset.asstr` to retrieve ``str`` objects.
+fixed-length strings. On NumPy >=2.0, use :meth:`Dataset.astype('T')<.Dataset.astype>`
+to read into an array of native NumPy strings. If you are using NumPy 1.x, you should
+instead call :meth:`.Dataset.asstr` to retrieve an array of ``str`` objects.
 
 Variable-length strings in attributes are read as ``str`` objects. These are
 decoded as UTF-8 with surrogate escaping for unrecognised bytes. Fixed-length
@@ -25,7 +27,9 @@ Storing strings
 
 When creating a new dataset or attribute, Python ``str`` or ``bytes`` objects
 will be treated as variable-length strings, marked as UTF-8 and ASCII respectively.
-Numpy bytes arrays (``'S'`` dtypes) make fixed-length strings.
+NumPy bytes arrays (``'S'`` dtypes) make fixed-length strings.
+NumPy NpyStrings arrays (``'T'`` dtypes) make native variable-length strings.
+
 You can use :func:`.string_dtype` to explicitly specify any HDF5 string datatype,
 as shown in the examples below::
 
@@ -49,6 +53,7 @@ When writing data to an existing dataset or attribute, data passed as bytes is
 written without checking the encoding. Data passed as Python ``str`` objects
 is encoded as either ASCII or UTF-8, based on the HDF5 datatype.
 In either case, null bytes (``'\x00'``) in the data will cause an error.
+Data passed as NpyStrings is always encoded as UTF-8.
 
 .. warning::
 
@@ -58,13 +63,35 @@ In either case, null bytes (``'\x00'``) in the data will cause an error.
    Fixed-length strings in HDF5 hold a set number of bytes.
    It may take multiple bytes to store one character.
 
+.. _npystrings:
+
+NumPy variable-width strings
+----------------------------
+
+Starting with NumPy 2.0 and h5py 3.14, you can also use ``dtype('T')`` to specify
+native NumPy variable-width strings. However, note that on a round-trip you will
+read object arrays of ``bytes`` again. This is to ensure that NumPy 1.x and 2.x
+behave in the same way unless the user explicitly requests native strings::
+
+    # NpyStrings (implicit)
+    npdata = np.asarray(string_data, dtype='T')
+    ds = f.create_dataset('npystrings1', data=npdata)
+
+    # NpyStrings (explicit)
+    ds = f.create_dataset('npystrings2', shape=4, dtype='T')
+    ds[:] = string_data
+
+    # On a round-trip, you're back to bytes by default.
+    # Calling .astype('T') creates a writeable view with NpyStrings dtype.
+    ds = f['npystrings1'].astype('T')
+
 What about NumPy's ``U`` type?
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-NumPy also has a Unicode type, a UTF-32 fixed-width format (4-byte characters).
-HDF5 has no support for wide characters.  Rather than trying to hack around
-this and "pretend" to support it, h5py will raise an error if you try to store
-data of this type.
+NumPy also has a Unicode fixed-width type, a UTF-32 fixed-width format
+(4-byte characters). HDF5 has no support for wide characters.
+Rather than trying to hack around this and "pretend" to support it,
+h5py will raise an error if you try to store data of this type.
 
 .. _str_binary:
 
--- a/docs/strings.rst
+++ b/docs/strings.rst
@@ -13,14 +13,15 @@ Reading strings
 ---------------
 
 String data in HDF5 datasets is read as bytes by default: ``bytes`` objects
-for variable-length strings, or numpy bytes arrays (``'S'`` dtypes) for
+for variable-length strings, or NumPy bytes arrays (``'S'`` dtypes) for
 fixed-length strings. On NumPy >=2.0, use :meth:`Dataset.astype('T')<.Dataset.astype>`
-to read into an array of native NumPy strings. If you are using NumPy 1.x, you should
-instead call :meth:`.Dataset.asstr` to retrieve an array of ``str`` objects.
+to read into an array of native variable-width NumPy strings.
+If you need backwards compatibility with NumPy 1.x or h5py <3.14, you should instead
+call :meth:`.Dataset.asstr` to retrieve an array of ``str`` objects.
 
 Variable-length strings in attributes are read as ``str`` objects. These are
 decoded as UTF-8 with surrogate escaping for unrecognised bytes. Fixed-length
-strings are read as numpy bytes arrays, the same as for datasets.
+strings are read as NumPy bytes arrays, the same as for datasets.
 
 Storing strings
 ---------------
@@ -28,7 +29,8 @@ Storing strings
 When creating a new dataset or attribute, Python ``str`` or ``bytes`` objects
 will be treated as variable-length strings, marked as UTF-8 and ASCII respectively.
 NumPy bytes arrays (``'S'`` dtypes) make fixed-length strings.
-NumPy NpyStrings arrays (``'T'`` dtypes) make native variable-length strings.
+NumPy NpyStrings arrays (``StringDType`` or ``'T'`` dtypes) make native variable-length
+strings.
 
 You can use :func:`.string_dtype` to explicitly specify any HDF5 string datatype,
 as shown in the examples below::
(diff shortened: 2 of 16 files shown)


crusaderky:
All CI failures are codecov crashes and unrelated with this PR.
@takluyver this is ready for review.

h5py/_proxy.pyx
```diff
@@ -138,7 +148,31 @@ cdef herr_t dset_rw(hid_t dset, hid_t mtype, hid_t mspace, hid_t fspace,
...
 
-            if read:
+            if H5Tis_variable_str(mtype):  # numpy.dtypes.StringDType
+                assert NUMPY_GE2
+                if read:
+                    H5Dread(dset, dstype, cspace, fspace, dxpl, conv_buf)
+                    if not H5Tis_variable_str(dstype):
+                        # Convert fixed-width bytes to char** in place.
+                        # When dstype is already char**, H5Tconvert is a very expensive no-op.
+                        H5Tconvert(dstype, mtype, npoints, conv_buf, back_buf, dxpl)
```
crusaderky, line 158: We could achieve a 25% speedup by writing a specialized variant of vstrings_scatter instead. I chose not to as I believe this should be a rather niche use case so I preferred to avoid code duplication.

crusaderky:
@takluyver gentle ping

ngoldbaum:
Hi all, I'm going to try to give this PR a thorough once-over sometime next week. I'm actually just learning today that Guido has been working on this and I'm really excited to see `StringDType` getting integrated into h5py.

takluyver:
Thanks @crusaderky,  I'm just starting to look at this.

First question, I was expecting that this would add conversion machinery in `_conv.pyx`. E.g. this is where we already have conversions between HDF5 variable-length data and numpy object arrays. Did you explore this possibility and decide that there was some reason it couldn't work?

The `py_create()` function to convert a numpy dtype to an HDF5 datatype has a parameter to distinguish between exact and 'logical' conversion:

> when this flag is set, instead of returning a byte-for-byte identical representation of the type, the function returns the closest logically appropriate HDF5 type.  For example, in the case of a "hinted" dtype of kind "O" representing a string, it would return an HDF5 variable-length string type.

If logical is not set, for object arrays it will always return the special `PYTHON_OBJECT` datatype. This should never* appear in an actual file, so it always needs a conversion on read/write. I would hope it's possible to work with Numpy strings in a similar fashion by making a new `NUMPY_STRINGS` opaque datatype object.

(*it's probably not impossible to make a dataset with this datatype and write useless pointers directly into it. But hopefully it will only happen if you really try to make that happen)

crusaderky:
Another major snag (possibly a showstopper?) is that NpyStrings need to access the `PyArray_Descr` from the input/output ndarray. By the time we're deep inside HDF5 machinery I don't think it's retrievable?

takluyver:
> otherwise you will no longer be able to import h5py when NumPy 1.x is installed.

How much easier would this work be without that requirement? All else being equal, I would probably keep supporting Numpy 1.x for a while, but I also have the impression that the Numpy 2.x release has gone pretty well. If this can be substantially simplified by dropping 1.x support - or demoting it to 'if you build from source' - then we can weigh that up. We will probably be supporting this for Numpy 2.x for much longer than we're trying to support 1.x and 2.x.

> I'm reading ndarray2vlen and observing that it calls emalloc once per point being written

I don't think that one is insurmountable; HDF5 calls the conversion function for multiple elements at a time.

> NpyStrings need to access the PyArray_Descr from the input/output ndarray. By the time we're deep inside HDF5 machinery I don't think it's retrievable?

That one might be insurmountable, though. I'll take a look.

In any case, I think that `py_create` with `logical=False` shouldn't say that the new numpy string dtype is equivalent to HDF5 variable length strings. That is meant to mean that the data type is "byte-for-byte memory-compatible", i.e. no conversion is needed. We can either have an opaque type like we do for object arrays, or simply raise an error that there's no matching HDF5 datatype.

crusaderky:
> How much easier would this work be without that requirement? All else being equal, I would probably keep supporting Numpy 1.x for a while, but I also have the impression that the Numpy 2.x release has gone pretty well. If this can be substantially simplified by dropping 1.x support - or demoting it to 'if you build from source' - then we can weigh that up. We will probably be supporting this for Numpy 2.x for much longer than we're trying to support 1.x and 2.x.

While SPEC0 states that NumPy 1 should exit is maintenance window in Q3 2025, I expect that *a lot* of people will become irate if you upgrade your minimum dependencies that quickly.
On the other hand, this is a killer feature for anybody working with strings, and limiting it to only people that build from sources (which is not an easy task particularly for less-than-expert users on windows) would severely impact its availability.

All in all I don't think that the solution I came up with is terrible in terms of readability/maintainability. Things would _not_ become particularly simpler if I could use the C interface.

h5py/_npystrings.pyx
```diff
@@ -0,0 +1,254 @@
...
+    """
+    return <size_t>_vstrings_gather(space, <void *>contig, <void *>noncontig,
+                                    <PyArray_Descr *>descr, npoints)
+
+
+cdef char * _vstrings_gather(hid_t space, void *contig, void *noncontig,
+                             PyArray_Descr *descr, size_t npoints):
+    """Convert a NpyString array (NumPy's native variable-width strings dtype)
+    to a zero-terminated char**, which is the in-memory representation for a
+    HDF5 variable-width string dataset.
```
ngoldbaum, line 176: It occurs to me that this is the same operation as asking NumPy to do a cast to a fixed-width byte array (which outputs UTF-8 data).

I wonder if that would be more maintainable and/or faster than trying to do the copies yourself.

ngoldbaum (commented):
Neat, seems like a clean design given that the copies are necessary.

As you see from the comments, it also makes me wonder if there are things we can do in NumPy to make this easier. 

I think probably storing NpyString as null terminated buffers in the arena might be the most useful, to avoid the bookkeeping outside of NumPy. The extra memory cost is probably worth it because many real-world cases need the null termination.

h5py/_hl/dataset.py
```diff
@@ -173,6 +175,9 @@ def make_new_dset(parent, shape=None, dtype=None, data=None, name=None,
         sid = h5s.create_simple(shape, maxshape)
 
     dset_id = h5d.create(parent.id, name, tid, sid, dcpl=dcpl, dapl=dapl)
+    # Multiple NumPy dtypes can map to the same hdf5 type; e.g.
+    # object strings or NpyStrings map to vlen strings.
+    dset_id._dtype = dtype
```
takluyver, line 180: Let's avoid this - a newly created dataset should not behave any differently from one retrieved from an existing file. I.e. it should not know that it was created with the new string dtype.

docs/high/dataset.rst
```diff
@@ -508,10 +508,23 @@ Reference
...
 
+        This is typically faster than reading the data and then
+        converting it with NumPy:
+
+            >>> out = dset[:500].astype('int16')  # slower
+
+        Calling ``.astype('T')`` (NumPy's native variable-width strings)
+        on a dataset returns a writable dataset with the specified dtype;
+        read more at :ref:`npystrings`.
+        For all other dtypes, this method returns a read-only view.
```
takluyver, line 523: Why is a writable view needed for this case? I think writing an array with "T" dtype into a variable length string dataset should work without needing a view.

h5py/_hl/dataset.py
```diff
@@ -436,7 +441,36 @@ def astype(self, dtype):
...
+                )
+            out = self.parent[self.name]
+            assert out.id is not self.id
+            out.id._dtype = dtype
+            return out
+        elif dtype == object and self.dtype.kind == "T":
+            out = self.parent[self.name]
+            assert out.id is not self.id
+            out.id._dtype = None
+            return out
```
takluyver, line 472: This seems rather convoluted, and I think it's in service of having writable `.astype('T')` views, which it's not clear to me are actually important. I think if we can get away from switching the dtype of dataset objects, quite a few of the later changes in this file can be avoided as well.

h5py/_npystrings.pyx
```diff
@@ -0,0 +1,254 @@
...
+    info.contig = <const char**>contig
+    info.allocator = NpyString_acquire_allocator(<PyArray_StringDTypeObject *>descr)
+    if info.allocator is NULL:
+        raise RuntimeError("Failed to acquire string allocator")
+
+    # Disregard mtype from _proxy::dset_rw.
+    # The memory type is actually npy_packed_static_string
+    # (an opaque struct *typically* 16 bytes per point),
+    # and not HDF5 variable length strings (char*; 8 bytes per point).
+    tid = H5Tcreate(H5T_OPAQUE, SIZEOF_NPY_PACKED_STATIC_STRING)
```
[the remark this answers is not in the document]
crusaderky: The tid is needed by H5TIterate a few lines later. I've added a comment.

h5py/_npystrings.pyx
```diff
@@ -0,0 +1,254 @@
...
+    info.allocator = NpyString_acquire_allocator(<PyArray_StringDTypeObject *>descr)
+    if info.allocator is NULL:
+        raise RuntimeError("Failed to acquire string allocator")
+
+    tid = H5Tcreate(H5T_OPAQUE, SIZEOF_NPY_PACKED_STATIC_STRING)
+    try:
+        # Multiple steps needed:
+        # 1. Read npy_packed_static_string[] from numpy and unpack
+        #    to npy_static_string[]; which is
+        #    {const char* buf, size_t size}[] - NOT zero-terminated
```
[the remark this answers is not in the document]
crusaderky: > Would it avoid copies and be generally helpful here if NumPy strings were null-terminated? It would make it more complicated to support older NumPy versions here, of course. We wanted to hear about use-cases before deciding to do that.

Yes, it would avoid a copy of the whole thing into the temporary buffer.

> I also wonder if it makes sense to have a NumPy C API function that does this conversion to a `char**` buffer, there might be optimizations that are possible if you have access to the NpyString internals.

It would probably be useful. From my understanding of NpyStrings internal structure, NumPy could add a null termination at no cost for all strings shorter than 15 bytes (the maximum size of an embedded string).

h5py/_hl/dataset.py
```diff
@@ -650,6 +693,12 @@ def maxshape(self):
     @with_phil
     def fillvalue(self):
         """Fill value for this dataset (0 by default)"""
+        if self.dtype.kind == "T":
+            # For the sake of simplicity, pass through object strings
+            arr = numpy.zeros((1,), dtype=self.astype(object).dtype)
```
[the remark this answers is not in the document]
crusaderky: I would have needed to replicate everything I did for datasets (or make it somehow generic and probably much less legible) around `dcpl.get_fill_value` -> `H5Pget_fill_value`.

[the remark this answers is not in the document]
crusaderky: `attrs` have the same issue - they would probably benefit from being read into a StringDType array and then converted to list instead of being read into a object array of bytes and then converted with `bytes.decode`. But performance wise they're inconsequential so I left them untouched.

h5py/_hl/dataset.py
```diff
@@ -1115,17 +1164,20 @@ def __array__(self, dtype=None, copy=None):
...
-                namestr, self.shape, self.dtype.str
-            )
-        return r
+            name = pp.basename(pp.normpath(self.name))
+            namestr = f'"{name}"' if name else "/"
+
+        if self.dtype.kind == "T":
+            typestr = str(self.dtype)  # StringDType()
+        else:
+            typestr = f'"{self.dtype.str}"'
```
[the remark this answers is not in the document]
crusaderky: https://github.com/numpy/numpy/issues/28670

h5py/_npystrings.pyx
crusaderky: I don't understand. Isn't a fixed-witdth array padded with NULLs? That would potentially cause a very high peak in memory usage?

docs/high/dataset.rst
```diff
@@ -508,10 +508,23 @@ Reference
         type.  Conversion is handled by HDF5 directly, on the fly::
 
             >>> dset = f.create_dataset("bigint", (1000,), dtype='int64')
-            >>> out = dset.astype('int16')[:]
+            >>> out = dset.astype('int16')[:500]
             >>> out.dtype
             dtype('int16')
 
+        This is typically faster than reading the data and then
+        converting it with NumPy:
```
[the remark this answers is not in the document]
crusaderky: I have not done timings for the numeric use case. It is _definitely_ much faster in the specific case of NpyStrings though (benchmark in the initial comment of this PR).

h5py/_hl/dataset.py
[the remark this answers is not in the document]
neutrinoceros: Thanks @crusaderky ! Can you also link the issue in an inline comment so it's easier to decide when to clean up in the future ?

h5py/_hl/dataset.py
```diff
@@ -124,10 +124,12 @@ def make_new_dset(parent, shape=None, dtype=None, data=None, name=None,
         if string_info is not None:
             # fake vlen dtype for fixed len string fillvalue
             # to not trigger unwanted encoding
-            dtype = h5t.string_dtype(string_info.encoding)
-            fillvalue = numpy.array(fillvalue, dtype=dtype)
+            fv_dtype = h5t.string_dtype(string_info.encoding)
+        elif dtype.kind == "T":
```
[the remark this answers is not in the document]
crusaderky: `h5t.check_string_dtype` and `h5t.string_dtype` are strictly for vlen dtypes - not for StringDType.

h5py/_hl/dataset.py
crusaderky: This is what makes it possible to write
```python
f.create_dataset("x", ["foo"], dtype="T")
```
or
```python
data = np.asarray(["foo"], dtype="T")
f.create_dataset("x", data)
```
otherwise, the dtype information would be immediately forgotten and I think you'd agree it would be mightily confusing for end users to have to call `.astype("T")` immediately afterwards.

docs/high/dataset.rst
crusaderky: Yes, writing a StringDType array into a vlen Dataset works.
However, if the user fancies _reading_ back from it, they would be surprised to get an object array.
Particularly (see my previous comment) if they just explicitly called `create_dataset(..., dtype="T")`.

h5py/_hl/dataset.py
crusaderky: Are you saying that `create_dataset(..., dtype="T")` should be disallowed?
Could you write a mockup of the end-user API that you have in mind?

h5py/_hl/dataset.py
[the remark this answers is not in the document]
ngoldbaum: Maybe leave a comment explaining?

h5py/_npystrings.pyx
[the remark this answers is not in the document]
ngoldbaum: I brought this up over on the StringDType tracking issue: https://github.com/numpy/numpy/issues/25693#issuecomment-2786673501

h5py/_hl/dataset.py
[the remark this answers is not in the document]
crusaderky: Done already :)

h5py/_hl/dataset.py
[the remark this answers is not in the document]
crusaderky: Done

docs/high/dataset.rst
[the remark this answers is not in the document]
crusaderky: Changed wording from _is typically faster_ to _can be faster_.

takluyver:
Sorry for the delay. I would like to avoid h5py's Dataset object having mutable state (except what is stored in the file, of course). I don't think it's a good design for a dataset newly created by specifying dtype "T" to behave differently to that same dataset when reopening the file.

I think logically the writing behaviour should not be an issue, because Numpy's variable length string dtype can only map to 1 HDF5 datatype, so any time we're trying to write data with this dtype, we should convert it to HDF5 variable-length strings. I don't believe we need to mutate the Dataset object to enable this.

Reading is a bit more complicated, because we want object arrays to remain the default (for now), but it would defeat the point to construct object arrays and then convert them to dtype "T" afterwards. But I think this is still possible without needing mutable state.

I'm going to check this branch out and play around a bit.

h5py/_hl/base.py
```diff
@@ -120,7 +120,7 @@ def array_for_new_object(data, specified_dtype=None):
     # In most cases, this does nothing. But if data was already an array,
     # and as_dtype is a tagged h5py dtype (e.g. for an object array of strings),
     # asarray() doesn't replace its dtype object. This gives it the tagged dtype:
-    if as_dtype is not None:
+    if as_dtype is not None and data.dtype.kind != "T":
```
aragilar, line 123: Maybe I missed it, but why are we special-casing the variable width strings here?

docs/high/dataset.rst
takluyver: I missed changing the docs, but since my changes that you merged into this PR, `.astype('T')` is now more like other `.as*` methods, giving you a read-only view, while you can write without needing astype.

h5py/h5t.pyx
```diff
@@ -266,6 +269,14 @@ H5Tlock(H5PY_OBJ)
 
 PYTHON_OBJECT = lockid(H5PY_OBJ)
 
+# HDF5 datatype to represent Numpy (>=2) variable-length strings
+# Should never be directly stored in files.
+cdef hid_t H5PY_NPYSTR = H5Tcreate(H5T_OPAQUE, 2 * sizeof(size_t))
```
takluyver, line 274: Just to note, the size I specified here kind of relies on the in-memory format of the numpy string datatype, which is [described in NEP 55](https://numpy.org/neps/nep-0055-string_dtype.html#memory-layout-and-managing-heap-allocations), but it's not part of the API. I don't *think* it should actually matter if it's wrong, since we're just using the datatype as a marker, and handling all the conversion ourselves, but I haven't checked that.

I guess there's an argument that we should set it to a totally incorrect value, like 1 byte, so that nothing can rely on the size. :shrug:

h5py/_npystrings.pyx
```diff
@@ -0,0 +1,255 @@
...
+
+def vstrings_scatter(hid_t space, size_t contig, size_t noncontig, size_t descr):
+    """Pure-python API interface to _proxy.pyx. This is necessary to
+    allow this module to be conditionally imported, allowing numpy 1.x to
+    continue working everywhere else.
+    """
+    _vstrings_scatter(space, <void *>contig, <void *>noncontig, <PyArray_Descr *>descr)
+
+
+cdef void _vstrings_scatter(hid_t space, void *contig, void *noncontig,
```
takluyver, line 106: It may just be me, but I'm not finding the names 'scatter' and 'gather' here easy to grasp at all - I can see where it comes from, but the core of 'scatter' is calling a 'pack' function, which as a word is kind of the opposite.

How about basing the names on something like pack/unpack, pack/load (following Numpy's API), or even something like `hdf5_strings_to_numpy` / `numpy_strings_to_hdf5`?

h5py/_npystrings.pyx
```diff
@@ -0,0 +1,255 @@
...
+    )
+    info[0].i += 1
+    return res
+
+
+def vstrings_scatter(hid_t space, size_t contig, size_t noncontig, size_t descr):
+    """Pure-python API interface to _proxy.pyx. This is necessary to
+    allow this module to be conditionally imported, allowing numpy 1.x to
+    continue working everywhere else.
+    """
```
takluyver, line 102: This is useful explanation, but I think this more public function should have the longer API docstring.

h5py/_npystrings.pyx
```diff
@@ -0,0 +1,255 @@
...
+cdef herr_t vstrings_gather_cb(
+    void *elem, hid_t type_id, unsigned ndim,
+    const hsize_t *point, void *operator_data
+) except -1:
+    cdef vstrings_gather_t *info = <vstrings_gather_t *>operator_data
+    cdef npy_static_string *unpacked = info[0].unpacked + info[0].i
+    # Obtain a reference to the string (NOT zero-terminated) and its size
+    res = NpyString_load(info[0].allocator, <npy_packed_static_string*>elem, unpacked)
+    info[0].i += 1
+    return res
```
takluyver, line 161: The NEP says that NpyString_load is allowed to return 1 to indicate a null string, and the [H5Diterate docs](https://support.hdfgroup.org/documentation/hdf5/latest/group___h5_d.html#ga71421c684884ab49765748720fe938db) imply that returning a non-zero value exits the loop, although this isn't actually specified. I couldn't trigger a problem in a brief test, but do we need any special handling around this?

takluyver:
I think we wouldn't be able to use a compound dtype containing T strings, because this is outside the more general conversion machinery, but I don't think there's a good way to solve this, given the need for the dtype object to pack & load strings. Numpy doesn't allow T within a compound dtype either, at least for now:

```text
In [35]: dt = np.dtype([('num', 'u2'), ('text', 'T')])
Cell In[35], line 1
----> 1 dt = np.dtype([('num', 'u2'), ('text', 'T')])

TypeError: StringDType is not currently supported for structured dtype fields.
```

We could also support reading & writing 'T' dtype data to attributes - I don't think there's any particular challenge for that, just some additional code to write. But attributes are also meant to be small - often just a single string - so it shouldn't be a big problem to convert to & from object arrays. So I think we can get the new functionality for datasets in & released and see how it goes in practice.

docs/strings.rst
```diff
@@ -58,13 +63,38 @@ In either case, null bytes (``'\x00'``) in the data will cause an error.
    Fixed-length strings in HDF5 hold a set number of bytes.
    It may take multiple bytes to store one character.
 
+.. _npystrings:
+
+NumPy variable-width strings
```
takluyver, line 68: This section also needs an update - datasets created with 'T' dtype are now regular variable-length string datasets, and reopening them doesn't doesn't matter.

h5py/_hl/base.py
crusaderky: There's some weird behaviour with literal string `'T'` dtype. I reworked it by ensuring it's a np.dtype object.

h5py/_npystrings.pyx
crusaderky: renamed to `npystrings_pack` / `npystrings_unpack`

h5py/_npystrings.pyx
crusaderky: Couldn't reproduce it either. For the sake of caution, I've added special handling for it and added a unit test.

h5py/_hl/base.py
```diff
@@ -102,6 +102,9 @@ def is_float16_dtype(dt):
 def array_for_new_object(data, specified_dtype=None):
     """Prepare an array from data used to create a new dataset or attribute"""
 
+    if not isinstance(specified_dtype, (np.dtype, type(None))):
+        specified_dtype = np.dtype(specified_dtype)
```
crusaderky, line 106: literal `'T'` caused weird behaviour in `data.view` below

crusaderky:
@takluyver I went through the documentation and everything else once more.
This is ready for final review and merge if there are no further comments.

docs/high/dataset.rst
```diff
@@ -504,14 +504,27 @@ Reference
...
             >>> out.dtype
             dtype('int16')
 
+        This can be faster than reading the data and then
+        converting it with NumPy:
+
+            >>> out = dset[:500].astype('int16')  # Typically slower
+
+        In case of variable-width strings, calling ``.astype('T')``
+        (NumPy's native variable-width strings) is zero cost and
```
takluyver, line 521: I'm not sure what 'zero cost' means here - clearly there is still a cost to reading and repacking data. Maybe we can just say it's more efficient than making Python string objects?

h5py/_npystrings.pyx
```diff
@@ -0,0 +1,255 @@
...
+        # 3. Copy to temporary buffer which is a concatenation of
+        #    zero-terminated char* and point to it from the
+        #    output char*[] for h5py
+        zero_terminated_buf = <char *>malloc(total_size)
+        if zero_terminated_buf is NULL:
+            raise MemoryError("Failed to allocate memory")
+        zero_terminated_cur = zero_terminated_buf
+        for i in range(npoints):
+            cur_size = info.unpacked[i].size
+            memcpy(zero_terminated_cur, info.unpacked[i].buf, cur_size)
```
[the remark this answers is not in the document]
takluyver: Aha, that makes sense, thank-you!

h5py/_npystrings.pyx
```diff
@@ -0,0 +1,262 @@
...
+    try:
+        # Multiple steps needed:
+        # 1. Read npy_packed_static_string[] from NumPy and unpack
+        #    to npy_static_string[]; which is
+        #    {const char* buf, size_t size}[] - NOT zero-terminated
+        info.unpacked = <npy_static_string*>malloc(
+            npoints * sizeof(npy_static_string)
+        )
+        if info.unpacked is NULL:
+            raise MemoryError("Failed to allocate memory")
```
takluyver, line 232: We have an `emalloc()` wrapper in `utils.pyx` that calls includes the error handling, if you want to save a few lines.

takluyver:
Thanks Guido! I think this is looking good - once we adjust the bit saying 'zero cost' in the docs I'm happy to merge it. I'll aim to do a new release fairly soon so people can easily start trying this.

ngoldbaum:
Thanks to you both for handling this. It's really awesome to see the ecosystem taking up support for StringDType. This is especially cool because StringDType is a "new-style" dtype that uses the infrastructure @seberg, @stefanv, and others got funded, implemented, and eventually shipped in NumPy 2.0.

If you notice bug reports related to this feature that look like bugs in NumPy or issues related to the use of new-style dtypes please reach out.

crusaderky:
@takluyver implemented suggestions

takluyver:
Thank you!

takluyver:
This is now out in 3.14, so it's easy for people to start trying it.

PR #2565, opened 2025-03-12
Add Path to the accepted types in the high-level doc of File

loichuder:
`pathlib.Path` is a valid type for `name` in `File` (see https://github.com/h5py/h5py/blob/187bb41ea2cbd0e78455d0b88de007354c3c8431/h5py/tests/test_file.py#L849) but it was not in the doc

### DIFF
--- a/docs/high/file.rst
+++ b/docs/high/file.rst
@@ -463,10 +463,9 @@ Reference
     listed below, :class:`File` objects inherit the full interface of
     :class:`Group`.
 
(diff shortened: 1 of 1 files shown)


aragilar:
I think this should be PathLike not Path?

neutrinoceros:
`os.PathLike[str]` is the thing we really test for, seems to me

neutrinoceros (approved):

PR #2580, opened 2025-05-09
Bump setuptools requirement in CIBW

### LINKED ISSUE #2579: WHL: wheel builds are broken on Linux
last success (April 30th): https://github.com/h5py/h5py/actions/runs/14745593606
first failure (May 1st): https://github.com/h5py/h5py/actions/runs/14768755146

The error is raised from `setuptools`
```
    configuration error: `project.license` must be valid exactly by one definition (2 matches found):
  
        - keys:
            'file': {type: string}
          required: ['file']
        - keys:
            'text': {type: string}
          required: ['text']
```

I failed to spot it last week because I assumed #2563 was the cause, but it's not.

aragilar:
The new license format requires setuptools>=77.

Closes #2579 (hopefully).

### DIFF
--- a/ci/triage_build.sh
+++ b/ci/triage_build.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+# REMEMBER TO KEEP THE VERSIONS IN THIS FILE THE SAME AS IN THE pyproject.toml
+
 set -eo pipefail
 
 if [[ "$1" == "" ]] || [[ "$2" == "" ]]; then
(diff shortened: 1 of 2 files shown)


neutrinoceros:
Thank you ! Can we also add inline comments to make sure we keep this file in sync with pyproject.toml in the future ?

neutrinoceros:
I don't think so. We must be building with build isolation disabled (`--no-build-isolation`), in which case duplication is the less harmful approach available.

aragilar:
Ah, ok, I guess if we need to avoid build isolation then that's a bit more painful. I wonder if someone's come up with a quick tool which installs build deps automatically...

neutrinoceros:
I believe it's somewhere on uv's roadmap.

takluyver:
Thanks! I'm somewhat confused why this makes a difference, since normally pip prefers the newest version of a dependency, specifying `>=` usually doesn't change anything. Maybe there's some environment variable somewhere that causes it to prefer the oldest version? :confused:

neutrinoceros:
I think what happened is that we ran in an env where setuptools was already present, but with a version that satisfied the old requirement, and not the new one.

takluyver:
Ah, that makes sense, thanks. :+1:

PR #2582, opened 2025-05-16
Tweak `Dataset.__repr__`

crusaderky:
Moved from #2557

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -1121,7 +1121,7 @@ class Dataset(HLObject):
             namestr = "(anonymous)"
         else:
             name = pp.basename(pp.normpath(self.name))
-            namestr = f'"{name}"' if name else "/"
+            namestr = f'"{name if name else "/"}"'
 
         return f'<HDF5 dataset {namestr}: shape {self.shape}, type "{self.dtype.str}">'
 
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -1120,7 +1120,7 @@ class Dataset(HLObject):
         if self.name is None:
             name = "(anonymous)"
         else:
-            name = pp.basename(pp.normpath(self.name)) or "/"
+            name = pp.basename(pp.normpath(self.name))
             name = f'"{name}"'
 
         return f'<HDF5 dataset {name}: shape {self.shape}, type "{self.dtype.str}">'
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -1115,17 +1115,15 @@ class Dataset(HLObject):
     @with_phil
     def __repr__(self):
         if not self:
-            r = '<Closed HDF5 dataset>'
+            return "<Closed HDF5 dataset>"
+
+        if self.name is None:
+            namestr = "(anonymous)"
         else:
-            if self.name is None:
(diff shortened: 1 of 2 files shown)


h5py/_hl/dataset.py
```diff
@@ -1115,17 +1115,15 @@ def __array__(self, dtype=None, copy=None):
...
-                namestr = '("anonymous")'
-            else:
-                name = pp.basename(pp.normpath(self.name))
-                namestr = '"%s"' % (name if name != '' else '/')
-            r = '<HDF5 dataset %s: shape %s, type "%s">' % (
-                namestr, self.shape, self.dtype.str
-            )
-        return r
+            name = pp.basename(pp.normpath(self.name))
+            namestr = f'"{name}"' if name else "/"
```
crusaderky, line 1124: I can't find a way to reproduce the use case for `/`?

tacaswell, line 1124:
```suggestion
            namestr = f'"{name if name else "/"}"'
```

to make sure we always have quotes in the repr.

tacaswell: wait, you are correct, a dataset can never be name '/' as '/' is by definition a group!

tacaswell:
Lifting from a folded comment, the name of a dataset can never be `/` as the root group is always (as the same suggests) a group!

crusaderky:
@tacaswell I've updated the code to reflect this

h5py/tests/test_dataset.py
```diff
@@ -49,12 +49,19 @@ class TestRepr(BaseDataset):
...
+    def test_repr_basic(self):
+        ds = self.f.create_dataset('foo', (4,), dtype='int32')
+        assert repr(ds) == '<HDF5 dataset "foo": shape (4,), type "<i4">'
+
+    def test_repr_closed(self):
         """ repr() works on live and dead datasets """
         ds = self.f.create_dataset('foo', (4,))
-        self.assertIsInstance(repr(ds), str)
         self.f.close()
-        self.assertIsInstance(repr(ds), str)
```
tacaswell, line 57: I assume these are left over from our py2/py3 transition days.

h5py/_hl/dataset.py
```diff
@@ -1115,17 +1115,15 @@ def __array__(self, dtype=None, copy=None):
...
-                name = pp.basename(pp.normpath(self.name))
-                namestr = '"%s"' % (name if name != '' else '/')
-            r = '<HDF5 dataset %s: shape %s, type "%s">' % (
-                namestr, self.shape, self.dtype.str
-            )
-        return r
+            name = pp.basename(pp.normpath(self.name))
+            name = f'"{name}"'
+
+        return f'<HDF5 dataset {name}: shape {self.shape}, type "{self.dtype.str}">'
```
tacaswell, line 1126: if we are playing code-golf I think this can all be folded into a ternary in the fstring 🤣

tacaswell: I om not sure we should be playing code golf though.

tacaswell (approved):
The number of lines of code can be reduced, it can be merged as-is.

h5py/_hl/dataset.py
crusaderky: I think it would become less readable.

PR #2585, opened 2025-05-23
Make phil Lock handling fork-safe

### LINKED ISSUE #2584: h5py hanging in subprocess with fork when multiple threads are present
When we combine multiple threads reading via `h5py` with forked subprocesses we sometimes observe the subprocess hanging. This happens because
1. in the parent process one of the threads currently holds the `phil` Lock
2. when we fork only the current thread is kept in the child process
3. if the child process now tries to acquire the `phil` Lock it will hang forever since the thread in the parent process which acquired it is not running in the child process and therefore the lock gets never released in the child process

Example code to trigger this issue:
```
def test_phil_fork_with_threads(self):
        # Test that handling of the phil Lock after fork is correct
        # even when multiple threads are present.

        with tempfile.TemporaryDirectory() as tmpdir:
            fns = []
            for i in range(10):
                fn = f'{tmpdir}/test{i}.h5'
                with h5py.File(fn, 'w') as f:
                    f.create_dataset('values', data=np.random.rand(1000, 1000))
                fns.append(fn)

            def f(fn):
                with h5py.File(fn, 'r') as f:
                    for _ in range(100):
                        _ = f['values'][:]

            # create 10 threads, each reading from an HDF5 file
            threads = []
            for fn in fns:
                thread = threading.Thread(target=f, args=(fn,))
                thread.start()
                threads.append(thread)

            # While the threads are running (and potentially holding the phil Lock)
            # create 10 processes, each also reading from an HDF5 file
            worker2pid = {}
            for worker_id, fn in enumerate(fns):
                pid = os.fork()
                if pid == 0:
                    # child process
                    f(fn)
                    os._exit(0)
                else:
                    # parent process
                    worker2pid[worker_id] = pid

            # Wait for all child processes to finish
            for worker_id, pid in worker2pid.items():
                os.waitpid(pid, 0)

            # Wait for all threads to finish
            for thread in threads:
                thread.join()
```

Here's a stack dump via py-spy:
```
Process 2389555: python /usr/local/python/python-3.11/std/bin/pytest --tb=native --import-mode=importlib h5py/tests/test_objects.py -k test_phil_fork_with_threads -s
Python v3.11.8 (/opt/python/python-3.11/bin/python3.11)

Thread 2389555 (idle): "MainThread"
    test_phil_fork_with_threads (h5py/tests/test_objects.py:85)
    _callTestMethod (unittest/case.py:579)
    run (unittest/case.py:623)
    __call__ (unittest/case.py:678)
    runtest (_pytest/unittest.py:333)
    pytest_runtest_call (_pytest/runner.py:169)
    _multicall (pluggy/_callers.py:77)
    _hookexec (pluggy/_manager.py:115)
    __call__ (pluggy/_hooks.py:493)
    <lambda> (_pytest/runner.py:262)
    from_call (_pytest/runner.py:341)
    call_runtest_hook (_pytest/runner.py:261)
    call_and_report (flaky/flaky_pytest_plugin.py:138)
    runtestprotocol (_pytest/runner.py:133)
    pytest_runtest_protocol (_pytest/runner.py:114)
    pytest_runtest_protocol (flaky/flaky_pytest_plugin.py:94)
    _multicall (pluggy/_callers.py:77)
    _hookexec (pluggy/_manager.py:115)
    __call__ (pluggy/_hooks.py:493)
    pytest_runtestloop (_pytest/main.py:350)
    _multicall (pluggy/_callers.py:77)
    _hookexec (pluggy/_manager.py:115)
    __call__ (pluggy/_hooks.py:493)
    _main (_pytest/main.py:325)
    wrap_session (_pytest/main.py:271)
    pytest_cmdline_main (_pytest/main.py:318)
    _multicall (pluggy/_callers.py:77)
    _hookexec (pluggy/_manager.py:115)
    __call__ (pluggy/_hooks.py:493)
    main (_pytest/config/__init__.py:169)
    console_main (_pytest/config/__init__.py:192)
    <module> (pytest:8)

---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Process 2389594: python /usr/local/python/python-3.11/std/bin/pytest --tb=native --import-mode=importlib h5py/tests/test_objects.py -k test_phil_fork_with_threads -s
Python v3.11.8 (/opt/python/python-3.11/bin/python3.11)
Parent Process 2389555: python /usr/local/python/python-3.11/std/bin/pytest --tb=native --import-mode=importlib h5py/tests/test_objects.py -k test_phil_fork_with_threads -s

Thread 2389594 (idle): "MainThread"
    __init__ (h5py/_hl/files.py:554)
    f (h5py/tests/test_objects.py:59)
    test_phil_fork_with_threads (h5py/tests/test_objects.py:77)
    _callTestMethod (unittest/case.py:579)
    run (unittest/case.py:623)
    __call__ (unittest/case.py:678)
    runtest (_pytest/unittest.py:333)
    pytest_runtest_call (_pytest/runner.py:169)
    _multicall (pluggy/_callers.py:77)
    _hookexec (pluggy/_manager.py:115)
    __call__ (pluggy/_hooks.py:493)
    <lambda> (_pytest/runner.py:262)
    from_call (_pytest/runner.py:341)
    call_runtest_hook (_pytest/runner.py:261)
    call_and_report (flaky/flaky_pytest_plugin.py:138)
    runtestprotocol (_pytest/runner.py:133)
    pytest_runtest_protocol (_pytest/runner.py:114)
    pytest_runtest_protocol (flaky/flaky_pytest_plugin.py:94)
    _multicall (pluggy/_callers.py:77)
    _hookexec (pluggy/_manager.py:115)
    __call__ (pluggy/_hooks.py:493)
    pytest_runtestloop (_pytest/main.py:350)
    _multicall (pluggy/_callers.py:77)
    _hookexec (pluggy/_manager.py:115)
    __call__ (pluggy/_hooks.py:493)
    _main (_pytest/main.py:325)
    wrap_session (_pytest/main.py:271)
    pytest_cmdline_main (_pytest/main.py:318)
    _multicall (pluggy/_callers.py:77)
    _hookexec (pluggy/_manager.py:115)
    __call__ (pluggy/_hooks.py:493)
    main (_pytest/config/__init__.py:169)
    console_main (_pytest/config/__init__.py:192)
    <module> (pytest:8)

---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Process 2389592: python /usr/local/python/python-3.11/std/bin/pytest --tb=native --import-mode=importlib h5py/tests/test_objects.py -k test_phil_fork_with_threads -s
Python v3.11.8 (/opt/python/python-3.11/bin/python3.11)
Parent Process 2389555: python /usr/local/python/python-3.11/std/bin/pytest --tb=native --import-mode=importlib h5py/tests/test_objects.py -k test_phil_fork_with_threads -s

Thread 2389592 (idle): "MainThread"
    __init__ (h5py/_hl/files.py:554)
    f (h5py/tests/test_objects.py:59)
    test_phil_fork_with_threads (h5py/tests/test_objects.py:77)
    _callTestMethod (unittest/case.py:579)
    run (unittest/case.py:623)
    __call__ (unittest/case.py:678)
    runtest (_pytest/unittest.py:333)
    pytest_runtest_call (_pytest/runner.py:169)
    _multicall (pluggy/_callers.py:77)
    _hookexec (pluggy/_manager.py:115)
    __call__ (pluggy/_hooks.py:493)
    <lambda> (_pytest/runner.py:262)
    from_call (_pytest/runner.py:341)
    call_runtest_hook (_pytest/runner.py:261)
    call_and_report (flaky/flaky_pytest_plugin.py:138)
    runtestprotocol (_pytest/runner.py:133)
    pytest_runtest_protocol (_pytest/runner.py:114)
    pytest_runtest_protocol (flaky/flaky_pytest_plugin.py:94)
    _multicall (pluggy/_callers.py:77)
    _hookexec (pluggy/_manager.py:115)
    __call__ (pluggy/_hooks.py:493)
    pytest_runtestloop (_pytest/main.py:350)
    _multicall (pluggy/_callers.py:77)
    _hookexec (pluggy/_manager.py:115)
    __call__ (pluggy/_hooks.py:493)
    _main (_pytest/main.py:325)
    wrap_session (_pytest/main.py:271)
    pytest_cmdline_main (_pytest/main.py:318)
    _multicall (pluggy/_callers.py:77)
    _hookexec (pluggy/_manager.py:115)
    __call__ (pluggy/_hooks.py:493)
    main (_pytest/config/__init__.py:169)
    console_main (_pytest/config/__init__.py:192)
    <module> (pytest:8)

.......
```

ArvidJB:
When we use multiple threads to read
via `h5py` with forked subprocesses we
can sometimes observe the subprocess hanging.
This happens because in the parent process one
of the threads currently holds the `phil` Lock
and when we fork only the current thread is kept
in the child process.
If the child process now tries to acquire the
`phil` Lock it will hang forever.

To work around this issue we explicitly acquire
the `phil` Lock in the thread which is about to
fork *before* forking and then release it again
*after* forking in both the parent and the child
process.

There are three scenarios:
1. no thread currently holds the `phil` Lock: in this case it is just acquired for the duration of the fork and immediately released after
2. the current thread holds the `phil` Lock: since the lock is reentrant this basically just increments the lock count before forking and decrements it after the fork
3. a different thread holds the `phil` Lock: in this case we would wait until the lock is released by the other thread and only then fork.

Closes #2584

### DIFF
--- a/h5py/_objects.pyx
+++ b/h5py/_objects.pyx
@@ -57,6 +57,25 @@ def with_phil(func):
     functools.update_wrapper(wrapper, func)
     return wrapper
 
+def _phil_before_fork():
+    """
+    Acquire the lock before forking so the current thread is the only one
+    holding the lock.
+    """
+    _phil.acquire()
+
+def _phil_after_fork():
+    """
+    Release the lock after forking in the child.
+    """
+    _phil.release()
+
+# Register fork handlers to safely handle forked child processes.
+import os
+os.register_at_fork(before=_phil_before_fork,
+                    after_in_child=_phil_after_fork,
+                    after_in_parent=_phil_after_fork)
+
 # --- End locking code --------------------------------------------------------
 
 
--- a/h5py/tests/test_objects.py
+++ b/h5py/tests/test_objects.py
@@ -6,10 +6,18 @@
 #
 # License:  Standard 3-clause BSD; see "license.txt" for full license terms
 #           and contributor agreement.
+import os
+import random
+import tempfile
+import threading
+from concurrent.futures import ThreadPoolExecutor
 
+import numpy as np
+
+import h5py
 from h5py import _objects as o
-
 from .common import TestCase
+
 
 class TestObjects(TestCase):
 
@@ -34,3 +42,48 @@ class TestObjects(TestCase):
         oid = o.ObjectID(42)
         with self.assertRaises(TypeError):
             hash(oid)
+
+    def test_phil_fork_with_threads(self):
+        # Test that handling of the phil Lock after fork is correct
+        # even when multiple threads are present.
+
+        with tempfile.TemporaryDirectory() as tmpdir:
+            fns = []
+            for i in range(10):
+                fn = f'{tmpdir}/test{i}.h5'
+                with h5py.File(fn, 'w') as f:
+                    f.create_dataset('values', data=np.random.rand(1000, 1000))
+                fns.append(fn)
+
+            def f(fn):
+                with h5py.File(fn, 'r') as f:
+                    for _ in range(100):
+                        _ = f['values'][:]
+
+            # create 10 threads, each reading from an HDF5 file
+            threads = []
+            for fn in fns:
+                thread = threading.Thread(target=f, args=(fn,))
+                thread.start()
+                threads.append(thread)
+
+            # While the threads are running (and potentially holding the phil Lock)
+            # create 10 processes, each also reading from an HDF5 file
+            worker2pid = {}
+            for worker_id, fn in enumerate(fns):
+                pid = os.fork()
+                if pid == 0:
+                    # child process
+                    f(fn)
+                    os._exit(0)
+                else:
+                    # parent process
+                    worker2pid[worker_id] = pid
+
+            # Wait for all child processes to finish
+            for worker_id, pid in worker2pid.items():
+                os.waitpid(pid, 0)
+
+            # Wait for all threads to finish
+            for thread in threads:
+                thread.join()


aragilar:
I *think* this is fine, but I want other opinions on this. I'm not sure if there are any HDF5 locks that should also be acquired/released in the same way.

neutrinoceros:
> 2. the current thread holds the phil Lock: since the lock is reentrant this basically just increments the lock count before forking and decrements it after the fork

I think I'm not getting how this is any different from what happens without your patch. Can you elaborate ?

ArvidJB:
> > 2. the current thread holds the phil Lock: since the lock is reentrant this basically just increments the lock count before forking and decrements it after the fork
> 
> I think I'm not getting how this is any different from what happens without your patch. Can you elaborate ?

Sorry if this was confusing. For this case the net end effect is the exact same behavior that we currently have, so no change in behavior.

ArvidJB:
> Rather than try to force the race, on a background thread write a helper function that uses `@phil` to hold the lock and sleep for 1s.

That's a great idea, I will try to add that, something like
```
def f_on_thread():
    phil.acquire()
    try:
        thread_acquired_phil_event.set()
        sleep(1)
    finally:
        phil.release()
```
and on the main thread
```
thread_acquired_phil_event.wait()
pid = os.fork()
if pid == 0:
    with phil:
        pass
```

h5py/tests/test_objects.py
```diff
@@ -34,3 +41,53 @@ def test_hash(self):
...
+
+            # create 10 threads, each reading from an HDF5 file
+            threads = []
+            for fn in fns:
+                thread = threading.Thread(target=f, args=(fn,))
+                thread.start()
+                threads.append(thread)
+
+            # While the threads are running (and potentially holding the phil Lock)
+            # create 10 processes, each also reading from an HDF5 file
```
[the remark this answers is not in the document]
ArvidJB: I added a test like that in `test_phil_fork_with_threads_2`.

h5py/_objects.pyx
```diff
@@ -57,6 +57,27 @@ def with_phil(func):
...
+
+def _phil_after_fork():
+    """
+    Release the lock after forking in both the parent and the child.
+    """
+    _phil.release()
+
+# Register fork handlers to safely handle `phil` Lock in forked child processes
+# in the presence of other threads which might potentially hold the lock.
+import os
```
neutrinoceros, line 75: can we move this import to the module top level ?

h5py/tests/test_objects.py
```diff
@@ -34,3 +42,91 @@ def test_hash(self):
...
+            try:
+                thread_acquired_phil_event.set()
+                time.sleep(1)
+            finally:
+                o.phil.release()
+
+        thread = threading.Thread(target=f)
+        thread.start()
+        try:
+            # on the main thread wait for the thread to have acquired the phil lock
```
neutrinoceros, line 120: this comment + your naming convention is confusing to me: which one is the "main" thread ? variable names would lead me to conclude that's `thread`, but the logic as a whole appears to imply the opposite.

h5py/tests/test_objects.py
```diff
@@ -34,3 +42,91 @@ def test_hash(self):
...
+
+            # Wait for all child processes to finish
+            for worker_id, pid in worker2pid.items():
+                os.waitpid(pid, 0)
+
+            # Wait for all threads to finish
+            for thread in threads:
+                thread.join()
+
+    def test_phil_fork_with_threads_2(self):
```
neutrinoceros, line 96: I would actually rename the other test and drop the `_2` suffix on this one; this test is actually checking the phil lock itself while the previous one was a higher level integration test, right ?

ArvidJB: Yes, the other test was my original reproducer for this issue we saw. I renamed it.

neutrinoceros:
A couple of jobs on Azure timed out. Re-trying them now.

takluyver:
> A couple of jobs on Azure timed out

Did it look like they were hanging in the newly added tests? If so, that might indicate that the issue is not reliably fixed. Or did this look more like a coincidental issue with the infrastructure.

The actual fix looks OK to me. Has someone run the new tests a few times *without* the fix to check that they ~reliably reproduce the problem?

If possible, I'd like to make the tests so they can actually fail rather than just getting stuck. For the lower-level test, I think we can do that in the child something like:

```python
if o.phil.acquire(blocking=True):
    o.phil.release()
    os._exit(0)
else:
    os._exit(1)
```

Plus checking the exit status in the parent process. It might also be worth having a timeout in the parent process, even if that means doing a non-blocking waitpid in a loop with small sleeps, rather than a single blocking waitpid.

neutrinoceros:
> Did it look like they were hanging in the newly added tests? If so, that might indicate that the issue is not reliably fixed. Or did this look more like a coincidental issue with the infrastructure?

Sorry, I didn't look closely enough to tell you :/

ArvidJB:
> If possible, I'd like to make the tests so they can actually fail rather than just getting stuck
That's a great idea. I modified the tests accordingly and (after commenting out the fork handlers) they now fail like this:
```
________________________________________________________________________________________________ TestObjects.test_fork_with_threads _________________________________________________________________________________________________
  File "/opt/python/python-3.11/lib64/python3.11/unittest/case.py", line 57, in testPartExecutor
    yield
  File "/opt/python/python-3.11/lib64/python3.11/unittest/case.py", line 623, in run
    self._callTestMethod(testMethod)
  File "/opt/python/python-3.11/lib64/python3.11/unittest/case.py", line 579, in _callTestMethod
    if method() is not None:
       ^^^^^^^^
  File "/codemill/bessen/opensource/h5py/h5py/tests/test_objects.py", line 111, in test_fork_with_threads
    assert False, "Some child processes did not finish and had to be killed"
AssertionError: Some child processes did not finish and had to be killed
assert False
______________________________________________________________________________________________ TestObjects.test_phil_fork_with_threads ______________________________________________________________________________________________
  File "/opt/python/python-3.11/lib64/python3.11/unittest/case.py", line 57, in testPartExecutor
    yield
  File "/opt/python/python-3.11/lib64/python3.11/unittest/case.py", line 623, in run
    self._callTestMethod(testMethod)
  File "/opt/python/python-3.11/lib64/python3.11/unittest/case.py", line 579, in _callTestMethod
    if method() is not None:
       ^^^^^^^^
  File "/codemill/bessen/opensource/h5py/h5py/tests/test_objects.py", line 160, in test_phil_fork_with_threads
    assert os.WEXITSTATUS(status) == 0
AssertionError: assert 1 == 0
 +  where 1 = <built-in function WEXITSTATUS>(256)
 +    where <built-in function WEXITSTATUS> = os.WEXITSTATUS
====================================================================================================== short test summary info ======================================================================================================
FAILED h5py/tests/test_objects.py::TestObjects::test_fork_with_threads - AssertionError: Some child processes did not finish and had to be killed
FAILED h5py/tests/test_objects.py::TestObjects::test_phil_fork_with_threads - assert 1 == 0
=========================================================================================== 2 failed, 809 deselected in 61.95s (0:01:01) ============================================================================================
```

> > Did it look like they were hanging in the newly added tests? If so, that might indicate that the issue is not reliably fixed. Or did this look more like a coincidental issue with the infrastructure?
> 
> Sorry, I didn't look closely enough to tell you :/

I ran these tests 100 times in a loop and was not able to get them to hang - is there a way to get the old azure test logs?

takluyver:
Thanks, I think that's looking good. :+1: 

Good point, it is possible to get the old Azure logs for a while. It looks like the Python 3.9 job on Mac OS was stuck:

```python
py39-test-mindeps/lib/python3.9/site-packages/h5py/tests/test_h5o.py .   [ 84%]
py39-test-mindeps/lib/python3.9/site-packages/h5py/tests/test_h5p.py ... [ 85%]
..........                                                               [ 86%]
py39-test-mindeps/lib/python3.9/site-packages/h5py/tests/test_h5pl.py .. [ 86%]
....                                                                     [ 87%]
py39-test-mindeps/lib/python3.9/site-packages/h5py/tests/test_h5s.py .   [ 87%]
py39-test-mindeps/lib/python3.9/site-packages/h5py/tests/test_h5t.py ... [ 87%]
                                                                         [ 87%]
py39-test-mindeps/lib/python3.9/site-packages/h5py/tests/test_h5z.py ..  [ 88%]
py39-test-mindeps/lib/python3.9/site-packages/h5py/tests/test_npystrings.py s [ 88%]
sssssssss                                                                [ 89%]
py39-test-mindeps/lib/python3.9/site-packages/h5py/tests/test_objects.py . [ 89%]
##[error]The Operation will be canceled. The next steps may not contain expected logs.
##[error]The operation was canceled.
```

That does look a bit like it could have been these new tests, if it was stuck in `test_objects`. I'm guessing from your paths that you're on Linux? If anyone on this thread has a Mac handy, it would be great if you could check the branch out and run the test suite a few times over (beats retrying CI :wink: ). If not, I think we can take our chances with it.

tacaswell:
I ran it 150 times without a hang with py3.12 and macOS 15.0.1

```
platform darwin -- Python 3.12.3, pytest-7.4.0, pluggy-1.5.0
rootdir: /Users/tcaswell/source/bnl/h5py/h5py
configfile: pytest.ini
plugins: mpi-0.6, rerunfailures-12.0, timeout-2.1.0, cov-4.1.0, xdist-3.3.1
collected 5 items

h5py/tests/test_objects.py .....     
```

ArvidJB:
> I ran it 150 times without a hang with py3.12 and macOS 15.0.1

I think this all indicates that if there is indeed another problem it is a very rare one. 

> That does look a bit like it could have been these new tests, if it was stuck in `test_objects`. I'm guessing from your paths that you're on Linux? If anyone on this thread has a Mac handy, it would be great if you could check the branch out and run the test suite a few times over (beats retrying CI 😉 ). If not, I think we can take our chances with it.

I think that since the tests no longer will get stuck but instead should time out and fail an assertion it's probably okay to merge this. If there's indeed another race condition / deadlock with forking then at least those tests should eventually discover it.

takluyver:
Yup, I agree. Thanks for your work on this!

takluyver:
And this is released as part of 3.14. :+1:

PR #2588, opened 2025-06-03
TST: improve parallel-robustness of the test suite itself using unique ids

neutrinoceros:
Got on a tagent working on #2478, and found that a large fraction of our test suite doesn't survive being run in parallel through `pytest-run-parallel` purely because of tests contain some hardcoded identifiers, creating races when run in multiple instances concurrently. I addressed a bunch of these here (sadly, not all), but I'm running out of time for today so I'm submitting this now because it could in principle be merged as is (nothing prevents doing the rest as a follow up) and on the off chance that I'm taking a wrong turn, I'd prefer getting feedback earlier rather than later :)


h5py/tests/test_base.py
```diff
@@ -106,18 +107,18 @@ def _check_type(self, obj):
...
         """ Dataset repr() with unicode """
-        dset = self.f.create_dataset(self.USTRING, (1,))
+        dset = self.f.create_dataset(f"{self.USTRING}{uuid4()}", (1,))
         self._check_type(dset)
 
     def test_namedtype(self):
         """ Named type repr() with unicode """
-        self.f['type'] = np.dtype('f')
-        typ = self.f['type']
+        self.f[uid:=str(uuid4())] = np.dtype('f')
```
neutrinoceros:
```suggestion
        uid = str(uuid4())
        self.f[uid] = np.dtype('f')
```

aragilar:
This seems pretty reasonable to me. I'm not clear if you can use pytest-run-parallel to automatically get a different ID per process, rather than requiring uuids (depending on the relative slowdown for generating all the different uuids, if it's not too large uuids seem reasonable to me). I'm also not suggesting this be added in this PR, but it's been on my todo list for a while to use hypothesis to try to test for more values (and work out how to manage the likely slowdown due to it), it might be worth watching out where it could be used, and where pytest-run-parallel would conflict with that.

neutrinoceros:
> I'm not clear if you can use pytest-run-parallel to automatically get a different ID per process

I'm not sure that's in scope for this plugin. And even if it was: I'd rather make this improvement in a way that doesn't require more (test) dependencies.


> it might be worth watching out where it could be used

In my (admittedly limited) experience with `hypothesis`, I found that pre-existing test cases rarely fit the "property-based" testing paradigm, and that building new, dedicated tests around it was preferable to retro-fitting it.

>  and where `pytest-run-parallel` would conflict with that.

If we find that they don't play nice with each other, it's always possible to mark `hypothesis` tests with `@pytest.mark.thread_unsafe` from `testes-run-parallel` later, so I'm not too worried.

takluyver:
I think it might be better to look at the test machinery and see if parallel instances of the same test can easily get different files to work on, rather than changing lots of tests. How does `pytest-run-parallel` interact with pytest fixtures like `tmp_path`? Does each parallel instance of a test get a separate path, or do they all get one?

If a test goes wrong and you have to look into the files to investigate what's happening, it's much easier to inspect short, readable, consistent names than long meaningless names which change every time you run the tests. So I'd prefer not to scatter UUIDs everywhere if it's avoidable.

tacaswell:
We should go after this by making sure tests get their own files not by making the names of objects in the files random.

neutrinoceros:
I agree this would be preferable (by a lot), but I haven't found a way to do that yet.

> How does `pytest-run-parallel` interact with `pytest` fixtures like `tmp_path`?

I couldn't tell: when I tried it, I immediately hit an issue with combining `unittest`-style test classes with `pytest` fixtures. Namely:
```
 TypeError: BaseAttrs.setUp() missing 1 required positional argument: 'tmp_path'
```
does anyone know if there's a way to make them work together, or would we need to rewrite our tests in a `pytest`-native style to find out ?

takluyver:
We have a mixture of unittest & pytest style tests already - I made a `writable_file` fixture` for the pytest style ones. So selecting just a couple of those should be enough to see how they behave with the parallel plugin.

I'm slightly reluctant to convert everything to pytest-style tests just because it would be a big diff for little immediate gain. But pytest's fixtures are nicer than setup/teardown methods, and you can still have [tests in classes](https://docs.pytest.org/en/7.1.x/getting-started.html#group-multiple-tests-in-a-class), which would mean only a big diff rather than a massive one. So if we do want to modernise our tests, it's doable.

tacaswell:
Refactoring tests suites is always a bit scary (what tests your tests!?!).  Moving mpl from nose -> pytest was a big effort.

ngoldbaum:
> I think it might be better to look at the test machinery and see if parallel instances of the same test can easily get different files to work on, rather than changing lots of tests. How does pytest-run-parallel interact with pytest fixtures like tmp_path? Does each parallel instance of a test get a separate path, or do they all get one?

Right now all threads share the same fixture, see https://github.com/Quansight-Labs/pytest-run-parallel/issues/14. This is annoying but unfortunately (at least as I understand it) is hard to fix in pytest-run-parallel due to the design of pytest.

That said, maybe there is something custom we could do in pytest-run-parallel to make this use-case nicer. Please feel free to reach out to the pytest-run-parallel developers (me and @lysnikolaou mostly) if you want to play around with this more.

crusaderky:
I've resurrected this PR in https://github.com/h5py/h5py/pull/2655. As per @takluyver's request, in my version the name remains deterministic when running regular serial tests.

neutrinoceros:
Thanks @crusaderky 
Closing in favor of https://github.com/h5py/h5py/pull/2655

PR #2590, opened 2025-06-05
MNT: bump windows image on azure (2019 -> 2022)

neutrinoceros:
xref: https://github.com/actions/runner-images/issues/12045


neutrinoceros:
This migration will be mandatory within a month and in the mean time, we might hit annoying brownout periods, so I don't think this should be controversial and I'll self-merge once CI passes.

PR #2592, opened 2025-06-06
Describe the current release process

### LINKED ISSUE #1300: Document release process

### LINKED ISSUE #1835: Replace rever for making releases?
I'm working on my third release using rever (see #1834), and I've run into a xonsh bug on Python 3.9 (https://github.com/xonsh/xonsh/issues/3888). With apologies to @scopatz, I've found rever a hard tool to love, and I'd like to propose moving away from it - though I don't have a concrete proposal for what to do instead as yet.

- Using xonsh for the config feels like an unnecessary complication - it could be Python code (our `rever.xsh` config file *is* Python code if you just replace `$VARIABLES` with `VARIABLES`), and avoid a whole pile of complex implementation (with the attendant potential for bugs like the one I ran into).
- New contributors are often get a CI failure demanding they add themselves to the `.authors.yml` file. It's not always obvious exactly what information is needed, and even when it is, it's another little obstacle to contributing.
- Typical use of rever is to run `rever 3.2.0` and let it do everything. I want a couple of stopping points for manual intervention in there - one to clean up release notes and make a 'prepare for release' PR, one to wait for CI to build wheels, then download them before uploading to PyPI (though this could be automated). I can go step-by-step manually, so this is a minor point, but it's not the way it's designed to be used.

What I do like:

- People adding news fragments in the PR with the change. There are tools like [towncrier](https://pypi.org/project/towncrier/) which focus on managing this.
- Automated updating of the version number. Again, there are more focused tools like [bump2version](https://pypi.org/project/bump2version/) that could do this.
- Having a defined process - even if I use `-a` to go through the actions manually, it's nice having the steps laid out. This could be in docs rather than code, given the way I do it.

I'm ambivalent about the authorship tracking. If we want to maintain a list of contributors sorted by num_commits, it's good to have it automated, and I think the way rever manages it is reasonable. But it is an extra bit of work for new contributors, and it means recording names & email addresses in a very obvious list. I know they're in the git log as well, but Github doesn't show email addresses on commits in the web UI, so it's a little bit more work to get them.

takluyver:
While it's fresh in my mind from releasing 3.14, I thought I'd write down how releases currently work. I've been thinking of dropping rever for some time (#1835) and it no longer works on Python 3.13, so for now this is largely a manual process, with little specific release tooling. I'm not sure how much can usefully be automated - even with tools to manage pulling together fragments into release notes, I always spend a while manually cleaning them up, reordering things, adding links, and so on.

I've left the `news` directory & the instructions to use it in the contributing docs for now. If people do write a concise release note, that can save a bit of work summarising the change, but I probably won't worry too much about asking people to do that.

Closes #1300
Closes #1835


neutrinoceros (approved):

tacaswell:
I have a followup question about why we do not use CI to make the sdist as well, but that is not relevant for documenting what is done!

takluyver:
Thanks!

It wouldn't be hard to build the sdist on CI as well, but there's not much benefit while we have the manual download + upload step to put things on PyPI. We could of course do that from CI as well, but setting it all up is just enough effort that I haven't got round to it yet. Of course it's not easy to test without actually doing a release. But if someone wants to sort out the Github actions side, I'm happy to set up the trusted publishing configuration on PyPI.

PR #2595, opened 2025-06-13
WHL: enable cp314 nightlies

neutrinoceros:

PR #2596, opened 2025-06-13
TST: broaden scope of a warning filter on a test that otherwise fails on CPython 3.14

neutrinoceros:
Discovered in #2595 (and reproduced locally), follow up to #2428


neutrinoceros:
for posterity; the issue is that the message attached to the emitted `ResourceWarning` changed in Python 3.14. Matching any message resolves the problem. We could instead match either variant, I'm just not sure it's worth it.

PR #2599, opened 2025-06-15
WHL: test our nightly builds against numpy's at runtime

neutrinoceros:
Extracted from #2595

### DIFF
--- a/ci/triage_build.sh
+++ b/ci/triage_build.sh
@@ -21,11 +21,9 @@ KIND="$RUNNER_OS $ARCH"
 CIBW_SKIP="*musllinux*"
 # If it's a scheduled build or [pip-pre] in commit message, use pip-pre
 if [[ "$GITHUB_EVENT_NAME" == "schedule" ]] || [[ "$MSG" = *'[pip-pre]'* ]]; then
-    echo "Using NumPy pip-pre wheel and (on Linux), setting CIBW_BEFORE_BUILD, CIBW_BUILD_FRONTEND and CIBW_BEFORE_TEST"
+    echo "Using NumPy pip-pre wheel and (on Linux), setting CIBW_BEFORE_BUILD, CIBW_BUILD_FRONTEND"
     echo "CIBW_BEFORE_BUILD=pip install --pre --only-binary numpy --extra-index-url https://pypi.anaconda.org/scientific-python-nightly-wheels/simple \"numpy>=2.0.0.dev0\" \"Cython>=0.29.31,<4\" pkgconfig \"setuptools>=77\" wheel" | tee -a $GITHUB_ENV
     echo "CIBW_BUILD_FRONTEND=pip; args: --no-build-isolation" | tee -a $GITHUB_ENV
-    # This is harder on other architectures, so only do it on Linux for now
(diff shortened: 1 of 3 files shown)


ci/triage_build.sh
```diff
@@ -21,11 +21,10 @@ KIND="$RUNNER_OS $ARCH"
 CIBW_SKIP="*musllinux*"
 # If it's a scheduled build or [pip-pre] in commit message, use pip-pre
 if [[ "$GITHUB_EVENT_NAME" == "schedule" ]] || [[ "$MSG" = *'[pip-pre]'* ]]; then
-    echo "Using NumPy pip-pre wheel and (on Linux), setting CIBW_BEFORE_BUILD, CIBW_BUILD_FRONTEND and CIBW_BEFORE_TEST"
+    echo "Using NumPy pip-pre wheel and (on Linux), setting CIBW_BEFORE_BUILD, CIBW_BUILD_FRONTEND"
     echo "CIBW_BEFORE_BUILD=pip install --pre --only-binary numpy --extra-index-url https://pypi.anaconda.org/scientific-python-nightly-wheels/simple \"numpy>=2.0.0.dev0\" \"Cython>=0.29.31,<4\" pkgconfig \"setuptools>=77\" wheel" | tee -a $GITHUB_ENV
     echo "CIBW_BUILD_FRONTEND=pip; args: --no-build-isolation" | tee -a $GITHUB_ENV
     # This is harder on other architectures, so only do it on Linux for now
-    echo "CIBW_BEFORE_TEST=pip install --pre --only-binary numpy --extra-index-url https://pypi.anaconda.org/scientific-python-nightly-wheels/simple \"numpy>=2.0.0.dev0\"" | tee -a $GITHUB_ENV
```
neutrinoceros, line 28: I noticed that this confusingly didn't affect what version is installed and tested against (tox does its own `pip` calls), so I'm cleaning this up

neutrinoceros:
This took me a while, but I'm closing in on it.
Notes:
- apparently if the first line in `tox.ini` after `setenv =` is a comment, any other line following it is silently ignored (might want to report this upstream)
- linux builds will not benefit from this change by itself because numpy has dropped support for `manylinux2014`, so we'll need to upgrade our own image at some point (but it doesn't need to be this PR)

neutrinoceros:
additional note: `mindeps` is broken because it relies on `oldest-supported-numpy` to pin numpy, but this metapackage hasn't been updated since 2023 so it doesn't have correct pinning information for Python 3.13 (or 3.14, for that matters). I'll report this as a proper issue later, as solutions do exist but are a little bit more involved and might require discussion

neutrinoceros:
Also I'm hitting what looks like a random fluke from #2585

```
2025-06-15T15:11:39.8511610Z >               assert False, "Some child processes did not finish and had to be killed"
2025-06-15T15:11:39.8512090Z E               AssertionError: Some child processes did not finish and had to be killed
2025-06-15T15:11:39.8512800Z E               assert False
```

I'll simply restart the job for now because it seems like a rare occurrence, but let me ping @takluyver and @ArvidJB for awareness (this too will require its own report, please don't reply directly to this thread)

for posterity, the complete log is at https://github.com/h5py/h5py/actions/runs/15664460454/job/44126741326?pr=2599

neutrinoceros:
Alright, merging now to unblock #2595, will create reports later (probably tomorrow)

Issue #2536, opened 2024-12-05
Fancy selection with many indices can be slow


takluyver:
I wanted to record this while I remember, although it's not exactly an h5py bug, h5py makes it quite easy to hit this.

We ran into a case where selecting 644_742 indices from a length 2_578_968 dimension (full shape `(2578968, 1, 256, 256)`) was extremely slow - the overall read operation taking over an hour. I investigated and got stack traces like these from py-spy:

<details><summary>Stack traces</summary>

```
Thread 2194161 (active+gil): "MainThread"
    H5S__hyper_merge_spans_helper (libhdf5-a6e30693.so.310.4.0)
    H5S__hyper_add_disjoint_spans (libhdf5-a6e30693.so.310.4.0)
    H5S__fill_in_new_space (libhdf5-a6e30693.so.310.4.0)
    H5S__generate_hyperslab (libhdf5-a6e30693.so.310.4.0)
    H5S_select_hyperslab.localalias (libhdf5-a6e30693.so.310.4.0)
    H5Sselect_hyperslab (libhdf5-a6e30693.so.310.4.0)
    defs_H5Sselect_hyperslab (defs.c:46320)
    Selector_select_fancy (_selector.c:9928)
    Selector_apply_args (_selector.c:9663)
    make_selection (_selector.c:10559)
    make_selection (_selector.c:10142)
```

```
Thread 2194161 (active+gil): "MainThread"
    _int_free (libc.so.6)
    free (libc.so.6)
    H5FL_reg_free (libhdf5-a6e30693.so.310.4.0)
    H5S__hyper_free_span_info (libhdf5-a6e30693.so.310.4.0)
    H5S__hyper_add_disjoint_spans (libhdf5-a6e30693.so.310.4.0)
    H5S__fill_in_new_space (libhdf5-a6e30693.so.310.4.0)
    H5S__generate_hyperslab (libhdf5-a6e30693.so.310.4.0)
    H5S_select_hyperslab.localalias (libhdf5-a6e30693.so.310.4.0)
    H5Sselect_hyperslab (libhdf5-a6e30693.so.310.4.0)
    defs_H5Sselect_hyperslab (defs.c:46320)
    Selector_select_fancy (_selector.c:9928)
```

```
Thread 2194161 (active+gil): "MainThread"
    _int_malloc (libc.so.6)
    malloc (libc.so.6)
    H5FL__malloc (libhdf5-a6e30693.so.310.4.0)
[...]

PR #2603, opened 2025-06-16
Speed up fancy indexing

iguinn:
See issue https://github.com/h5py/h5py/issues/2536

Fancy indexing goes very slowly when using a large number of indices. This is because the underlying hdf5 algorithm for adding hyperslab to a selection scales as O(n^2) since each new block must be checked for overlaps with every other block. This is especially noticeable once you have thousands of elements in a fancy index array. The problem and fortunately also a solution are laid out here: https://forum.hdfgroup.org/t/quadratic-runtime-of-selecting-n-hyperslabs/12555

Instead of iterating through each element in the fancy index array and adding a hyperslab one at a time, we can use a merge algorithm, since combining two non-overlapping selections is still linear in the size of the selection. This results in  O(nlogn) complexity.

To implement this, I exposed H5Scombine_hyperslab, H5Scombine_select, H5Smodify_select, and H5Sselect_copy to the low level h5s interface (only modify_select and select_copy were needed for this speedup though). These functions were introduced in hdf5 version 1.10.6. I then modified the Selector to use a recursive algorithm to build up the selection by splitting the array in two and using combine_select to merge the resulting h5s objects. If the size of the fancy index is <16, it will iterate through the list of indices as previously (I tested a few limits here and found 16 to perform best).

FancyIndexTiming_slow
FancyIndexTiming_fast

The resulting algorithm is dramatically faster than the old one. The above test used a dataspace of size 1000000 and tested different sizes of randomly selected fancy index arrays. This test can be repeated with:
```
def test_fancy_index(n_space, n_samps):
    entries = sorted(random.sample(range(n_space), n_samps))
    h5s_src = h5py.h5s.create_simple((n_space,))
    h5s_src.select_none()
    
    t0 = time.time()
    sel = h5py._selector.Selector(h5s_src)
    sel.make_selection((entries,))
    return time.time() - t0
```

I have not yet added/modified any tests; I'm unsure of what is needed for exposed low level functions, and if any changes are needed to existing tests of fancy indexing. As suggested, I tried running `tox -e pre-commit` but got an error immediately due to distutils being removed. Running the tests, all passed except test_file.TestDrivers, which was not caused by this change.

### DIFF
--- a/h5py/_selector.pyx
+++ b/h5py/_selector.pyx
@@ -242,23 +242,22 @@ cdef class Selector:
             self.is_fancy = False
         return True
 
-    cdef build_fancy_hyperslab(self, ndarray array_arg, int array_ix, hsize_t* tmp_start, hsize_t* tmp_count):
+    cdef build_fancy_hyperslab(self, space, ndarray array_arg, int array_ix, hsize_t* tmp_start, hsize_t* tmp_count):
         """Recursive merge algorithm to help select_fancy quickly apply selection to the dataspace"""
 
         # With fewer than 16 elements, looping through fancy indices is faster
         if len(array_arg)<16:
-            space = H5Scopy(self.space)
-            H5Sselect_none(space)
             for i in array_arg:
                 tmp_start[array_ix] = i
                 H5Sselect_hyperslab(space, H5S_SELECT_OR, tmp_start, self.stride, tmp_count, self.block)
             return space
         else:
-            space1 = self.build_fancy_hyperslab(array_arg[:len(array_arg)//2], array_ix, tmp_start, tmp_count)
-            space2 = self.build_fancy_hyperslab(array_arg[len(array_arg)//2:], array_ix, tmp_start, tmp_count)
-            H5Smodify_select(space1, H5S_SELECT_OR, space2)
+            self.build_fancy_hyperslab(space, array_arg[:len(array_arg)//2], array_ix, tmp_start, tmp_count)
+            space2 = H5Screate_simple(self.rank, self.dims, NULL)
+            H5Sselect_none(space2)
+            self.build_fancy_hyperslab(space2, array_arg[len(array_arg)//2:], array_ix, tmp_start, tmp_count)
+            H5Smodify_select(space, H5S_SELECT_OR, space2)
             H5Sclose(space2)
-            return space1
 
     cdef select_fancy(self, int array_ix, ndarray array_arg):
         """Apply a 'fancy' selection (array of indices) to the dataspace"""
@@ -275,9 +274,7 @@ cdef class Selector:
             memcpy(tmp_count, self.count, sizeof(hsize_t) * self.rank)
             tmp_count[array_ix] = 1
 
-            new_space = self.build_fancy_hyperslab(array_arg, array_ix, tmp_start, tmp_count)
-            H5Sselect_copy(self.space, new_space)
-            H5Sclose(new_space)
+            self.build_fancy_hyperslab(self.space, array_arg, array_ix, tmp_start, tmp_count)
         finally:
             efree(tmp_start)
             efree(tmp_count)
--- a/h5py/_selector.pyx
+++ b/h5py/_selector.pyx
@@ -242,6 +242,24 @@ cdef class Selector:
             self.is_fancy = False
         return True
 
+    cdef build_fancy_hyperslab(self, ndarray array_arg, int array_ix, hsize_t* tmp_start, hsize_t* tmp_count):
+        """Recursive merge algorithm to help select_fancy quickly apply selection to the dataspace"""
+
+        # With fewer than 16 elements, looping through fancy indices is faster
+        if len(array_arg)<16:
+            space = H5Scopy(self.space)
+            H5Sselect_none(space)
+            for i in array_arg:
+                tmp_start[array_ix] = i
+                H5Sselect_hyperslab(space, H5S_SELECT_OR, tmp_start, self.stride, tmp_count, self.block)
+            return space
+        else:
+            space1 = self.build_fancy_hyperslab(array_arg[:len(array_arg)//2], array_ix, tmp_start, tmp_count)
+            space2 = self.build_fancy_hyperslab(array_arg[len(array_arg)//2:], array_ix, tmp_start, tmp_count)
+            H5Smodify_select(space1, H5S_SELECT_OR, space2)
+            H5Sclose(space2)
+            return space1
+
     cdef select_fancy(self, int array_ix, ndarray array_arg):
         """Apply a 'fancy' selection (array of indices) to the dataspace"""
         cdef hsize_t* tmp_start
@@ -257,10 +275,9 @@ cdef class Selector:
             memcpy(tmp_count, self.count, sizeof(hsize_t) * self.rank)
             tmp_count[array_ix] = 1
 
-            # Iterate over the array of indices, add each hyperslab to the selection
-            for i in array_arg:
-                tmp_start[array_ix] = i
-                H5Sselect_hyperslab(self.space, H5S_SELECT_OR, tmp_start, self.stride, tmp_count, self.block)
+            new_space = self.build_fancy_hyperslab(array_arg, array_ix, tmp_start, tmp_count)
+            H5Sselect_copy(self.space, new_space)
+            H5Sclose(new_space)
         finally:
             efree(tmp_start)
             efree(tmp_count)
--- a/h5py/api_functions.txt
+++ b/h5py/api_functions.txt
@@ -489,7 +489,7 @@ hdf5:
 
   hid_t     H5Scombine_hyperslab(hid_t space_id, H5S_seloper_t op,  hsize_t *start, hsize_t *_stride, hsize_t *count, hsize_t *_block)
   hid_t     H5Scombine_select(hid_t space1_id, H5S_seloper_t op, hid_t space2_id)
-  hid_t     H5Smodify_select(hid_t space1_id, H5S_seloper_t op, hid_t space2_id)
+  herr_t     H5Smodify_select(hid_t space1_id, H5S_seloper_t op, hid_t space2_id)
 
   herr_t    H5Sencode(hid_t obj_id, void *buf, size_t *nalloc)
   hid_t     H5Sdecode(void *buf)
--- a/h5py/api_functions.txt
+++ b/h5py/api_functions.txt
@@ -487,7 +487,6 @@ hdf5:
   herr_t    H5Sget_select_hyper_blocklist(hid_t space_id,  hsize_t startblock, hsize_t numblocks, hsize_t *buf )
   herr_t    H5Sselect_hyperslab(hid_t space_id, H5S_seloper_t op,  hsize_t *start, hsize_t *_stride, hsize_t *count, hsize_t *_block)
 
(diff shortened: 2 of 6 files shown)


neutrinoceros:
> These functions were introduced in hdf5 version 1.10.6

We still support 1.10.4, so we'll probably need to raise our requirements. In my naive opinion this seems worth it, but @takluyver knows best.

takluyver:
> We still support 1.10.4

The minimum is 1.10.6 since last year (#2486) - did I forget to update some bit of documentation?

neutrinoceros:
Not that I know of ! Sorry, I was just misremembering, it seems !

takluyver:
Thanks, nice work! I hadn't spotted when I opened #2536 that there was a way for h5py to get round the performance issue.

I've not looked through the changes in detail yet, but please do have a look through the tests and see if there's already a case that covers the new code paths or if we need to add one - I'm also not sure. I'd start with files like `test_dataset` and `test_dataset_getitem`, but indexing logic is a major part of h5py, so it could be in other modules.

h5py/api_functions.txt
```diff
@@ -488,6 +489,9 @@ hdf5:
   herr_t    H5Sget_select_hyper_blocklist(hid_t space_id,  hsize_t startblock, hsize_t numblocks, hsize_t *buf )
   herr_t    H5Sselect_hyperslab(hid_t space_id, H5S_seloper_t op,  hsize_t *start, hsize_t *_stride, hsize_t *count, hsize_t *_block)
 
+  hid_t     H5Scombine_hyperslab(hid_t space_id, H5S_seloper_t op,  hsize_t *start, hsize_t *_stride, hsize_t *count, hsize_t *_block)
+  hid_t     H5Scombine_select(hid_t space1_id, H5S_seloper_t op, hid_t space2_id)
+  hid_t     H5Smodify_select(hid_t space1_id, H5S_seloper_t op, hid_t space2_id)
```
takluyver:
```suggestion
  herr_t     H5Smodify_select(hid_t space1_id, H5S_seloper_t op, hid_t space2_id)
```

According to the docs, anyway. If I've understood correctly, this is the difference between combine_select and modify_select - combine returns a new dataspace (all HDF5 objects are `hid_t` at the C level), and modify modifies the existing one.

h5py/h5s.pyx
```diff
@@ -584,6 +593,69 @@ cdef class SpaceID(ObjectID):
...
+        return SpaceID(H5Scombine_select(self.id, <H5S_seloper_t>op, space2.id))
+
+    @with_phil
+    def modify_select(self, SpaceID space2, int op=H5S_SELECT_OR):
+        """(SpaceID space2, INT op=SELECT_OR) => SpaceID
+
+        Refines a hyperslab selection with an operation, using a second hyperslab
+        to modify it.
+        """
+        return SpaceID(H5Smodify_select(self.id, <H5S_seloper_t>op, space2.id))
```
takluyver:
```suggestion
    def modify_select(self, SpaceID space2, int op=H5S_SELECT_OR):
        """(SpaceID space2, INT op=SELECT_OR)

        Refines a hyperslab selection with an operation, using a second hyperslab
        to modify it.
        """
        H5Smodify_select(self.id, <H5S_seloper_t>op, space2.id)
```

h5py/_selector.pyx
```diff
@@ -258,10 +276,9 @@ cdef class Selector:
             memcpy(tmp_count, self.count, sizeof(hsize_t) * self.rank)
             tmp_count[array_ix] = 1
 
-            # Iterate over the array of indices, add each hyperslab to the selection
-            for i in array_arg:
-                tmp_start[array_ix] = i
-                H5Sselect_hyperslab(self.space, H5S_SELECT_OR, tmp_start, self.stride, tmp_count, self.block)
+            new_space = self.build_fancy_hyperslab(array_arg, array_ix, tmp_start, tmp_count)
+            H5Sselect_copy(self.space, new_space)
```
takluyver: I'm not totally sure about this, but would it make sense for `build_fancy_hyperslab` to take space as a parameter and modify it, rather than always creating a new space and then copying the selection back to the original. For a large number of indices it wouldn't make much difference, because you'd have to create most of the spaces anyway, but it would presumably be a bit less overhead for smaller cases. :shrug: 

The divide-and-conquer part would then look something like this:

```cython
space2 = H5Scopy(space)
H5Sselect_none(space2)
self.build_fancy_hyperslab(space,  array_arg[:len(array_arg)//2], array_ix, tmp_start, tmp_count)
self.build_fancy_hyperslab(space2, array_arg[len(array_arg)//2:], array_ix, tmp_start, tmp_count)
H5Smodify_select(space, H5S_SELECT_OR, space2)
H5Sclose(space2)

**Review by takluyver — COMMENTED** [review]
Thanks, this is mostly looking good. I have a few specific comments inline.

I would also like basic tests for the added low-level methods, as well as making sure we've got a fancy indexing test that uses the new code pathway. One way to check the latter is to intentionally raise an error in that branch and see if the tests fail.

**Comment by takluyver** [comment]
It looks like, although these H5S functions are documented as present since 1.10.6, they're not there, or not usable somehow, until 1.10.7. The [detailed release information for 1.10.7](https://support.hdfgroup.org/documentation/hdf5/latest/rel_spec_110_change.html#subsec_rel_spec_110_change_7versus6) lists them as added then.

I'll prepare a PR to bump the minimum supported HDF5 up again; it seems reasonable at this point.

File: h5py/_selector.pyx (hunk 3 above)
Comment (iguinn) [reply h5py/_selector.pyx]: I did this with one small difference. Instead of calling H5Scopy, which takes a noticeable amount of time to copy the selections existing in space (and then gets rid of it with select_none), it creates a new space.

**Comment by iguinn** [comment]
Thanks for the feedback! I've added a couple of tests to test_dataset_getitem (and tested the test by inserting an error as suggested). I'll try to get to adding tests for the low level functions over the next few days.

**Comment by iguinn** [comment]
I've added tests now. When testing `h5s.combine_hyperslab` I found that it seems to be buggy; it seemed to work as expected only some of the time. This was with hdf5 v1.14.3. So I removed this function, which isn't important to the overall PR, and seems to be of limited utility since it is expected to be equivalent to `copy` followed by `select_hyperslab`. For the other functions, I only added very simple tests for the other low level functions for their basic functionality.

**Comment by iguinn** [comment]
Hi @takluyver and @neutrinoceros,
Are there plans on merging this into h5py? I can see how it would make sense to take it slow to avoid breaking an existing feature; if that's a concern, would it be worth splitting this into two separate PRs, where one would expose the needed hdf5 and the other would update the fancy-indexed reading? For the project I'm working on, we already use some of the low-level hdf5 commands, so we could, for the time being, write our own version of this based on these, if they were added to h5py.

File: h5py/h5s.pyx [hunk 4]
```diff
@@ -412,6 +412,15 @@ cdef class SpaceID(ObjectID):
         H5Sselect_all(self.id)
 
 
+    @with_phil
+    def select_copy(self, SpaceID src_id):
+        """(SpaceID src_id)
+
+        Copies a selection from one dataspace to another.
```
takluyver: The parameter name implies it, but it would be good for the docstring to also be clear which way this goes - it modifies this dataspace, the one you're calling the method on.

h5py/h5s.pyx
```diff
@@ -583,6 +592,24 @@ cdef class SpaceID(ObjectID):
...
+        with the resulting selection.
+        """
+        return SpaceID(H5Scombine_select(self.id, <H5S_seloper_t>op, space2.id))
+
+    @with_phil
+    def modify_select(self, SpaceID space2, int op=H5S_SELECT_OR):
+        """(SpaceID space2, INT op=SELECT_OR)
+
+        Refines a hyperslab selection with an operation, using a second hyperslab
+        to modify it.
```
takluyver: As with `select_copy`, it would be good to make it explicit that the dataspace whose method you're calling is the one that gets modified.

(To check I've understood correctly, `select_copy` is equivalent to `modify_select`  with the `H5S_SELECT_SET` operator, right?)

h5py/api_functions.txt
```diff
@@ -486,6 +487,9 @@ hdf5:
   herr_t    H5Sget_select_hyper_blocklist(hid_t space_id,  hsize_t startblock, hsize_t numblocks, hsize_t *buf )
   herr_t    H5Sselect_hyperslab(hid_t space_id, H5S_seloper_t op,  hsize_t *start, hsize_t *_stride, hsize_t *count, hsize_t *_block)
 
+  hid_t     H5Scombine_hyperslab(hid_t space_id, H5S_seloper_t op,  hsize_t *start, hsize_t *_stride, hsize_t *count, hsize_t *_block)
```
takluyver:
```suggestion
```

As we're no longer wrapping this, there's no point including it in the generated code.

h5py/tests/test_h5s.py
```diff
@@ -23,3 +23,35 @@ def test_same_shape():
 
     s3 = Helper((5, 6))[:4, :3]
     assert not s1.select_shape_same(s3)
+
+def test_select_copy():
+    s1 = h5s.create_simple((50,))
+    s1.select_hyperslab((5,), (1,), block=(5,), op=h5s.SELECT_SET)
+    s2 = h5s.create_simple((50,))
+    s2.select_copy(s1)
```
takluyver, line 32: Maybe also do a different selection on s2 first, to verify that `select_copy()` replaces the previous selection?

h5py/tests/test_h5s.py
```diff
@@ -23,3 +23,35 @@ def test_same_shape():
...
+    assert (s1.get_select_hyper_blocklist() == s2.get_select_hyper_blocklist()).all()
```
takluyver:
```suggestion
    np.testing.assert_array_equal(s1.get_select_hyper_blocklist(), s2.get_select_hyper_blocklist())
```

For numpy arrays, this can give more informative error messages if the assertion fails. (Same for the other 2 tests)

h5py/tests/test_h5s.py
```diff
@@ -23,3 +23,35 @@ def test_same_shape():
...
+    s2 = h5s.create_simple((50,))
+    s2.select_hyperslab((20,), (1,), block=(2,), op=h5s.SELECT_SET)
+
+    s3 = s1.combine_select(s2, op=h5s.SELECT_OR)
+    s2.select_hyperslab((5,), (1,), block=(5,), op=h5s.SELECT_OR)
+    assert (s3.get_select_hyper_blocklist() == s2.get_select_hyper_blocklist()).all()
+    assert not (s3.get_select_hyper_blocklist() == s1.get_select_hyper_blocklist()).all()
+
+def test_modify_select():
+    # combine_select should create a new space without modifying the original
```
takluyver, line 55:
```suggestion
```

takluyver (commented):
Thanks @iguinn ! I think this is small enough to land in one go, we don't need to break it down further, I just haven't got round to looking at it for a long time. Now that I have done so, I'm pretty happy with it, I just have a few more minor comments.

takluyver:
My suggestions were all pretty small things, so I've taken the liberty of just making changes directly on this branch. I also had to rebase it onto the master branch.

takluyver:
Thanks @iguinn !

gipert:
Hi, may I ask what's the timeline for releasing a new version that includes this patch?

takluyver:
I'm aiming to do a new release pretty soon - hopefully within a week or so, but I've been hoping that for a few weeks :wink: . At present, #2694 is top of my list to finish & land before releasing.

gipert:
Thank you and @iguinn for this work!

Issue #2601, opened 2025-06-15
TST: `mindeps` tox factor is "broken" on Python 3.13 and newer


neutrinoceros:
> `mindeps` is broken because it relies on `oldest-supported-numpy` to pin numpy, but this metapackage hasn't been updated since 2023 so it doesn't have correct pinning information for Python 3.13 (or 3.14, for that matters).

_Originally posted by @neutrinoceros in https://github.com/h5py/h5py/issues/2599#issuecomment-2974081128_

It's possible to fix this *without* duplicating the work there: we have the option of using a dependency resolver that supports this workflow. Specifically, I'm thinking of `uv`, which supports `UV_RESOLUTION=lowest` and `UV_RESOLUTION=lowest-direct`, and can be used withing tox via `tox-uv`. I implemented this solution on `astropy` already so I'm game to do this here too, unless this proposal meets objections or if another solution is preferred.
I'm *assuming* numpy devs don't want to maintain `oldest-supported-numpy` anymore (because it's original purpose is now fulfilled by numpy itself since 1.25), so I don't think upstreaming missing metadata is an option.

takluyver:
I see [oldest-supported-numpy is marked as deprecated](https://github.com/scipy/oldest-supported-numpy). It was maintained for scipy, but clearly that isn't needed any more.

I might be inclined to go back to pinning specific versions of numpy for the `mindeps` tests, because in practice we want to test with the oldest version of numpy which has wheels for each Python version, not the oldest one which the metadata claims is compatible, and I don't think uv has an option for that (I'm happy to be proved wrong 😉 ). At present, we specify `numpy>=1.19.3`, which has wheels for Python up to 3.9, so it would try to build numpy from source for 3.10-3.13, which is at least going to take a while, and possibly fail.

https://github.com/h5py/h5py/blob/107b582abca804ef9d3c152c60ee9a5a1287931d/setup.py#L27-L35

neutrinoceros:
> I don't think uv has an option for that (I'm happy to be proved wrong 😉 )

`UV_RESOLUTION=lowest uv pip install ...` will still prefer oldest versions that have matching binaries over *exact* versions specified as lower bounds, unless of course if the `--no-binary` flag is also provided !
I think it does exactly what you (and me, both) want here :)

takluyver:
In that case, I like your plan 👍

neutrinoceros:
great, I'll take a go at it when I get a minute !

neutrinoceros:
[...]

PR #2604, opened 2025-06-17
DEP: declare dev-only dependencies as PEP 735 dependency-groups

neutrinoceros:
Both `tox` and `cibuildwheel` support this standard. This move improves integration with other standard-compliant tools like `pip` or `uv`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -82,6 +82,37 @@ packages = [
     "h5py.tests.test_vds",
 ]
 
+[dependency-groups]
+check-manifest = [
+    "check-manifest>=0.50",
+]
+check-readme = [
+    "build>=1.2.2",
+    "twine>=6.1.0",
+]
+docs = [
+    "sphinx>=7.4.7",
+]
+lint = [
+    "pre-commit>=4.2.0",
+]
+mpi = [
+    "mpi4py>=3.1.1",
+]
+tables = [
+    "tables>=3.4.4",
+]
+test = [
+    "pytest>=8.2.2",
+    "pytest-cov>=5.0.0",
+    "pytest-mpi>=0.2",
+]
+wheels = [
+    "codecov>=2.1.13",
+    "coverage>=7.9.1",
+    "tox>=4.26.0",
+]
+
 [tool.cibuildwheel]
 skip = [
     "*musllinux*",
@@ -91,11 +122,7 @@ build-verbosity = 1
 build-frontend = { name = "pip", args = ["--only-binary", "numpy"] }
 manylinux-x86_64-image = "ghcr.io/h5py/manylinux_2_28_x86_64-hdf5"
 manylinux-aarch64-image = "ghcr.io/h5py/manylinux_2_28_aarch64-hdf5"
-test-requires = [
-    "tox",
-    "codecov",
-    "coverage",
-]
+test-groups = ["wheels"]
 test-command = "bash {project}/ci/cibw_test_command.sh {project} {wheel}"
 
 [tool.cibuildwheel.linux]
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -7,22 +7,14 @@ isolated_build = True
...
 deps =
-    test: pytest
-    test: pytest-cov
-    test: pytest-mpi>=0.2
-
-    # For --pre, 2.0.0b1 has a header size difference that has been reverted, so avoid it
-    py39-deps: numpy>=1.19.3,!=2.0.0b1
-    py310-deps: numpy>=1.21.3,!=2.0.0b1
-    py311-deps: numpy>=1.23.2,!=2.0.0b1
-    py312-deps: numpy>=1.26.0,!=2.0.0b1
```
neutrinoceros, line 18: this part is not replaced, but simply removed as not needed anymore: numpy 2.0.0b2 is long superseded and wouldn't be resolved to in any env anyway.

pyproject.toml
```diff
@@ -82,16 +82,30 @@ packages = [
...
 ]
 
+[dependency-groups]
+mpi = [
+    "mpi4py>=3.1.1",
+]
+tables = [
+    "tables>=3.4.4",
+]
+test = [
```
neutrinoceros, line 105: I've set arbitrary lower bounds here to roughly one-year-old versions. Having lower bounds will help with #2601, and we don't need to support very old versions for dev-only dependencies.

pyproject.toml
```diff
@@ -82,16 +82,30 @@ packages = [
...
+    "pytest>=8.2.2",
+    "pytest-cov>=5.0.0",
+    "pytest-mpi>=0.2",
+]
+wheels = [
```
neutrinoceros, line 110: here I just used current-latest versions as we don't need to check anything older: downstream distributors should be able to safely ignore this

tox.ini
```diff
@@ -4,25 +4,16 @@
...
-
     mindeps: oldest-supported-numpy
```
neutrinoceros, line 11: I'm not replacing this part yet because it needs a different change to fix #2601

neutrinoceros (commented):
Additional note: I'm intentionally not replacing `deps =` in the `apidocs` env yet because rtd doesn't have first-party support for dependency groups at the moment https://github.com/readthedocs/readthedocs.org/issues/11766
Setting it up with uv is doable, but out of scope here.

neutrinoceros:
I don't expect the idea to be controversial but maybe others would like to discuss arbitrary names I picked for some groups. Point is, I don't want to merge this without some feedback.

aragilar:
Seems fine to me? My only question is how new this support is in tox i.e. are we going to have issues that the tox required is too new (or versions of tox that support it but have issues)? I will say though this is a nice change, I hadn't realised this was now an option with tox.

neutrinoceros:
The PEP itself is relatively new (accepted in oct 2024), but tox added support for it almost immediately (tox 4.22 was released on oct 15, 2024)

aragilar:
Oh, and I guess the minversion in the tox file should mean there's a helpful error message? I doubt downstreams are using tox for testing (and maybe the dependency-groups metadata would actually make it easier for them?), so I guess the only people who will see this is contributors. Dependency-groups has had about 9 months of support in tox which isn't that long, but maybe if there's some an issue down the line we revert then if we have to (I don't expect there to be one, I've not seen any internet complaints about this yet).

+1 from me to merge this (not sure if @tacaswell or @takluyver have any objections?).

tacaswell:
I'm ok with a very aggressive minimum versions of dev tool (like tox or linters).

neutrinoceros:
This doesn't seem to be controversial, so I'm inclined to self-merge it soon if there are no clear objections.

aragilar:
Looks like there's a conflict currently, but unless @takluyver has objections, I think fix the conflict then go ahead and merge?

neutrinoceros:
conflict solved. Thanks for pointing it out !

PR #2605, opened 2025-06-17
Require HDF5 1.10.7+ or 1.12.1+

takluyver:
From #2603, some API functions we'd like to use are not usable until 1.10.7. It seems reasonable to bump the minimum supported HDF5 up again, rather than introducing more conditions. While looking at it, I also chose to drop support for 1.12.0, so we can assume the file locking APIs are present.

1.10.7 was released in September 2020, 1.12.1 in July 2021. Debian 11 has 1.10.6 and is still supported, but Debian 12 has been out for 2 years with HDF5 1.10.8. The other major [Linux distros I checked](https://pkgs.org/search/?q=hdf5) aren't affected by this move: either we already dropped compatibility last year (e.g. RHEL8), or their HDF5 is still compatible (e.g. Ubuntu 22.04).

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -92,7 +92,7 @@ jobs:
         TOXENV: py39-test-deps,py39-test-mindeps,py39-test-deps-pre
         HDF5_VERSION: 1.10.7
         HDF5_DIR: $(HDF5_CACHE_DIR)/$(HDF5_VERSION)
-        HDF5_VSVERSION: "16-64"
+        HDF5_VSVERSION: "17-64"
         H5PY_ENFORCE_COVERAGE: yes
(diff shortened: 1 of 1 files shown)


takluyver:
@neutrinoceros I think the change in 747692b was needed following #2590, but the cache meant it didn't need to rebuild HDF5 until I changed the version. I guess the version of Windows should be included in the cache key so moving to a new version triggers a rebuild, but I can't see a [predefined variable](https://learn.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#buildclean) for that. :shrug:

neutrinoceros:
Good to know ! I'll take another look later to see if we missed anything. Thanks for the ping.

PR #2607, opened 2025-06-18
Remove check-manifest pre-commit hook

takluyver:
The old version we're stuck on doesn't work with recent Python versions (as mentioned again in #2603), and newer versions aren't great in CI (see #2046) as well as being slower to run locally. I don't remember this hook actually catching any problems, so let's drop it.

Closes #2046.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,10 +7,5 @@ repos:
     -   id: trailing-whitespace
         exclude: .+\.patch
 
--   repo: https://github.com/mgedmin/check-manifest
-    rev: "0.39"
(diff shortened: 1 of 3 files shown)


neutrinoceros:
I regularly see `check-manifest` catch issues in CI (although not for `h5py`), most of the time because dev-only files are added. I agree we should drop the broken pre-commit hook; but I'd rather replace it with a simple CI job.

takluyver:
Fair enough, I've added it in to the existing `static_checks` job on Azure. I also took the API docs build out of that, since there's a separate job on Github actions which does that.

PR #2608, opened 2025-06-18
Run pre-commit in static_checks on Azure pipelines again

takluyver:
So we can avoid having a service (pre-commit.ci) with push access to the repo.

cc @tacaswell


tacaswell:
I have suspended preccommit.ci's access to the h5py org, if nothing breaks I'll remove it in a week or two.

Issue #756, opened 2016-09-22
Region references do not work when stored in datasets


jjhelmus:
Region references do not appear to work when they are stored as a dataset element but work fine when they are stored as attributes.  Not sure if this is an issue with h5py or HDF5 itself.

Minimal example:

``` Python
import h5py

f = h5py.File('test.h5', 'w')
dset1 = f.create_dataset('dset1', data=range(10))
region_ref = dset1.regionref[::2]

f.attrs['region_reference_attr'] = region_ref

ref_dtype = h5py.special_dtype(ref=h5py.Reference)
ref_dataset = f.create_dataset("ref_dataset", (1, ), dtype=ref_dtype)
ref_dataset[0] = region_ref
f.close()

f = h5py.File('test.h5', 'r')

region_ref_from_attr = f.attrs['region_reference_attr']
region_ref_from_dataset = f['ref_dataset'][0]

dset1 = f['dset1']
print(dset1[region_ref_from_attr])         # works print [0 2 4 6 8]
print(dset1[region_ref_from_dataset])  # fails 

f.close()
```

This raises a TypeError

```
[0 2 4 6 8]
Traceback (most recent call last):
  File "/home/jhelmus/anaconda/lib/python3.5/site-packages/h5py/_hl/selections.py", line 85, in select
    int(a)
TypeError: int() argument must be a string, a bytes-like object or a number, not 'h5py.h5r.Reference'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "bug.py", line 23, in <module>
    print(dset1[region_ref_from_dataset])
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper (/home/ilan/minonda/conda-bld/work/h5py/_objects.c:2696)
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper (/home/ilan/minonda/conda-bld/work/h5py/_objects.c:2654)
  File "/home/jhelmus/anaconda/lib/python3.5/site-packages/h5py/_hl/dataset.py", line 462, in __getitem__
    selection = sel.select(self.shape, args, dsid=self.id)
  File "/home/jhelmus/anaconda/lib/python3.5/site-packages/h5py/_hl/selections.py", line 88, in select
    sel[args]
  File "/home/jhelmus/anaconda/lib/python3.5/site-packages/h5py/_hl/selections.py", line 362, in __getitem__
    raise TypeError("Advanced selection inappropriate")
[...]

PR #2609, opened 2025-06-18
Documentation update for clarity in references

### LINKED ISSUE #949: DOC: regionrefs preserve shape
Just noticed that the docs still note that [regionrefs do not preserve shape]( https://github.com/h5py/h5py/blob/2.7.1/docs/refs.rst#using-region-references ) despite this having been fixed in PR ( https://github.com/h5py/h5py/pull/340 ) as part of `2.2.0`. Would be good to have the docs updated to reflect this new behavior of `regionrefs`.

### LINKED ISSUE #2594: confusion (resolved) regarding saving a region reference as a dataset (TypeError: 'h5py.h5r.RegionReference' object is not iterable)
Hi all!

I started writing out a bug issue before figuring out what I was doing wrong. I figured I would share it anyway in case anyone else runs into the same problem. I also think it could be an area where documentation could be improved a bit. The documentation has otherwise been very helpful as I get used to writing my own hdf5 files.

Below, I've included the issue I wrote before figuring it out. In short, I was getting a TypeError when trying to save a region reference because I didn't specify the keywords properly. I was trying `grp.create_dataset('ref', RefObj, dtype=h5py.regionref_dtype)` when what I should have been trying was `grp.create_dataset('ref', data=RefObj, dtype=h5py.regionref_dtype)`. The code was interpreting the latter as `grp.create_dataset('ref', shape=RefObj, dtype=h5py.regionref_dtype)`. I'm usually saving smaller files with numpy.save, so I naively assumed that the first argument after the name of the dataset would be the data itself.

The [documentation on saving reference objects to a dataset](https://docs.h5py.org/en/stable/refs.html) currently includes the example:
`ref_dataset = myfile.create_dataset("MyRefs", (100,), dtype=h5py.ref_dtype)`

I think it would be clearer to new users if it read instead: 
`ref_dataset = myfile.create_dataset("MyRefs", shape=(100,), dtype=h5py.ref_dtype)`

It wouldn't hurt to also include a specific example where the dataset is intialized with the region reference object already defined on the same page:
`regref_dataset = myfile.create_dataset("MyRegRefs", data=regref, dtype=h5py.regionref_dtype)`

I can put together a PR if this seems reasonable. I think it would also be nice for the reference documentation to briefly discuss what a user should consider before deciding whether to store a reference as an attribute or a dataset, but I'm unsure if the only consideration there is the size of the reference object.

Anyway, I assume I won't be the only one to run into this problem, so I hope this helps!

-------------------------

**Issue before I understood what I was doing wrong**

When I try to store a region reference as a dataset, I get a TypeError thrown because the region reference is not an iterable. I am able to save region references as attributes, but since my data and the regions I am referencing are very large (of order 10^8-10^9 elements), I figured it was better to store the references as a dataset. 

The error can be reproduced with this code: 

```
import h5py
import numpy as np

data = np.random.rand(10, 1)
print(data)

with h5py.File('testfile.h5', 'w') as f:
    grp = f.create_group('Group')
    ds = grp.create_dataset('data', data)
    ref = ds.regionref[:5]
    grp.create_dataset('ref', ref, dtype=h5py.regionref_dtype)
```

which throws this error:

```
Cell In[6], line 13
     11 ref = ds.regionref[:5]
     12 #print(ref.shape)
---> 13 grp.create_dataset('ref', ref, dtype=h5py.regionref_dtype)

File ~\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\local-packages\Python312\site-packages\h5py\_hl\group.py:186, in Group.create_dataset(self, name, shape, dtype, data, **kwds)
    183         parent_path, name = name.rsplit(b'/', 1)
    184         group = self.require_group(parent_path)
--> 186 dsid = dataset.make_new_dset(group, shape, dtype, data, name, **kwds)
    187 dset = dataset.Dataset(dsid)
    188 return dset

File ~\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0\LocalCache\local-packages\Python312\site-packages\h5py\_hl\dataset.py:58, in make_new_dset(parent, shape, dtype, data, name, chunks, compression, shuffle, fletcher32, maxshape, compression_opts, fillvalue, scaleoffset, track_times, external, track_order, dcpl, dapl, efile_prefix, virtual_prefix, allow_unknown_filter, rdcc_nslots, rdcc_nbytes, rdcc_w0, fill_time)
     56     shape = data.shape
     57 else:
---> 58     shape = (shape,) if isinstance(shape, int) else tuple(shape)
     59     if data is not None and (product(shape) != product(data.shape)):
     60         raise ValueError("Shape tuple is incompatible with data")

TypeError: 'h5py.h5r.RegionReference' object is not iterable
```

This error is also raised when saving object references, which can be reproduced with this code:
```
with h5py.File('myfile.hdf5', 'w') as myfile:
    mygroup = myfile.create_group('group')
    ref = mygroup.ref
    print(ref)
    ref_dataset = myfile.create_dataset("MyRefs", ref, dtype=h5py.ref_dtype)
```

In the documentation example, they use: ` ref_dataset = myfile.create_dataset("MyRefs", (100,), dtype=h5py.ref_dtype)`, which does not raise the error.

**Version Info**

I'm running into this issue on my personal machine and an HPC envrionment that are running slightly different versions:

```
Summary of the h5py configuration (personal)

h5py    3.14.0
HDF5    1.14.6
Python  3.12.10 (tags/v3.12.10:0cc8128, Apr  8 2025, 12:21:36) [MSC v.1943 64 bit (AMD64)]
sys.platform    win32
numpy   2.1.3
cython (built with) 3.1.1
numpy (built against) 2.2.6
HDF5 (built against) 1.14.6
```

```

Summary of the h5py configuration (HPC)

h5py    3.12.1
HDF5    1.14.5
Python  3.11.11 (main, Dec 11 2024, 16:28:39) [GCC 11.2.0]
sys.platform    linux
numpy   1.26.4
cython (built with) 3.0.11
numpy (built against) 2.2.1
HDF5 (built against) 1.14.5
```

bryannemcd:
Minor updates to the documentation of references, resolves #2594.

I also added a note regarding the error that gets raised when the proper dtype is not specified, related to #756 

We should also close #949, as that has been fixed for some time (Commit 935e01a)

### DIFF
--- a/docs/refs.rst
+++ b/docs/refs.rst
@@ -98,9 +98,10 @@ Here is an example with the region reference ``regref`` defined above:
 
     >>> regref_dataset = myfile.create_dataset("MyRegRefs", data=regref, dtype=h5py.regionref_dtype)
 
-Note that this specifies the data type as a region reference.
-If you do not specify the correct data type, 
-you will get a ``TypeError`` when trying to index a dataset with a region reference.
+To read a reference from a dataset:
+    >>> myregref = regref_dataset[()]
+Then, the reference can be used to access the region within the dataset it points to:
+    >>> myds[myregref]
 
 Storing references in an attribute
 ----------------------------------
--- a/docs/refs.rst
+++ b/docs/refs.rst
@@ -93,15 +93,10 @@ You can write the reference to the dataset and retrieve the reference as you wou
     >>> print(ref_dataset[0])
     <HDF5 object reference>
 
-You can also write the reference to the dataset in a single line, without specifying the shape. 
+You can also create a dataset containing references in a single line. 
 Here is an example with the region reference ``regref`` defined above:
 
-    >>> regref_dataset = myfile.create_dataset("MyRegRefs", data=regref, dtype=h5py.regionref_dtype)
-
-To read a reference from a dataset:
-    >>> myregref = regref_dataset[()]
-Then, the reference can be used to access the region within the dataset it points to:
-    >>> myds[myregref]
+    >>> regref_dataset = myfile.create_dataset("MyRegRefs", data=regref)
 
 Storing references in an attribute
 ----------------------------------
--- a/docs/refs.rst
+++ b/docs/refs.rst
@@ -82,7 +82,7 @@ These dtypes are available from h5py for references and region references:
 * ``h5py.ref_dtype`` - for object references
 * ``h5py.regionref_dtype`` - for region references
 
-To store an array of references, use the appropriate dtype when initalizing the
+To store an array of references, use the appropriate dtype when initializing the
 dataset:
 
     >>> ref_dataset = myfile.create_dataset("MyRefs", shape=(100,), dtype=h5py.ref_dtype)
--- a/docs/refs.rst
+++ b/docs/refs.rst
@@ -48,7 +48,7 @@ Using region references
 Region references always contain a selection.  You create them using the
 dataset property "regionref" and standard NumPy slicing syntax:
 
-    >>> myds = myfile.create_dataset('dset', (200,200))
+    >>> myds = myfile.create_dataset('dset', shape=(200,200))
     >>> regref = myds.regionref[0:10, 0:5]
     >>> print(regref)
     <HDF5 region reference>
@@ -82,16 +82,25 @@ These dtypes are available from h5py for references and region references:
 * ``h5py.ref_dtype`` - for object references
 * ``h5py.regionref_dtype`` - for region references
 
-To store an array of references, use the appropriate dtype when creating the
+To store an array of references, use the appropriate dtype when initalizing the
 dataset:
 
-    >>> ref_dataset = myfile.create_dataset("MyRefs", (100,), dtype=h5py.ref_dtype)
+    >>> ref_dataset = myfile.create_dataset("MyRefs", shape=(100,), dtype=h5py.ref_dtype)
 
-You can read from and write to the array as normal:
+You can write the reference to the dataset and retrieve the reference as you would any other dataset:
 
     >>> ref_dataset[0] = myfile.ref
     >>> print(ref_dataset[0])
     <HDF5 object reference>
+
+You can also write the reference to the dataset in a single line, without specifying the shape. 
+Here is an example with the region reference ``regref`` defined above:
+
+    >>> regref_dataset = myfile.create_dataset("MyRegRefs", data=regref, dtype=h5py.regionref_dtype)
+
+Note that this specifies the data type as a region reference.
+If you do not specify the correct data type, 
+you will get a TypeError when trying to index a dataset with a region reference.
 
 Storing references in an attribute
 ----------------------------------
--- a/docs/refs.rst
+++ b/docs/refs.rst
@@ -100,7 +100,7 @@ Here is an example with the region reference ``regref`` defined above:
 
 Note that this specifies the data type as a region reference.
 If you do not specify the correct data type, 
-you will get a TypeError when trying to index a dataset with a region reference.
+you will get a ``TypeError`` when trying to index a dataset with a region reference.
 
 Storing references in an attribute
 ----------------------------------
--- a/docs/refs.rst
+++ b/docs/refs.rst
@@ -93,7 +93,7 @@ You can write the reference to the dataset and retrieve the reference as you wou
     >>> print(ref_dataset[0])
     <HDF5 object reference>
 
-You can also create a dataset containing references in a single line. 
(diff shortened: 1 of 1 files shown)


docs/refs.rst
```diff
@@ -82,17 +82,26 @@ These dtypes are available from h5py for references and region references:
...
     <HDF5 object reference>
 
+You can also write the reference to the dataset in a single line, without specifying the shape. 
+Here is an example with the region reference ``regref`` defined above:
+
+    >>> regref_dataset = myfile.create_dataset("MyRegRefs", data=regref, dtype=h5py.regionref_dtype)
+
+Note that this specifies the data type as a region reference.
+If you do not specify the correct data type, 
+you will get a ``TypeError`` when trying to index a dataset with a region reference.
```
takluyver: I don't think you do need to explicitly specify the dtype in this case. I've just tried:

```
In [11]: rrds = f.create_dataset('region_ref', data=regref)

In [15]: rr_out = rrds[()]

In [16]: f['int_seq'][rr_out]
Out[16]: array([10, 11, 12, 13, 14, 15, 16, 17, 18, 19])
```

It doesn't like a list of references without a dtype (`data=[regref]`), but that fails directly on creating the dataset - and could probably be fixed. I'm not sure why you'd get a TypeError later when using a retrieved region reference.

bryannemcd: I was basing the TypeError off #756, and from what I remembered seeing when running into similar issues. But I understand now that the problem is how one reads the region reference from a dataset. The TypeError is raised when the dataset holding the region reference is used in place of the region reference itself:

    >>> f = h5py.File('myfile.hdf5','w')
    >>> dset = f.create_dataset('myds', (200,200))
    >>> regref = dset.regionref[0:10,0:5]
    >>> regref_ds = f.create_dataset('myrefs', data=regref)
    >>> regref_ds[()]
    <HDF5 region reference>
    >>> regref_ds
    <HDF5 dataset "myrefs": shape (), type "|O">
    >>> dset[regref_ds[()]]
    array([[0., 0., 0., 0., 0.],
        [0., 0., 0., 0., 0.],
        [0., 0., 0., 0., 0.],
        [0., 0., 0., 0., 0.],
        [0., 0., 0., 0., 0.],
        [0., 0., 0., 0., 0.],
        [0., 0., 0., 0., 0.],
        [0., 0., 0., 0., 0.],
        [0., 0., 0., 0., 0.],
        [0., 0., 0., 0., 0.]], dtype=float32)
    >>> dset[regref_ds]
    File "/projects/mccleary_group/mcdonough.b/TNG_dust/TNG-DUST-env/lib/python3.11/site-packages/h5py/_hl/dataset.py", line 854, in __getitem__
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/projects/mccleary_group/mcdonough.b/TNG_dust/TNG-DUST-env/lib/python3.11/site-packages/h5py/_hl/selections.py", line 82, in select
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "h5py/_selector.pyx", line 282, in h5py._selector.Selector.make_selection
    File "h5py/_selector.pyx", line 181, in h5py._selector.Selector.apply_args
    TypeError: Selection can't process <HDF5 dataset "myrefs": shape (), type "|O">

This error is of course also raised even when the correct dtype is specfied. I can correct the statement in the PR, and add an example of reading references from datasets.

```suggestion
    >>> regref_dataset = myfile.create_dataset("MyRegRefs", data=regref, dtype=h5py.regionref_dtype)

To read a reference from a dataset:
    >>> myregref = regref_dataset[()]
Then, the reference can be used to access the region within the dataset it points to:
    >>>myds[myregref]
```

takluyver:
I've made a separate PR - #2790 - to fix a couple of flaws around storing references in datasets that were part of this confusion. Then I've simplified this change a bit, on the basis that if there are fewer quirks, there's less need to document them.

docs/refs.rst
```diff
@@ -82,17 +82,22 @@ These dtypes are available from h5py for references and region references:
 * ``h5py.ref_dtype`` - for object references
 * ``h5py.regionref_dtype`` - for region references
 
-To store an array of references, use the appropriate dtype when creating the
+To store an array of references, use the appropriate dtype when initalizing the
```
takluyver:
```suggestion
To store an array of references, use the appropriate dtype when initializing the
```

PR #2611, opened 2025-06-23
Disable timestamps on group objects by default

### LINKED ISSUE #2610: Not possible to turn time tracking off on file and group level with `tack_order=True`
I am not able to generate identical data by running the same code multiple times.

This was stated as the reason for setting `track_times=False` by default in `create_dataset`:
- https://github.com/h5py/h5py/issues/1953 (merged 2021-09)

The following code generates different data each time it is run, only when I set `track_order=False` I get what I want:
```python
import h5py
import hashlib

with h5py.File('data.h5', 'w', track_order=True) as hf:
    g = hf.create_group("group", track_order=True)

with open('data.h5', "rb") as f:
    digest = hashlib.md5(f.read()).hexdigest()
print(digest)
```

Chances are that the same problem is mentioned here:
- https://github.com/h5py/h5py/issues/1919#issuecomment-874355675 (2021-07)

I can't turn `track_order` off because I don't use `h5py` directly, but through `h5netcdf` and `xarray`. Related issues form there:
- https://github.com/h5netcdf/h5netcdf/issues/211
- https://github.com/pydata/xarray/issues/7680

I was able to fix the issue by directly changing the code in `h5py` by calling `set_obj_track_times(False)` on the results of `h5p.create(h5p.FILE_CREATE)` and `h5p.create(h5p.GROUP_CREATE)` it fixes the example code above, but I am not sure that these are the only places where this is needed. And obviously there is no API right now to do this properly, so it is not a solution but a quick hack.

Version info:
```
$ python -c 'import h5py; print(h5py.version.info)'

h5py    3.14.0
HDF5    1.14.6
Python  3.13.5 (main, Jun 12 2025, 08:56:14) [GCC 13.3.0]
sys.platform    linux
numpy   2.3.0
cython (built with) 3.1.1
numpy (built against) 2.2.6
HDF5 (built against) 1.14.6
```

takluyver:
The default settings cause groups (including the root group) to be created with v1 object headers and no timestamps by default, but setting `track_order=True` creates a v2 object header, and then timestamps are included by default. This is an unexpected inconsistency, and timestamps mean that the file is not reproducible.

In line with what we already did for datasets (#1953), this explicitly disables timestamps by default, so creating files is more reproducible, and adds a `track_times` parameter to turn timestamps on.

I added a test with a 1.1 second sleep; the timestamps that HDF5 stores are only to the second, so I don't see any better way to test this.

Closes #2610

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,7 +1,7 @@
 [pytest]
 markers =
     network: Tests requiring network access, skipped with --no-network
-    slow: Slow tests, skipped unless --run-slow is used
+    slow: Slow tests
 filterwarnings =
     error
 required_plugins = pytest-mpi


takluyver:
I'll merge this in a couple more days if there are no objections.

h5py/_hl/group.py
```diff
@@ -54,13 +48,25 @@ def create_group(self, name, track_order=None):
         track_order
             Track dataset/group/attribute creation order under this group
             if True. If None use global default h5.get_config().track_order.
+        track_times
+            Store timestamps for this group in the file.
```
neutrinoceros:
```suggestion
        track_times: bool, default: False
            If True, store timestamps for this group in the file.
```

h5py/tests/test_file.py
```diff
@@ -1065,3 +1068,18 @@ def test_close_gc(writable_file):
...
+
+def test_reproducible_file(tmp_path):
+    def write(path):
+        with File(path, 'w', track_order=True) as hf:
+            g = hf.create_group("group", track_order=True)
+            g.create_dataset("dset", shape=10, dtype='f4')
+
+    f1 = tmp_path / "f1.h5"
+    write(f1)
+    time.sleep(1.1)  # Ensure any timestamps are different
```
neutrinoceros, line 1082: does it really need to be that long to avoid flakyness ?

neutrinoceros (commented):
a quick suggestion and a minor question

h5py/tests/test_file.py
takluyver: The timestamps in the HDF5 format only have second precision, so anything less than 1 second delay is unreliable because you'd sometimes create both files in the same clock second. The extra .1 is probably  unnecessary; I just want to make sure that if the platform fudges the timer a bit, it's still over a second.

It is annoying to have a delay in every test run for something that may well never be an issue again (:crossed_fingers:). Better ideas welcome. Maybe we should make it skipped by default and use some option to enable it.

neutrinoceros: Ah, I see. Then we could define a custom pytest marker `@pytest.mark.slow` for long-running tests that would be skipped unless explicitly enabled via the command line. This could easily be done in a separate PR, but ideally we would give it a try first. Do you have time for it or would you like me to step in ? I could probably get an implementation in later today.

takluyver: I can tackle that. :slightly_smiling_face:

takluyver:
:face_exhaling: It seems like pytest [won't fully use a conftest.py file](https://github.com/pytest-dev/pytest/issues/1596) found inside a package with the `--pyargs` option, which is how we generally run h5py's tests. Putting a `conftest.py` file at the project root (next to `pyproject.toml`)  allows it to see the option, but then it tries to import h5py from the source folder instead of the installed copy, which fails.

I'll leave the marker in place but go back to running all the tests for now.

takluyver:
One way round that would be to use a plugin like [pytest-optional-tests](https://github.com/reece/pytest-optional-tests) or [pytest-explicit](https://github.com/cryptaliagy/pytest-explicit), rather than defining the logic for skipping tests in our own `conftest.py`. I think that can be a separate PR, though.

h5py/_hl/files.py
```diff
@@ -429,6 +433,8 @@ def __init__(self, name, mode='r', driver=None, libver=None, userblock_size=None
         track_order
             Track dataset/group/attribute creation order under root group
             if True. If None use global default h5.get_config().track_order.
+        track_times
+            Store timestamps for the root group in the file.
```
neutrinoceros: I should have seen this the first time around, sorry
```suggestion
        track_times: bool or None, default: False
            If True, store timestamps for this group in the file.
            If None, fall back to the default value.
```

h5py/_hl/group.py
```diff
@@ -54,13 +48,25 @@ def create_group(self, name, track_order=None):
         track_order
             Track dataset/group/attribute creation order under this group
             if True. If None use global default h5.get_config().track_order.
+        track_times: bool, default: False
+            If True, store timestamps for this group in the file.
```
neutrinoceros:
```suggestion
        track_times: bool or None, default: False
            If True, store timestamps for this group in the file.
            If None, fall back to the default value.
```

h5py/_hl/group.py
```diff
@@ -54,13 +48,25 @@ def create_group(self, name, track_order=None):
...
+            if track_order:
+                order_flags = h5p.CRT_ORDER_TRACKED | h5p.CRT_ORDER_INDEXED
+                gcpl.set_link_creation_order(order_flags)
+                gcpl.set_attr_creation_order(order_flags)
+            if track_times is None:
+                track_times = False  # Allow explicit None to mean h5py's default
+            if track_times in (True, False):
+                gcpl.set_obj_track_times(track_times)
+            else:
+                raise TypeError("track_times must be either True or False")
```
neutrinoceros:
```suggestion
                raise TypeError("track_times must be either True, False, or None")
```

neutrinoceros (commented):
If we can't disable slow test by default here I'd be in favor of dropping the logic immediately and enabling `pytest-optional-tests` (which appears to be actively maintained) in a follow up PR. Thanks for exploring the simple implementation first though !

takluyver:
Thanks, I've now got rid of the logic to skip slow tests by default.

tacaswell:
~~Do we need to add `--run-slow` to at least some of our CI jobs?~~

Sorry, was reading the code and commented before catching up on the other comments 🤦🏻

pytest.ini
```diff
@@ -1,6 +1,7 @@
 [pytest]
 markers =
-    nonetwork: Tests to skip when no network access
+    network: Tests requiring network access, skipped with --no-network
+    slow: Slow tests, skipped unless --run-slow is used
```
tacaswell: How do we skip the slow tests now?

takluyver: At present, we don't - we decided to leave skipping them by default for a separate PR. I should update this note, though.

```suggestion
    slow: Slow tests
```

neutrinoceros (approved):

PR #2617, opened 2025-07-25
WHL: add wheels for CPython 3.14 (GIL enabled)

neutrinoceros:
Python 3.14.0rc1 is now ABI stable, and numpy already has wheels for it on PyPI so those wheels should be ready for release (on the condition that no bug is discovered while trying this out)


neutrinoceros:
I was actually going to merge now, if you don't mind :)

neutrinoceros:
(why Aug 1st specifically though ? just curious)

tacaswell:
nothing in particular about Aug 1.  Far enough in the future that there is time to sort out any problem, but soon enough that we will definitely get it done before 3.14 final!

PR #2619, opened 2025-07-25
WHL: cross compile macos x86_64 wheels

neutrinoceros:
[GitHub is planning to retire the `macos-13` image](
https://github.blog/changelog/2025-07-11-upcoming-changes-to-macos-hosted-runners-macos-latest-migration-and-xcode-support-policy-updates/#macos-13-is-closing-down), but luckily we should be able to get away with cross-compilation now, so the migration is actually trivial (famous last words...)


tacaswell:
>   ImportError: dlopen(/Users/runner/work/h5py/h5py/.tox/py39-test-deps-pre/lib/python3.9/site-packages/h5py/_errors.cpython-39-darwin.so, 0x0002): symbol not found in flat namespace '_H5E_ALREADYEXISTS_g'

That seems bad...

neutrinoceros:
ooch, indeed. I had no problems in other repos, so I know `cibuildwheel` is capable of handling testing in this specific case. I suspect `tox` might be getting in the way and somehow break containerization ? (this seems far fetched, admittedly I have a bias against tox). In my opinion, skipping the test phase for this cross-compiled binary should be fine (it's rarely even available, so it doesn't look like the end of the world), but maybe others disagree.

neutrinoceros:
Trying to skip tests for now, but I'll run them locally on a mac-intel machine later today (possibly tomorrow).

neutrinoceros:
hum, I can actually reproduce the crash locally with one of the wheels produced

```
❯ python -c "import h5py"
  File "/Users/clm/dev/python/h5py/test-x86/39/.venv/lib/python3.9/site-packages/h5py/__init__.py", line 25, in <module>
    from . import _errors
ImportError: dlopen(/Users/clm/dev/python/h5py/test-x86/39/.venv/lib/python3.9/site-packages/h5py/_errors.cpython-39-darwin.so, 0x0002): symbol not found in flat namespace '_H5E_ALREADYEXISTS_g'
```

so:
- this isn't tox's fault
- skipping tests was a bad idea
- we need to figure this out

I don't have enough time to dig into it now, but I'll try to come back to this problem soon (preferably before `macos-13` is pulled from under us)

neutrinoceros:
In order to try and understand what's going on, I'm attempting to first create a *short* iteration loop. Since the crash is seen in CI, and `cibuildwheel`'s documentation explicitly states that `x86_64` wheels can be tested on `arm64` via Rosetta 2, I'm assuming it's possible to reproduce this with a single Apple Silicon machine, which is what I'm currently trying to do.

I overcame a couple immediate hurdles, but I'm not there yet. Let me share my progress so far:

Using a `homebrew`-installed `gcc-15` (this is important, because somehow, and seemingly without an actual shell alias, I have the `gcc` command linked to `clang`), I can cross compile locally as

```shell
CC=gcc-15 CFLAGS="-arch x86_64" uv pip install -e .
```
(I confirmed that the binary `*.so` files produced are different with and without `CFLAGS="-arch x86_64"`)
The issue I'm having now is that `uv run --no-sync python -c 'import h5py'` completes without raising an exception; meaning I do not effectively reproduce the error...
I could still try sending a locally compiled wheel to my mac-intel machine and see what happens, but that's already a much more involved loop than I'd like. Still, right now I don't have a better idea.

neutrinoceros:
Having a closer look at logs, I understand now that I still need to run tests with a `x86_64` python interpreter.
I also found that `gcc-15` did not actually support `-arch x86_64` on my system. Instead, I find using clang is recommended for cross-compilation: https://github.com/orgs/Homebrew/discussions/2499

Presumably I should be able to reproduce with something along the lines of
```
mkdir build && cd build
export UV_PYTHON=... 
uv venv --clear
CC=clang CFLAGS='-target x86_64-apple-macos11' uv pip install ../ --offline --reinstall
uv run python -c "import h5py"
```

I just need to figure out the second line:
- have a x86_64 python interpreter available on my machine
- set `UV_PYTHON` to its resolved path

neutrinoceros:
I don't know how I didn't think of that earlier, but I'm now reasonably convinced the issue is that in cross-compiling on mac, we link to HDF5 dynamically, making the resulting wheel not portable. I was hoping I could quickly confirm this hypothesis by forcing static linking in this context, using windows as an example since it is (I think ?) always how it's done there, but I realize now this is a more substantial task than I initially thought, so before I try again more seriously, does this sound like a reasonable lead to you, @tacaswell ?

aragilar:
@neutrinoceros Maybe it's worth asking the HDF group if they've seen this issue yet with their builds, and if they have any advice on the static linking?

neutrinoceros:
I've been dragging my feet on this one. But trying to take a look now I found a mistake in my approach, so trying to iron it out before I commit to reporting upstream.

neutrinoceros:
I made some progress towards cross-compiling HDF5 itself, however it's still not working as I'd expect, and while I'm learning a lot through this trial-and-error process, I feel like there must be something obvious I'm still missing. Basically, when HDF5 is cross compiled using `--host=x86_64-apple` on Apple Silicon, the result is a *static* library `libhdf5.a` together with a `.la` metadata file. Meanwhile, our build pipeline expects a `libhdf5.dylib` dynamic library, which is missing, crashing `h5py`'s build.
Any idea what's missing ? Otherwise, I'll report upstream.

neutrinoceros:
Looks like I've gone full circle on this 🙃
The builds *seemingly* succeeds, albeit with these somewhat concerning linker warnings

```
2025-08-07T14:58:01.2952470Z ld: warning: ignoring file '/Users/runner/work/h5py/h5py/cache/hdf5/1.14.6-x86_64/lib/libhdf5.310.dylib': found architecture 'arm64', required architecture 'x86_64'
2025-08-07T14:58:01.2953770Z ld: warning: ignoring file '/Users/runner/work/h5py/h5py/cache/hdf5/1.14.6-x86_64/lib/libhdf5_hl.310.dylib': found architecture 'arm64', required architecture 'x86_64'
```

and then we crash at import time:

```
2025-08-07T14:59:17.1213230Z py39-test-deps-pre: commands[2] /Users/runner/work/h5py/h5py/.tox> python -c 'from h5py.version import info; print(info)'
2025-08-07T14:59:17.4031760Z Traceback (most recent call last):
2025-08-07T14:59:17.4034530Z   File "<string>", line 1, in <module>
2025-08-07T14:59:17.4046090Z   File "/Users/runner/work/h5py/h5py/.tox/py39-test-deps-pre/lib/python3.9/site-packages/h5py/__init__.py", line 25, in <module>
2025-08-07T14:59:17.4048310Z     from . import _errors
2025-08-07T14:59:17.4050620Z ImportError: dlopen(/Users/runner/work/h5py/h5py/.tox/py39-test-deps-pre/lib/python3.9/site-packages/h5py/_errors.cpython-39-darwin.so, 0x0002): symbol not found in flat namespace '_H5E_ALREADYEXISTS_g'
```

neutrinoceros:
this upstream issue looks related, and already has a patch up for review https://github.com/HDFGroup/hdf5/issues/5709

hmaarrfk:
my patch hit a roadblock, don't get your hopes up. help would be appreciated.

But they recommended simply using github's native arm runners. I'm just not sure how to enable them at conda-forge.

hmaarrfk:
The big question is honestly: do you need fortran support for h5py on pypi? if not, just disable fortran, and the unpatched builds on HDF5-2.0 work fine, and if I recall, so do the ones on the 1.14.6

neutrinoceros:
I'm not sure what Fortran supports entails (hence, wether we need it), or how to disable it.

hmaarrfk:
You are currently using autotools, they are getting deprecated.

With `cmake` you can use the flag `HDF5_BUILD_FORTRAN‎` 
https://github.com/HDFGroup/hdf5/blob/497b65ad0f21433601e75c328a40bd83e112da1a/CMakeBuildOptions.cmake#L77C9-L77C27


For autotools, we have a bunch of patches at conda-forge that we apply that might be helpful
https://github.com/conda-forge/hdf5-feedstock/tree/main/recipe/patches

neutrinoceros:
I'm hoping switching the build system we use for HDF5 should be easy (famous last words ?), so I'd rather go straight to the long term solution, but thanks a lot for linking these patches !

takluyver:
To repeat myself from #2651, I wouldn't spend much time to keep making pre-built packages for Intel Macs if they're not readily available to run free CI on. I'm not a Mac user, but my impression is that there was a lot of enthusiasm for the ARM Macs and thus uptake was pretty fast. Remaining Intel Mac users have alternatives including conda packages, building from source and using older versions. We're at the whim of two staggeringly valuable companies here, I have no problem pointing at them if people are disappointed.

If someone personally wants to make Intel Mac builds work, and it doesn't make CI too much slower or more awkward, I'm also fine with keeping it around for a year or two more. But I don't think anyone should feel obliged to.

neutrinoceros:
I agree with everything you said. For full disclosure: I own a macbook from 2017 which I would like to keep relevant for development as long as I can, including to work on projects that depend on h5py. So, I think I *am* the person who really wants this. Keeping support for other users is just a side benefit the way I see it 😅

takluyver:
Great, of course if you have a personal use for it, go ahead! I'm still
keeping a laptop from 2015 going, albeit not a Mac.

neutrinoceros:
closing in favor of #2704

PR #2620, opened 2025-07-28
doc improvements

chris2crawford:
small improvements to the documentation:
formatting fix in file.rst
and some fixes to quick.rst to allow users to follow along on the command-line

### DIFF
--- a/docs/quick.rst
+++ b/docs/quick.rst
@@ -34,10 +34,13 @@ when using h5py is:
 
     **Groups work like dictionaries, and datasets work like NumPy arrays**
 
-Suppose someone has sent you a HDF5 file, :code:`mytestfile.hdf5`. (To create this file, read `Appendix: Creating a file`_.) The very first thing you'll need to do is to open the file for reading::
+Suppose someone has sent you a HDF5 file, :code:`mytestfile.hdf5`. (To create
+this file, follow the commands in `Appendix: Creating a file`_.) The very first
+thing you'll need to do is to open the file for reading and writing (omit the
+:code:`'a'` or replace it with the option :code:`'r'` to read only)::
 
     >>> import h5py
-    >>> f = h5py.File('mytestfile.hdf5', 'r')
+    >>> f = h5py.File('mytestfile.hdf5', 'a')
 
 The :ref:`File object <file>` is your starting point. What is stored in this file? Remember :py:class:`h5py.File` acts like a Python dictionary, thus we can check the keys,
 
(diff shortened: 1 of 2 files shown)


docs/quick.rst
```diff
@@ -73,21 +76,23 @@ For more, see :ref:`file` and :ref:`dataset`.
...
+:code:`r` (for read only access).
 A full list of file access modes and their meanings is at :ref:`file`. ::
 
     >>> import h5py
     >>> import numpy as np
     >>> f = h5py.File("mytestfile.hdf5", "w")
 
 The :ref:`File object <file>` has a couple of methods which look interesting. One of them is ``create_dataset``, which
-as the name suggests, creates a data set of given shape and dtype ::
+as the name suggests, creates a data set of given shape and dtype and then close the file::
```
takluyver, line 92: This wording implies that `create_dataset` closes the file for you, which is not the case (although your example just below is clear that it's a separate step). I'd separate this out a bit more, add a bit talking about closing the file after the `create_dataset` example and before the context manager bit.

docs/quick.rst
```diff
@@ -191,3 +195,7 @@ implements the dictionary interface::
     True
 
 For more, see :ref:`attributes`.
+
+Save the file to finalize changes::
+
+    >>> f.close()
```
takluyver, line 201: Let's leave this out from here, and just describe it in 'creating a file' - it's a bit strange to put it in the attributes section but not e.g. groups, and if we repeat it everywhere then the quickstart isn't so quick.

It's also not really accurate to describe this command as 'sav[ing]' the file - that implies we're constructing something in memory and then writing it out on close, which is not really the case (although close might do some writes as it flushes caches).

docs/quick.rst
```diff
@@ -34,10 +34,13 @@ when using h5py is:
...
 
-Suppose someone has sent you a HDF5 file, :code:`mytestfile.hdf5`. (To create this file, read `Appendix: Creating a file`_.) The very first thing you'll need to do is to open the file for reading::
+Suppose someone has sent you a HDF5 file, :code:`mytestfile.hdf5`. (To create
+this file, follow the commands in `Appendix: Creating a file`_.) The very first
+thing you'll need to do is to open the file for reading and writing (omit the
+:code:`'a'` or replace it with the option :code:`'r'` to read only)::
 
     >>> import h5py
-    >>> f = h5py.File('mytestfile.hdf5', 'r')
+    >>> f = h5py.File('mytestfile.hdf5', 'a')
```
takluyver, line 43: I'm not sure it makes sense to open read-write here, as this first section is all about reading.

If we're assuming someone follows this page all the way through, they'll reopen `f` read-write in the 'creating a file' section before any write operations. If we're assuming someone jumps to the section they're interested in, then the current structure which points out the switch to read-write mode seems preferable. The current structure of the page isn't really clear what pattern it's written for, though - maybe we need a bigger overhaul to rearrange it into a reading part and a writing part. :shrug:

chris2crawford:
I agree with your comments, except where the file is opened with 'r', and then the data is changed:
>>> dset[...] = np.arange(100)

The separation into a reading part followed by a writing part could be simplified and linearized by providing a link to a 'sample.hf5' file to open, which would also emphasize reading others' data.  Then move the dset[...]= command to the second section where you show how the datafile was created.

The quickstart would have been a lot quicker for me to follow if I could type one command after another to see the effect.

If you agree, I'll take a stab at it.


--Chris

Christopher Crawford, Professor

Chair, Faculty Senate; Director, REU/IRES Programs
Dept. Physics and Astronomy, University of Kentucky

CP281 uky.zoom.us/my/c.crawford<https://uky.zoom.us/my/c.crawford> cell: 859-421-9612

schedule: www.pa.uky.edu/~crawford<http://www.pa.uky.edu/~crawford>

________________________________

PR #2623, opened 2025-07-29
MNT: upgrade cibuildwheel to 3.1.2

neutrinoceros:
This is a non-critical upgrade, but it fixes a mistake I made in 3.1.1, so I care about it :-)

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -88,7 +88,7 @@ jobs:
       - run: bash ./ci/triage_build.sh "${{ matrix.arch }}" "${{ github.event.pull_request.head.sha || github.sha }}" "${{ matrix.python }}"
 
       # Now actually build the wheels
-      - uses: pypa/cibuildwheel@v3.1.1
+      - uses: pypa/cibuildwheel@v3.1.2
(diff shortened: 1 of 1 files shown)


.github/workflows/build_wheels.yml
```diff
@@ -88,7 +88,7 @@ jobs:
       - run: bash ./ci/triage_build.sh "${{ matrix.arch }}" "${{ github.event.pull_request.head.sha || github.sha }}" "${{ matrix.python }}"
 
       # Now actually build the wheels
-      - uses: pypa/cibuildwheel@v3.1.1
+      - uses: pypa/cibuildwheel@v3.1.2
```
tacaswell, line 91:
```suggestion
      - uses: pypa/cibuildwheel@9e4e50bd76b3190f55304387e333f6234823ea9b  # 3.1.2
```

Can we move to pinning to SHAs?

neutrinoceros: Good idea. Let me do this globally in a different PR

neutrinoceros: https://github.com/h5py/h5py/pull/2624

PR #2624, opened 2025-07-29
MNT: migrate all github actions pins to shas

neutrinoceros:
Replace #2623

For reference, this patch was performed automatically using [`gha-update`](https://pypi.org/project/gha-update/)


tacaswell:
oh, nice!  I have a pile of half-baked scripts to do the same pinning work.

PR #2625, opened 2025-07-30
Update SWMR example in documentation

### LINKED ISSUE #2622: SWMR documentation example doesn't work
The reader example on the [SWMR documentation page](https://docs.h5py.org/en/stable/swmr.html#using-the-swmr-feature-from-h5py) doesn't work:

```python
f = h5py.File("swmr.h5", 'r', libver='latest', swmr=True)
dset = f["data"]
while True:
    dset.id.refresh()
    shape = dset.shape
    print( shape )
```

Will keep on reporting the initial shape (`(4,)`) even after flushing the file in the writer process.
`dset.id.shape` correctly reports the updated shape. Similarly, replacing `dset.id.refresh()` by `dset.refresh()` makes this code work as well.

The problem is likely with some caching mechanism (since `Dataset.refresh()` also calls `self._cache_props.clear()` in addition to `id.refresh()` https://github.com/h5py/h5py/blob/master/h5py/_hl/dataset.py#L1148-L1155 ).

Could you update the documentation please?

Thanks!
Maarten

---
### h5py version info
```text

h5py    3.13.0
HDF5    1.14.6
Python  3.12.3 (main, Jun 18 2025, 17:59:45) [GCC 13.3.0]
sys.platform    linux
numpy   1.26.4
cython (built with) 3.0.12
numpy (built against) 2.2.3
HDF5 (built against) 1.14.6
```

maarten-ic:
Fixes #2622

### DIFF
--- a/docs/swmr.rst
+++ b/docs/swmr.rst
@@ -78,7 +78,7 @@ The following snippet demonstrate how to monitor a dataset as a SWMR reader::
     f = h5py.File("swmr.h5", 'r', libver='latest', swmr=True)
     dset = f["data"]
     while True:
-        dset.id.refresh()
+        dset.refresh()
         shape = dset.shape
         print( shape )
 


takluyver:
Thanks @maarten-ic !

PR #2626, opened 2025-07-30
Add dataset flush/refresh SWMR methods to reference docs

takluyver:
While looking at #2625, I noticed that the SWMR methods on datasets aren't in the reference section of the documentation.

I also removed some notes in docstrings about them being conditional on the HDF5 version, since we no longer support pre-SWMR versions, and cleared up conditions in the code which are now always true.

### DIFF
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -588,6 +588,18 @@ Reference
 
         Return the size of the first axis.
 
+    .. method:: refresh
+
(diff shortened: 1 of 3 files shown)


takluyver:
I don't think this should be controversial at all, so I'm going to merge it myself.

neutrinoceros:
Indeed. FWIW I'm happy to approve this post-mortem 💀

PR #2628, opened 2025-07-30
Try to fix mpirun 'available slots' CI problem

### LINKED ISSUE #2627: MPI tests failing with error about available slots
E.g. on #2626. This seems to be repeatable. Maybe we need to do something to make the 'slots' available before running? Or maybe this job should be moved over to Github actions?

cc @aragilar 

```
There are not enough slots available in the system to satisfy the 2
slots that were requested by the application:

  /home/vsts/work/1/s/.tox/py39-test-mindeps-mpi4py/bin/python

Either request fewer slots for your application, or make more slots
available for use.

A "slot" is the Open MPI term for an allocatable unit where we can
launch a process.  The number of slots available are defined by the
environment in which Open MPI processes are run:

  1. Hostfile, via "slots=N" clauses (N defaults to number of
     processor cores if not provided)
  2. The --host command line parameter, via a ":N" suffix on the
     hostname (N defaults to 1 if not provided)
  3. Resource manager (e.g., SLURM, PBS/Torque, LSF, etc.)
  4. If none of a hostfile, the --host command line parameter, or an
     RM is present, Open MPI defaults to the number of processor cores

In all the above cases, if you want Open MPI to default to the number
of hardware threads instead of the number of processor cores, use the
--use-hwthread-cpus option.

Alternatively, you can use the --oversubscribe option to ignore the
number of available slots when deciding the number of processes to
launch.
```

takluyver:
Closes #2627.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -30,7 +30,7 @@ commands =
     test: python {toxinidir}/ci/fix_paths.py
     test: python -c "from h5py.version import info; print(info)"
     test-!mpi4py: python -m pytest --pyargs h5py --cov=h5py -rxXs --cov-config={toxinidir}/.coveragerc {posargs}
-    test-mpi4py: mpirun -n {env:MPI_N_PROCS:2} {envpython} -m pytest --pyargs h5py -rxXs --with-mpi {posargs}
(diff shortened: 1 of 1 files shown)


takluyver:
The `--use-hwthread-cpus` option seems to have fixed it. Thanks @aragilar !

PR #2630, opened 2025-08-04
Fix segfault on converting unallocated string data to Numpy strings.

### LINKED ISSUE #2615: Seg fault on resizing unlimited dataset and then viewing as numpy variable-length strings
If I create a dataset of strings with `maxshape=(None,)`, then resize it larger, then attempt to view it as a numpy variable-length string array, I get a seg fault. Here's a minimal example:

This code seg faults for me on the last line:
```python
import h5py
import numpy as np

f = h5py.File("example.h5", "w")

data = np.array(["string1", "string2", "string3"], dtype="T")
d = f.create_dataset("dset", data=data, maxshape=(None,))

d.resize((10,))
print(d[:])
print(d.astype("T")[:])
```
Removing the resize removes the seg fault.

python version: 3.11 (linux)
h5py version: 3.14.0
numpy version: 2.2.6

takluyver:
Fixes #2615.

@crusaderky could you have a look and see if this fix looks right? I don't feel like I have a great understanding of the numpy strings API, but this change seems to prevent the segfault reported.

### DIFF
--- a/h5py/_npystrings.pyx
+++ b/h5py/_npystrings.pyx
@@ -87,7 +87,9 @@ cdef herr_t npystrings_pack_cb(
     cdef npystrings_pack_t* info = <npystrings_pack_t*>operator_data
     cdef const char* buf = info[0].contig[info[0].i]
     if buf is NULL:
-        NpyString_pack_null(info[0].allocator, <npy_packed_static_string*>elem)
+        # Special case of default values created by resize(). Regular empty strings
+        # are valid char* with strlen(buf) == 0.
+        res = NpyString_pack_null(info[0].allocator, <npy_packed_static_string*>elem)
     else:
         # Deep copy char* from h5py into the NumPy array
         res = NpyString_pack(
--- a/h5py/_npystrings.pyx
+++ b/h5py/_npystrings.pyx
@@ -58,6 +58,7 @@ ELSE:
         npy_static_string,
         npy_packed_static_string,
         NpyString_pack,
+        NpyString_pack_null,
         NpyString_load,
         NpyString_acquire_allocator,
         NpyString_release_allocator,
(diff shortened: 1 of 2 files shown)


takluyver:
The Windows build seems to be repeatedly failing the new test and then hanging. It seems it's not entirely consistent, because sometimes it hangs on Python 3.9, other times it gets past that and hangs on 3.10, but I've restarted it several times and it always seems to hang. I'm not sure why this would cause a hang, or why it's specific to Windows, and I can't easily debug it...

h5py/tests/test_npystrings.py
```diff
@@ -161,3 +161,15 @@ def test_astype_nonstring(writable_file):
     x = writable_file.create_dataset("x", shape=(2, ), dtype="i8")
     with pytest.raises(TypeError, match="HDF5 string datatype"):
         x.astype("T")
+
+
+def test_resized_read(writable_file):
```
crusaderky, line 166:
```suggestion
def test_resized_read(writable_file):
    """Read default values created by resize(). This triggers a special case
    where libhdf5 returns a char** containing NULL pointers.
    """
```

h5py/_npystrings.pyx
```diff
@@ -85,13 +86,16 @@ cdef herr_t npystrings_pack_cb(
...
     cdef const char* buf = info[0].contig[info[0].i]
-    # Deep copy char* from h5py into the NumPy array
-    res = NpyString_pack(
-        info[0].allocator,
-        <npy_packed_static_string*>elem,
-        buf,
-        strlen(buf),
-    )
+    if buf is NULL:
+        NpyString_pack_null(info[0].allocator, <npy_packed_static_string*>elem)
```
crusaderky:
```suggestion
        # Special case of default values created by resize(). Regular empty strings
        # are valid char* with strlen(buf) == 0.
        res = NpyString_pack_null(info[0].allocator, <npy_packed_static_string*>elem)
```

crusaderky: Fixes Windows hanging due to uninitialised variable. I think the reason it only hangs on Windows is that on both Linux and MacOS the stack is always initialised at 0 whereas on Windows it's random garbage.

takluyver: Aha, thank you, I totally overlooked that the return value was doing something. :sweat_smile:

crusaderky (approved):

tacaswell (approved):

PR #2631, opened 2025-08-06
TST: fix minimal env testing using `uv` + `tox-uv`

### LINKED ISSUE #2601: TST: `mindeps` tox factor is "broken" on Python 3.13 and newer

neutrinoceros:
I think I found the missing piece of the puzzle in uv's static configuration. Let's see how this pans out.
If this works, close #2601

### DIFF
--- a/ci/cibw_test_command.sh
+++ b/ci/cibw_test_command.sh
@@ -12,14 +12,7 @@ WHEEL_PATH=$2
 echo "WHEEL_PATH=$WHEEL_PATH"
 
 export PYVER=$(python -c "import sys; print(''.join(map(str, sys.version_info[:2])))")
-ENVLIST="py$PYVER-test-deps,py$PYVER-test-deps-pre"
-
-if ! ( [[ "$AUDITWHEEL_PLAT" == musllinux_* ]] && [[ "$PYVER" == "39" || "$PYVER" == "310" || "$PYVER" == "311" ]] ); then
-    # skip mindeps on musllinux + python 3.9 to 3.11 because oldest supported numpy versions
-    # are higher for this target (1.25.0) and there's no way (that I found) to specify it
-    # directly in package metadata.
-    ENVLIST="py$PYVER-test-mindeps,$ENVLIST"
-fi
+ENVLIST="py$PYVER-test-mindeps,py$PYVER-test-deps,py$PYVER-test-deps-pre"
 
 export H5PY_TEST_CHECK_FILTERS=1
 echo "ENVLIST=$ENVLIST"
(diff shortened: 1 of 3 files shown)


neutrinoceros:
It *almost* works, except for the fact that it seemingly breaks running tox on Windows via azure CI. The error message isn't really helpful though, and I don't see anything beyond this:
```
error: command 'C:\\Program Files\\Microsoft Visual
      Studio\\2022\\Enterprise\\VC\\Tools\\MSVC\\14.44.35207\\bin\\HostX86\\x64\\cl.exe'
      failed with exit code 1

      hint: This usually indicates a problem with the package or the build
      environment.
```

neutrinoceros:
Actually, grepping for `error` in raw logs reveals an actual (but well hidden in the surrounding warning noise) error:
```
2025-08-07T10:18:47.0952646Z       C:\Users\VssAdministrator\AppData\Local\uv\cache\sdists-v9\path\95d5ce2ea0d2d66c\GTtsp5zg5io8f7gNAP04F\src\h5py\defs.c
2025-08-07T10:18:47.0952981Z       : fatal error C1083: Cannot open compiler generated file: '': Invalid
2025-08-07T10:18:47.0953337Z       argument
```

neutrinoceros:
I worked around the Azure issue by simply skipping `tox-uv` installation there. Maybe that's a bit hacky but I think it's still a step in the right direction, so let's open this for review !

neutrinoceros:
Tentatively triaging this for 3.15 because other useful PRs are starting to include workarounds that this makes unnecessary

neutrinoceros:
I merged both PRs my previous message alluded to. Now these workarounds are cleaned up here.

neutrinoceros:
actually there is still another workaround in #2646 that I'd like to avoid merging.
@tacaswell @takluyver, what do you think of this patch ?

neutrinoceros:
I'm again hitting cases that really make me wish uv was used in CI, and I'm not seeing any objections yet so I'm now planning to self merge in a couple days unless someone speaks up.

PR #2632, opened 2025-08-06
WHL: enable musllinux wheels

neutrinoceros:
This is an experiment. I suspsect our only reason for skipping these was that NumPy *used to* not ship `musllinux` wheels, but that hasn't been true for a little while so let's try it out.

### DIFF
--- a/ci/cibw_test_command.sh
+++ b/ci/cibw_test_command.sh
@@ -12,7 +12,14 @@ WHEEL_PATH=$2
 echo "WHEEL_PATH=$WHEEL_PATH"
 
 export PYVER=$(python -c "import sys; print(''.join(map(str, sys.version_info[:2])))")
-ENVLIST="py$PYVER-test-deps,py$PYVER-test-mindeps,py$PYVER-test-deps-pre"
+ENVLIST="py$PYVER-test-deps,py$PYVER-test-deps-pre"
+
+if ! ( [[ "$AUDITWHEEL_PLAT" == musllinux_* ]] && [[ "$PYVER" == "39" || "$PYVER" == "310" || "$PYVER" == "311" ]] ); then
+    # skip mindeps on musllinux + python 3.9 to 3.11 because oldest supported numpy versions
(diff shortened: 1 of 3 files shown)


neutrinoceros:
Well we also need dedicated images for this job, so here's my attempt to build some: https://github.com/h5py/hdf5-manylinux/pull/18

neutrinoceros:
The companion PR with docker image builds is ready !
ping @takluyver

neutrinoceros:
Testing reveals another flaw: numpy 1.25 is the oldest version with musllinux wheels, so `mindeps` will not work out of the box for Python 3.9, 3.10 or 3.11 (3.12, 3.13 and 3.14 are fine because they already require more recent versions).

neutrinoceros:
(I'll update this once https://github.com/h5py/h5py/pull/2634 is dealt with)

neutrinoceros:
This should now be stable and ready for review !

ci/cibw_test_command.sh
```diff
@@ -12,7 +12,14 @@ WHEEL_PATH=$2
 echo "WHEEL_PATH=$WHEEL_PATH"
 
 export PYVER=$(python -c "import sys; print(''.join(map(str, sys.version_info[:2])))")
-ENVLIST="py$PYVER-test-deps,py$PYVER-test-mindeps,py$PYVER-test-deps-pre"
+ENVLIST="py$PYVER-test-deps,py$PYVER-test-deps-pre"
+
+if ! ( [[ "$AUDITWHEEL_PLAT" == musllinux_* ]] && [[ "$PYVER" == "39" || "$PYVER" == "310" || "$PYVER" == "311" ]] ); then
```
neutrinoceros, line 17: Note to reviewers: this might become unnecessary when combined with #2631

neutrinoceros: I just confirmed it in #2635

neutrinoceros:
Actually, I don't expect this to be controversial at all, so I'll just self-merge after CI passes again, and I'll clean up the temporary workaround in #2631

PR #2636, opened 2025-08-08
Initial work for HDF5 2.0 support

ajelenak:
Although HDF5 2.0 is still in development, the changes in this PR add the initial support:

* New libver constant `v200`.
* Added new libver constants since HDF5 1.10 to the docs.
* Updated ros3 driver tests and its documentation to reflect the use of the Amazon C S3 library.

PR #2637, opened 2025-08-08
BLD: replace deprecated Cython preproc APIs with `tempita`-based templating

neutrinoceros:
Address most of what's left from #2445 using [`tempita`](https://github.com/TurboGears/tempita/blob/main/README.rst) as suggested in the upstream thread.
Although I'm getting close to a working build locally with surprisingly little effort, I'm not there yet, but I'm opening this already as a proof of concept, hoping that CI might yield additional information regarding what's left to do.


neutrinoceros:
The diff is completely bloated by one mistake I made where I apparently didn't *move* `_errors.pxd` properly and instead erased the existing file and added the modified version back. I don't know how to fix this easily, my previous attempt resulted in a very painful rebase where I essentially had to re-move everything manually.
Any way, this is a detail that can be resolved later when everything is working, which isn't the case now.
I'm going to take a break from this for now.

neutrinoceros:
Spite driven development: I couldn't actually take this off my mind while it was broken. I'll try again now that it's working.

neutrinoceros:
Never mind about the diff, I think it wasn't a mistake after all, but rather, a limitation from git itself. Despite a couple attempts to fix it carefully locally, the end result doesn't change.

neutrinoceros:
This now closes #2445 ! :tada:

Additional notes to reviewers:
- I initially wanted to name template files as, e.g., `*.pyx.templ`, which I think is more common in the wild, but I ended up using `*.templ.pyx` instead to avoid messing up with syntax highlighting in everybody's editors as well as in GH's UI
- I deliberately chose to prepend tempita conditionals with `###`, which isn't part of the language, so the code remained valid Cython and there would be an easily greppable trace of the templating in pre-processed Cython files

pyproject.toml
```diff
@@ -6,6 +6,7 @@ requires = [
     "numpy >=2.0.0, <3",
     "pkgconfig",
     "setuptools >=77",
+    "tempita>=0.6.0"
```
tacaswell: Remember to update ci/triage_build.sh as well.

I think we also need a release note on this (for the down-stream packagers).

tacaswell:
> Never mind about the diff, I think it wasn't a mistake after all, but rather, a limitation from git itself. Despite a couple attempts to fix it carefully locally, the end result doesn't change.

Git has heuristics for identifying renames that you can control locally, but in the case of the view on GH is something they control (because of how git internal work it maps trees of file names to shas and then has a file on disk for each sha (named by the sha) [modulo a bunch of compression / implementation details]) so removing the leading space from ~ every line of a file fails the "just a rename" test.

Setting GH to ignore whitespace makes this easier to review.

pyproject.toml
neutrinoceros: nice catch. Done !

tacaswell:
To check my understanding, we are doing the templating ourselves (via an additional build time dependency) so we don't pick up a (high) floor on cython verions?

If we control the string would sneaking in `h5py` or `template` somehow be good?  Something that makes it obvious to someone who opens the source files cold is given a hint that they are an marker of our invention? `#H5PYT#` or soething?

neutrinoceros:
> To check my understanding, we are doing the templating ourselves (via an additional build time dependency) so we don't pick up a (high) floor on cython verions?

correct.

> If we control the string would sneaking in `h5py` or `template` somehow be good? Something that makes it obvious to someone who opens the source files cold is given a hint that they are an marker of our invention? #H5PYT# or something?

I'm not sure I understand your question. Is this about the file naming convention, or the intentional string remnants that I padded conditionals with (`###`) ?

tacaswell:
> the intentional string remnants that I padded conditionals with (###) ?


Sorry, typed this as I was running out the door and was not clear.  This was referring to the `###`.  I incorrectly thought that it was involved in identifying where to template, but on reading a bit more I think it is just invisible to both cython and the templating engine?  There is not anyway to enforce it going forward, but makes this easy to track?


I think the file names are clear enough, although I'm of two minds of the order.  Leaving `.pyx` etc last helps the editors, but in other cases where this comes up (e.g. ansible templates) the extra extension is last so you can search for the template extension and then write out the target with it removed. 


----

Is tempita the right choice?  The readme suggests it is a ~1 person project, should we go with something like jinja2 ("no one gets fired for buying IBM")?  On the other hand, it looks like cython ships a copy of it (https://github.com/pandas-dev/pandas/blob/7863029c3084c79b226c0a1c3daeac82929534f1/generate_pxi.py#L4 ) ( Can we use that instead of adding a dependency ?) and pandas uses it so if it goes sideways we are in a boat with others.

neutrinoceros:
> I think it is just invisible to both cython and the templating engine?

correct 

> There is not anyway to enforce it going forward, but makes this easy to track?

hardly. Nothing out of the box, at least.

I didn't know Cython.Tempita was a thing, I agree it's better than adding a dependency. Full disclosure: I didn't really stop to think wether tempita was the right choice, I just started experimenting with it and constantly felt like I was 2min away from solving the problem so I didn't realize how much of an investment it'd turn out to be. Now that I have a working branch I'm strongly biased to prefer it over anything still unknown to me, like jinja2, so I don't think I can make that call (at least not now).

tacaswell:
If cython ships a version of tempatia that is a pretty strong argument for using it and suggests that there will be multiple projects using it for this sort of templating.

There are some things where we want to be out on the edge and doing something interesting and some things (like this) where we want to what ever the most boring widely used thing that everyone else uses ;)

neutrinoceros:
I strongly agree. Boring, simple and *stable* for the win.

ajelenak:
Is the templating method in this PR accepted as the right one? No one is objecting to the use of `tempita` compared to something else? I am asking because I'd like to start switching to `tempita` in my PR #2636.

neutrinoceros:
(rebased and force-pushed as a single commit to resolve merge conflicts)

tacaswell:
I am :+1: to say we go with tempatia as the templating engine of choice, but I think that @takluyver and @aragilar should weigh in as well.

aragilar:
I'm happy with it too. I think really all we needed from the template system was conditionals (so jinja2 is way more featurefull than we need), and if cython has adopted it and integrated it in then I think go with it.

takluyver:
I'm fine with using Tempita - we only need very basic templating features, performance is unlikely to be an issue, we know it's already used for similar scenarios, and we can hopefully rely on Cython to maintain it. The one thing I'd like is for it to be mentioned in the Cython docs that we can rely on importing it from Cython - but in practice I don't think it's likely to go away.

I like the pragmatic path of keeping the file extensions and putting the template markers in comment lines so the files remain valid Cython syntax and get highlighted by default.

setup_build.py
```diff
@@ -117,6 +120,20 @@ def _make_extensions(cls, config):
...
+                if not (templ := raw_path.with_suffix(f'.templ{ext}')).exists():
+                    continue
+
+                if (target := raw_path.with_suffix(ext)).exists():
+                    current_text = target.read_text()
+                else:
+                    current_text = ""
+                new_text = tempita.sub(templ.read_text(), **templ_config)
+                if new_text != current_text:
+                    target.write_text(new_text)
```
takluyver:
```suggestion
                    current_text = target.read_text('utf-8')
                else:
                    current_text = ""
                new_text = tempita.sub(templ.read_text(), **templ_config)
                if new_text != current_text:
                    target.write_text(new_text, 'utf-8')
```

The default encoding is unfortunately platform dependent - IMO a missed opportunity when Python 3 came out.

Maybe everything in our Cython files is ASCII anyway, in which case the default will work, but it's easier to do it right now than figure out a build problem on Windows later.

neutrinoceros: Good point. Thank you !

h5py/h5t.templ.pyx
```diff
@@ -44,7 +44,7 @@ _IS_PPC64LE = _UNAME_MACHINE == "ppc64le"
 cdef char* H5PY_PYTHON_OPAQUE_TAG = "PYTHON:OBJECT"
 cdef char* H5PY_NUMPY_STRING_TAG = "NUMPY:STRING"
 
-NUMPY_GE2 = int(np.__version__.split('.')[0]) >= 2
+NUMPY_RUNTIME_VERSION_TUPLE = tuple(int(x) for x in np.__version__.split('.')[:3])
```
takluyver: I think this could fail with pre-release versions of Numpy, if `np.__version__` is like `2.3.0rc1`, which is the canonical format for pre-release versions.

`re.findall(r'\d+', np.__version__)[:3]` might be a slightly more robust way to find the numeric groups, without getting too far into parsing version numbers.

neutrinoceros: nice catch, thanks again

neutrinoceros:
Good to go ?

PR #2471, opened 2024-08-12
DEP: drop support for CPython 3.8

neutrinoceros:
Python 3.8 will be EOL in less than 2 months, and the previous such bump happened more than a year ago (https://github.com/h5py/h5py/pull/2262), so I figured I would go for it now.

PR #2640, opened 2025-08-09
DEP: drop support for Python 3.9, require 3.10 or newer

neutrinoceros:
Nothing urgent, but as I'm currently spending lots of time waiting for wheels to build, I figure it may be time to remove these.

For reference, the previous such PR, #2471, was merged almost a year ago.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -14,7 +14,7 @@ jobs:
   steps:
     - task: UsePythonVersion@0
       inputs:
-        versionSpec: '3.9'
+        versionSpec: '3.10'
         architecture: 'x64'
     - script: |
         pip install tox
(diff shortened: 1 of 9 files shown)


tacaswell (approved):

tacaswell:
Something in not right, azure did not fire on this PR.

neutrinoceros:
Is there a way to retrigger it, other than a power-cycle ?

tacaswell:
> Is there a way to retrigger it, other than a power-cycle ?

for individual jobs there are buttons in the UI, but given we missed a whole CI service, just power-cycled.

neutrinoceros:
> AppVeyor was unable to build non-mergeable pull request

Hum, weird, I see no conflicts here. Let's try a rebase still

tacaswell:
This makes we worried the changes broke the azure configuration.

azure-pipelines.yml
```diff
@@ -129,9 +129,9 @@ jobs:
...
       # -mindeps : test with oldest supported version of (python) dependencies
       # -deps-pre : test pre-release versions of (python) dependencies)
-      py39:
-        python.version: '3.9'
-        TOXENV: py39-test-deps,py39-test-mindeps,py39-test-deps-pre
+      py310:
+        python.version: '3.10'
+        TOXENV: py310-test-deps,py310-test-mindeps,py310-test-deps-pre
         H5PY_ENFORCE_COVERAGE: yes
       py310:
```
takluyver: We've got a duplicate key here, that might be what's breaking Azure.

```suggestion
      py310:
```

azure-pipelines.yml
```diff
@@ -87,18 +87,18 @@ jobs:
     # -deps-pre : test pre-release versions of (python) dependencies)
     #
     # 64 bit - HDF5 1.10
-      py39-hdf5110:
-        python.version: '3.9'
-        TOXENV: py39-test-deps,py39-test-mindeps,py39-test-deps-pre
+      py310-hdf5110:
```
takluyver, line 90: We were roughly taking a diagonal across the matrix of possible dependencies - testing a few different Python versions and HDF5 versions at the same time. Now they're getting all bunched up on the same Python version. We might want to shuffle some along to newer Python versions - or we could decide it's redundant with the wheel builds on Github actions and cut some of them out. :shrug:

neutrinoceros: your call ! I wasn't aware this was the strategy, but I think I'd be comfortable with dropping some of these.

takluyver: I'm comfortable dropping some at this point - I think we've got pretty good coverage by building and testing wheels on GHA. There are a couple of corner cases it would be good to keep, though, like the C locale tests and the MPI tests. The Mac tests here also install HDF5 from Homebrew, whereas on GHA we directly build it from source, so it's probably worth keeping one or two jobs doing that.

takluyver:
And I agree that we can reasonably drop Python 3.9 now. We'll be adding 3.14 before long.

neutrinoceros:
Actually I already added support for 3.14, the only things we're missing are
- stable wheels (next release will have them)
- non-nightly tests, which require 3.14.0 to be available on GHA

neutrinoceros:
Ok I think this one is go now. Thanks all !

takluyver:
Thanks, nice one!

PR #2646, opened 2025-08-14
BLD: add back support for building against numpy 1.x

### LINKED ISSUE #2644: 3.14 enforces numpy >= 2, despite comment in setup.py
The `pyproject.toml` clearly requires `numpy >=2` to build, [however setup.py notes](https://github.com/h5py/h5py/blob/3.14.0/setup.py#L31):
```
 # here, and if you can build an older NumPy on a newer Python, h5py probably
    # NumPy 1.19.3 is the first with wheels for Python 3.9, our minimum Python.
```

This worked pre-3.14, however, [3.14 looks to enforce](https://github.com/h5py/h5py/blob/3.14.0/setup_build.py#L130) the build with `numpy >= 2`
```
assert numpy.__version__ >= '2.0'  # See build dependencies in pyproject.toml
```

I wanted to confirm that builds of >=3.14 really do require `numpy >= 2`

neutrinoceros:
Close #2644
At the time of opening, there's a missing piece: I haven't implemented a warning for `h5py._npystrings` being missing in this configuration yet.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -38,7 +38,7 @@ permissions:
 
 jobs:
   build_wheels:
-    name: 'Wheels: ${{ matrix.os }} ${{ matrix.arch }} ${{ matrix.python }}'
+    name: 'Wheels: ${{ matrix.os }} ${{ matrix.arch }}'
     runs-on: ${{ matrix.os }}
     continue-on-error: true
     strategy:
@@ -90,7 +90,7 @@ jobs:
         if: runner.os == 'macOS'
 
       # Triage build
-      - run: bash ./ci/triage_build.sh "${{ matrix.arch }}" "${{ github.event.pull_request.head.sha || github.sha }}" "${{ matrix.python }}"
+      - run: bash ./ci/triage_build.sh "${{ matrix.arch }}" "${{ github.event.pull_request.head.sha || github.sha }}"
 
       # Now actually build the wheels
       - uses: pypa/cibuildwheel@c923d83ad9c1bc00211c5041d0c3f73294ff88f6 # v3.1.4
@@ -104,9 +104,44 @@ jobs:
           name: cibw-wheels-${{ matrix.os }}-${{ strategy.job-index }}
           path: ./wheelhouse/*.whl
 
+
+  check-minimal-build:
+    # this job isn't meant to upload anything, it's only a sanity check
+    name: Check minimal build requirements (${{ matrix.label }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        include:
+          - os: ubuntu-latest
+            arch: x86_64
+            cibw_before_build: pip install --only-binary ':all:' 'numpy==1.25.0' 'Cython==3.0.0' 'setuptools==77.0.1' 'pkgconfig==1.5.5'
+            label: 'numpy-min'
+          - os: ubuntu-latest
+            arch: x86_64
+            cibw_before_build: pip install --only-binary ':all:' 'numpy==2.0.0' 'Cython==3.0.0' 'setuptools==77.0.1' 'pkgconfig==1.5.5'
+            label: 'cython-min + numpy 2.0'
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          fetch-depth: 0  # get the non-merge commit for PRs
+
+      # Triage build
+      - run: bash ./ci/triage_build.sh "${{ matrix.arch }}" "${{ github.event.pull_request.head.sha || github.sha }}"
+
+      # Now actually build the wheels
+      - uses: pypa/cibuildwheel@352e01339f0a173aa2a3eb57f01492e341e83865 # v3.1.3
+        env:
+          CIBW_ARCHS: ${{ matrix.arch }}
+          CIBW_BUILD: "cp310-*"
+          CIBW_BEFORE_BUILD: ${{ matrix.cibw_before_build }}
+          CIBW_BUILD_FRONTEND: 'pip; args: --no-build-isolation'
+          ONLY_MINDEPS_TESTS: "1"
+
   upload_wheels:
     name: Upload wheels
-    needs: build_wheels
+    needs:
+      - build_wheels
+      - check-minimal-build
     runs-on: ubuntu-latest
     defaults:
       run:
--- a/ci/cibw_test_command.sh
+++ b/ci/cibw_test_command.sh
@@ -12,7 +12,11 @@ WHEEL_PATH=$2
 echo "WHEEL_PATH=$WHEEL_PATH"
 
 export PYVER=$(python -c "import sys; print(''.join(map(str, sys.version_info[:2])))")
-ENVLIST="py$PYVER-test-mindeps,py$PYVER-test-deps,py$PYVER-test-deps-pre"
+
+ENVLIST="py$PYVER-test-mindeps"
+if [[ "$ONLY_MINDEPS_TESTS" != "1" ]] ; then
+    ENVLIST="$ENVLIST,py$PYVER-test-deps,py$PYVER-test-deps-pre"
+fi
 
 export H5PY_TEST_CHECK_FILTERS=1
 echo "ENVLIST=$ENVLIST"
--- a/news/bld-new-mindeps.rst
+++ b/news/bld-new-mindeps.rst
@@ -1,0 +1,9 @@
+Building h5py
+-------------
+
+* The minimum versions build-time Python requirements were updated to
+  ``Cython==3.0.0`` (up from ``0.29.1``), and ``numpy==1.25.0`` (down from
+  ``2.0.0``). We still recommend building with numpy 2 or newer whenever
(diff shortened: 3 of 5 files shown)


neutrinoceros:
I'll need to iron out a couple defects too:
- mindeps is broken on every normal build
- pre-deps cannot possibly work with this build, I think

neutrinoceros:
I think I managed to resolve immediate problems. Now I should only need to add and tests for the warning.
I'll be back to this soon.

crusaderky:
The intersection of cython==0.29.37 and numpy>=2 segfaults on npystrings (see https://github.com/h5py/h5py/pull/2647).
CI passes because it tests cython==0.29.37 and numpy<2.
Cython 3.0 fixes it.

I'd recommend adding a test for minimum version of cython + numpy==2.0.0 in addition to numpy 1.x.

setup_build.py
```diff
@@ -117,22 +117,27 @@ def _make_extensions(cls, config):
...
             settings['runtime_library_dirs'] = settings['library_dirs']
 
-        return [cls._make_extension(m, settings) for m in MODULES]
+        # 'NPY_1_21_API_VERSION' == 0x0000000e
+        settings['define_macros'].append(('NPY_TARGET_VERSION', 'NPY_1_21_API_VERSION'))
+        extensions = [cls._make_extension(m, settings) for m in MODULES]
+
+        if int(numpy.__version__.split('.')[0]) >= 2:
+            # Enable NumPy 2.0 C API for modules that require it.
+            # NUMPY2_MODULES will not be importable when NumPy 1.x is installed.
```
crusaderky, line 126:
```suggestion
            # MODULES_NUMPY2_ONLY will not be importable when NumPy 1.x is installed.
```

setup_build.py
```diff
@@ -117,22 +117,27 @@ def _make_extensions(cls, config):
...
+            # 'NPY_2_0_API_VERSION' == 0x00000012
+            settings['define_macros'].append(('NPY_TARGET_VERSION', 'NPY_2_0_API_VERSION'))
+            extensions.extend(cls._make_extension(m, settings) for m in MODULES_NUMPY2_ONLY)
```
crusaderky:
```suggestion
        extensions = [cls._make_extension(m, settings) for m in MODULES]

        if int(numpy.__version__.split('.')[0]) >= 2:
            # Enable NumPy 2.0 C API for modules that require it.
            # MODULES_NUMPY2_ONLY will not be importable when NumPy 1.x is installed.
            # 'NPY_2_0_API_VERSION' == 0x00000012
            settings['define_macros'].append(('NPY_TARGET_VERSION', 'NPY_2_0_API_VERSION'))
            extensions.extend(cls._make_extension(m, settings) for m in MODULES_NUMPY2_ONLY)
        else:
            # 'NPY_1_21_API_VERSION' == 0x0000000e
            settings['define_macros'].append(('NPY_TARGET_VERSION', 'NPY_1_21_API_VERSION'))
```

neutrinoceros: Good point, better build all extensions with a consistent `NPY_TARGET_VERSION`. Not sure your suggestion applies as is though so I'll write my own version of it. Thanks !

neutrinoceros:
> I'd recommend adding a test for minimum version of cython + numpy==2.0.0 in addition to numpy 1.x.

great suggestion, thanks a lot. I'll let CI run its course without it first since I just did a "risky" rebase, but I'll include this in the next iteration !

setup_build.py
neutrinoceros: Actually this doesn't seem to work; the resulting wheels are not compatible with numpy 1.21, so I reverted this change.

https://github.com/h5py/h5py/actions/runs/16965494246/job/48088254143?pr=2646

neutrinoceros:
So I started working on implementing a warning as suggested in #2644, until I realized that importing h5py, built with numpy 1.x, and with numpy 2.x present at runtime, would already raise an exception:
```python-traceback
  File "/Users/clm/dev/h5py/h5py/__init__.py", line 45, in <module>
    from ._conv import register_converters as _register_converters, \
  File "h5py/_conv.pyx", line 1, in init h5py._conv
    # cython: profile=False
  File "h5py/h5r.pyx", line 1, in init h5py.h5r
    # cython: language_level=3
  File "h5py/h5p.pyx", line 1, in init h5py.h5p
    # cython: language_level=3
ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject
```

with this in mind, I don't think adding a warning is necessary or helpful; this setup is clearly broken, we shouldn't need to tell users that it is.
It's also clear from the traceback where the problem is coming from.

ci/cibw_test_command.sh
```diff
@@ -12,15 +12,20 @@ WHEEL_PATH=$2
...
+if [[ "$ONLY_MINDEPS_TESTS" == "1" ]] ; then
+    ENVLIST="py$PYVER-test-mindeps"
+else
+    ENVLIST="py$PYVER-test-deps,py$PYVER-test-deps-pre"
+    if ! ( [[ "$AUDITWHEEL_PLAT" == musllinux_* ]] && [[ "$PYVER" == "39" || "$PYVER" == "310" || "$PYVER" == "311" ]] ); then
+        # skip mindeps on musllinux + python 3.9 to 3.11 because oldest supported numpy versions
+        # are higher for this target (1.25.0) and there's no way (that I found) to specify it
+        # directly in package metadata.
+        ENVLIST="py$PYVER-test-mindeps,$ENVLIST"
+    fi
```
neutrinoceros: this can be simplified a lot with #2631

neutrinoceros:
@takluyver what do you think ?

neutrinoceros:
rebased to resolve a merge conflict

.github/workflows/build_wheels.yml
```diff
@@ -103,9 +103,55 @@ jobs:
...
+      # Get HDF5 outside of cibuildwheel so that we can set env vars via GITHUB_ENV
+      # inside the HDF5-getting scripts and have them be in effect for the
+      # cibuildwheel step.
+
+      # Cache HDF5
+      - name: Cache built HDF5
+        uses: actions/cache@0400d5f644dc74513175e3cd8d07132dd4860809 # v4.2.4
+        with:
+          key: ${{ matrix.os }}-${{ matrix.arch }}-6
+          path: cache/hdf5
```
takluyver: I think this cache step is doing nothing here - in the main `build_wheels` job it caches HDF5 built on Windows & Mac, but this job only runs on Linux, and it gets HDF5 from the container image we point cibuildwheel at.

takluyver:
> It's also clear from the traceback where the problem is coming from.

I don't think the `numpy.dtype size changed` error message is clear at all, except that people may have seen it before and remember what sort of thing caused it. But I don't expect this scenario to come up often, since we'll continue building packages with Numpy 2.x, so I agree the error is good enough and we don't need to add a warning.

.github/workflows/build_wheels.yml
neutrinoceros: oh, nice catch ! it looks like I just copy-pasted this bit from above, where it's not conditional on the os (but could be). I'll delete it here and open a separate PR to skip this step where we don't need it !

neutrinoceros:
> I don't think the numpy.dtype size changed error message is clear at all

you're right, I wasn't realising I made an assumption in my interpretation of it.

.github/workflows/build_wheels.yml
neutrinoceros: (cleaned up here)

neutrinoceros: and other PR at https://github.com/h5py/h5py/pull/2695

news/bld-new-mindeps.rst
```diff
@@ -0,0 +1,6 @@
+Building h5py
+-------------
+
+* The minimum versions build-time Python requirements were updated to
+  ``Cython==3.0.0`` (up from ``0.29.1``), and ``numpy==1.25.0`` (down from
+  ``2.0.0``).
```
takluyver: Let's mention here that we still recommend building with Numpy 2.x in general, this is just to allow ecosystems where you build with the same version you want to use to still use Numpy 1.x.

takluyver:
Other than that, I think this looks good.

news/bld-new-mindeps.rst
neutrinoceros: good idea. How's my attempt at it ?

news/bld-new-mindeps.rst
```diff
@@ -0,0 +1,9 @@
+Building h5py
+-------------
+
+* The minimum versions build-time Python requirements were updated to
+  ``Cython==3.0.0`` (up from ``0.29.1``), and ``numpy==1.25.0`` (down from
+  ``2.0.0``). We still recommend building with numpy 2 or newer whenever
+  possible, this is done to improve support for external package ecosystems,
+  not controlled by h5py's maintainers, where numpy 2 might not be available
+  yet.
```
takluyver: I think it's not so much about numpy 2 not being *available*, more:

```suggestion
  where the system uses the same version of packages for building and
  installation.
```

(You might want/need to install with Numpy 1.x, even if the ecosystem has 2.x available, because of some other dependency. The big conda environments we set up at work still get numpy 1.x, for instance, although with conda that doesn't mean you need to build with 1.x.)

PR #2647, opened 2025-08-14
NpyStrings segfault on Cython <3

crusaderky:
`test_npystrings.py` segfaults on

python=3.10
numpy=2.0.0
cython=0.29.31 ~ 0.29.37

Everything works correctly with Cython=3.0.0.
I did not spend time investigating the issue. cython 3.0 is over 2 years old now so it should be a non-issue to upgrade the minimum dependency?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
 
 # REMEMBER TO KEEP THE VERSIONS IN THIS FILE THE SAME AS IN ci/triage_build.sh
 requires = [
-    "Cython >=0.29.31,<4",
+    "Cython >=3.0.0,<4",
     "numpy >=2.0.0, <3",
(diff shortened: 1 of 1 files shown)


neutrinoceros (commented):
This seems fine to me, but since we're in the process of improving support for existing distributors, I'd like to wait for them to greenlight this upgrade first.
See #2644

neutrinoceros:
All good, thank you !

neutrinoceros (approved):

neutrinoceros:
I'm just thinking about the changelog now... @takluyver do you want this mentionned ? I could easily fix it up in #2646 too

tacaswell:
We should add a change note for this.

neutrinoceros:
done in #2646

PR #2648, opened 2025-08-14
MAINT: Remove redundant `#cython: language_level=3`

crusaderky:
This flag is already defined centrally in setup_build.py:
https://github.com/h5py/h5py/blob/ad92ec229f19646b32428d793c064054aff48a5e/setup_build.py#L189


tacaswell:
Thank you!

PR #2650, opened 2025-08-14
BLD: add freethreading_compatible Cython markers (v2)

crusaderky:
This is a rework of #2478 (with thanks to @neutrinoceros) that should allow merging the PR for now while we work on unit testing (#2588 and more PRs after that).

- Set up CI infrastructure to produce 3.13t wheels; however disable it for now by commenting out a line
- When h5py is compiled with python 3.13t, mark it as noGIL-capable. This will make the Python interpreter release the GIL after import.
- Add documentation stating that multi-threading and even more so noGIL support is highly experimental and untested
- Add Python 3.13 and 3.13t (single-threaded) to CI. Multi-threading tests are in scope for later PR.


crusaderky:
@takluyver @neutrinoceros @rgommers this is ready to get your feedback

tacaswell:
If cibildwheel is building 13t, why do we want to turn in off before merging?

crusaderky:
> If cibildwheel is building 13t, why do we want to turn in off before merging?

@takluyver in #2478 expressed his opposition to shipping a wheel until we have some degree of confidence (read: tests) that it is mostly fit for purpose.

djhoese:
I say this as a downstream library maintainer and as a user. Obviously h5py maintainers, you make your own decisions, but:

The classifier already says it is Unstable. The whole feature is experimental in 3.13. It is opt-in for 3.14. We know. The worst thing that should reasonably happen is that you get told about thread-based bugs sooner. Sure repeat issues about the same thing will be annoying, but let's find out sooner rather than later. If someone is running free-threaded python operationally then they should know the risks (jeez I really hope so). Obviously if the existing tests fail then that's not good, but this may be super difficult to write tests for.

rgommers:
The approach here seems good to me, and the added docs are helpful.

Re wheels or no wheels: downstream packages will know it's experimental yes, but if the state is still what @neutrinoceros said [here](https://github.com/h5py/h5py/pull/2478#issuecomment-2935121260) ("seg faults all over the place") then there isn't much point in letting downstream users run into those segfaults. 

Not sure y'all discussed a plan elsewhere @crusaderky and @neutrinoceros, but something like seems pragmatic:
1. Merge this PR
2. Get `pytest-run-parallel` to run without basic issues (xref gh-2588)
3. Add more locking and/or fixes so that doing basic things like "read separate HDF5 files from multiple threads in parallel" works as expected.
4. Set up a regular CI job with `cp314t` if not yet present (no parallel testing, just the regular test suite)
5. Release `cp314t` wheels together with `cp314` in the next release (which I assume will come soon since it's 3.14 RC time); amend the docs added in this PR to be up-to-date while doing so.
6. In a tracking issue, add an overview of remaining harder thread-safety issues and tasks (e.g., run everything under TSan, see what needs fixing) and then work on that.

All of that can be done relatively fast (6 should be the actually time-consuming stuff I'd think) and it will be helpful for the downstream packages that are now starting to be blocked on not having any `h5py` support at all.

pyproject.toml
```diff
@@ -119,10 +119,17 @@ wheels = [
...
+# https://github.com/h5py/h5py/pull/2478
+# https://github.com/h5py/h5py/pull/2588
 [tool.cibuildwheel]
 skip = [
     "*musllinux*",
-    "cp314t-*",
+    # "cp314t-*",  # DNM uncomment before merging
+]
+enable = [
+    "cpython-freethreading",  # build cp313t wheels  # DNM comment out before merging
```
crusaderky: DNM

crusaderky:
Added cp314t wheels (also to be disabled before merging). Now both cp313t and cp314t wheel creation is controlled from a single location in pyproject.toml.

docs/threads.rst
```diff
@@ -0,0 +1,36 @@
+Multi-threading
+===============
+libhdf5, at least with the default compilation flags, is not thread-safe.
+
+h5py supports multi-threading, although its actual robustness is, at the moment
+of writing, largely untested. Multi-threading support is enabled by wrapping
```
neutrinoceros:
```suggestion
h5py supports multi-threading, although its actual robustness is, as of v3.15.0,
largely untested. Multi-threading support is enabled by wrapping
```

pyproject.toml
neutrinoceros: ?

ci/triage_build.sh
```diff
@@ -20,8 +20,8 @@ KIND="$RUNNER_OS $ARCH"
 
 # If it's a scheduled build or [pip-pre] in commit message, use pip-pre
 if [[ "$GITHUB_EVENT_NAME" == "schedule" ]] || [[ "$MSG" = *'[pip-pre]'* ]]; then
-    echo "Using NumPy pip-pre wheel and, setting CIBW_BEFORE_BUILD, CIBW_BUILD_FRONTEND and CIBW_ENABLE"
-    echo "CIBW_BEFORE_BUILD=pip install --pre --only-binary numpy --extra-index-url https://pypi.anaconda.org/scientific-python-nightly-wheels/simple \"numpy>=2.0.0.dev0\" \"Cython>=0.29.31,<4\" pkgconfig \"setuptools>=77\" wheel" | tee -a $GITHUB_ENV
+    echo "Using NumPy pip-pre wheel; setting CIBW_BEFORE_BUILD and CIBW_BUILD_FRONTEND"
```
neutrinoceros, line 23: we need to keep this in sync with the `[build-system]` table in `pyproject.toml`. Could you also set an explicit lower bound for `packaging` while you're at it ? (using a recent version is fine)

neutrinoceros (commented):
So, do I understand correctly that with this, we'd be publishing `cp314t` nightlies (and *only* nightlies), and still require the GIL to be explicitly disabled by downstream testers ? I think this is exactly what we want, but just checking.

We should probably also include the following PyPI classifier
`Programming Language :: Python :: Free Threading :: 1 - Unstable`
(assuming my understand that it does *not* imply we already ship wheels on PyPI is correct)

neutrinoceros:
Thanks for reviving this !

> but if the state is still what @neutrinoceros said https://github.com/h5py/h5py/pull/2478#issuecomment-2935121260 ("seg faults all over the place") then there isn't much point in letting downstream users run into those segfaults.

for the record, nothing changed on this front yet.

> Not sure y'all discussed a plan elsewhere @crusaderky and @neutrinoceros, but something like seems pragmatic:

we didn't, but that's a great plan to get this rolling again, thanks Ralph !

@crusaderky could you squash your commits ? (possibly mine too, although I'd prefer if we keep our respective contributions separated, just so it's easier for me to see exactly what you added and changed) 

(I meant to post this earlier but apparently didn't)

pyproject.toml
crusaderky: This PR is currently building cp313t and cp314t wheels, both nightlies and final releases. You need to change those 2 lines before merging to disable them.
@takluyver in https://github.com/h5py/h5py/pull/2478 expressed his opposition to shipping wheels to final users for the time being.

crusaderky:
> So, do I understand correctly that with this, we'd be publishing `cp314t` nightlies (and _only_ nightlies)

The current PR does not build cp313t or cp314t wheels, neither nightlies nor published pypi versions (once you comment out/uncomment the DNM lines).
IMHO it should not be a big problem for advanced users who want to experiment with free-threading to build from sources.

> and still require the GIL to be explicitly disabled by downstream testers ?

No, the GIL will be already disabled if you build from sources (no wheels published on pypi) on 3.13t or 3.14t.

> We should probably also include the following PyPI classifier `Programming Language :: Python :: Free Threading :: 1 - Unstable`

~~I've already added it.~~ [EDIT] Bizarre. It got lost at some point. Added again.

> (assuming my understand that it does _not_ imply we already ship wheels on PyPI is correct)

Correct

crusaderky:
> @crusaderky could you squash your commits ? (possibly mine too, although I'd prefer if we keep our respective contributions separated, just so it's easier for me to see exactly what you added and changed)

I've partially squashed it to make your review easier. Your changes remain in separate commits.

neutrinoceros:
> No, the GIL will be already disabled if you build from sources (no wheels published on pypi) on 3.13t or 3.14t.

I was refering to how free-threading builds will still re-enable the GIL at runtime as soon as one of the following condition is met:
1) python was started with `-Xgil=1`
2)  else, if `PYTHON_GIL=1` is defined
3) else, when importing any extension module that is not explicitly declared as free-threading compatible

specifically, @tacaswell previously raised concerns that we shouldn't allow testing without explicit GIL disabling (via either 1) or 2)), so we should *not* mark our modules as "compatible" just yet.
I missed it earlier but I see now that you've added thus marker. I now agree with Thomas that this shoudn't be included in this PR, but reserved for when we have decent test coverage ourselves. Correspondingly, the title of the PR should be changed.

setup_build.py
```diff
@@ -182,10 +184,15 @@ def run(self):
 """
         write_if_changed(config_file, s)
 
+        compiler_directives = {}
+        if Version(cython_version) >= Version("3.1.0b1"):
+            compiler_directives["freethreading_compatible"] = True
```
neutrinoceros: let's drop this, please

crusaderky:
@neutrinoceros I've done as you asked. This is ready to be merged if there are no further objections

neutrinoceros:
Thank you ! I can review this again tomorrow (or later). In the mean time, could you rebase and resolve merge conflicts please ? Sorry for the inconvenience !

docs/threads.rst
```diff
@@ -0,0 +1,36 @@
+Multi-threading
+===============
+libhdf5, at least with the default compilation flags, is not thread-safe.
+
+h5py supports multi-threading, although its actual robustness is, as of v3.15.0,
+largely untested. Multi-threading support is enabled by wrapping
+all critical sections that invoke the libhdf5 C API with an internal
+*interpreter-wide reentrant lock*, to be used either as a context manager or
```
takluyver: I think this introduction is technically correct but somewhat confusing. I think the key points to convey up front are 1) you can use h5py from multiple threads without taking any special precautions, but 2) only one thread can do HDF5 operations at a time, so it's not that useful.

I think/hope that the `phil` lock is pretty robust, at least before free-threading Python is involved. One big lock is a pretty simple design, and it's been in h5py for a long time for people to find and fix issues. So I'd say calling it 'largely untested' is overly pessimistic.

crusaderky: I've reworked the wording; see if it's OK now

azure-pipelines.yml
```diff
@@ -36,38 +36,66 @@ jobs:
...
+        python.version: '3.13'
+        python.architecture: 'x64'
+        TOXENV: py313-test-deps
+        HDF5_VERSION: 1.14.6
+        HDF5_DIR: $(HDF5_CACHE_DIR)/$(HDF5_VERSION)
+        H5PY_ENFORCE_COVERAGE: yes
+    # Free-threading (noGIL) Python
+      py313t-deps-hdf51146:
+        python.version: '3.13'
+        python.architecture: 'x64-freethreaded'
```
neutrinoceros, line 87: I don't see this value mentionned on https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/use-python-version-v0, could you point me to where you found this setting ?

azure-pipelines.yml
```diff
@@ -36,38 +36,66 @@ jobs:
...
+
+        # FIXME it was decided to not mark the module as freethreading-compatible
+        # for the time being due to instability concerns.
+        # See matching FIXMEs in pyproject.toml and setup_build.py.
+        PYTHON_GIL: 0
+
+        TOXENV: py313t-test-deps
```
neutrinoceros: I'd suggest jumping straight to 3.14t. I think all our dependencies support it already and it and given the timing, I don't see us retro-actively adding wheels for 3.13t when anyone interested will already have access to the significantly superior 3.14t
```suggestion
        TOXENV: py314t-test-deps
```

ci/cibw_test_command.sh
```diff
@@ -11,7 +11,7 @@ echo "PROJECT_PATH=$PROJECT_PATH"
 WHEEL_PATH=$2
 echo "WHEEL_PATH=$WHEEL_PATH"
 
-export PYVER=$(python -c "import sys; print(''.join(map(str, sys.version_info[:2])))")
+export PYVER=$(python -c "import sys; print(''.join(map(str, sys.version_info[:2])) + ('t' if (sys.version_info>=(3,13) and not sys._is_gil_enabled()) else ''))")
```
neutrinoceros: Might as well future proof this now that I still understand what this line does
```suggestion
export PYVER=$(python -c "import sys; print(''.join(map(str, sys.version_info[:2])) + ('t' if (sys.version_info>=(3,13) and sys.implementation.name=='cpython' and not sys._is_gil_enabled()) else ''))")
```

setup_build.py
```diff
@@ -16,6 +16,8 @@
 import os.path as op
 from pathlib import Path
 
+from packaging.version import Version
```
neutrinoceros: this looks like a dangling dependency to me.

pyproject.toml
```diff
@@ -4,6 +4,7 @@
 requires = [
     "Cython >=3.0.0,<4",
     "numpy >=2.0.0, <3",
+    "packaging >=23.0",
```
neutrinoceros: is it actually used ?

pyproject.toml
```diff
@@ -119,10 +121,15 @@ wheels = [
     "tox>=4.26.0",
 ]
 
+# FIXME Temporarily disable cp313t/cp314t wheels due to instability concerns.
+ # See matching FIXMEs in azure-pipelines.yml and setup_build.py.
```
neutrinoceros:
```suggestion
# See matching FIXMEs in azure-pipelines.yml and setup_build.py.
```

pyproject.toml
```diff
@@ -119,10 +121,15 @@ wheels = [
...
 [tool.cibuildwheel]
 skip = [
-    "cp314t-*",
     "cp310-win_arm64", #Skipped due to unavailability of Python binary and NumPy for Win-ARM64
+    "cp314t-*",  # Comment out to build cp314t wheels
+]
+enable = [
+    # "cpython-freethreading",  # Uncomment to build cp313t wheels
```
neutrinoceros: I'd rather not add commented code for now. I'm well aware of this option any way, I'm confident we won't forget about it :)

docs/threads.rst
```diff
@@ -0,0 +1,35 @@
...
+libhdf5, at least with the default compilation flags, is not thread-safe.
+
+h5py supports multi-threading, in the sense that you can use h5py from multiple threads
+without taking any special precautions. This is enabled by wrapping all critical
+sections that invoke the libhdf5 C API with an internal *interpreter-wide reentrant
+lock*, to be used either as a context manager or as a function decorator:
+
+.. code-block:: python
+
+    from h5py._objects import phil, with_phil
```
neutrinoceros, line 12: maybe more of a question for more senior maintainers but since it comes up a lot in discussions lately: why is this lock named Phil ? 😅

docs/threads.rst
```diff
@@ -0,0 +1,35 @@
...
+
+The global lock means that, while it is possible to pipeline h5py and non-h5py
+work across multiple threads (e.g. read/write to a file and process the data),
+multiple calls to the h5py API will not run in parallel - not even if they operate
+on different datasets or different files.
+
+
+free-threading
+--------------
+h5py can be compiled for free-threading Python interpreters (e.g. 3.13t), where the
```
neutrinoceros: just an example, but I'd prefer advertising the newer version
```suggestion
h5py can be compiled for free-threading Python interpreters (e.g. 3.14t), where the
```

docs/threads.rst
```diff
@@ -0,0 +1,35 @@
...
+GIL has been disabled. This does not disable the ``phil`` global lock, though, as
+such lock protects against race conditions in libhdf5 and not in the Python interpreter.
+
+Free-threading stability is, at the moment, largely untested. For this reason,
```
neutrinoceros: Bit-rot protection included
```suggestion
Free-threading stability is, as of h5py 3.15.0, largely untested. For this reason,
```

docs/threads.rst
```diff
@@ -0,0 +1,35 @@
...
+free-threading wheels (e.g. cp313t) are not published on PyPI and the functionality is
```
neutrinoceros:
```suggestion
free-threading wheels (e.g. ``cp314t``) are not published on PyPI and the functionality is
```

neutrinoceros (commented):
Sorry for my delayed response. Only minor remarks left !

azure-pipelines.yml
crusaderky: It's quite well hidden. Took me a while to find it too.
The page you linked -> GitHub actions registry -> [versions-manifest.json](https://github.com/actions/python-versions/blob/main/versions-manifest.json)

azure-pipelines.yml
crusaderky: I tried that already. Doesn't work on Azure CI yet.

pyproject.toml
crusaderky: It is used by the commented out code.
I've removed the commented out code for now and I'll add a DNM PR once this is merged so that we don't forget where to re-add stuff.

crusaderky:
@neutrinoceros I've addressed all your concerns

azure-pipelines.yml
neutrinoceros: fair enough. Thanks !

azure-pipelines.yml
neutrinoceros: oh, cool. I was aware of this version manifest, just not that Azure used it at all ! so I actually don't need to remember *much* more than I already knew, that's a relief !

neutrinoceros:
I'm sorry, another merge conflict slipped in ...

docs/threads.rst
tacaswell: It was named that when I took over from @andrewcollette .

docs/threads.rst
crusaderky: @takluyver gentle ping. This PR is blocked by you on this point.

crusaderky:
@neutrinoceros fixed and retested.

docs/threads.rst
crusaderky: @takluyver gentle ping

neutrinoceros:
I've taken the liberty to create a dedicated label, given how much attention this is getting these days. Thanks for championing this @crusaderky !

crusaderky:
@takluyver gentle ping

neutrinoceros:
(I fixed a couple merge conflicts I just created)

neutrinoceros:
(and another one)

crusaderky:
@neutrinoceros we've been waiting for @takluyver to confirm he likes the reviewed documentation paragraph, but he has not interacted for over a month. Can we just merge this PR?

takluyver:
Sorry @crusaderky, I get a ton of notifications from Github, and real life has been fairly distracting recently, so I haven't really been noticing this thread. I'll take a look now.

takluyver:
I notice Azure has a warning saying:

> You should provide GitHub token if you want to download a python release. Otherwise you may hit the GitHub anonymous download limit.

It looks like the regular non-freethreading builds are already available on Azure, but the freethreading one has to be downloaded during the build. Could I ask you to figure out what we need to do for this.

Otherwise, I think this looks fine now, thanks (both for the work and the patience!).

neutrinoceros:
failures are likely due to an on-going GH incident: https://www.githubstatus.com/incidents/zlzvt6h3ck58

crusaderky:
> I notice Azure has a warning saying:
> 
> > You should provide GitHub token if you want to download a python release. Otherwise you may hit the GitHub anonymous download limit.
> 
> It looks like the regular non-freethreading builds are already available on Azure, but the freethreading one has to be downloaded during the build. Could I ask you to figure out what we need to do for this.
> 
> Otherwise, I think this looks fine now, thanks (both for the work and the patience!).

I've managed to run Azure pipelines on 3.14 and 3.14t. Feel free to revert the commit if you prefer 3.13/3.13t.

The warning above is fixable by enabling PAT authentication:
- https://learn.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#personal-access-token-pat-authentication
- https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens

After you've set up the token in both the github and the azure control panels, you can uncomment the line I added in the latest commit
```
    # githubToken: '$(githubToken)'
```
I _think_ the above means that the name of the service connection in the Azure portal must be called `githubToken`. But I may be wrong and there may be a missing prefix in the variable name above (I found both documentation and logging to be awfully light).

neutrinoceros:
Thank you ! I think we can merge now (I'm okay with testing only 3.14t, as opposed to 3.13t). @takluyver, I assume you'll want to setup GitHub and Azure as described, but that can more easily be addressed in a follow up PR. Thanks again to @crusaderky for stepping up on this and for your patience with the review process !

takluyver:
Thanks! Yes, the Azure pipelines stuff is pretty confusing. I've got the token working in #2701.

Issue #2253, opened 2023-04-24
h5py.ipy_completer compatibility with ipython 8.12


anntzer:
```
Summary of the h5py configuration
---------------------------------

h5py    3.8.0
HDF5    1.12.2
Python  3.11.3 | packaged by conda-forge | (main, Apr  6 2023, 08:58:31) [Clang 14.0.6 ]
sys.platform    darwin
sys.maxsize     9223372036854775807
numpy   1.24.2
cython (built with) 0.29.33
numpy (built against) 1.23.2
HDF5 (built against) 1.12.2
```

h5py.ipy_completer crashes ipython 8.12 when trying to tab-complete a dict key:
```
$ ipython --profile=null
Python 3.11.3 | packaged by conda-forge | (main, Apr  6 2023, 08:58:31) [Clang 14.0.6 ]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.12.0 -- An enhanced Interactive Python. Type '?' for help.

IPython profile: null

In [1]: %load_ext h5py.ipy_completer

In [2]: d = {}

Traceback (most recent call last):
  File "/Users/antony/.local/lib/python3.11/site-packages/IPython/core/completer.py", line 3171, in _complete
    result = matcher(context)
             ^^^^^^^^^^^^^^^^
  File "/Users/antony/.local/lib/python3.11/site-packages/IPython/core/completer.py", line 2707, in custom_completer_matcher
    matches = self.dispatch_custom_completer(context.token) or []
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/antony/.local/lib/python3.11/site-packages/IPython/core/completer.py", line 2747, in dispatch_custom_completer
    res = c(event)
          ^^^^^^^^
  File "/Users/antony/.local/lib/python3.11/site-packages/h5py/ipy_completer.py", line 131, in h5py_completer
    if not isinstance(self._ofind(base).get('obj'), (AttributeManager, HLObject)):
                      ^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'OInfo' object has no attribute 'get'

If you suspect this is an IPython 8.12.0 bug, please report it at:
    https://github.com/ipython/ipython/issues
or send an email to the mailing list at ipython-dev@python.org

You can print a more detailed traceback right now with "%tb", or use "%debug"
to interactively debug it.

Extra-detailed tracebacks for bug-reporting purposes can be enabled via:
[...]

PR #2652, opened 2025-08-15
ENH: automatically load ipython completion on import, deprecate `h5py.enable_ipython_completer`

neutrinoceros:
A surprisingly long-lived yet low-hanging fruit.
I wrote this patch on a very tight schedule. I'll need to actually check that it works before undrafting, but I'm opening the PR now so I don't forget about this branch.

### DIFF
--- a/h5py/__init__.py
+++ b/h5py/__init__.py
@@ -92,24 +92,31 @@ def run_tests(args=''):
     return run_tests(args)
 
 
+
+try:
+    # this name is only defined if running within ipython/jupyter
+    __IPYTHON__
+except NameError:
+    pass
+else:
+    from . import ipy_completer
+    ipy_completer.load_ipython_extension()
+    del ipy_completer
+
+
 def enable_ipython_completer():
     """ Call this from an interactive IPython session to enable tab-completion
     of group and attribute names.
-    """
-    import sys
-    if 'IPython' in sys.modules:
-        ip_running = False
-        try:
-            from IPython.core.interactiveshell import InteractiveShell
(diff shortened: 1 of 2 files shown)


neutrinoceros:
Actually, testing it made me think that the entire module is likely unneeded any more: it looks like autocompletion built into ipython now supports every feature that `h5py.ipy_completer`'s docstring claims it adds.

What's puzzling though is that the module was still modified a mere 2 years ago, when I expected it to be irrelevant for much longer, see https://github.com/h5py/h5py/issues/2253

Am I missing anything, @tacaswell ?

tacaswell:
I'm against auto-installing on import (import time global side effects always seem nice but are ultimately a bad idea in my experience), but if IPython now provides all the useful tab-complete out of the box I am :+1: on deprecating this helper.

takluyver:
Can IPython's built-in completion mechanism go into subgroups, as in `f['foo/bar/b<TAB>`?

I think when I last looked into it, the completions in IPython worked by retrieving a complete list of possible keys, so you were either limited to a single level (`f['fo<TAB>`), or something has to walk all the groups in the file to get all the possible paths, which is slow and probably not a good user experience.  IIRC, this was why the custom IPython completer logic was kept. But I also suspect it's not used very often and maybe not very good, so I might hesitate to turn it on automatically.

neutrinoceros:
> Can IPython's built-in completion mechanism go into subgroups ?

I don't think it can. That actually sounds like a show stopper for this PR. Thanks for your input !

Issue #2578, opened 2025-05-08
Provide wheel for Windows ARM64


khmyznikov:
### What is the current behavior?

Unable to install the package from pip.

### What is the desired behavior?

To have [native wheel for WoA](https://blogs.windows.com/windowsdeveloper/2025/04/14/github-actions-now-supports-windows-on-arm-runners-for-all-public-repos/). GitHub Actions now supports win-arm64 for free.

### How would this improve `h5py`?

Due to the library's popularity, a native version for the growing number of Windows on ARM (WoA) devices offers a better user experience.

<details>
<summary>Install log</summary>

```pwsh
X:\GitHub\Windows-DevRel\Tools\ML-tests> pip install h5py
Collecting h5py
  Using cached h5py-3.13.0.tar.gz (414 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Collecting numpy>=1.19.3 (from h5py)
  Using cached numpy-2.2.5-cp312-cp312-win_arm64.whl
Building wheels for collected packages: h5py
  Building wheel for h5py (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Building wheel for h5py (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [115 lines of output]
      C:\temp\pip-build-env-5nea3cwn\overlay\Lib\site-packages\setuptools\config\_apply_pyprojecttoml.py:82: SetuptoolsDeprecationWarning: `project.license` as a TOML table is deprecated
      !!
      
              ********************************************************************************
              Please use a simple string containing a SPDX expression for `project.license`. You can also use `project.license-files`. (Both options available on setuptools>=77.0.0).

              By 2026-Feb-18, you need to update your project and remove deprecated calls
              or your builds will no longer be supported.

              See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
              ********************************************************************************

      !!
        corresp(dist, value, root_dir)
      C:\temp\pip-build-env-5nea3cwn\overlay\Lib\site-packages\setuptools\config\_apply_pyprojecttoml.py:61: SetuptoolsDeprecationWarning: License classifiers are deprecated.
      !!

              ********************************************************************************
              Please consider removing the following classifiers in favor of a SPDX license expression:

              License :: OSI Approved :: BSD License

              See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
[...]

PR #2653, opened 2025-08-16
WHL: Enable support for building windows arm64 wheels

lowkeyrossi:
PR Description:

* The adoption of Windows on ARM (WoA) devices is steadily increasing, yet many Python wheels are still not available for this platform.
* GitHub Actions now offer native CI runners for Windows on ARM devices (windows-11-arm), enabling automated builds and testing.
* Currently, official h5py Python wheels are not provided for Windows ARM64 and thus users/developers were facing difficulties using popular h5py library natively.
* This PR introduces support for building h5py wheels on Windows ARM64, improving accessibility for developers and end users on this emerging platform.

Closes: https://github.com/h5py/h5py/issues/2578

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -50,6 +50,8 @@ jobs:
             arch: aarch64
           - os: windows-latest
             arch: AMD64
+          - os: windows-11-arm
+            arch: ARM64
           - os: macos-14
             arch: arm64
           - os: macos-13
@@ -78,6 +80,8 @@ jobs:
           python-version: 3.12
         if: runner.os == 'Windows'
       - run: bash ./ci/cibw_before_all_windows.sh "${{ github.workspace }}"
+        env:
+          ARCH: ${{ runner.arch }}
         if: runner.os == 'Windows'
 
       # macOS HDF5
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -81,7 +81,7 @@ jobs:
         if: runner.os == 'Windows'
       - run: bash ./ci/cibw_before_all_windows.sh "${{ github.workspace }}"
         env:
-          ARCH: ${{ runner.arch }}
+          ARCH: ${{ matrix.arch }}
         if: runner.os == 'Windows'
 
       # macOS HDF5
--- a/ci/bundle_hdf5_whl.py
+++ b/ci/bundle_hdf5_whl.py
@@ -28,7 +28,7 @@ def find_dlls():
         elif arch in ("amd64", "x86_64"):
             yield os.path.join(zlib_root, 'bin_release', 'zlib.dll')
         else:
-            raise RuntimeError(f"Unexpected architecture detected: {arch}")
+            raise RuntimeError(f"Unexpected architecture detected: {platform.machine()=}")
 
 def file_sha256(path):
     h = hashlib.sha256()
--- a/ci/bundle_hdf5_whl.py
+++ b/ci/bundle_hdf5_whl.py
@@ -11,6 +11,7 @@ import os
 import os.path as osp
 import shutil
 import sys
+import platform
 import tempfile
 from zipfile import ZipFile, ZIP_DEFLATED
 
@@ -21,7 +22,11 @@ def find_dlls():
     zlib_root = os.environ.get("ZLIB_ROOT")
     if zlib_root:
         print("ZLIB_ROOT", zlib_root)
-        yield os.path.join(zlib_root, 'bin_release', 'zlib.dll')
+        arch = platform.machine().lower()
+        if arch in ("arm64", "aarch64"):
+            yield os.path.join(zlib_root, 'bin', 'zlib1.dll')
+        elif arch in ("amd64", "x86_64"):
+            yield os.path.join(zlib_root, 'bin_release', 'zlib.dll')
 
 def file_sha256(path):
     h = hashlib.sha256()
--- a/ci/bundle_hdf5_whl.py
+++ b/ci/bundle_hdf5_whl.py
@@ -27,6 +27,8 @@ def find_dlls():
             yield os.path.join(zlib_root, 'bin', 'zlib1.dll')
         elif arch in ("amd64", "x86_64"):
             yield os.path.join(zlib_root, 'bin_release', 'zlib.dll')
+        else:
+            raise RuntimeError(f"Unexpected architecture detected: {arch}")
 
 def file_sha256(path):
     h = hashlib.sha256()
--- a/ci/cibw_before_all_windows.sh
+++ b/ci/cibw_before_all_windows.sh
@@ -31,7 +31,7 @@ elif [[ "$ARCH" == "x86_64" ]]; then
     export HDF5_VSVERSION="17-64"
 else
     echo "Got unexpected arch $ARCH"
-    exit(1)
+    exit 1
 fi
 
 export PATH="$PATH:$EXTRA_PATH"
--- a/ci/cibw_before_all_windows.sh
+++ b/ci/cibw_before_all_windows.sh
@@ -29,6 +29,9 @@ else
     export CL="/I$ZLIB_ROOT/include"
     export LINK="/LIBPATH:$ZLIB_ROOT/lib_release"
     export HDF5_VSVERSION="17-64"
+else
+    echo "Got unexpected arch $ARCH"
+    exit(1)
 fi
 
 export PATH="$PATH:$EXTRA_PATH"
--- a/ci/cibw_before_all_windows.sh
+++ b/ci/cibw_before_all_windows.sh
@@ -12,7 +12,8 @@ if [[ "$ARCH" == "ARM64" ]]; then
     #Use vcpkg for Windows ARM64, since Nuget\Chocolatey doesn't provide zlib package for Windows ARM64
     VCPKG_ROOT="$PROJECT_PATH/vcpkg"
     VCPKG_SHA="dd3097e305afa53f7b4312371f62058d2e665320"  # 2025.07.25
-    git clone https://github.com/microsoft/vcpkg@"$VCPKG_SHA" "$VCPKG_ROOT"
+    git clone https://github.com/microsoft/vcpkg.git "$VCPKG_ROOT"
+    (cd "$VCPKG_ROOT" && git checkout "$VCPKG_SHA")
     $VCPKG_ROOT/bootstrap-vcpkg.bat -disableMetrics
     VCPKG_TRIPLET="arm64-windows"
     $VCPKG_ROOT/vcpkg.exe install zlib:$VCPKG_TRIPLET
--- a/ci/cibw_before_all_windows.sh
+++ b/ci/cibw_before_all_windows.sh
@@ -9,7 +9,7 @@ fi
 PROJECT_PATH="$1"
 
 if [[ "$ARCH" == "ARM64" ]]; then
-    # vcpkg for Windows ARM64
+    #Use vcpkg for Windows ARM64, since Nuget\Chocolatey doesn't provide zlib package for Windows ARM64
     VCPKG_ROOT="$PROJECT_PATH/vcpkg"
     VCPKG_SHA="dd3097e305afa53f7b4312371f62058d2e665320"  # 2025.07.25
     git clone https://github.com/microsoft/vcpkg@"$VCPKG_SHA" "$VCPKG_ROOT"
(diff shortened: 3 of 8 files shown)


pyproject.toml
```diff
@@ -123,6 +123,9 @@ wheels = [
 skip = [
     "*musllinux*",
     "cp314t-*",
+    "cp38-win_arm64", 
+    "cp39-win_arm64",
```
neutrinoceros: These are not necessary, cibuildwheel already skips 3.8 and 3.9 as per `project.python-requires`

tox.ini
```diff
@@ -8,7 +8,8 @@ minversion = 4.22.0 # for dependency_groups
 
 [testenv]
 deps =
-    mindeps: oldest-supported-numpy
+    mindeps: numpy; platform_machine == "ARM64" and sys_platform == "win32"
+    mindeps: oldest-supported-numpy; platform_machine != "ARM64" and sys_platform != "win32"
```
neutrinoceros, line 12: Note to other reviewers : this part becomes obsolete with #2631

neutrinoceros (commented):
Thanks for tackling this ! I rapidly glossed over the changes and I'll approve CI to run to see where things are before I take a closer look !

pyproject.toml
```diff
@@ -123,6 +123,9 @@ wheels = [
...
+    "cp310-win_arm64",
```
neutrinoceros: I guess the reason for skipping this build is that numpy doesn't have a wheel for it (and likely never will) ? Can you add an inline comment so we don't need to remember or look it up again ?

MGN-GIT: @neutrinoceros That is one of the reason and another potential issue is that python distributions for these versions are stopped for Window arm64 officially. So, there is no use of providing these builds.

pyproject.toml
MGN-GIT: @neutrinoceros thanks for the info! Removed the redundant skips

.github/workflows/build_wheels.yml
```diff
@@ -78,6 +80,8 @@ jobs:
           python-version: 3.12
         if: runner.os == 'Windows'
       - run: bash ./ci/cibw_before_all_windows.sh "${{ github.workspace }}"
+        env:
+          ARCH: ${{ runner.arch }}
```
neutrinoceros: Not sure it makes a difference (if it does, please explain), but this would keep the style consistent across the file
```suggestion
          ARCH: ${{ matrix.arch }}
```

ci/cibw_before_all_windows.sh
```diff
@@ -8,17 +8,33 @@ if [[ "$1" == "" ]] ; then
...
+    git clone https://github.com/microsoft/vcpkg "$VCPKG_ROOT"
+    $VCPKG_ROOT/bootstrap-vcpkg.bat -disableMetrics
+    VCPKG_TRIPLET="arm64-windows"
+    $VCPKG_ROOT/vcpkg.exe install zlib:$VCPKG_TRIPLET
+    ZLIB_ROOT="$VCPKG_ROOT/installed/$VCPKG_TRIPLET"
+    EXTRA_PATH="$ZLIB_ROOT/bin"
+    export CL="/I$ZLIB_ROOT/include"
+    export LINK="/LIBPATH:$ZLIB_ROOT/lib"
+    export HDF5_VSVERSION="17-arm64"
+else
```
neutrinoceros:
```suggestion
elif [[ "$ARCH" == "x86_64" ]]; then
```

ci/cibw_before_all_windows.sh
```diff
@@ -8,17 +8,33 @@ if [[ "$1" == "" ]] ; then
...
+    # NuGet for Windows x64
+    nuget install zlib-msvc-x64 -ExcludeVersion -OutputDirectory "$PROJECT_PATH"
+    ZLIB_ROOT="$PROJECT_PATH/zlib-msvc-x64/build/native"
+    EXTRA_PATH="$ZLIB_ROOT/bin_release"
+    export CL="/I$ZLIB_ROOT/include"
+    export LINK="/LIBPATH:$ZLIB_ROOT/lib_release"
+    export HDF5_VSVERSION="17-64"
+fi
```
neutrinoceros, line 36:
```suggestion
else
    echo "Got unexpected arch $ARCH"
    exit(1)
fi
```

ci/cibw_before_all_windows.sh
```diff
@@ -8,17 +8,33 @@ if [[ "$1" == "" ]] ; then
 fi
 PROJECT_PATH="$1"
 
-# nuget
-nuget install zlib-msvc-x64 -ExcludeVersion -OutputDirectory "$PROJECT_PATH"
-EXTRA_PATH="$PROJECT_PATH\zlib-msvc-x64\build\native\bin_release"
+if [[ "$ARCH" == "ARM64" ]]; then
+    # vcpkg for Windows ARM64
+    VCPKG_ROOT="$PROJECT_PATH/vcpkg"
+    git clone https://github.com/microsoft/vcpkg "$VCPKG_ROOT"
```
neutrinoceros: I think we should pin a specific commit here. Preferably, one that's tagged, and add a comment to specify which tag it corresponds to

```suggestion
    VCPKG_SHA="dd3097e305afa53f7b4312371f62058d2e665320"  # 2025.07.25
    git clone https://github.com/microsoft/vcpkg@"$VCPKG_SHA" "$VCPKG_ROOT"
```

neutrinoceros: better yet, windows images from github come with the `Chocolatey` package manager (https://github.com/actions/runner-images/), would it be possible to use that instead ?

ci/fix_paths.py
```diff
@@ -17,15 +18,19 @@ def main():
...
         zlib_root = os.environ.get("ZLIB_ROOT")
         if zlib_root:
-            f = pjoin(zlib_root, 'bin_release', 'zlib.dll')
-            copy(f, pjoin(sitepackagesdir, 'h5py', 'zlib.dll'))
+            arch = platform.machine().lower()
+            if arch in ("arm64", "aarch64"):
+                f = pjoin(zlib_root, 'bin', 'zlib1.dll')
+            elif arch in ("amd64", "x86_64"):
+                f = pjoin(zlib_root, 'bin_release', 'zlib.dll')
+            copy(f, pjoin(sitepackagesdir, 'h5py', basename(f)))
```
neutrinoceros, line 35: there should be an `else` branch to, raising a clear exception if we get an unexpected arch.

ci/get_hdf5_win.py
```diff
@@ -28,12 +30,22 @@
...
-    ]
+    if arch in ("arm64", "aarch64"):
+        # ZLIB includes based on vcpkg layout
+        CMAKE_CONFIGURE_CMD += [
+            "-DHDF5_ENABLE_Z_LIB_SUPPORT=ON",
+            f"-DZLIB_INCLUDE_DIR={ZLIB_ROOT}\\include",
+            f"-DZLIB_LIBRARY_RELEASE={ZLIB_ROOT}\\lib\\zlib.lib",
+            f"-DZLIB_LIBRARY_DEBUG={ZLIB_ROOT}\\debug\\lib\\zlibd.lib",
+        ]
+    else:
```
neutrinoceros: should this be 
```suggestion
    elif arch in ("amd64", "x86_64"):
```

?
also, here too, please add a third, catch-all branch to raise a clear exception if we get an unexpected value, this will help finding the places that need an update in the future

neutrinoceros (changes requested):
This looks mostly good to me. Thanks a lot !
I have a couple minor comments that I'd like addressed before merging though

ci/cibw_before_all_windows.sh
MGN-GIT: @neutrinoceros I am not sure would choco provides packages for win-arm64, but I can check this out from my side and get back on this. Thanks for the suggestion!

.github/workflows/build_wheels.yml
MGN-GIT: @neutrinoceros this is looks meaningful! thanks

ci/fix_paths.py
MGN-GIT: @neutrinoceros Made the relevant change, please let me if the exception message needs to modified further

ci/cibw_before_all_windows.sh
neutrinoceros: sounds good. I'll "unresolve" this thread until then, to make sure we don't forget about it. Thank you !

ci/fix_paths.py
```diff
@@ -17,15 +18,21 @@ def main():
...
         if zlib_root:
-            f = pjoin(zlib_root, 'bin_release', 'zlib.dll')
-            copy(f, pjoin(sitepackagesdir, 'h5py', 'zlib.dll'))
+            arch = platform.machine().lower()
+            if arch in ("arm64", "aarch64"):
+                f = pjoin(zlib_root, 'bin', 'zlib1.dll')
+            elif arch in ("amd64", "x86_64"):
+                f = pjoin(zlib_root, 'bin_release', 'zlib.dll')
+            else:
+                raise RuntimeError(f"Unexpected architecture detected: {arch}")
```
neutrinoceros: So the error message clearly shows the unaltered value and where it comes from
```suggestion
                raise RuntimeError(f"Unexpected architecture detected: {platform.machine()=}")
```

ci/bundle_hdf5_whl.py
```diff
@@ -21,7 +22,11 @@ def find_dlls():
     zlib_root = os.environ.get("ZLIB_ROOT")
     if zlib_root:
         print("ZLIB_ROOT", zlib_root)
-        yield os.path.join(zlib_root, 'bin_release', 'zlib.dll')
+        arch = platform.machine().lower()
+        if arch in ("arm64", "aarch64"):
+            yield os.path.join(zlib_root, 'bin', 'zlib1.dll')
+        elif arch in ("amd64", "x86_64"):
+            yield os.path.join(zlib_root, 'bin_release', 'zlib.dll')
```
neutrinoceros, line 29: seems like I missed this one. Please also add a third branch here.

ci/cibw_before_all_windows.sh
MGN-GIT: @neutrinoceros I just tested this on my local Windows on ARM device. It looks like Chocolatey doesn’t provide zlib binaries for Windows ARM64. However, on x64, zlib installs successfully via Chocolatey, which confirms that zlib packages are not available for Windows ARM64 via Chocolatey. This is because Chocolatey internally relies on NuGet to install zlib.

ci/cibw_before_all_windows.sh
```diff
@@ -8,17 +8,37 @@ if [[ "$1" == "" ]] ; then
...
+    # NuGet for Windows x64
+    nuget install zlib-msvc-x64 -ExcludeVersion -OutputDirectory "$PROJECT_PATH"
+    ZLIB_ROOT="$PROJECT_PATH/zlib-msvc-x64/build/native"
+    EXTRA_PATH="$ZLIB_ROOT/bin_release"
+    export CL="/I$ZLIB_ROOT/include"
+    export LINK="/LIBPATH:$ZLIB_ROOT/lib_release"
+    export HDF5_VSVERSION="17-64"
+else
+    echo "Got unexpected arch $ARCH"
+    exit(1)
```
neutrinoceros: whoops, looks like my suggestion contained a syntax error
```suggestion
    echo "Got unexpected arch '$ARCH'"
    exit 1
```

ci/cibw_before_all_windows.sh
neutrinoceros: thanks a lot for checking. Can you add this explanation as inline comments so it doesn't get lost ?

ci/cibw_before_all_windows.sh
MGN-GIT: Yeah, just now I figured out from CI and fixed that!

ci/cibw_before_all_windows.sh
MGN-GIT: @neutrinoceros sure!

MGN-GIT: @neutrinoceros Can you retrigger the CI with the latest commit. Made a fix with respect to vcpkg commit.

neutrinoceros:
Looking good. I think we also should add a `news/*.rst` fragment for this. Beyond that, it looks ready to go !

MGN-GIT:
> Looking good. I think we also should add a `news/*.rst` fragment for this. Beyond that, it looks ready to go !

Please let me know, whether it should be added from my side or the respective reviewer would be adding that?

neutrinoceros:
Any of us can, but I wouldn't push to your branch without asking first. Any way I cannot do it right now. Your call !

MGN-GIT:
> Any of us can, but I wouldn't push to your branch without asking first. Any way I cannot do it right now. Your call !

I have created and added some description, please feel free to add your suggestion. I am open to rephrase accordingly!

PR #2655, opened 2025-08-18
TST: improve parallel-robustness of the test suite itself using unique ids (v2)

crusaderky:
Supersedes #2588

pytest-run-parallel operates as follows:
1. run setUp / setup_method, once
2. spawn many threads
3. run a single test in parallel. All variables created by setUp (`self.XXX`) are shared.
4. join threads
5. run tearDown / teardown_method, once

In the vast majority of h5py tests, `setUp` is used to open a `h5py.File` (and sometimes populate it with test data).

In preparation to running the whole test suite with pytest-run-parallel, preempt name collisions when two threads operate on the same file; this is done by randomizing dataset/group/attribute names.
As per @takluyver's request (#2588), on regular serial pytest runs these names will instead be deterministic to simplify debugging.

Read https://github.com/h5py/h5py/issues/2475#issuecomment-3197457594 for more context and follow-up work.

### DIFF
--- a/h5py/tests/common.py
+++ b/h5py/tests/common.py
@@ -12,6 +12,7 @@ import os
 import shutil
 import inspect
 import tempfile
+import threading
 import subprocess
 from contextlib import contextmanager
 from functools import wraps
@@ -19,7 +20,7 @@ from functools import wraps
 import numpy as np
 from numpy.lib.recfunctions import repack_fields
 import h5py
-
+from h5py._objects import phil
 import unittest as ut
 
 
@@ -53,10 +54,10 @@ class TestCase(ut.TestCase):
     def tearDownClass(cls):
         shutil.rmtree(cls.tempdir)
 
-    def mktemp(self, suffix='.hdf5', prefix='', dir=None):
+    def mktemp(self, suffix='.hdf5', prefix='tmp', dir=None):
         if dir is None:
             dir = self.tempdir
-        return tempfile.mktemp(suffix, prefix, dir=dir)
+        return tempfile.mktemp(suffix, make_name(prefix), dir=dir)
 
     def mktemp_mpi(self, comm=None, suffix='.hdf5', prefix='', dir=None):
         if comm is None:
@@ -167,16 +168,17 @@ class TestCase(ut.TestCase):
             self.assertArrayEqual(dset[s], arr_result)
 
             if not skip_fast_reader:
-                self.assertArrayEqual(
-                    dset._fast_reader.read(s_fast),
-                    arr_result,
-                )
+                with phil:
+                    self.assertArrayEqual(
+                        dset._fast_reader.read(s_fast),
+                        arr_result,
+                    )
         else:
             with self.assertRaises(exc):
                 dset[s]
 
             if not skip_fast_reader:
-                with self.assertRaises(exc):
+                with self.assertRaises(exc), phil:
                     dset._fast_reader.read(s_fast)
 
 NUMPY_RELEASE_VERSION = tuple([int(i) for i in np.__version__.split(".")[0:2]])
@@ -209,11 +211,11 @@ def insubprocess(f):
         for c in "/\\,:.":
             insub = insub.replace(c, "_")
         defined = os.environ.get(insub, None)
-        # fork process
         if defined:
+            # We're already running in a subprocess
             return f(request, *args, **kwargs)
         else:
-            os.environ[insub] = '1'
+            # Spawn a new interpreter and run pytest in it
             env = os.environ.copy()
             env[insub] = '1'
             env.update(getattr(f, 'subproc_env', {}))
@@ -236,3 +238,40 @@ def subproc_env(d):
         return f
 
     return decorator
+
+
+MAIN_THREAD_ID = threading.get_ident()
+
+
+def make_name(template_or_prefix: str = "foo", /) -> str:
+    """Return a static name, to be used e.g. as dataset name.
+
+    When running in pytest-run-parallel, append a thread ID to the name.
+    This allows running tests on shared resources, e.g. two threads can attempt to write
+    to separate datasets on the same File at the same time (even though the actual
+    writes will be serialized by the `phil` lock).
+
+    Calling this function twice from the same thread will return the same name.
+
+    Parameters
+    ----------
+    template_or_prefix
+        Either a prefix to which potentially append the thread ID, or a template
+        containing exactly one "{}" to be replaced with the thread ID.
+    """
+    tid = threading.get_ident()
+    suffix = "" if tid == MAIN_THREAD_ID else f"-{tid}"
+    if "{}" in template_or_prefix:
+        return template_or_prefix.format(suffix)
+    else:
+        return template_or_prefix + suffix
+
+
+def is_main_thread() -> bool:
+    """Return True if the test calling this function is being executed
+    in the main thread; False otherwise.
+    This can be used to detect when a test is running in pytest-run-parallel.
+    that spawns multiple separate threads to run the tests.
+    """
+    tid = threading.get_ident()
+    return tid == MAIN_THREAD_ID
--- a/h5py/tests/test_attribute_create.py
+++ b/h5py/tests/test_attribute_create.py
@@ -14,7 +14,7 @@
 import numpy as np
 from .. import h5t, h5a
 
-from .common import TestCase
+from .common import TestCase, make_name
 
 class TestArray(TestCase):
 
@@ -24,63 +24,68 @@ class TestArray(TestCase):
 
     def test_int(self):
         # See issue 498
-
+        name = make_name()
         dt = np.dtype('(3,)i')
         data = np.arange(3, dtype='i')
 
-        self.f.attrs.create('x', data=data, dtype=dt)
+        self.f.attrs.create(name, data=data, dtype=dt)
 
-        aid = h5a.open(self.f.id, b'x')
+        aid = h5a.open(self.f.id, name.encode('utf-8'))
 
         htype = aid.get_type()
         self.assertEqual(htype.get_class(), h5t.ARRAY)
 
-        out = self.f.attrs['x']
+        out = self.f.attrs[name]
 
         self.assertArrayEqual(out, data)
 
     def test_string_dtype(self):
         # See issue 498 discussion
 
-        self.f.attrs.create('x', data=42, dtype='i8')
+        self.f.attrs.create(make_name(), data=42, dtype='i8')
 
     def test_str(self):
         # See issue 1057
-        self.f.attrs.create('x', chr(0x03A9))
-        out = self.f.attrs['x']
+        name = make_name()
+        self.f.attrs.create(name, chr(0x03A9))
+        out = self.f.attrs[name]
         self.assertEqual(out, chr(0x03A9))
         self.assertIsInstance(out, str)
 
     def test_tuple_of_unicode(self):
         # Test that a tuple of unicode strings can be set as an attribute. It will
         # be converted to a numpy array of vlen unicode type:
+        name = make_name()
         data = ('a', 'b')
-        self.f.attrs.create('x', data=data)
-        result = self.f.attrs['x']
+        self.f.attrs.create(name, data=data)
+        result = self.f.attrs[name]
         self.assertTrue(all(result == data))
         self.assertEqual(result.dtype, np.dtype('O'))
 
+    def test_unicode_np_array(self):
         # However, a numpy array of type U being passed in will not be
         # automatically converted, and should raise an error as it does
         # not map to a h5py dtype
-        data_as_U_array = np.array(data)
-        self.assertEqual(data_as_U_array.dtype, np.dtype('U1'))
+        data = np.array(['a', 'b'], dtype='U1')
         with self.assertRaises(TypeError):
-            self.f.attrs.create('y', data=data_as_U_array)
+            self.f.attrs.create('x', data=data)
 
-    def test_shape(self):
-        self.f.attrs.create('x', data=42, shape=1)
-        result = self.f.attrs['x']
+    def test_shape_scalar(self):
+        name = make_name()
+        self.f.attrs.create(name, data=42, shape=1)
+        result = self.f.attrs[name]
         self.assertEqual(result.shape, (1,))
 
-        self.f.attrs.create('y', data=np.arange(3), shape=3)
-        result = self.f.attrs['y']
+    def test_shape_array(self):
+        name = make_name()
+        self.f.attrs.create(name, data=np.arange(3), shape=3)
+        result = self.f.attrs[name]
         self.assertEqual(result.shape, (3,))
 
     def test_dtype(self):
         dt = np.dtype('(3,)i')
         array = np.arange(3, dtype='i')
-        self.f.attrs.create('x', data=array, dtype=dt)
+        self.f.attrs.create(make_name(), data=array, dtype=dt)
         # Array dtype shape is incompatible with data shape
         array = np.arange(4, dtype='i')
         with self.assertRaises(ValueError):
--- a/h5py/tests/test_attrs.py
+++ b/h5py/tests/test_attrs.py
@@ -19,7 +19,7 @@ import numpy as np
 
 from collections.abc import MutableMapping
 
-from .common import TestCase
+from .common import TestCase, make_name
 
 import h5py
 from h5py import File
@@ -43,7 +43,7 @@ class TestRepr(TestCase):
     """
 
     def test_repr(self):
-        grp = self.f.create_group('grp')
+        grp = self.f.create_group(make_name())
         grp.attrs.create('att', 1)
         self.assertIsInstance(repr(grp.attrs), str)
         grp.id.close()
@@ -58,77 +58,88 @@ class TestAccess(BaseAttrs):
 
     def test_create(self):
         """ Attribute creation by direct assignment """
-        self.f.attrs['a'] = 4.0
-        self.assertEqual(list(self.f.attrs.keys()), ['a'])
-        self.assertEqual(self.f.attrs['a'], 4.0)
+        name = make_name()
+        self.f.attrs[name] = 4.0
+        assert name in self.f.attrs
+        self.assertEqual(self.f.attrs[name], 4.0)
 
     def test_create_2(self):
         """ Attribute creation by create() method """
-        self.f.attrs.create('a', 4.0)
-        self.assertEqual(list(self.f.attrs.keys()), ['a'])
-        self.assertEqual(self.f.attrs['a'], 4.0)
+        name = make_name()
+        self.f.attrs.create(name, 4.0)
+        assert name in self.f.attrs
+        self.assertEqual(self.f.attrs[name], 4.0)
 
     def test_modify(self):
         """ Attributes are modified by direct assignment"""
-        self.f.attrs['a'] = 3
-        self.assertEqual(list(self.f.attrs.keys()), ['a'])
-        self.assertEqual(self.f.attrs['a'], 3)
-        self.f.attrs['a'] = 4
-        self.assertEqual(list(self.f.attrs.keys()), ['a'])
-        self.assertEqual(self.f.attrs['a'], 4)
+        name = make_name()
+        self.f.attrs[name] = 3
+        assert name in self.f.attrs
+        self.assertEqual(self.f.attrs[name], 3)
+        self.f.attrs[name] = 4
+        self.assertEqual(self.f.attrs[name], 4)
 
     def test_modify_2(self):
         """ Attributes are modified by modify() method """
-        self.f.attrs.modify('a',3)
-        self.assertEqual(list(self.f.attrs.keys()), ['a'])
-        self.assertEqual(self.f.attrs['a'], 3)
+        a = make_name("a")
+        b = make_name("b")
 
-        self.f.attrs.modify('a', 4)
-        self.assertEqual(list(self.f.attrs.keys()), ['a'])
-        self.assertEqual(self.f.attrs['a'], 4)
+        self.f.attrs.modify(a, 3)
+        assert a in self.f.attrs
+        self.assertEqual(self.f.attrs[a], 3)
+
+        self.f.attrs.modify(a, 4)
+        assert a in self.f.attrs
+        self.assertEqual(self.f.attrs[a], 4)
 
         # If the attribute doesn't exist, create new
-        self.f.attrs.modify('b', 5)
-        self.assertEqual(list(self.f.attrs.keys()), ['a', 'b'])
-        self.assertEqual(self.f.attrs['a'], 4)
-        self.assertEqual(self.f.attrs['b'], 5)
+        self.f.attrs.modify(b, 5)
+        assert a in self.f.attrs
+        assert b in self.f.attrs
+        self.assertEqual(self.f.attrs[a], 4)
+        self.assertEqual(self.f.attrs[b], 5)
 
         # Shape of new value is incompatible with the previous
         new_value = np.arange(5)
         with self.assertRaises(TypeError):
-            self.f.attrs.modify('b', new_value)
+            self.f.attrs.modify(b, new_value)
 
     def test_overwrite(self):
         """ Attributes are silently overwritten """
-        self.f.attrs['a'] = 4.0
-        self.f.attrs['a'] = 5.0
-        self.assertEqual(self.f.attrs['a'], 5.0)
+        name = make_name()
+        self.f.attrs[name] = 4.0
+        self.f.attrs[name] = 5.0
+        self.assertEqual(self.f.attrs[name], 5.0)
 
     def test_rank(self):
         """ Attribute rank is preserved """
-        self.f.attrs['a'] = (4.0, 5.0)
-        self.assertEqual(self.f.attrs['a'].shape, (2,))
-        self.assertArrayEqual(self.f.attrs['a'], np.array((4.0,5.0)))
+        name = make_name()
+        self.f.attrs[name] = (4.0, 5.0)
+        self.assertEqual(self.f.attrs[name].shape, (2,))
+        self.assertArrayEqual(self.f.attrs[name], np.array((4.0,5.0)))
 
     def test_single(self):
         """ Attributes of shape (1,) don't become scalars """
-        self.f.attrs['a'] = np.ones((1,))
-        out = self.f.attrs['a']
+        name = make_name()
+        self.f.attrs[name] = np.ones((1,))
+        out = self.f.attrs[name]
         self.assertEqual(out.shape, (1,))
         self.assertEqual(out[()], 1)
 
     def test_access_exc(self):
         """ Attempt to access missing item raises KeyError """
         with self.assertRaises(KeyError):
-            self.f.attrs['a']
+            self.f.attrs['notexist']
 
     def test_get_id(self):
-        self.f.attrs['a'] = 4.0
-        aid = self.f.attrs.get_id('a')
+        name = make_name()
+        self.f.attrs[name] = 4.0
+        aid = self.f.attrs.get_id(name)
         assert isinstance(aid, h5a.AttrID)
 
         with self.assertRaises(KeyError):
-            self.f.attrs.get_id('b')
+            self.f.attrs.get_id("notexist")
+
 
 class TestDelete(BaseAttrs):
 
@@ -138,15 +149,16 @@ class TestDelete(BaseAttrs):
 
     def test_delete(self):
         """ Deletion via "del" """
-        self.f.attrs['a'] = 4.0
-        self.assertIn('a', self.f.attrs)
-        del self.f.attrs['a']
-        self.assertNotIn('a', self.f.attrs)
+        name = make_name()
+        self.f.attrs[name] = 4.0
+        self.assertIn(name, self.f.attrs)
+        del self.f.attrs[name]
+        self.assertNotIn(name, self.f.attrs)
 
     def test_delete_exc(self):
         """ Attempt to delete missing item raises KeyError """
         with self.assertRaises(KeyError):
-            del self.f.attrs['a']
+            del self.f.attrs['notexist']
 
 
 class TestUnicode(BaseAttrs):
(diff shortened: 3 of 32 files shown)


h5py/tests/common.py
```diff
@@ -236,3 +237,22 @@ def decorator(f):
...
+    This allows running tests with pytest-run-parallel on shared resources, e.g. two
+    threads can write to separate datasets on the same File at the same time.
+
+    Calling this functionn twice from the same thread will return the same name.
+    """
+    if GIL_ENABLED:
+        return prefix
+    else:
+        tid = threading.get_ident()
+        return f"{prefix}-{tid}"
```
crusaderky: This code path was tested locally (on a single thread). #2650 adds CI coverage.

h5py/tests/test_dataset.py
```diff
@@ -72,58 +72,58 @@ class TestCreateShape(BaseDataset):
...
     def test_create_integer_extended(self):
-        """ Create an extended dataset """
-        dset = self.f.create_dataset('foo', 63)
+        """ Create an extended dataset with integer shape """
+        dset = self.f.create_dataset(name(), 63)
         self.assertEqual(dset.shape, (63,))
         self.assertEqual(dset.size, 63)
-        dset = self.f.create_dataset('bar', (6, 10))
-        self.assertEqual(dset.shape, (6, 10))
-        self.assertEqual(dset.size, (60))
```
crusaderky, line 103: Redunant with `test_create_extended` a few lines above

aragilar:
Is there any possibility of this causing deadlocks when using the `@insubprocess` decorator (or testing MPI)? Do we need to add additional guards around such tests (or skip them in such cases)?

crusaderky:
> Is there any possibility of this causing deadlocks when using the `@insubprocess` decorator (or testing MPI)? Do we need to add additional guards around such tests (or skip them in such cases)?

This PR does not add pytest-run-parallel.

For when we will add pytest-run-parallel in a later PR: by reading the implementation of the `@insubprocess` decorator, I understand that pytest-run-parallel would spawn N threads, each of which would execute `subprocess.call([sys.executable, '-m', 'pytest'`, notably without the `--parallel-threads` flag. This is a (somewhat weirdly engineered) spawn, not a fork, so there will be no problem there. 

> (or testing MPI)?

Unless you explicitly call `mpirun -n 2 python -m pytest --with-mpi --parallel-threads=4`, the two pytest plugins won't interact with each other.

h5py/tests/common.py
```diff
@@ -236,3 +237,22 @@ def decorator(f):
...
+
+GIL_ENABLED = sys.version_info < (3, 13) or sys._is_gil_enabled()
+
+
+def name(prefix: str = "foo") -> str:
+    """Return a static name, to be used e.g. as dataset name.
+
+    When running on a free-threading interpreter, append a thread ID to the name.
+    This allows running tests with pytest-run-parallel on shared resources, e.g. two
+    threads can write to separate datasets on the same File at the same time.
```
tacaswell:
```suggestion
    threads can attempt to write to separate datasets on the same File at the same
    time (even though the actually writes will be serialized).
```

h5py/tests/common.py
```diff
@@ -236,3 +237,23 @@ def decorator(f):
...
+GIL_ENABLED = sys.version_info < (3, 13) or sys._is_gil_enabled()
+
+
+def name(prefix: str = "foo") -> str:
+    """Return a static name, to be used e.g. as dataset name.
+
+    When running on a free-threading interpreter, append a thread ID to the name.
+    This allows running tests with pytest-run-parallel on shared resources, e.g. two
+    threads can attempt to write to separate datasets on the same File at the same
+    time (even though the actually writes will be serialized).
```
crusaderky:
```suggestion
    time (even though the actual writes will be serialized by the `phil` lock).
```

tacaswell:
This is a lot of changes to the test suite, but I'm still not sure I understand the goal.

We serialize all IO to the file via libhdf5 (which is not threadsafe) and drop the GIL so if we had significant bugs there I think we would have found them (as the consequence of messing that up is segfault).  If we want to really hammer that with free-threading can we get away with a more minimal change of adding a specific set of tests for that?

crusaderky:
> We serialize all IO to the file via libhdf5 (which is not threadsafe) and drop the GIL so if we had significant bugs there I think we would have found them (as the consequence of messing that up is segfault).

Not really. You would have found them if your user base had been historically heavily involved in hammering h5py from multiple threads. Which, given the GIL situation, I strongly suspect has not been the case until now. This will change rapidly with 3.14 as free-threading becomes mainstream.

> If we want to really hammer that with free-threading can we get away with a more minimal change of adding a specific set of tests for that?

This kind of objection applies to all python projects that exist, and not just h5py. I suspect this was discussed quite a lot while designing pytest-run-parallel. The short of it is that writing ad-hoc free-threading tests that are as thorough as the existing test suite is very slim. There is immense value in running everything in parallel, including functionality that is seldom used and that one would not think of when considering the typical parallel use cases.

crusaderky:
> We [...] drop the GIL

I can't find it? e.g. in all the code that leads to H5Dread / H5Dwrite I can't see any point that releases the GIL. Am I missing anything?

tacaswell:
Given the bug reports we have gotten I think we have historically been hammered from multiple threads regularly.  We release the GIL when we can so if you are doing a mix of IO and compute with numpy/pandas (which also release the GIL) threads are worth doing.   We always release the GIL inside of phil so I don't think there is a way for the GIL to have been protecting us [on application level logic!.

If we are going to have issues I expect them to be in things like iterating over the contents of a group/attrs while adding/removing elements from another thread, reading and writing to the same dataset from multiple thread, reading from a dataset while truncating it from another thread, etc.  Basically anyplace where we can have a Python side cache of the state of the underlying file that can get out-of-sync with our view in Python. 

Despite my skepticism this approach did find a fatal flaw for h5py + freethreading!  While I mostly stand by my analysis above, the detail I missed is that the re-entrant counting in the lock object we use for `phil` was being protected by the GIL and without it we are getting the classic reference counting bugs: https://github.com/h5py/h5py/blob/f44f1114c649dfae6276c723978a222427fb56af/h5py/_locks.pxi#L66-L127

The fix is we need a lock on our lock to protect its reference counting.

---

> I can't find it? 

We use the cython `nogil` syntax.

----

I have also pushed a commit to this branch the fixes a segfault when running with the GIL (something that needed phil held was not holding it because we use private methods directly in the tests).  Running with 2 threads I got 97 failures, almost all of which look like they are due to the test trying to create files or tests that mutate-and-check some state on the unit test object.

Can you add a CI job in this PR that uses pytest-run-parallel with the GIL to flush out all of those issues in one shot?   It some cases it may be better just to mark the test as not thread safe.

crusaderky:
> > I can't find it?
> 
> We use the cython `nogil` syntax.

I still can't find it? I see a bunch of functions declared as `nogil`, which means that they can be run without the GIL. But I can't find `with nogil:` anywhere.

> Can you add a CI job in this PR that uses pytest-run-parallel with the GIL to flush out all of those issues in one shot? It some cases it may be better just to mark the test as not thread safe.

I've started the work for this, and I would advise to keep it to a follow-up PR, as it will require quite a few changes. I think that reviewing this PR in its current state is onerous enough as it is, and it _mostly_ applies the same change over and over.
Likewise I think it would be a better idea to move your latest change to a separate PR.

tacaswell:
I think I finally put my finger on what bothers me `pytest-run-parallell`: it is testing that the test suite -- as an application -- is thread safe which is a different question is the library threadsafe.  While the library does need to be thread safe for the test-suite-as-application to be threadsafe, the library can be thread safe without the test-suite being written that way.  

If we are going to use pytest-run-parallel going forward, then I think the first step is to get the test suite to be a thread safe application and I would rather review one large PR that makes all of the required changes rather than a bunch of big but individually incomplete PRs that do not add any value in the interim state (and likely still require the context of all of the previous PRs to review).  It keeps all of the context together and if we do accidentally break the tests in the process of effectively re-writing all of them we only have one place to go back and check later.  Put another way, I do not think that this is can be seen as an incremental process, it is an all-or-nothing step change to the test suite.

With the goal of getting the h5py to work with freethreaded builds I think that fixing the lock-on-the-lock is a far higher priority and that con be exercised with a handful of tests (it is driving the simplest of race conditions incrementing/decrementing a counter).  I'll make a new issue to track that.

----


See the `api_gen.py` logic : https://github.com/h5py/h5py/blob/f44f1114c649dfae6276c723978a222427fb56af/api_gen.py#L268-L286  We have a text file that specifies how to wrap the underlying libhdf5 functions and then auto-generate the lowest level bindings, we have hand-written cython that calls those, and then a user-facing Python level on top of that.

tacaswell (dismissed):
I do not think this should be merged until pytest-run-parallel runs cleanly with the GIL enabled.

The project here is to make the test suite a threaded application and I want all of that work to be merged together.

pytest.ini
```diff
@@ -2,6 +2,8 @@
 markers =
     network: Tests requiring network access, skipped with --no-network
     slow: Slow tests
+    parallel_threads: Number of threads to use in pytest-run-parallel
+    thread_unsafe: Tests that must run on a single thread in pytest-run-parallel
```
crusaderky, line 8: These two lines are necessary when you don't have pytest-run-parallel installed

h5py/tests/test_group.py
```diff
@@ -269,33 +277,35 @@ def test_reference_numpyobj(self):
...
-
         ref = h5py.h5r.Reference()
 
         with self.assertRaises(ValueError):
             self.f[ref]
 
-        self.f.create_group('x')
-        ref = self.f['x'].ref
-        del self.f['x']
+    @pytest.mark.thread_unsafe(reason="FIXME #2672 does not raise")
```
crusaderky, line 305: https://github.com/h5py/h5py/issues/2672

crusaderky:
@tacaswell as requested I've expanded the scope of the PR.
Now, pytest-run-parallel terminates successfully on both 3.13 and 3.13t.
Actually running pytest-run-parallel in CI remains out of scope and will be the target of a follow-up PR.

Additional follow-ups for non-trivial issues:
- https://github.com/h5py/h5py/issues/2673
- https://github.com/h5py/h5py/issues/2672

This is again ready for review.

crusaderky:
Everything runs successfully on 3.14t too, which fixes #2673.

Reproducer:
```bash
#!/bin/bash
set -o errexit

eval "$(mamba shell hook --shell bash)"
mamba create -y -n h5py-314t python-freethreading=3.14 numpy hdf5 python-build pytest pytest-mpi pytest-run-parallel -c conda-forge -c conda-forge/label/python_rc
mamba activate h5py-314t

rm -rf dist
python -m build
pip install dist/*.whl --force-reinstall --no-deps

mkdir -p /tmp/h5py
cp pytest.ini /tmp/h5py/
cd /tmp/h5py
export PYTHON_GIL=0
python -m pytest --pyargs h5py -v --durations 10 --parallel-threads=32
```

crusaderky:
@tacaswell gentle ping

crusaderky:
@tacaswell one more gentle ping. This has been waiting for review since Sep 1.

neutrinoceros:
@crusaderky can you rebase this branch please ?
There are no conflicts (yet), but a clean history might help with reviews. Thanks !

crusaderky:
> @crusaderky can you rebase this branch please ? There are no conflicts (yet), but a clean history might help with reviews. Thanks !

done

crusaderky:
Hi @neutrinoceros @tacaswell , any chance there can be a review soon? This has been waiting since Sep 1st.

h5py/tests/common.py
```diff
@@ -236,3 +238,38 @@ def decorator(f):
         return f
 
     return decorator
+
+
+MAIN_THREAD = threading.get_ident()
```
neutrinoceros:
```suggestion
MAIN_THREAD_ID = threading.get_ident()
```

h5py/tests/common.py
```diff
@@ -236,3 +238,38 @@ def decorator(f):
...
+    Calling this function twice from the same thread will return the same name.
+
+    Parameters
+    ----------
+    template_or_prefix
+        Either a prefix to which potentially append the thread ID, or a template
+        containing exactly one "{}" to be replaced with the thread ID.
+    """
+    tid = threading.get_ident()
+    suffix = "" if tid == MAIN_THREAD else f"-{tid}"
```
neutrinoceros:
```suggestion
    suffix = "" if tid == MAIN_THREAD_ID else f"-{tid}"
```

h5py/tests/common.py
```diff
@@ -236,3 +238,38 @@ def decorator(f):
         return f
 
     return decorator
+
+
+MAIN_THREAD = threading.get_ident()
+
+
+def name(template_or_prefix: str = "foo", /) -> str:
```
neutrinoceros: I'd suggest either renaming the function to `make_name` or to remove the default value, as the effect of `name()` is (IMO) impossible to guess from a call site. Ideally I would even prefer doing both but that might also make call sites harder to read...

h5py/tests/common.py
```diff
@@ -236,3 +238,38 @@ def decorator(f):
...
+    tid = threading.get_ident()
+    suffix = "" if tid == MAIN_THREAD else f"-{tid}"
+    if "{}" in template_or_prefix:
+        return template_or_prefix.format(suffix)
+    else:
+        return template_or_prefix + suffix
+
+
+def is_parallel_test() -> bool:
+    """Return True if the test calling this function is likely being executed
```
neutrinoceros: "likely" makes me worried it might produce subtly broken results. Maybe we could invert the logic and use a `is_main_thread` function instead ? (I think the id of the main thread cannot be recycled in practice because it'll only be freed when the program exits ?)

h5py/tests/test_vds/test_highlevel_vds.py
```diff
@@ -316,7 +318,7 @@ def test_index_layout(self):
         vsource = h5.VirtualSource(filename, 'data', shape=(10,))
         layout[inds] = vsource
 
-        outfile = osp.join(self.tmpdir, 'VDS.h5')
+        outfile = osp.join(self.tmpdir, f'{name("VDS")}.h5')
```
neutrinoceros: this looks like it predates the current version of the `name` function
```suggestion
        outfile = osp.join(self.tmpdir, name('VDS{}.h5'))
```

pytest.ini
neutrinoceros: neat. Actually I wouldn't mind adding `pytest-parallel-run` to a new `concurrency` dependency-group (which should include `{include-group="test"}`). We should add an inline comment here with your explanation though

h5py/tests/test_vds/test_lowlevel_vds.py
```diff
@@ -233,7 +230,11 @@ def test_percival_low_level(self):
             # Create the source dataset dataspace
             k = 0
             for foo in self.fname:
-                in_data = h5.File(foo, 'r')['data']
+                # FIXME can't open the same file twice from multiple threads,
+                # even if read-only
```
neutrinoceros: I think this comment should link to issue on the bug tracker

h5py/tests/test_attribute_create.py
```diff
@@ -28,59 +28,60 @@ def test_int(self):
...
         with self.assertRaises(TypeError):
-            self.f.attrs.create('y', data=data_as_U_array)
+            self.f.attrs.create('x', data=data)
 
-    def test_shape(self):
-        self.f.attrs.create('x', data=42, shape=1)
-        result = self.f.attrs['x']
+    def test_shape_scalar(self):
+        self.f.attrs.create(name(), data=42, shape=1)
+        result = self.f.attrs[name()]
```
neutrinoceros: I know `name()` returns a deterministic result so this isn't *broken* per say, but it would be clearer from the outside if we stored and reused the result instead of calling the function twice

h5py/tests/test_attribute_create.py
```diff
@@ -28,59 +28,60 @@ def test_int(self):
...
         self.assertEqual(result.shape, (1,))
 
-        self.f.attrs.create('y', data=np.arange(3), shape=3)
-        result = self.f.attrs['y']
+    def test_shape_array(self):
+        self.f.attrs.create(name(), data=np.arange(3), shape=3)
+        result = self.f.attrs[name()]
```
neutrinoceros: same here

h5py/tests/test_attrs.py
```diff
@@ -157,22 +161,23 @@ class TestUnicode(BaseAttrs):
 
     def test_ascii(self):
         """ Access via pure-ASCII byte string """
-        self.f.attrs[b"ascii"] = 42
-        out = self.f.attrs[b"ascii"]
+        n = name("ascii").encode("ascii")
```
neutrinoceros: neat, that's a bugfix in my book

h5py/tests/test_attrs_data.py
```diff
@@ -156,8 +156,8 @@ def test_float(self):
         for dt in dtypes:
             data = np.ndarray((1,), dtype=dt)
             data[...] = 42.3
-            self.f.attrs['x'] = data
-            out = self.f.attrs['x']
+            self.f.attrs[name()] = data
+            out = self.f.attrs[name()]
             # TODO: Clean up after issue addressed !
             print("dtype: ", out.dtype, dt)
             print("value: ", out, data)
```
neutrinoceros, line 163: might be a as good a time as any to honor that TODO

h5py/tests/test_attrs_data.py
```diff
@@ -206,22 +206,22 @@ def test_vlen_string_array(self):
...
         self.assertEqual(out[0], data[0])
         self.assertEqual(out[1], data[1])
 
     def test_string_scalar(self):
         """ Storage of variable-length byte string scalars (auto-creation) """
 
-        self.f.attrs['x'] = b'Hello'
-        out = self.f.attrs['x']
+        self.f.attrs[name()] = b'Hello'
+        out = self.f.attrs[name()]
```
neutrinoceros: again here, I'd prefer we store and reuse the result

h5py/tests/test_attrs_data.py
```diff
@@ -230,12 +230,12 @@ def test_string_scalar(self):
     def test_unicode_scalar(self):
         """ Storage of variable-length unicode strings (auto-creation) """
 
-        self.f.attrs['x'] = u"Hello" + chr(0x2340) + u"!!"
-        out = self.f.attrs['x']
+        self.f.attrs[name()] = u"Hello" + chr(0x2340) + u"!!"
+        out = self.f.attrs[name()]
```
neutrinoceros: here too

h5py/tests/test_base.py
```diff
@@ -106,18 +108,18 @@ def _check_type(self, obj):
...
-        dset = self.f.create_dataset(self.USTRING, (1,))
+        dset = self.f.create_dataset(name(self.USTRING), (1,))
         self._check_type(dset)
 
     def test_namedtype(self):
         """ Named type repr() with unicode """
-        self.f['type'] = np.dtype('f')
-        typ = self.f['type']
+        self.f[name()] = np.dtype('f')
+        typ = self.f[name()]
```
neutrinoceros: and here

h5py/tests/test_completions.py
```diff
@@ -1,50 +1,42 @@
...
+        attrs = self.f.create_group(name()).attrs
 
         # Write out of alphabetical order to test that completions come back in
         # alphabetical order, as opposed to, say, insertion order.
         attrs['b'] = 1
         attrs['a'] = 2
-        self.assertEqual(
-            attrs._ipython_key_completions_(),
-            ['a', 'b']
-        )
```
neutrinoceros, line 46: not sure why this was removed, but if that's intentional, we should maybe update the comment too

h5py/tests/test_dataset.py
```diff
@@ -49,12 +50,13 @@ class TestRepr(BaseDataset):
     endian_mark = '>' if sys.byteorder=='big' else '<'
 
     def test_repr_basic(self):
-        ds = self.f.create_dataset('foo', (4,), dtype='int32')
-        assert repr(ds) == f'<HDF5 dataset "foo": shape (4,), type "{self.endian_mark}i4">'
+        ds = self.f.create_dataset(name(), (4,), dtype='int32')
+        assert repr(ds) == f'<HDF5 dataset "{name()}": shape (4,), type "{self.endian_mark}i4">'
```
neutrinoceros: store+reuse

h5py/tests/test_dataset.py
```diff
@@ -330,61 +333,70 @@ class TestCreateRequire(BaseDataset):
...
         self.assertIsInstance(dset, Dataset)
         self.assertEqual(dset.shape, (10, 3))
 
     def test_create_existing(self):
         """ require_dataset yields existing dataset """
-        dset = self.f.require_dataset('foo', (10, 3), 'f')
-        dset2 = self.f.require_dataset('foo', (10, 3), 'f')
+        dset = self.f.require_dataset(name(), (10, 3), 'f')
+        dset[0, 0] = 123
+        dset2 = self.f.require_dataset(name(), (10, 3), 'f')
```
neutrinoceros: store+reuse here and in the immediate following test cases

h5py/tests/test_dataset.py
```diff
@@ -663,7 +677,7 @@ class TestCreateCompressionNumber(BaseDataset):
     """
         Feature: Datasets created with a compression code
     """
-
+    @pytest.mark.thread_unsafe(reason="monkey-patch")
```
neutrinoceros, line 695: can you elaborate ? is monkey-patching *always* thread unsafe somehow ?

h5py/tests/common.py
crusaderky: Changed to is_main_thread

h5py/tests/test_attrs_data.py
crusaderky: I'm being _very_ careful not to introduce scope creep in this already humongous PR.

h5py/tests/test_dataset.py
crusaderky: Any monkey-patching that 
1. samples the previous global states and stores it in a backup local variable
2. monkey-patches the global state
3. does its thing
4. reinstates the backup

is thread-unsafe even if all threads set the same temporary value, because when each thread executes the backup (1) the global state may have already been monkey-patched by another thread.

Various context managers like the one here specifically execute steps 1 to 4 in parallel over multiple threads.
The `monkeypatch` pytest fixture executes steps 1-3 in parallel, then collapses the thread pool, and finally executes step 4 - with the same issue. For this reason the fixture causes pytest-run-parallel to automatically label the test as thread-unsafe.

Monkey-patching from a fixture with `scope="interpreter"` is thread-safe because steps 1, 2, and 4 happen all in a single thread.

h5py/tests/test_completions.py
crusaderky: I found it redundant with the test immediately afterwards. The comment is still valid.
In the interest of reducing scope creep, I will revert.

h5py/tests/test_vds/test_lowlevel_vds.py
crusaderky: https://github.com/h5py/h5py/issues/2714

crusaderky:
@neutrinoceros I've addressed all comments; this is ready for the next round of review.

h5py/tests/test_vds/test_lowlevel_vds.py
crusaderky: Fixed underlying issue in master

crusaderky:
@neutrinoceros @tacaswell any update?

neutrinoceros:
in the middle of a sprint this week, I'll get back to this as soon as I'm able !

crusaderky:
@neutrinoceros @tacaswell this PR is now 3 months old and blocks any further work on free-threading. Keeping it up-to-date with main is also time consuming. Can we push it through the finishing line?

neutrinoceros:
My understanding is that @tacaswell's concerns were indeed addressed, so I plan to merge this on Friday at the latest, unless he corrects me in the interval. Thanks again !

crusaderky:
> I'm really sorry about the delay. I have a single suggestion left: let's add a `concurrency` dependency group with `pytest-run-parallel>=0.7.1` in it. That'll help keeping things tidy. Otherwise, LGTM. I'll just make sure others' concerns were addressed.

I was planning to do that as part as the followup PR that run pytest-run-parallel in CI

neutrinoceros:
That's fine, no worries. It's really a detail, not a blocking request.

neutrinoceros:
Ah. I'm finding out now that @tacaswell's request for changes is disabling the merge button. I could "dismiss" his review but that does feel a little too norm-breaking for me, so I prefer to wait for his explicit signing off.

h5py/tests/common.py
```diff
@@ -209,11 +211,11 @@ def wrapper(request, *args, **kwargs):
         for c in "/\\,:.":
             insub = insub.replace(c, "_")
         defined = os.environ.get(insub, None)
-        # fork process
         if defined:
+            # We're already running in a subprocess
             return f(request, *args, **kwargs)
         else:
-            os.environ[insub] = '1'
```
Carreau, line 216: +1, looking at commit history it was an oversight to have it twice.

h5py/tests/test_attrs.py
```diff
@@ -185,24 +198,27 @@ class TestCreate(BaseAttrs):
     def test_named(self):
         """ Attributes created from named types link to the source type object
         """
-        self.f['type'] = np.dtype('u8')
-        self.f.attrs.create('x', 42, dtype=self.f['type'])
-        self.assertEqual(self.f.attrs['x'], 42)
-        aid = h5a.open(self.f.id, b'x')
+        t = make_name("t")
+        x = make_name("x")
+        self.f[t] = np.dtype('u8')
```
Carreau, line 203: for any other reviewer, I was wondering if `'type'` had any specific meaning, it does not AFAICT, it can be renamed.

h5py/tests/test_completions.py
```diff
@@ -1,40 +1,37 @@
...
-
-        self.f.create_group('data2', [1, 2, 3])
-        self.assertEqual(
-            self.f._ipython_key_completions_(),
-            ['data', 'data2', 'g', 'h'],
-        )
-
-        # Test on subgroup.
-        g.create_dataset('g_data1', [1, 2, 3])
+        with h5py.File(self.mktemp(), 'w') as f:
```
Carreau, line 10: maybe and even likeley dumb question, why not rely on `self.f` and the fact that to avoid sharing the file with the other test in this class ?

h5py/tests/test_completions.py
```diff
@@ -1,40 +1,37 @@
...
+            )
+
+    def test_subgroup_completions(self):
+        g = self.f.create_group(make_name())
         g.create_dataset('g_data2', [4, 5, 6])
-        self.assertEqual(
-            g._ipython_key_completions_(),
-            ['g_data1', 'g_data2'],
-        )
-
```
Carreau, line 29: I think the testing here was about caching and that creating a 3rd group would not still return `['g_data1', 'g_data2']` and indeed have             `'g_data3'`

h5py/tests/test_completions.py
```diff
@@ -1,40 +1,37 @@
...
+        g.create_dataset('g_data1', [1, 2, 3])
         g.create_dataset('g_data3', [7, 8, 9])
         self.assertEqual(
             g._ipython_key_completions_(),
             ['g_data1', 'g_data2', 'g_data3'],
         )
 
     def test_attrs_completions(self):
-        attrs = self.f.attrs
```
Carreau, line 37: +1 to create a name group and test on this subgroup.\Maybe a comment that's because of concurrent access?

h5py/tests/test_dataset.py
```diff
@@ -330,61 +336,78 @@ class TestCreateRequire(BaseDataset):
...
         self.assertIsInstance(dset, Dataset)
         self.assertEqual(dset.shape, (10, 3))
 
     def test_create_existing(self):
         """ require_dataset yields existing dataset """
-        dset = self.f.require_dataset('foo', (10, 3), 'f')
-        dset2 = self.f.require_dataset('foo', (10, 3), 'f')
+        name = make_name()
+        dset = self.f.require_dataset(name, (10, 3), 'f')
+        dset[0, 0] = 123
```
Carreau, line 347: +1

h5py/tests/test_dataset.py
```diff
@@ -621,39 +647,42 @@ class TestCreateGzip(BaseDataset):
...
         self.assertEqual(dset.compression_opts, 9)
 
     def test_gzip_implicit(self):
         """ Create with implicit gzip level (level 4) """
-        dset = self.f.create_dataset('foo', (20, 30), compression='gzip')
+        dset = self.f.create_dataset(make_name(), (20, 30), compression='gzip')
         self.assertEqual(dset.compression, 'gzip')
         self.assertEqual(dset.compression_opts, 4)
 
+    @pytest.mark.thread_unsafe(reason="monkey-patch")
```
Carreau, line 661: 👍

h5py/tests/test_dataset.py
```diff
@@ -1871,7 +1931,8 @@ def _help_float_testing(self, np_dt, dataset_name='vlen'):
 
         # Make sure we can close the file.
         self.f.flush()
-        self.f.close()
```
Carreau, line 1874: This appear to be wrong even on base branch when not having threads, if it's closed, should each test create a new File ? (potentially to fix separately, but keeping a note for myself)

h5py/tests/test_dataset.py
```diff
@@ -1915,28 +1981,30 @@ class TestLowOpen(BaseDataset):
...
 
     def test_dapl(self):
         """ Test the dapl keyword to h5d.open """
+        name = make_name()
         dapl = h5py.h5p.create(h5py.h5p.DATASET_ACCESS)
-        dset = self.f.create_dataset('x', (100,))
+        dset = self.f.create_dataset(name, (100,))
         del dset
-        dsid = h5py.h5d.open(self.f.id, b'x', dapl)
+        dsid = h5py.h5d.open(self.f.id, bytes(name, 'utf-8'), dapl)
```
Carreau: I'm curious, any reason you use bytes(name, utf8) while in ohter file it's .encode()?

h5py/tests/test_completions.py
crusaderky: it would make it more complicated to test `f._ipython_key_completions_()` a few lines below. If you used `self.f.create_group(name("g"))`, you would end up with a more or less arbitrary number of elements returned by `f._ipython_key_completions_()` starting with `g-`, which makes the test less readable and less robust in the use case of single thread.

Carreau:
+1 to latest commit.

tacaswell:
I'm OK with this being merged on @Carreau 's and @neutrinoceros  's reviews.

Do we want to clean the history up to eliminate the extranious merges (either squash-merge or a rebase before merge)?

crusaderky:
Rebased

Carreau:
🥳

neutrinoceros:
Thanks @crusaderky and @Carreau for your help landing this. Much appreciated.

PR #2657, opened 2025-08-20
ChunkIterator support for `None` in slice objects

ajelenak:
ChunkIterator currently does not support selections where slice objects contain `None` or are integer numbers. Example:

```python
>>> import numpy
>>> numpy.index_exp[:,19]
(slice(None, None, None), 19)
```

This PR will transform the above tuple into: `(slice(0, dset.shape[dim], None), slice(19, 20, None))`.

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -391,7 +391,7 @@ class ChunkIterator:
                         step = s.step
                     case _:
                         # TODO: use typing.assert_never when Python 3.10 is dropped
-                        raise TypeError(f'{s}: Selection object must be a slice or integer')
+                        raise AssertionError(f'{s}: Selection object must be a slice or integer')
                 sel[dim] = slice(start, stop, step)
                 self._sel = tuple(sel)
 


h5py/_hl/dataset.py
```diff
@@ -367,16 +367,26 @@ def __init__(self, dset, source_sel=None):
...
         else:
-            if isinstance(source_sel, slice):
+            if isinstance(source_sel, (slice, int)):
                 self._sel = (source_sel,)
             else:
                 self._sel = source_sel
-        if len(self._sel) != rank:
-            raise ValueError("Invalid selection - selection region must have same rank as dataset")
+            if len(self._sel) != rank:
+                raise ValueError("Invalid selection - selection region must have same rank as dataset")
```
neutrinoceros: the additional indentation looked like a smell to me at first, because it implies changes in semantics, but then I realized that the only case where this would cause a change in behavior was if the `if source_sel is None` branch was buggy and defined a `_sel` attribute with an incorrect shape, which currently is clearly not the case. Maybe we could add a simple `assert` there so we'd still catch a mistake if it ever slipped in in the future, but that doesn't seem necessary atm.

h5py/_hl/dataset.py
```diff
@@ -367,16 +367,26 @@ def __init__(self, dset, source_sel=None):
         if source_sel is None:
             # select over entire dataset
             self._sel = tuple(
-                slice(0, self._shape[dim])
-                for dim in range(rank)
+                slice(0, self._shape[dim]) for dim in range(rank)
             )
         else:
-            if isinstance(source_sel, slice):
+            if isinstance(source_sel, (slice, int)):
```
neutrinoceros, line 373: since we end up binding `self._sel` later anyway, maybe we should bind the local `sel` list here instead ?

h5py/_hl/dataset.py
```diff
@@ -367,16 +367,26 @@ def __init__(self, dset, source_sel=None):
...
+                raise ValueError("Invalid selection - selection region must have same rank as dataset")
+            sel = list()
+            for dim in range(rank):
+                s = self._sel[dim]
+                if isinstance(s, int):
+                    sel.append(slice(s, s + 1))
+                else:
+                    sel.append(slice(s.start or 0,
+                                     s.stop or self._shape[dim],
+                                     s.step))
```
neutrinoceros: trying to future proof the logic here for type-checking
```suggestion
            for dim, s in enumerate(sel):
                start: int | None
                stop: int | None
                step: int | None
                match s:
                    case int():
                        start = s
                        stop = s + 1
                        step = None
                    case slice():
                        start = s.start or 0
                        stop = s.stop or self._shape[dim]
                        step = s.step
                    case _:
                        # should be unreachable
                        # TODO: use typing.assert_never when Python 3.10 is dropped
                        raise RuntimeError
                sel[dim] = slice(start, stop, step)
```

h5py/_hl/dataset.py
ajelenak: Yes, I noticed the same.

h5py/_hl/dataset.py
ajelenak: See my changes in 3b0791de.

h5py/_hl/dataset.py
ajelenak: I added this code block, see https://github.com/h5py/h5py/commit/3b0791decc434c64cd9460adcc649eb1adb37fe3.

neutrinoceros (approved):

h5py/_hl/dataset.py
```diff
@@ -367,16 +367,34 @@ def __init__(self, dset, source_sel=None):
...
+                        start = s
+                        stop = s + 1
+                        step = None
+                    case slice():
+                        start = s.start or 0
+                        stop = s.stop or self._shape[dim]
+                        step = s.step
+                    case _:
+                        # TODO: use typing.assert_never when Python 3.10 is dropped
+                        raise TypeError(f'{s}: Selection object must be a slice or integer')
```
neutrinoceros: I think this should be `raise AssertionError(...)` to match what `typing.assert_never` will do when we have access to it

neutrinoceros:
@ajelenak this grew a merge conflict. Can you rebase ? I'm happy to merge after that.

h5py/_hl/dataset.py
ajelenak: Done. You can merge this PR now.

Issue #2656, opened 2025-08-20
phil lock incorrect under freethreaded usage


tacaswell:
There is a fatal flaw for h5py + freethreading!   Although we expect `phil` to protect the user, the reference counting in `phil` to be re-entrant was depending on the GIL to protect its counters: https://github.com/h5py/h5py/blob/f44f1114c649dfae6276c723978a222427fb56af/h5py/_locks.pxi#L66-L127

The fix is we need a lock on our lock to protect its reference counting, possibly contingent on if we are in a freethreaded build or not.  The current implementation has some performance optimizations, we need to make sure that we do not cause regressions in the GIL case and minimal regression in the freethread case.

_Originally posted by @tacaswell in https://github.com/h5py/h5py/issues/2655#issuecomment-3202051533_

takluyver:
> The fix is we need a lock on our lock to protect its reference counting...

Maybe it's simpler to just swap out the FastRLock implementation for a regular reentrant lock in free-threading builds? I think the point of FastRLock is that it speeds up the common case by assuming the GIL exists. If we add a lock around its counter, its speed advantage probably disappears, and we've just got an extra layer of complexity over a regular lock.

https://github.com/h5py/h5py/blob/f44f1114c649dfae6276c723978a222427fb56af/h5py/_locks.pxi#L16-L20

We could also do performance tests between FastRLock and a standard RLock - the FastRLock code is over a decade old at this point, it's possible the advantage of its shortcut path has diminished over time. 🤷

tacaswell:
Falling back to the standard rlock in probably the better choice!

Poking around I found https://github.com/scoder/fastrlock/issues/19 which is on a version of our lock as a more general purpose package where Stefan says "just use threading.Rlock" and points out that upstream did this with atomics rather than a second lock. 

That Stefan is still maintaining that package makes me optimistic that there is still a speed win.

neutrinoceros:
Something like this seems to work (I ran the test suite with it locally) (patch based of #2637)

```patch
diff --git a/h5py/_objects.templ.pyx b/h5py/_objects.templ.pyx
index 29bd5247..54c3c76b 100644
--- a/h5py/_objects.templ.pyx
+++ b/h5py/_objects.templ.pyx
@@ -37,7 +37,12 @@ import os
 # but not both at the same time.
 
 ### {{if OBJECTS_USE_LOCKING}}
+    ### {{if FREE_THREADING}}
+from threading import RLock
+_phil = RLock()
[...]

PR #2658, opened 2025-08-21
ENH: make `phil` a `threading.RLock` in free-threaded CPython

neutrinoceros:
My interpretation of the proposed fix for #2656

### DIFF
--- a/h5py/_objects.templ.pyx
+++ b/h5py/_objects.templ.pyx
@@ -37,7 +37,12 @@ import os
 # but not both at the same time.
 
 ### {{if OBJECTS_USE_LOCKING}}
+  ### {{if FREE_THREADING}}
+from threading import RLock
+_phil = RLock()
+  ### {{else}}
 cdef FastRLock _phil = FastRLock()
+  ### {{endif}}
 ### {{else}}
 cdef BogoLock _phil = BogoLock()
 ### {{endif}}
--- a/setup_build.py
+++ b/setup_build.py
@@ -12,6 +12,7 @@ except ImportError:
 from distutils.command.build_ext import build_ext
 import copy
 import sys
+import sysconfig
 import os
 import os.path as op
 import platform
@@ -195,6 +196,7 @@ class h5py_build_ext(build_ext):
             "PLATFORM_SYSTEM": platform.system(),
             "OBJECTS_USE_LOCKING": True,
             "OBJECTS_DEBUG_ID": False,
+            "FREE_THREADING": sysconfig.get_config_var("Py_GIL_DISABLED") == 1,
         }
         # Run Cython
         print("Executing cythonize()")


tacaswell:
Can you also add a minimal test?  I think 5 threads all trying to create 5 datasets in a loop in the same file at the same time should be enough.

neutrinoceros:
Sure, though we're not testing on free-threading builds yet. Should this wait for #2650 ?

tacaswell:
I'm happy to merge this on a test + local testing.

neutrinoceros:
Cool. Here you go ! Is this what you had in mind ?

I tested locally with the following command
```
PYTHON_GIL=0 uv run -p 3.14t --group test pytest h5py/tests/test_dataset.py::test_concurrent_dataset_creation
```

h5py/tests/test_dataset.py
```diff
@@ -2196,3 +2197,23 @@ def test_view_properties(view_getter, make_ds, writable_file):
...
+        # Ensure that all threads reach this point before concurrent execution.
+        barrier.wait()
+        for j in range(25):
+            writable_file.create_dataset(f'concurrent_{ithread:02d}_{j:02d}', (1000,), dtype='i4')
+
+    with ThreadPoolExecutor(max_workers=N_THREADS) as executor:
+        futures = [executor.submit(closure, ithread) for ithread in range(N_THREADS)]
+
+    [f.result() for f in futures]
+    print(set(writable_file))
```
neutrinoceros:
```suggestion
```

tacaswell:
I made the test more aggressive as it was not failing if I switched back to our FastRLock.

h5py/tests/test_dataset.py
tacaswell: oops, fixed.

h5py/tests/test_dataset.py
```diff
@@ -2196,3 +2197,22 @@ def test_view_properties(view_getter, make_ds, writable_file):
...
+def test_concurrent_dataset_creation(writable_file):
+    N_THREADS = 25
+    # Defines a thread barrier that will be spawned before parallel execution
+    # this increases the probability of concurrent access clashes.
+    barrier = threading.Barrier(N_THREADS)
+
+    def closure(ithread):
+        # Ensure that all threads reach this point before concurrent execution.
+        barrier.wait()
+        for j in range(25):
```
neutrinoceros: could we use another variable here to clarify that this 25 is the same as the one on line 2218, and unrelated to `N_THREADS` ?

tacaswell (approved):

h5py/tests/test_dataset.py
tacaswell: That seems like over-kill for the internals of a test, should have used a different number.

neutrinoceros: (I'll push another commit in a minute to this effect)

neutrinoceros:
Thanks for checking the deadlock without the patch, I should have thought of that. I reduced the number of datasets per threads and verified that I still got the test to deadlock (seemingly 100% of the time) without the patch.

tacaswell:
Did you verify that with the reduced number of datasets it will reliably trigger the deadlock?

ngoldbaum:
I'm not sure if it might be helpful for a followup, but it turns out you can implement a reentrant lock using PyMutex and some atomic operations. This does require depending on C standard library atomics, which only has experimental support under MSVC. But if that's OK, you can avoid all the overhead of calling into Python (besides PyMutex):

https://github.com/nascheme/py_rlock

No idea if the performance of the `phil` lock is critical enough to justify vendoring some C code.

tacaswell:
I promoted @ngoldbaum 's comment to a new issue, we need some numbers to make any decision other than "change nothing" ;)

PR #2659, opened 2025-08-23
BLD: replace deprecated `setuptools.setup` parameter `setup_requires` with in-tree build backend

neutrinoceros:
Implements a TODO comment from `setup.py`. See documentation linked as in-line comments

### DIFF
--- a/_custom_build/backend.py
+++ b/_custom_build/backend.py
@@ -18,7 +18,7 @@ def get_requires_for_build_wheel(config_settings=None):
     requires = _orig.get_requires_for_build_wheel(config_settings)
 
     # Set the environment variable H5PY_SETUP_REQUIRES=0 if we need to skip
-    # setup_requires for any reason.
+    # these build requirements for any reason.
     if os.getenv('HDF5_MPI') == 'ON' and os.getenv('H5PY_SETUP_REQUIRES') != '0':
         requires.extend([
             "mpi4py ==3.1.2; python_version=='3.10.*'",


_custom_build/backend.py
```diff
@@ -0,0 +1,36 @@
...
+        "mpi4py ==4.0.1; python_version=='3.13.*'",
+        "mpi4py ==4.1.0; python_version=='3.14.*'",
+        # leave dependency unpinned for unstable Python versions
+        "mpi4py",
+    ]
+else:
+    SETUP_REQUIRES = []
+
+def get_requires_for_build_sdist(config_settings=None):
+    return _orig.get_requires_for_build_sdist(config_settings) + SETUP_REQUIRES
```
neutrinoceros: I don't think this is actually needed, in retrospect 

```suggestion
    return _orig.get_requires_for_build_sdist(config_settings)
```

neutrinoceros: I could just remove the entire function wrapper...

neutrinoceros:
Tentatively triaging this to 3.15, although there's no actual urgency, but I'd like to have as much future proofing as we can in there

tacaswell:
We should also consider moving to meson instead of playing whack-a-mole with setuptools issues.

neutrinoceros:
Agreed. Although I'm a bit reluctant to try it *now* because I only managed to migrate a single (extremely simple) package to meson so far and only did so with a lot of help from meson experts. I'm sure they'll be happy to help us further with h5py, as important as it is, but that'll still be a sizeable effort.

takluyver:
Is there some indication that `setup_requires` is going to break soon? I'm somewhat tempted to leave it alone as long as it's working, avoid introducing yet another part, and hopefully get round to a bigger rework of how h5py is built before it becomes a problem in practice. :crossed_fingers: There's obviously some risk to that plan, though.

neutrinoceros:
> Is there some indication that `setup_requires` is going to break soon?

not that I know of, other than the deprecation warning, of course.

> I'm somewhat tempted to leave it alone as long as it's working, avoid introducing yet another part, and hopefully get round to a bigger rework of how h5py is built before it becomes a problem in practice. 🤞 There's obviously some risk to that plan, though.

Well, I would actually like to put "migrate to meson" on the roadmap for next cycle/year, but I need to get familiar with how our *current* build system currently works (which is what I've been doing). In the process, I want to make incremental changes that can also land in intermediate releases that will iron out the existing (minor) defects we have that could turn into catastrophic situations, so when we start making the transition, we don't feel rushed by a collapsing status quo. Anyway, all that is to say: I'd like to do *both*, and reduce all known risks to 0 if I can :)

takluyver:
Fair enough. I think this looks good, then - I'll give it a few more days in case @tacaswell or anyone else wants to weigh in, otherwise I think we can merge it.

neutrinoceros:
Good to go ?

_custom_build/backend.py
```diff
@@ -0,0 +1,33 @@
...
+    # return is a list of packages needed to build h5py
+    # (in addition to static list in pyproject.toml)
+    # For mpi4py, we build against the oldest supported version;
+    # h5py wheels should then work with newer versions of these.
+    # Downstream packagers - e.g. Linux distros - can safely build with newer
+    # versions.
+    requires = _orig.get_requires_for_build_wheel(config_settings)
+
+    # Set the environment variable H5PY_SETUP_REQUIRES=0 if we need to skip
+    # setup_requires for any reason.
```
takluyver:
```suggestion
    # these build requirements for any reason.
```

PR #2662, opened 2025-08-24
ENH: add a `filepath` attribute to `File` and `ExternalLink` instances

neutrinoceros:
Seems to complete the API quite nicely

### DIFF
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -14,6 +14,7 @@
 import inspect
 import os
 import sys
+from pathlib import Path
 from warnings import warn
 
 from .compat import filename_decode, filename_encode
@@ -293,9 +294,15 @@ class File(Group):
 
     @property
     @with_phil
-    def filename(self):
+    def filename(self) -> str:
         """File name on disk"""
         return filename_decode(h5f.get_name(self.id))
+
+    @property
+    @with_phil
+    def filepath(self) -> Path:
+        """File path on disk"""
+        return Path(self.filename).resolve()
 
     @property
     @with_phil
--- a/h5py/_hl/group.py
+++ b/h5py/_hl/group.py
@@ -12,6 +12,7 @@
 """
 
 from contextlib import contextmanager
+from pathlib import Path
 import posixpath as pp
 import numpy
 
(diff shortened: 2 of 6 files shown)


h5py/tests/test_group.py
```diff
@@ -19,7 +19,7 @@
 
 import numpy as np
 import os
-import os.path
```
neutrinoceros, line 22: to be clear I'm removing this import because it's no-op (with `os` already imported). I'm not replacing it with pathlib

h5py/_hl/group.py
```diff
@@ -793,16 +794,21 @@ class ExternalLink:
...
-        """ Path to the external HDF5 file in the filesystem. """
+    def filename(self) -> str:
+        """ Path to the external HDF5 file in the filesystem (as a str)"""
         return self._filename
 
-    def __init__(self, filename, path):
+    @property
+    def filepath(self) -> Path:
+        """ Path to the external HDF5 file in the filesystem. """
+        return Path(self._path, self._filename).resolve()
```
takluyver, line 809: I think `self._path` is the path to the object within the HDF5 file, so it shouldn't be part of this.

takluyver: Also, I think calling `.resolve()` on the path from an external link is not quite right. HDF5 has its own rules for how relative paths in external links are resolved, documented under [H5Lcreate_external](https://support.hdfgroup.org/documentation/hdf5/latest/group___h5_l.html#title9). Most obviously, relative paths are relative to the directory of the file the link is in, not the CWD, but there are also a couple of overrides.

(And confusingly, if it's an absolute path but the file doesn't exist, HDF5 can treat it as a relative path :confused: )

So I think we should either just return the path as it was given, or ensure it's resolved following HDF5's rules, preferably by calling an HDF5 function to resolve it.

h5py/_hl/files.py
```diff
@@ -293,10 +294,16 @@ def attrs(self):
...
-    def filename(self):
+    def filename(self) -> str:
         """File name on disk"""
         return filename_decode(h5f.get_name(self.id))
 
+    @property
+    @with_phil
+    def filepath(self) -> Path:
+        """File path on disk"""
+        return Path(self.filename).resolve()
```
takluyver, line 305: Does `H5Fget_name()` always give us an absolute path, or can it give a relative path? If it can be relative, is there a way to get the right absolute path if we open a file and then `os.chdir()` somewhere else?

takluyver:
I guess the summary of my review could be: resolving paths seems like a straightforward win, but it's actually a can of worms. I've focused on making relative paths absolute, but resolving symlinks also comes with complexities. It might be best to leave that up to the caller.

h5py/_hl/files.py
takluyver: I've just tested, and it does just store whatever you passed when you opened the file, so if this is a relative pathname, we don't know what the CWD was when it was opened.

takluyver:
I think this shouldn't go in in its current form, because it implies we can reliably resolve these paths, but we can't.

What I think could belong in h5py is a function/method to resolve paths for external links following HDF5's documented rules. Taking the relevant paths as parameters hopefully makes the limitations around relative paths etc. clearer.

I have coincidentally [written basically this](https://github.com/European-XFEL/hdf5-vds-check/blob/c8330613736b5644e4432a41bfd0359beee1f09b/hdf5_vds_check.py#L26-L61) for virtual dataset paths recently, which follow similar but slightly different rules to external links. I skipped the bit about setting a prefix in the library with `H5Pset_virtual_prefix()`, since I know that doesn't apply in context. And there are probably some corner cases that go wrong on Windows paths, since I only need it for Linux.

PR #2663, opened 2025-08-24
BUG: fix an unreleased bug where the `track_times` argument to `File` was ignored

neutrinoceros:
This looks like an oversight from #2611

We could easily catch such mistakes before merge in the future by enabling a linter. Specifically I'm thinking of https://docs.astral.sh/ruff/rules/#flake8-unused-arguments-arg
Any objections ?

Furthermore, this should have been caught by the test suite, but it wasn't, which is a strong hint that we're missing a test case. I don't know the details enough to do that myself quite yet, and I actually caught this while working on something else. Suggestions welcome. Otherwise, I'll circle back to it eventually.

### DIFF
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -556,9 +556,9 @@ class File(Group):
                                  alignment_interval=alignment_interval,
                                  meta_block_size=meta_block_size,
                                  **kwds)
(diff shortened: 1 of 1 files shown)


neutrinoceros:
the fix itself should be uncontroversial, and I don't mean to setup a linter within the same PR anyway so let's get this in.

PR #2664, opened 2025-08-24
TYP: annotate `h5py._hl.compat` and `h5py._hl.files` modules

neutrinoceros:
This is a work in progress (I have a couple TODO comments to address before I undraft), but I'm opening early for feedback on the approach itself:

I'm trying a new approach to start getting type annotations into `h5py`, where I focus first on first class public APIs. Here, I'm doing `h5py.File` and its helper functions.
There's no systematic type checking yet. My goal is to first try to provide immediate value to users by supporting IDE autocompleters for most common APIs. Hopefully this approach is sound.


h5py/_hl/files.py
```diff
@@ -23,6 +24,48 @@
...
+    else:
+        from typing_extensions import Self
+
+    if sys.version_info >= (3, 12):
+        from collections.abc import Buffer
+    else:
+        from typing_extensions import Buffer
+
+    OpenFileMode: TypeAlias = Literal['r', 'r+', 'w', 'w-', 'x', 'a']
+    Driver: TypeAlias = Literal['sec2', 'stdio', 'core', 'family', 'fileobj', 'split', 'ros3']
```
neutrinoceros, line 45: This should be 
```suggestion
    BuiltinDriver: TypeAlias = Literal['sec2', 'stdio', 'core', 'family', 'fileobj', 'split', 'ros3']
```

h5py/_hl/files.py
```diff
@@ -349,38 +408,59 @@ def meta_block_size(self):
...
-                 rdcc_nslots=None, rdcc_nbytes=None, rdcc_w0=None, track_order=None,
-                 fs_strategy=None, fs_persist=False, fs_threshold=1, fs_page_size=None,
-                 page_buf_size=None, min_meta_keep=0, min_raw_keep=0, locking=None,
-                 alignment_threshold=1, alignment_interval=1, meta_block_size=None,
-                 *, track_times=False, **kwds):
+    def __init__(
+        self,
+        name: "os.PathLike[str | bytes] | FileLike",
+        mode: "OpenFileMode" = 'r',
+        driver: "Driver | None" = None,
```
neutrinoceros, line 445:
```suggestion
        driver: "Driver | str | None" = None,
```

h5py/_hl/files.py
```diff
@@ -502,6 +582,9 @@ def __init__(self, name, mode='r', driver=None, libver=None, userblock_size=None
             Passed on to the selected file driver.
         """
         if driver == 'ros3':
+            if not isinstance(name, (str, bytes)):
+                raise TypeError(f"with {driver=!r}, name must be str or bytes, not {type(name).__name__}")
```
neutrinoceros, line 593: This will directly conflict with https://github.com/h5py/h5py/pull/2636, but an easy way out is to integrate it directly there. Ping @ajelenak

setup.py
```diff
@@ -30,6 +30,7 @@
     # works (assuming you build it from source too).
     # NumPy 1.21.2 is the first release with any wheels for Python 3.10, our minimum Python.
     "numpy >=1.21.2",
+    "typing-extensions >= 4.6.0 ; python_version < '3.12'",
```
neutrinoceros, line 33: In the strict scope of this PR in isolation, `typing-extension` isn't actually needed at runtime, so it could be a dev-only dependency. But in some instances (e.g. `typing.overload`, `typing.assert_never` ...), which I anticipate I'll need in further iterations, it is necessary to have it available at runtime, so I'm making it a "hard" requirement from the get go (with appropriate env markers), to avoid needless churn.

tacaswell:
Have we had a discussion about inline typing vs adding pyi files?

We did pyi with Matplotlib and it has worked out pretty well.

neutrinoceros:
Admittedly this topic is under discussed: I tried adding stubs in https://github.com/h5py/h5py/pull/2438 but didn't get any feedback at the time. For what it's worth I do see a value for files that are either extension code, or have more than one implementation available (in which case a stub file avoids duplication), and I understand how this approach might have been easier for matplotlib where there's (I think) a lot more activity than in h5py, but for this package specifically I'm not sure I see where the value would be; I find it's generally easier to write and maintain type annotations the closer they live to the source code, and we don't get enough churn that separating the two would compensate for the added maintenance burden, though that's just my opinion.

h5py/_hl/files.py
```diff
@@ -23,6 +24,48 @@
...
+        from typing_extensions import Buffer
+
+    OpenFileMode: TypeAlias = Literal['r', 'r+', 'w', 'w-', 'x', 'a']
+    Driver: TypeAlias = Literal['sec2', 'stdio', 'core', 'family', 'fileobj', 'split', 'ros3']
+    DriverFunction: TypeAlias = Callable[Concatenate[PropFAID, ...], None]
+    LibVer: TypeAlias = Literal['earliest', 'v108', 'v110', 'v112', 'v114' 'latest']
+    FileSpaceStrategy: TypeAlias = Literal['fsm', 'page', 'aggregate', 'none']
+
+    T = TypeVar("T", str, bytes)
+    class FileObj(Protocol, Generic[T]):
```
takluyver, line 51: Honestly, this is a pretty good illustration of the kind of scenario where I *dislike* type hinting: there's a whole separate set of concepts to learn - almost a second language - on top of 'regular' Python.

I totally understand how it can prevent bugs, how IDEs can give more useful hints, etc. But not having to do this kind of stuff was part of what made Python so attractive in the first place. Type hints are a neat idea because we can use them as much or as little as makes sense, but now that we have them it feels essentially mandatory to use them as much as possible, with as much detail as possible. :disappointed: 

Even with a hierarchy of classes and a union to define the acceptable interface as closely as possible, it's still not 100% accurate - a file-like object without `.write()` or `.truncate()` methods would work in read mode.

Maybe the benefits are worth it and this is all just my personal grumble. But I'd prefer to keep the type hints a bit simpler, even if that's less precise. :shrug:

h5py/_hl/files.py
takluyver: As the list of drivers is extensible, (almost) everywhere this is used it has to have a `| str` alternative, which seems to negate the benefit of having a fixed list in the first place. How about we drop this and just hint it as `str`?

h5py/_hl/compat.py
```diff
@@ -29,7 +29,7 @@ def filename_encode(filename):
         return os.fsencode(filename)
 
 
-def filename_decode(filename):
+def filename_decode(filename: str | bytes) -> os.PathLike[str]:
```
takluyver, line 32: I think this one is wrong? It returns a `str`, not a path object.

```suggestion
def filename_decode(filename: str | bytes) -> str:
```

h5py/_hl/files.py
```diff
@@ -353,38 +412,59 @@ def meta_block_size(self):
...
-        def atomic(self):
+        def atomic(self) -> bool:
             """ Set/get MPI-IO atomic mode
             """
             return self.id.get_mpi_atomicity()
 
         @atomic.setter
         @with_phil
-        def atomic(self, value):
+        def atomic(self, value: bool) -> None:
```
takluyver, line 422: This is probably just another pet peeve of mine, but I dislike having `-> None` scattered around methods that you wouldn't expect to return anything anyway, like this property setter. I'd probably extend that to any short method that doesn't have a return statement.

takluyver:
> Have we had a discussion about inline typing vs adding pyi files?

If we're doing type hints, and especially if we're keeping them light (as I prefer), I lean towards inline. But I haven't really worked with .pyi files, and I'm open to being persuaded on that.

I guess we should also start running some type checker in CI, if only to ensure that the type hints we do have stay accurate?

neutrinoceros:
> If we're doing type hints, and especially if we're keeping them light (as I prefer), I lean towards inline. But I haven't really worked with .pyi files, and I'm open to being persuaded on that.

Hum, upon reading your review I get the feeling that maybe stub files would actually be a great compromise; you seem to feel strongly about verbose type hints degrading readability in source code, and I feel strongly that we should be making them as accurate as possible (because h5py is so low in the stack and used by so many), so if we write them in separate files maybe everyone gets what they want ? What do you think ?

> I guess we should also start running some type checker in CI, if only to ensure that the type hints we do have stay accurate?

yes. It's a bit of a wild west right now (many type checkers available, with slight incompatibilities), but I guess the one that most our users might want supported right now is mypy, so I can try to set it up with maximal flexibility so it doesn't *pressure* us to add *more* hints yet. How about I do this in a separate PR (no idea when though) ?

PR #2665, opened 2025-08-25
RFC: use direct `os` import statements and reduce complexity in `filename_(en|de)code` helper functions

neutrinoceros:
A small quality upgrade, inspired by #2664

### DIFF
--- a/h5py/_hl/compat.py
+++ b/h5py/_hl/compat.py
@@ -1,8 +1,8 @@
 """
 Compatibility module for high-level h5py
 """
+import os
 import sys
-from os import fspath, fsencode, fsdecode
 from ..version import hdf5_built_version_tuple
 
 # HDF5 supported passing paths as UTF-8 for Windows from 1.10.6, but this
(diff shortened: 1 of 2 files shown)


h5py/_hl/compat.py
```diff
@@ -38,11 +37,10 @@ def filename_decode(filename):
...
-    if sys.platform == "win32":
-        if isinstance(filename, bytes):
-            return filename.decode(WINDOWS_ENCODING, "strict")
-        elif isinstance(filename, str):
-            return filename
-        else:
-            raise TypeError("expect bytes or str, not %s" % type(filename).__name__)
-    return fsdecode(filename)
+    if not isinstance(filename, (str, bytes)):
+        raise TypeError(f"expect bytes or str, not {type(filename).__name__}")
```
neutrinoceros, line 41: moving this `raise` statement up here isn't a API-breaking change: `os.fsdecode` would already have raised a `TypeError` on any platform in the same conditions, but this change makes the exact error message and traceback consistent on all platforms.

neutrinoceros:
this is supposed to be a *pure* refactor, with 100% backwards compatibility, so I don't expect it to be controversial if I self merge here.

PR #2666, opened 2025-08-25
DOC: fix a couple quirks in `h5py.File`'s docstring

neutrinoceros:
- `'v114'` was missing from `libver`'s list of valid values
- `rdcc_nslot`'s position in the docstring didn't match the one it occupies in the signature

### DIFF
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -400,7 +400,8 @@ class File(Group):
             recommended), 'core', 'sec2', 'direct', 'stdio', 'mpio', 'ros3'.
         libver
             Library version bounds.  Supported values: 'earliest', 'v108',
-            'v110', 'v112', 'v114' and 'latest'.
+            'v110', 'v112', 'v114', 'v200' and 'latest' depending on the
+            version of libhdf5 h5py is built against.
         userblock_size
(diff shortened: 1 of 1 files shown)


h5py/_hl/files.py
```diff
@@ -400,12 +400,21 @@ def __init__(self, name, mode='r', driver=None, libver=None, userblock_size=None
             recommended), 'core', 'sec2', 'direct', 'stdio', 'mpio', 'ros3'.
         libver
             Library version bounds.  Supported values: 'earliest', 'v108',
-            'v110', 'v112'  and 'latest'.
+            'v110', 'v112', 'v114' and 'latest'.
```
ajelenak: You can add `v200` here, too, as it will soon be available.

tacaswell:
```suggestion
            'v110', 'v112', 'v114', 'v200' and 'latest' depending on the
            version of libhdf5 h5py is built against.
```

neutrinoceros: done with @tacaswell's suggestion !

neutrinoceros:
I addressed both (identical) suggestions. Let's get this in.

PR #2675, opened 2025-09-02
RFC: set explicit strict arguments in `zip` calls

neutrinoceros:
This is a pet peeve of mine; now that we dropped Python 3.9, we have access to the `strict` kwarg in `zip`, which, in a better world, would default to `strict=True`, but doesn't for backward compat reasons. I want to stress test against CI and see if *any* of these actually need the implicit `strict=False` behavior.

for reference, I ensured my search was thorough by seeding the patch with
```
ruff check --select B905 --fix
```
but then changed `strict=False` to `strict=True`

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -66,6 +66,9 @@ def make_new_dset(parent, shape=None, dtype=None, data=None, name=None,
     # Validate chunk shape
     if isinstance(chunks, int) and not isinstance(chunks, bool):
         chunks = (chunks,)
+    # Logically, the following `zip`` could be strict, but it's happening
+    # before we've done checks elsewhere that raise more descriptive errors
+    # (like "chunks" must have same rank as dataset shape).
     if isinstance(chunks, tuple) and any(
         chunk > dim for dim, chunk in zip(tmp_shape, chunks, strict=False) if dim is not None
     ):
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -67,7 +67,7 @@ def make_new_dset(parent, shape=None, dtype=None, data=None, name=None,
     if isinstance(chunks, int) and not isinstance(chunks, bool):
         chunks = (chunks,)
     if isinstance(chunks, tuple) and any(
-        chunk > dim for dim, chunk in zip(tmp_shape, chunks) if dim is not None
+        chunk > dim for dim, chunk in zip(tmp_shape, chunks, strict=False) if dim is not None
     ):
         errmsg = "Chunk shape must not be greater than data shape in any dimension. "\
                  "{} is not compatible with {}".format(chunks, shape)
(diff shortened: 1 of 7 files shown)


neutrinoceros:
clearly this doesn't work out of the box. From a very quick look, I'm not sure the one failure means I should loosen the one `zip` call, or if I should change the test. I'll dig into this more seriously tomorrow.

tacaswell:
```
____________________ TestCreateChunked.test_chunks_scalar _____________________

self = <h5py.tests.test_dataset.TestCreateChunked testMethod=test_chunks_scalar>

    def test_chunks_scalar(self):
        """ Attempting to create chunked scalar dataset raises TypeError """
>           self.f.create_dataset('foo', shape=(), chunks=(50,))

py310-test-deps-pre\lib\site-packages\h5py\tests\test_dataset.py:421: 
py310-test-deps-pre\lib\site-packages\h5py\_hl\group.py:193: in create_dataset
    dsid = dataset.make_new_dset(group, shape, dtype, data, name, **kwds)
py310-test-deps-pre\lib\site-packages\h5py\_hl\dataset.py:69: in make_new_dset
    if isinstance(chunks, tuple) and any(

>   if isinstance(chunks, tuple) and any(
        chunk > dim for dim, chunk in zip(tmp_shape, chunks, strict=True) if dim is not None
    ):
E   ValueError: zip() argument 2 is longer than argument 1

py310-test-deps-pre\lib\site-packages\h5py\_hl\dataset.py:69: ValueError

```

It now fails in the wrong way

neutrinoceros:
Ok I think this is actually pretty straight forward with a rested mind and a couple coffees; that one `zip` call is *intended* as not strict, so let's just make it explicit.

h5py/_hl/dataset.py
```diff
@@ -67,7 +67,7 @@ def make_new_dset(parent, shape=None, dtype=None, data=None, name=None,
     if isinstance(chunks, int) and not isinstance(chunks, bool):
         chunks = (chunks,)
     if isinstance(chunks, tuple) and any(
-        chunk > dim for dim, chunk in zip(tmp_shape, chunks) if dim is not None
+        chunk > dim for dim, chunk in zip(tmp_shape, chunks, strict=False) if dim is not None
```
takluyver, line 72: Logically, this one could be strict, but it's happening before we've done checks elsewhere that raise more descriptive errors (like `"chunks" must have same rank as dataset shape`). I'd rather not get into restructuring the code to deal with that, so +1 to leaving it non-strict for now, but maybe it's worth adding a comment to clarify why it's like that.

takluyver:
Other than that, LGTM

h5py/_hl/dataset.py
neutrinoceros: thank you ! I added your explanation more or less verbatim as an in line comment

h5py/_hl/dataset.py
```diff
@@ -66,8 +66,11 @@ def make_new_dset(parent, shape=None, dtype=None, data=None, name=None,
     # Validate chunk shape
     if isinstance(chunks, int) and not isinstance(chunks, bool):
         chunks = (chunks,)
+    # Logically, the following `zip`` could be strict, but it's happening
+    # before we've done checks elsewhere that raise more descriptive errors
+    # (like "chunks" must have same rank as dataset shape).
```
takluyver: Balancing backticks, and we probably don't need the example of the error message.

```suggestion
    # Logically, the following `zip` could be strict, but it's happening
    # before we've done checks elsewhere that raise more descriptive errors
```

PR #2676, opened 2025-09-02
RFC: set explicit exception chains in raise-in-except-blocks

neutrinoceros:
A small quality of life improvement that I found while setting up a linter.
These were detected with
```
ruff check --select B904
```

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -751,7 +751,7 @@ class Dataset(HLObject):
                 try:
                     newlen = int(size)
                 except TypeError:
-                    raise TypeError("Argument must be a single int if axis is specified")
+                    raise TypeError("Argument must be a single int if axis is specified") from None
                 size = list(self.shape)
                 size[axis] = newlen
 
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -170,8 +170,8 @@ def make_fapl(
 
     try:
         set_fapl = _drivers[driver]
-    except KeyError:
-        raise ValueError('Unknown driver type "%s"' % driver)
+    except KeyError as exc:
+        raise ValueError(f'Unknown driver type {driver!r}') from exc
     else:
         if driver == 'ros3':
             token = kwds.pop('session_token', None)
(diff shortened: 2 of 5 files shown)


tacaswell:
By what metric are these being identified as confusing?

neutrinoceros:
There are three ways to chain exceptions, both with associated tracebacks


### option 1: no explicit chaining
```python
>>> def complex_function():
...     ...
...     raise TypeError("original exception")
...     ...
...
>>> try:
...     complex_function()
... except TypeError:
...     # default chaining: 'another exception occured'
...     raise ValueError("replacing exception")
...
```
```python-traceback
  File "<python-input-1>", line 2, in <module>
    complex_function()
    ~~~~~~~~~~~~~~~~^^
  File "<python-input-0>", line 3, in complex_function
    raise TypeError("original exception")
TypeError: original exception

During handling of the above exception, another exception occurred:

  File "<python-input-1>", line 5, in <module>
    raise ValueError("replacing exception")
ValueError: replacing exception
```

### option 2: ditch original exception as irrelevant to the reader
see [PEP 409](https://peps.python.org/pep-0409/)
```py
>>> try:
...     complex_function()
... except TypeError:
...     # we don't care about the original exception, throw it out and make our own instead
...     raise ValueError("replacing exception") from None
...
```
```python-traceback
  File "<python-input-2>", line 5, in <module>
    raise ValueError("replacing exception") from None
ValueError: replacing exception
```

### option 3: properly chain exceptions
(I believe this one comes from [PEP 3134](https://peps.python.org/pep-3134/))
```py
>>> try:
...     complex_function()
... except TypeError as exc:
...     # raise a new, context-specific exception, but keep the original one around
...     raise ValueError("replacing exception") from exc
...
```

```python-traceback
  File "<python-input-3>", line 2, in <module>
    complex_function()
    ~~~~~~~~~~~~~~~~^^
  File "<python-input-0>", line 3, in complex_function
    raise TypeError("original exception")
TypeError: original exception

The above exception was the direct cause of the following exception:

  File "<python-input-3>", line 5, in <module>
    raise ValueError("replacing exception") from exc
ValueError: replacing exception
```

In my opinion, option 2 and 3 are each clearer than option 1, because they let the exception traceback carry the original intention, while option 1 just reads like something completely unexpected happened, and should be reserved for this exceptional case (but happens to be the default because... backwards compat I guess ? chaining is clearly an after thought in the language)

tacaswell:
Can we go with option 3 and only go with option 2 when we are sure it is safe.  I personally find a long error message less confusing than a short error message that has removed any information needed to debug the problem.

neutrinoceros:
sounds good. I rewrote this with `from exc` everywhere but 2 cases where the `try` block only contains a single line, analogous to `b = int(a)`, where I think `from None` is perfectly safe. Let me know if you have any objection there.

PR #2678, opened 2025-09-03
Free-threading: avoid `AttributeError: __exit__` on interpreter shutdown

### LINKED ISSUE #2673: Free-threading: unclean pytest shutdown
Running https://github.com/h5py/h5py/pull/2655 with Python 3.13t and pytest-run-parallel, I get a lot of 

```
Exception ignored in: 'h5py._objects.ObjectID.__dealloc__'
AttributeError: __exit__
AttributeError: __exit__
```

at interpreter exit, followed by a 0 exit code. No traceback is available.

Reading the code,
https://github.com/h5py/h5py/blob/9eff7b8feeb5825ca2493d2a29b7cb7979a92323/h5py/_objects.templ.pyx#L224-L237

it seems that the issue is that `_phil.__exit__` is not available anymore at interpreter shutdown.
Note that `_phil` in 3.13t is a `threading.RLock`, whereas normally it is a FastRLock.

Related: #2671

crusaderky:
Closes #2673

### DIFF
--- a/h5py/_objects.templ.pyx
+++ b/h5py/_objects.templ.pyx
@@ -224,18 +224,16 @@ cdef class ObjectID:
     def __dealloc__(self):
         self._dealloc()
 
-    # Hack to increase the reference counter of _phil by 1.
-    # In free-threading interpreters, this prevents a race condition during interpreter
-    # shutdown where the global _phil is dereferenced by the main thread while
-    # ObjectID.__dealloc__ is being executed in another thread.
-    def _dealloc(self, _phil=_phil):
+    # During interpreter shutdown, module attributes are set to None
+    # before __dealloc__ and __del__ methods are executed.
+    def _dealloc(self, _phil=_phil, warn=warnings.warn, registry=registry):
         with _phil:
             ### {{if OBJECTS_DEBUG_ID}}
             print("DEALLOC - unregistering %d HDF5 id %d" % (self._pyid, self.id))
             ### {{endif}}
             if is_h5py_obj_valid(self) and (not self.locked):
                 if H5Idec_ref(self.id) < 0:
-                    warnings.warn(
+                    warn(
                         "Reference counting issue with HDF5 id {}".format(
                             self.id
                         )
@@ -332,7 +330,7 @@ cdef hid_t pdefault(ObjectID pid):
     return pid.id
 
 
-# Note: _phil=_phil is a hack to increase the reference counter of _phil by 1.
+# Note: _phil=_phil allows this to work during interpreter shutdown.
 # Read note at ObjectID.__dealloc__.
 cdef int is_h5py_obj_valid(ObjectID obj, _phil=_phil):
     """


h5py/_objects.templ.pyx
```diff
@@ -111,15 +111,15 @@ if hasattr(os, "register_at_fork"):
...
+# See also __cinit__ and __del__ for class ObjectID.
 
 import gc
 import weakref
 import warnings
 
 # Will map id(obj) -> weakref(obj), where obj is an ObjectID instance.
 # Objects are added only via ObjectID.__cinit__, and removed only by
-# ObjectID.__dealloc__.
+# ObjectID.__del__.
```
tacaswell: I am concerned about changing from `__dealloc__` to `__del__`.

tacaswell:
The issue is that in the non-free threaded case `_phil` is not Python reference counted but in the free-threaded case it is.

Can you please use the templates to keep this hack only in effect in the free-threading build?  Either than or we do need to implement a cython-only version of the RLock.

h5py/_objects.templ.pyx
```diff
@@ -221,10 +221,17 @@ cdef class ObjectID:
...
 
 
-    def __dealloc__(self):
+    def __del__(self):
+        self._del()
+
+    # Hack to increase the reference counter of _phil by 1.
+    # In free-threading interpreters, this prevents a race condition during interpreter
+    # shutdown where the global _phil is dereferenced by the main thread while
+    # ObjectID.__del__ is being executed in another thread.
```
ngoldbaum: Is it still problematic to call `__dealloc__` if you have a strong reference? I thought you only had to either have a strong reference or switch from `__dealloc__` to `__del__` to avoid the race at shutdown. In this PR you're doing both of those.

h5py/_objects.templ.pyx
ngoldbaum: Can you elaborate a little bit more what the concern is?

tacaswell: It changes when in the lifecycle the delete happens.

tacaswell:
Per https://cython.readthedocs.io/en/latest/src/userguide/special_methods.html#finalization-methods-dealloc-and-del we should not expect to be use a Python object is `__dealloc__`.

On reflection, I stand by my comment from yesterday.  I think we should template this such that for the non-freethreaded case we use `__dealloc__` and for the free-threaded case we use `__del__`.   I do not have a counter example of when it breaks, but I'm worried about moving around the clean up code as it is generically a bit brittle and has been a source of subtle bugs.

ngoldbaum:
> I do not have a counter example of when it breaks, but I'm worried about moving around the clean up code as it is generically a bit brittle and has been a source of subtle bugs.

Seems fair and I understand the desire to be conservative.

That said, you're iterating through a python dict and deleting items from it - this seems like the sort of thing the docs for `__dealloc__` warn against and say to use `__del__` for instead:

https://github.com/h5py/h5py/blob/7a9fa23782a1a9cb04c42ce2d2fcacec26d6dce0/h5py/_objects.templ.pyx#L236-L237

neutrinoceros:
> That said, you're iterating through a python dict and deleting items from it - this seems like the sort of thing the docs for `__dealloc__` warn against and say to use `__del__` for instead

That is also how I understand [the following excerpt](https://cython.readthedocs.io/en/latest/src/userguide/special_methods.html#finalization-methods-dealloc-and-del):

> you should avoid invoking any Python operations which might touch the object.

This being a given, I still share @tacaswell's concern that we might create subtle bugs out of purity principles, and I'd rather avoid that. Free-threaded land seems like ideally suited to experiment though: we don't have a history of backward compatibility for it yet so I think it's okay that we try to adhere to better quality standards from the get go, and if it's ever going to become the defaut build (even, the only one), then we'll just end up getting rid of `__dealloc__` "for free".

h5py/_objects.templ.pyx
neutrinoceros: +1

tacaswell:
I read that excerpt as "the `self` may be half torn down don't trust the Python parts" (and trying to revive a cython object which is half deallocated sounds like a way to get segfaults).  I am less clear on if we should not do _other_ Python things (either for normal "tear down happens in its own order and you can assume nothing" reasons or "thar be dragons" reasons).

I was worried that we had a latent bug with `registry`, however I do not think that is the case as 

https://github.com/h5py/h5py/blob/7a9fa23782a1a9cb04c42ce2d2fcacec26d6dce0/h5py/_objects.templ.pyx#L123

so it is a cython dict (and is not accessible from Python).  There may still be a race on teardown here, but I think it is a race controlled by cython.   That said, there may be a problem with `warnings.warn` on shutdown, but think the use of `registry` is safe.

Maybe the minimal fix is to do something like

```
cdef dict keep_the_lock_alive = {'lock': _phil}
```

in the RLock case?

h5py/_objects.templ.pyx
crusaderky: Switching from `__dealloc__` to `__del__` does not fix the problem.
However, it does not seem to be necessary either now that there is a strong reference. Reverted.

crusaderky:
> Maybe the minimal fix is to do something like
> 
> ```
> cdef dict keep_the_lock_alive = {'lock': _phil}
> ```
> 
> in the RLock case?

Tried this; doesn't seem to have any benefit.

h5py/_objects.templ.pyx
```diff
@@ -326,7 +332,9 @@ cdef hid_t pdefault(ObjectID pid):
     return pid.id
 
 
-cdef int is_h5py_obj_valid(ObjectID obj):
+# Note: _phil=_phil is a hack to increase the reference counter of _phil by 1.
+# Read note at ObjectID.__dealloc__.
+cdef int is_h5py_obj_valid(ObjectID obj, _phil=_phil):
```
crusaderky: I cannot make sense of this.
Logic would say that if you have `_phil=_phil` in both `_dealloc` and `is_h5py_obj_valid`, when you are inside `is_h5py_obj_valid`  then `_phil` should have a reference counter of 2.
However, if I have `_phil=_phil` exclusively in `_dealloc`, I get this:

```python
Exception ignored in: 'h5py._objects.ObjectID.__dealloc__'
  File "h5py/_objects.pyx", line 222, in h5py._objects.ObjectID._dealloc
  File "h5py/_objects.pyx", line 224, in h5py._objects.ObjectID._dealloc
  File "h5py/_objects.pyx", line 339, in h5py._objects.is_h5py_obj_valid
AttributeError: __exit__
```
@ngoldbaum any idea what's happening here?

crusaderky:
I've reached a minimal reproducer. To get this issue you must

- run on Python 3.13t, and
- start (and cleanly rejoin and dereference) at least one thread, and
- have a variable pointing to a h5py.File at the moment of interpreter shutdown, even if it was previously properly closed.

```python
import tempfile
import threading
import h5py

global_file = None

def main():
    with tempfile.TemporaryFile() as fh:
        with h5py.File(fh, 'w') as f:
            pass

    global global_file
    global_file = f

    t = threading.Thread(target=lambda: None)
    t.start()
    t.join()


if __name__ == "__main__":
    main()
```
```
$ python demo.py 
AttributeError: __exit__
Exception ignored in: 'h5py._objects.ObjectID.__dealloc__'
AttributeError: __exit__
```

In the case of the test suite, this is triggered by the many `self.f = h5py.File(...)` variables created by `setUp`. Even if they are all closed cleanly by `tearDown`, the instances of the test classes are not dereferenced until interpreter shutdown.

The `Object` instances that trigger this issue were never owned nor interacted with the separate threads. In fact, you can swap out the two blocks of code, the one that creates and destroys the thread and the one that creates and closes the File, and the issue persists.

crusaderky:
Changing
```patch
< cdef int is_h5py_obj_valid(ObjectID obj):
> cdef int is_h5py_obj_valid(ObjectID obj) except -1:
```
does not cause a hard failure - the interpreter still says Exception ignored.

crusaderky:
If I change objects.templ.pyx to use `_phil = RLock()` on 3.13 (not free-threaded), the issue cannot be reproduced.

Even more minimal reproducer:
```python
import threading
from h5py._objects import ObjectID

t = threading.Thread(target=lambda: None)
t.start()
t.join()
del t

obj = ObjectID(123)
# del obj
```
Uncommenting `del obj` makes the issue disappear.

crusaderky:
I reproduced this in pure python.

```python
import threading

class C:
    def f(self):
        pass

c = C()

class D:
    def __del__(self):
        c.f()

D = D()
# del d

t = threading.Thread(target=lambda: None)
t.start()
t.join()
del t
```
```
(h5py-313t) crusaderky@melody:/tmp/h5py$ python demo.py 
Exception ignored in: <function D.__del__ at 0x51292b24ce0>
  File "demo.py", line 11, in __del__
AttributeError: 'NoneType' object has no attribute 'f'
```

The problem disappears in Python 3.14t RC2.
I'm inclined to close both this PR and the tracking issue and mark everything as already fixed upstream.
The issue on 3.13t is purely cosmetic after all (the program exits 0 after the warnings).
What are your opinions?

ngoldbaum:
I wonder if we can track down the fix and if it can be backported in CPython to 3.13. I agree that this shouldn't be patched over in h5py.

I also think it might be reasonable if h5py decides to only support 3.14t and newer, but I'll wait for Kumar and the maintainers to weigh in on whether they think this is a show-stopper on 3.13t or not...

h5py/_objects.templ.pyx
```diff
@@ -222,6 +222,13 @@ cdef class ObjectID:
 
 
     def __dealloc__(self):
+        self._dealloc()
+
+    # Hack to increase the reference counter of _phil by 1.
```
kumaraditya303: This is not correct reasoning.

The issue is not because of reference counting, but because at runtime shutdown the module attributes are set to `None` hence if any finalizer tries to access module's global attribute, it can crash on `AttributeError`. 

I had suggested to add `_phil=_phil` to function to effectively turn `_phil` into a local variable for the function so that even when finalizer runs after module has`_phil=None`, it will not error with `AttributeError`.

h5py/_objects.templ.pyx
kumaraditya303: See my reasoning above

kumaraditya303:
> The problem disappears in Python 3.14t RC2.

That might be because the order of execution of finalizer changed, but it is still incorrect to access `c` from `__del__`.  It is not safe to rely on implicit gc behavior as it is not stable across versions. 

See https://github.com/python/cpython/blob/6b7b9d00a9dd9cf66e3ec38b4c4b8a385b16b453/Lib/multiprocessing/pool.py#L266 where globals are turned into locals in CPython too

neutrinoceros:
As far as I'm concerned I feel perfectly fine not ever supporting 3.13t at this point; I expect anyone relying on that version will promptly upgrade to 3.14t anyway.

h5py/_objects.templ.pyx
crusaderky: The current PR follows your suggestion. So are you saying this PR should be merged, with just a tweak to the comments?

kumaraditya303: Yes, but you should do the same for `warnings.warn` as that can have the same issue.

tacaswell: In the non-threaded case I do not think Python knows about `_phil`, but adding the closure seems benign enough to do it all cases as I agree we should be doing the same for warnings.warn.  I could also see being defensive about things being `None` and moving on though.

crusaderky: `register` should have the same problem too.

crusaderky: I've updated the PR.

h5py/_objects.templ.pyx
```diff
@@ -222,21 +222,25 @@ cdef class ObjectID:
 
 
     def __dealloc__(self):
+        self._dealloc()
+
+    # During interpreter shutdown, module attributes are set to None
+    # before __dealloc__ and __del__ methods are executed.
+    def _dealloc(self, _phil=_phil, warn=warnings.warn, registry=registry):
```
tacaswell, line 229: I continue to have doubts that we need this for `registry` and `_phil` in the non-freethreaded case (as they are `cdef`'d and to my understanding not managed/accessible to Python so I don't think Python's shut-down order can mess with us).

However if Cython is happy to compile this I do not see any harm in being extra defensive and forcing them to be locals.

kumaraditya303 (approved):

neutrinoceros:
@crusaderky Happy to merge if you're okay to undraft this now !

crusaderky:
@neutrinoceros ready to merge!

PR #2684, opened 2025-09-10
REL: automate sdist creation alongside wheels

neutrinoceros:
A quick follow up to https://github.com/h5py/h5py/pull/2592, simplifying the release process a bit

I can also set up auto-publication as proposed there, but I'll reserve this for a later PR
update: this is #2685

PR #2685, opened 2025-09-10
REL: implement auto publication to PyPI

neutrinoceros:
Builds on top of #2684
Requires actions on the PyPI side to properly setup trusted publishing.


takluyver:
I think we should probably go for this - it seems to be the way we're all encouraged to go for PyPI uploads, and it does mean it's more transparent what's getting uploaded. But it's worth noting that it increases the potential damage an attacker can do through Github. We've been able to relax a bit after at least [one compromise of another package](https://blog.yossarian.net/2024/12/06/zizmor-ultralytics-injection) simply because we knew our Github workflows can't automatically publish to PyPI. So we should plan to be even more careful with security.

PyPI [recommends](https://docs.pypi.org/trusted-publishers/security-model/#provider-specific-considerations) that we set up trusted publishing from Github using a dedicated environment, which can have manual approval and a wait timer. I haven't done that so far on smaller projects, but it's probably worth doing here.

This might also be a good time to trim down the list of people with push access on Github - there are currently 11, and I think most of them are not active.

takluyver:
I've sent an email to the people whose commit access we might remove, asking for responses in 2 weeks.

We could remove commit access by simply removing people from the h5py Github org entirely, or we could set the base role for the org to 'read', and use the teams feature to control finer-grained permissions, e.g. allowing some people to close & relabel issues & PRs (Triage role) without push access. I lean towards keeping it simple, but not strongly.

neutrinoceros:
@takluyver I rebased this and added it to the milestone just in case you'd like to try it out on 3.15, but feel free to postpone if not.

EpicWink:
To use credential-less publishing to PyPI, you will need to give the job the `id-token: write` permission.

Adding a GitHub environment (eg `pypi`, also registered with PyPI) with approval required by specific people is the way to go I think.

You may want to test this with a pre-release tag (eg `v3.15.1a0`), in case something was misconfigured and you don't want to spam downstream subscribers.

neutrinoceros:
thanks for pointing this out. I added the required permission

takluyver:
Thanks, I've set up the trusted publishing on PyPI, so this should work on the next release. I'm happy to wait until then to try it - collecting the packages together for the Github release is already working, this is hopefully a fairly simple addition. :crossed_fingers:

PR #2686, opened 2025-09-11
Set HDF5 plugin path to one value for the `h5pl.remove()` test

ajelenak:
This test was failing in HDF Group's CI using a [Spack](https://spack.io/) virtual environment. Since all the other h5py tests of the H5PL API set the value of the `HDF5_PLUGIN_PATH` env. variable, the PR is just doing the same for this one failing test so the total number of plugin paths is predictable and set to one.


ajelenak:
Self-merging as this is not a controversial PR and all the tests passed.

PR #2689, opened 2025-09-12
MNT: simplify cibuildwheel setup

neutrinoceros:
Explanation: unconditionally setting `CIBW_*` env variables is redundant and a footgun.

- redundant, because all variable we *want* to set dynamically via, e.g., `ci/triage_build.sh`, are already exported to `$GITHUB_ENV`
- a footgun, because if there's nothing to pass, we still define env vars as empty strings, which will take precedent over any corresponding option defined in the `[tool.cibuildwheel]` static table. This costed me longer than I care to admit to realise, so I'd like to simplify the setup before I forget and become subject to the same costly mistake again.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -93,12 +93,6 @@ jobs:
 
       # Now actually build the wheels
       - uses: pypa/cibuildwheel@c923d83ad9c1bc00211c5041d0c3f73294ff88f6 # v3.1.4
-        env:
-          CIBW_ARCHS: ${{ matrix.arch }}
-          CIBW_BUILD: ${{ env.CIBW_BUILD }}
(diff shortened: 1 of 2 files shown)


.github/workflows/build_wheels.yml
```diff
@@ -93,12 +93,6 @@ jobs:
 
       # Now actually build the wheels
       - uses: pypa/cibuildwheel@c923d83ad9c1bc00211c5041d0c3f73294ff88f6 # v3.1.4
-        env:
-          CIBW_ARCHS: ${{ matrix.arch }}
```
neutrinoceros, line 97: this is the only env var that was *always* set as non-empty, however I still think it's redundant given that this information is already contained in `CIBW_BUILD`, which is *also* unconditionally set via `ci/triage_build.sh`

neutrinoceros:
This shouldn't be controversial so I'll self merge here. I checked that the resulting wheels are identical to the ones we get without this patch.

PR #2691, opened 2025-09-14
BLD: cleanup setuptools subclassing, drop unused and deprecated build dependencies on distutils and wheel

neutrinoceros:
This contains a handful of small improvements I found while working on #2589 that can be merged independently.

Most importantly: moving `cythonize` call from `bdist_ext.run` to `bdist_ext.finalize_options` is crucial to avoid hacky patches in #2589, because `finalize_options` is called first, and is responsible for setting the `py_limited_api` attribute on `self.extensions` (which need to be already set at this point).

Furthermore, doing the rounds there allowed me to rid the code of both `wheel` and `distutils` as direct build dependencies, as they are both vendored by setuptools already *and* deprecated as APIs.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ requires = [
     "Cython >=3.0.0,<4",
     "numpy >=2.0.0, <3",
     "pkgconfig",
-    "setuptools >=77",
+    "setuptools >=77.0.1",
 ]
 build-backend = "setuptools.build_meta"
 
--- a/setup.py
+++ b/setup.py
@@ -58,7 +58,7 @@ if setup_configure.mpi_enabled():
 if os.environ.get('H5PY_SETUP_REQUIRES', '1') == '0':
     SETUP_REQUIRES = []
 
-# --- Custom Distutils commands -----------------------------------------------
+# --- Custom setuptools commands -----------------------------------------------
 
 CMDCLASS = {'build_ext': setup_build.h5py_build_ext}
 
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -5,7 +5,7 @@ requires = [
     "Cython >=3.0.0,<4",
     "numpy >=2.0.0, <3",
     "pkgconfig",
-    "setuptools >=77",
+    "setuptools >=77.0.1",
```
neutrinoceros, line 8: Kinda unrelated to the rest of the PR: setuptools 77 (or 77.0 or 77.0.0) doesn't exist, but `asv` will use the minimal version specified here exactly, so this allows running `asv run` once again

neutrinoceros:
I think failures on azure are unrelated, but I'm only 99% sure, and this is touching the build system, so I'm not going to self merge without confirmation

neutrinoceros:
Ok I'm now certain the issue is specific to this branch, however I can't reproduce locally and I'm struggling to configure tox correctly, so this might take some time.

pyproject.toml
```diff
@@ -149,3 +150,11 @@ delocate-wheel --require-archs {delocate_archs} -w {dest_dir} -v {wheel} \
 [tool.cibuildwheel.macos.environment]
 H5PY_ROS3 = "0"
 H5PY_DIRECT_VFD = "0"
+
+[tool.uv]
+# never build numpy from source
+no-build-package = ["numpy"]
+
+[tool.uv.sources]
```
neutrinoceros: this section is the reason why I rebased this branch on top of #2631

neutrinoceros:
Alright, this now works with a very minimal (temporary, and documented) hack, so I don't imagine this would be controversial. I'll merge to unblock #2589 once CI pass

PR #2694, opened 2025-09-16
Support for native complex number datatypes in HDF5-2.0

### LINKED ISSUE #2680: Support for HDF5 complex number datatypes
HDF5-2.0 comes with support for native complex number datatypes. The details are explained [here](https://github.com/HDFGroup/hdf5/blob/68264204214a373abd4eb099a05805efb75cab87/release_docs/RELEASE.txt#L442-L634).

I've started working on support for these datatypes in h5py but need the feedback how to handle the current way of storing complex numbers using a two-field HDF5 compound datatype. What I have done so far is [here](https://github.com/h5py/h5py/compare/master...ajelenak:h5py:native-complex?expand=1). I would like to keep this feature.

One way of achieving this is to use the `logical` flag in the `h5t.py_create(object dtype_in, bint logical=0, bint aligned=0)` [function](https://github.com/h5py/h5py/blob/7a9fa23782a1a9cb04c42ce2d2fcacec26d6dce0/h5py/h5t.templ.pyx#L1688):

* `logical=0` for native HDF5 complex number datatypes (new)
* `logical=1` is for the two-field HDF5 compound datatype (old)

The value of `logical` must be added to the internal `h5t._c_complex(cnp.dtype dt)` [function](https://github.com/h5py/h5py/blob/7a9fa23782a1a9cb04c42ce2d2fcacec26d6dce0/h5py/h5t.templ.pyx#L1568) where the actual work on an appropriate datatype happens.

Does this sound like a reasonable plan?

ajelenak:
Adds native complex number datatypes available in HDF5-2.0. Closes #2680.

The approach here is that the new complex number datatypes are used if the HDF5 library has them. The fallback option is the current 2-field HDF5 compound datatype.


h5py/api_types_hdf5.templ.pxd
```diff
@@ -171,8 +171,8 @@ cdef extern from "hdf5.h":
     H5F_LIBVER_V112 = 3,
     H5F_LIBVER_V114 = 4,
     H5F_LIBVER_V200 = 5,
+    H5F_LIBVER_LATEST   = 5,
     H5F_LIBVER_NBOUNDS
-  int H5F_LIBVER_LATEST  # Use the latest possible format available for storing objects
```
neutrinoceros, line 175: any reason not to make the same change to other variants of this enum ?

h5py/h5t.templ.pyx
```diff
@@ -306,6 +335,27 @@ cdef tuple _get_available_ftypes():
 
 cdef tuple _available_ftypes = _get_available_ftypes()
 
+### {{if HDF5_VERSION >= (2, 0, 0)}}
+
+# Available NumPy complex datatypes
+cdef tuple _get_available_ctypes():
```
neutrinoceros: Is "ctypes" an idiomatic shorthand for complex types in HDF5 ? I'm used to see this name refer to [Python's stdlib `ctypes` module](https://docs.python.org/3/library/ctypes.html), so it looks confusing to me. If it's already idiomatic I can live with it though.

ajelenak: There was already a helper function for the NumPy float dtypes with "ftypes" in its name so I just decided to use "ctypes". I think the comment clarifies what this function is for but what do you suggest for the new name?

h5py/api_types_hdf5.templ.pxd
ajelenak: Yes. `H5F_LIBVER_LATEST` has become an enum value starting with 2.0. It is a `#define` constant in all the prior versions.

h5py/h5t.templ.pyx
neutrinoceros: I would just go with `complex_types`, but for sure, the comment helps already, it's a *really* minor thing that it costs me a double take, but I can get used to it, especially if it matches a pre-existing convention.

takluyver: Yeah, I'd agree, `ctypes` is a familiar enough name in Python that it's worth using something else.

takluyver:
I think this looks broadly reasonable. We should probably have a test that complex numbers saved by h5py in HDF5 1.x - i.e. compound dtypes with fields named `r` and `i` - are still read in as complex numbers, so accessing existing files doesn't change. I'm fairly sure this will work, but I'd like to make sure it's tested.

I wonder if we should provide APIs either to conveniently save complex numbers the old way, for compatibility with code still using HDF5 1.x, or to opt out of the translation of `r` & `i` to complex numbers, reading them instead as numpy compound dtypes. That can be done separately if we do decide to, I'm just thinking aloud.

Do you know what the status of HDF5 2.0 is? Will there be betas? I'm a bit wary of merging things we can't yet test in CI. And I'd rather not set up our regular CI to build HDF5 from git, because then new failures can crop up at any time regardless of changes here. Though we should maybe have a nightly job to test with HDF5 master.

neutrinoceros:
> I wonder if we should provide APIs either to conveniently save complex numbers the old way, for compatibility with code still using HDF5 1.x, or to opt out of the translation of r & i to complex numbers, reading them instead as numpy compound dtypes.

It's a double edge sword, in my opinion; it sure sounds convenient to have such an API, but it'll beg the question of which behavior should be the default: maximize backward compatibility, or abide to the "current" format, as defined by the version of HDF5 h5py was compiled against ? I think each could be argued as the natural continuation of the *existing* behavior, and clearly we cannot add new *required* arguments to existing functions, because that'd be a major breaking change... I think overall I would lean towards "not worth it"; users who are somehow bound to HDF5 1 should be able to re-compile h5py where needed.
Interestingly, this would be a cool application to [wheel variants](https://astral.sh/blog/wheel-variants), but I don't anticipate they'll get standardized any time soon.

> Though we should maybe have a nightly job to test with HDF5 master.

sounds like a good idea to me.

takluyver:
The default is also a good question. :+1: I wonder if we should require opt-in to native complex storage for now, and plan for h5py 4.0 to switch the default. It could also switch to reading string data as numpy 2.x `T` strings by default - we said that had to stay backwards compatible when it was added.

CC @tacaswell

h5py/h5t.templ.pyx
ajelenak: Changed to `_get_available_cmplx_types`.

ajelenak:
> We should probably have a test that complex numbers saved by h5py in HDF5 1.x - i.e. compound dtypes with fields named `r` and `i` - are still read in as complex numbers, so accessing existing files doesn't change. I'm fairly sure this will work, but I'd like to make sure it's tested.

I've seen a few complex number dtype tests that passed on my laptop for both HDF5 1.14.6 and new 2.0, so I think this is tested at a basic level. Perhaps there could be a specific test file like I created for native complex dtypes but for the complex compound dtypes.

> I wonder if we should provide APIs either to conveniently save complex numbers the old way, for compatibility with code still using HDF5 1.x, or to opt out of the translation of `r` & `i` to complex numbers, reading them instead as numpy compound dtypes. That can be done separately if we do decide to, I'm just thinking aloud.

I'd wait with this a little to see if there are any user requests.

> Do you know what the status of HDF5 2.0 is? Will there be betas?

The plan is still to release in a couple of weeks. I think there will be pre-releases soon.

> I'm a bit wary of merging things we can't yet test in CI. And I'd rather not set up our regular CI to build HDF5 from git, because then new failures can crop up at any time regardless of changes here. Though we should maybe have a nightly job to test with HDF5 master.

HDFG is doing it with the h5py master and HDF5 develop branches: https://github.com/HDFGroup/hdf5/actions/workflows/h5py.yml.

ajelenak:
I've added a test that uses an HDF5 file where complex numbers are stored as two-filed compound datatype.

takluyver:
Sorry that I've left this for a couple of months.

On reflection, I would like h5py to do what I suggested in a previous comment: support reading & writing with the new HDF5 complex datatype, but not yet change the default behaviour when creating a new dataset from numpy complex data. Then at some point before too long, we will change the default in h5py 4.0. I think that we should treat deliberately changing what is written in files by already-working code as a breaking API change, especially when that means the resulting files can't be read by older versions of HDF5.

I'm going to open a tracking issue for h5py 4.0.

takluyver:
I've updated this in the direction I think we want for h5py 3.x, but I think we'll need some conversion machinery to enable writing to the new native complex dtype within this scheme, and I didn't have time to tackle that yet, so one of the tests is failing.

h5py/_conv.templ.pyx
```diff
@@ -906,6 +906,101 @@ cdef herr_t uint2bitfield(hid_t src_id, hid_t dst_id, H5T_cdata_t *cdata,
...
+            and H5Tget_size(H5Tget_member_type(tid, 0)) == member_size
+            and H5Tget_size(H5Tget_member_type(tid, 1)) == member_size
+            and H5Tget_member_offset(tid, 0) == 0
+            and H5Tget_member_offset(tid, 1) == <int>member_size
+    ):
+        return 1
+    return 0
+
+### {{if HDF5_VERSION >= (2, 0, 0)}}
+cdef herr_t complex2compound(hid_t src_id,
```
jhendersonHDF, line 932: Hi @takluyver,

I see that you added these compatibility no-op conversions for handling the case between the compound format and the native format. I added similar to HDF5 when implementing support for native complex types and tested them with h5py to be sure "old format" data could be read correctly, but I may have missed something. Did you happen to have trouble with the built-in conversion routines? See https://github.com/HDFGroup/hdf5/blob/develop/src/H5Tconv_complex.c#L1275 and more specifically https://github.com/HDFGroup/hdf5/blob/develop/src/H5Tconv_complex.c#L1318-L1397. I also added support for the case where the type is an array of 2 floating point types.

takluyver: Thank-you, and well spotted that I had duplicated this.

There's one tiny catch, which I'm trying to work out how much we should care about. You check for field names which are any prefix of 'real' & 'imaginary'. h5py allows people to customise the names that it uses & recognises for these fields, via `h5py.get_config().complex_names = ('foo', 'bar')`. So in theory, someone could be using totally different field names. My conversion routine can check against h5py's config to cope with that, whereas the one built into HDF5 obviously can't.

I was about to say that this is almost certainly entirely theoretical, because even if a handful of people customise it, all the likely names are covered. Then I did a Github search, and it appears that someone has used capital letters: `'Re', 'Im'`. There are a few different versions of the Crikit2 package floating around, here's [one clear example](https://github.com/HSIYJND/CRIkit2/blob/4378b2fe27fdb182c623acd74ee98afd4b0a71b0/crikit/utils/h5.py#L22), and here's the [authoritative version with an extra layer of indirection](https://github.com/CCampJr/CRIkit2/blob/fd4aecdd467c3d2a6544ae1c6c308ad90497ce86/crikit/io/lazy5/create.py#L9).

So, we could get rid of our own copy of the conversion routines, at a cost of breaking one moderately popular package which hasn't been updated in a few years (and *maybe* some other code that I haven't found). :thinking:

jhendersonHDF: > My conversion routine can check against h5py's config to cope with that, whereas the one built into HDF5 obviously can't.

This makes sense to me as a reason for custom conversion functions. I wonder if there's anything we can do in HDF5 in the future to track the associations?

> it appears that someone has used capital letters: 'Re', 'Im'

This is a good point and I had intended to account for this, but had used `strncmp` in the comparisons. Perhaps it would be better to use `strncasecmp` instead.

takluyver: The other thing that I wondered about after writing this code was whether we should ignore the names entirely,  allowing it to convert complex numbers to & from any compound datatype of two suitable floats. The conversion machinery is separate from how h5py picks a type if you create a new dataset or read without specifying a datatype, so the only drawback of being less specific is that some specific operations might confusingly *work* when an error might be more informative.

takluyver: I decided I would get rid of the field names check, because then in h5py 4.0 we can get rid of the config option for field names as well (#2756), and it will just work. It seems unlikely that someone would set up a compound datatype of exactly 2 floats of the right size and try to use it with complex numbers if they're not expecting it to work.

I've just realised I've missed checking endianness, though.

takluyver:
Close & reopen to re-run with master, which should test on HDF5 2.0

takluyver:
I'd like to ask one of the other maintainers to take a look - I've made enough changes in this PR myself that I'm no longer really a reviewer. It would be good to get this in for 3.16.

neutrinoceros:
Happy to give this a look over the weekend

takluyver:
Thanks!

h5py/tests/test_native_complex.py
```diff
@@ -0,0 +1,141 @@
...
+import numpy as np
+import pytest
+import h5py
+from h5py.h5 import get_config  # type: ignore
+from h5py import h5t
+from .data_files import get_data_file_path
+
+cfg = get_config()
+
+pytestmark = [
```
neutrinoceros, line 27: is this some left over from a previous iteration ? it doesn't seem to be used

h5py/tests/test_native_complex.py
```diff
@@ -0,0 +1,141 @@
...
+        (h5t.COMPLEX_IEEE_F64LE, 16, h5t.ORDER_LE),
+        (h5t.COMPLEX_IEEE_F64BE, 16, h5t.ORDER_BE),
+        (h5t.NATIVE_FLOAT_COMPLEX, 8, sys_bo),
+        (h5t.NATIVE_DOUBLE_COMPLEX, 16, sys_bo),
+    ]
+    if hasattr(np, "complex256"):
+        cases.append((h5t.NATIVE_LDOUBLE_COMPLEX, 32, sys_bo))
+    else:
+        cases.append((h5t.NATIVE_LDOUBLE_COMPLEX, 16, sys_bo))
+    for h5type, size, order in cases:
```
neutrinoceros: could we rewrite this as a parametrized test ? happy to do it myself if desired

h5py/tests/test_native_complex.py
```diff
@@ -0,0 +1,141 @@
...
+        np.testing.assert_array_equal(ds[...], complex_array)
+        np.testing.assert_array_equal(ds.attrs["c"], c)
+
+
+def test_dtype_cmpnd_cmplx():
+    """Check the resulting dtype of the two-field compound data as complex numbers"""
+    with h5py.File(get_data_file_path("compound-dtype-complex.h5"), mode="r") as f:
+        for obj in f.values():
+            if isinstance(obj, h5py.Dataset):
+                check_compound_complex_datatype(obj.id.get_type(), obj.dtype)
```
neutrinoceros:
```suggestion
        for obj in filter(lambda obj: isinstance(obj, h5py.Dataset), f.values()):
            check_compound_complex_datatype(obj.id.get_type(), obj.dtype)
```

h5py/tests/test_native_complex.py
```diff
@@ -0,0 +1,141 @@
...
+        (h5t.COMPLEX_IEEE_F32BE, np.dtype(">c8")),
+        (h5t.COMPLEX_IEEE_F64LE, np.dtype("<c16")),
+        (h5t.COMPLEX_IEEE_F64BE, np.dtype(">c16")),
+        (h5t.NATIVE_FLOAT_COMPLEX, np.dtype("=c8")),
+        (h5t.NATIVE_DOUBLE_COMPLEX, np.dtype("=c16")),
+    ]
+    if hasattr(np, "complex256"):
+        cases.append((h5t.NATIVE_LDOUBLE_COMPLEX, np.dtype("=c32")))
+    for h5type, dt in cases:
+        assert h5type.dtype == dt
```
neutrinoceros: Was going to suggest parametrization here too but then I realized the way to do it might not be obvious, so here's a complete suggestion
```suggestion
@pytest.mark.parametrize(
    "h5type, dt",
    [
        pytest.param(h5t.COMPLEX_IEEE_F32LE, "<c8", id="<c8"),
        pytest.param(h5t.COMPLEX_IEEE_F32BE, ">c8", id=">c8"),
        pytest.param(h5t.COMPLEX_IEEE_F64LE, "<c16", id="<c16"),
        pytest.param(h5t.COMPLEX_IEEE_F64BE, ">c16", id=">c16"),
        pytest.param(h5t.NATIVE_FLOAT_COMPLEX, "=c8", id="=c8"),
        pytest.param(h5t.NATIVE_DOUBLE_COMPLEX, "=c16", id="=c16"),
        pytest.param(
            h5t.NATIVE_LDOUBLE_COMPLEX, "=c32", id="=c32"),
            marks=pytest.mark.skipif(not hasattr(np, "complex256"),
        ),
    ],
)
def test_cmplx_type_trnslt(h5type, dt):
    """Translate native HDF5 complex number datatype to NumPy dtype"""
    assert h5type.dtype == np.dtype(dt)
```

h5py/tests/test_native_complex.py
```diff
@@ -0,0 +1,141 @@
...
+    if np_dtype.byteorder == ">" or (
+            sys.byteorder == "big" and np_dtype.byteorder == "="
+    ):
+        assert datatype.get_member_type(0).get_order() == h5t.ORDER_BE
+        assert datatype.get_member_type(1).get_order() == h5t.ORDER_BE
+    elif np_dtype.byteorder == "<" or (
+            sys.byteorder == "little" and np_dtype.byteorder == "="
+    ):
+        assert datatype.get_member_type(0).get_order() == h5t.ORDER_LE
+        assert datatype.get_member_type(1).get_order() == h5t.ORDER_LE
```
neutrinoceros: making the test less verbose by avoiding repetitions and ensuring that it cannot silently pass if broken
```suggestion
    match (np_dtype.byteorder, sys.byteorder):
        case (">", _) | ("=", "big"):
            expected = h5t.ORDER_BE
        case ("<", _) | ("=", "small"):
            expected = h5t.ORDER_LE
        case _ as _unreachable:
            raise AssertionError

    assert datatype.get_member_type(0).get_order() == expected
    assert datatype.get_member_type(1).get_order() == expected
```

h5py/tests/test_native_complex.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+def test_default_create(writable_file):
+    """Test default translation when creating datasets & attributes"""
+    for dt in ("<c8", ">c8", "<c16", ">c16"):
```
neutrinoceros: this should be trivial to parametrize

h5py/tests/test_native_complex.py
```diff
@@ -0,0 +1,141 @@
...
+        ds.attrs["c"] = c
+        check_compound_complex_datatype(ds.id.get_type(), np.dtype(dt))
+        check_compound_complex_datatype(ds.attrs.get_id("c").get_type(), np.dtype(dt))
+        np.testing.assert_array_equal(ds[...], complex_array)
+        np.testing.assert_array_equal(ds.attrs["c"], c)
+
+
+def test_explicit_create(writable_file):
+    """Explicitly use native complex datatype to create datasets & attributes"""
+    for h5_dt, np_dt_s in [
```
neutrinoceros: this too

h5py/h5.templ.pyx
```diff
@@ -61,10 +61,22 @@ cdef class H5PYConfig:
...
         self._track_order = False
+        ### {{if HDF5_VERSION >= (2, 0, 0)}}
+        self._native_complex = bool(
+            H5_SIZEOF_FLOAT_COMPLEX + H5_SIZEOF_DOUBLE_COMPLEX + H5_SIZEOF_LONG_DOUBLE_COMPLEX)
+        ### {{else}}
+        self._native_complex = False
+        ### {{endif}}
+
+    @property
+    def native_complex(self):
```
neutrinoceros: I think this would align better with our existing naming conventions
```suggestion
    def has_native_complex(self):
```

h5py/_conv.templ.pyx
```diff
@@ -906,6 +906,101 @@ cdef herr_t uint2bitfield(hid_t src_id, hid_t dst_id, H5T_cdata_t *cdata,
     return 0
 
 # =============================================================================
+# Complex to Compound routines
+#
+#   these are no-ops, but needed for interoperability between the native complex
+#   datatype (HDF5 2.0+) and h5py's older imitation of complex data as a compound
+#   datatype. In h5py 3.x, numpy complex dtypes are presented to HDF5 as compound
+#   to preserve existing behaviour. We expect to switch this in h5py 4.0.
```
neutrinoceros, line 913: I'm not entirely sure I understand what would break if we made the switch now. Is it reproducibility with HDF5 1.x VS 2.x ? is the plan to require HDF5 2.x in h5py 4 ?

neutrinoceros (commented):
I have a couple questions + suggestions to improve independence between test cases (so each test corresponds to a single failure mode). Happy to perform my suggested changes myself if you'd like ! I also want to re-review _conv.templ.pyx once my questions are answered.

h5py/tests/test_native_complex.py
ajelenak: I don't remember any more exactly why but I tried a parametrized test first and then it failed to work for one of the cases I added so reverted to this approach. Perhaps pytest improved in the meantime.

neutrinoceros: let me know if you mind me pushing to your branch or not :)

ajelenak: Go ahead. No problem.

neutrinoceros: thanks. Done !

h5py/tests/test_native_complex.py
neutrinoceros: resolved: I didn't know about this, but `pytest` will look for this exact variable name and apply markers to the entire module. Neat.

h5py/_conv.templ.pyx
takluyver: Switching would mean changing what HDF5 is produced if you create a dataset or an attribute from a numpy array of complex numbers. It's debatable whether that counts as a breaking change, but as we were starting to think about a 4.0 version with breaking changes anyway, and the complex dtype would presumably be unreadable in HDF5 1.x, my call was that we should delay changing the default.

> is the plan to require HDF5 2.x in h5py 4 ?

Good question. I lean towards no, but maybe we can pick that up on #2756.

h5py/tests/test_native_complex.py
```diff
@@ -0,0 +1,146 @@
...
+
+cfg = get_config()
+
+pytestmark = [
+    pytest.mark.skipif(
+        h5py.version.hdf5_version_tuple < (2, 0, 0),
+        reason="Requires HDF5 >= 2.0",
+    ),
+    pytest.mark.skipif(
+        not cfg.native_complex,
```
takluyver:
```suggestion
        not cfg.has_native_complex,
```

tacaswell:
I would consider changing the type/structure of a data set to be a breaking API change (e.g. someone has a Python process that writes the files and a c++ program that consumes them and/or having downstream tools that are not yet rebuilt with hdf5 2). 

I only quickly skimmed the technical details, but it is not clear to me if we also have a way to at dataset creation time to force the current structured encoding.  Ensuring we have that and documenting it are critical steps to being able to switch the default eventually (so there are some pre-default-change versions of h5py available when people can start forcing the old default before they _must_ force the old default).

takluyver:
Good point, thanks.

You *can* force the current structure by explicitly making a compound dtype with two float fields called `r` & `i`, but it's not obvious or documented how to do that. I propose we add & document a function to do that, in a similar style to things like `string_dtype()` and `enum_dtype()`.

takluyver:
I have added a function to create the compatible compound datatype for complex numbers, as a numpy dtype object. You can use it like this:

```python
f.create_dataset('c', shape=(100,), dtype=h5py.complex_compat_dtype('<c8'))
```

I deliberately didn't use the globally configured field names in this, as I'd like to move away from global mutable state where possible. You can pass `names=` to the function if you need names different from the default.

I'm thinking about making a corresponding `check_...` function like we have for other special dtypes. But I'm not sure yet if it should be specifically for the compatibility type, or recognise the native complex dtype as well. It's a bit more complicated because making a dataset with the compatibility dtype will still give a Numpy complex dtype in `dset.dtype`.

takluyver:
Now I've also added `check_complex_dtype()`, which checks for either numpy complex dtypes, or the dtypes returned from `complex_compat_dtype()`, and returns the equivalent numpy complex dtype for a match.

tacaswell (approved):

tacaswell:
I reviewed @takluyver 's commit at the end related to back-compatibility, have not reviewed the cython changes in detail so not merging.

takluyver:
I think this has had sufficient review, I'll merge it tomorrow if there are no objections.

The failing Windows build is just a failure to upload to codecov. I figure I won't waste time and energy rerunning it just to get a green tick.

takluyver:
Thanks everyone!

PR #2695, opened 2025-09-18
TST/MNT: skip caching HDF5 on Linux CI

neutrinoceros:
As noted by @takluyver in #2646, we don't actually install HDF5 on Linux CI because it's already present in our custum manylinux/musllinux images, so we can also skip caching.


takluyver:
Thanks!

PR #2696, opened 2025-09-21
SEC: fix security issues with GHA detected with zizmor

neutrinoceros:
In #2685, @takluyver suggested that we should be even more cautious on security if we're going to automate most of the release process with GHA using PyPI'ss trusted publishing. Here I'm setting up `zizmor` as a pre-commit hook for continuous auditing and using it to resolve existing issues in this area, including but not limited to, excessively permissive permission settings.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -43,6 +43,8 @@ jobs:
     steps:
     - name: Checkout repo
       uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      with:
+        persist-credentials: false
     - uses: actions/setup-python@e797f83bcb11b83ae66e0230d6156d7c80228e7c # v6.0.0
       with:
         python-version: 3.13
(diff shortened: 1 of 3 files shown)


tacaswell:
Thank you!

Doing this on h5py was someplace in my to-do backlog.

tacaswell:
This added a bug ~where the check on "is this a tag" seems to be too permissive~: https://github.com/h5py/h5py/actions/runs/17894803441/job/50880501354 working on a fix.

tacaswell:
It added a problem, but I miss-identified it.

I had not realized we (re)create a draft GH release on every push.  The issue is instead that we don't give this job enough power.

.github/workflows/build_wheels.yml
```diff
@@ -90,6 +92,7 @@ jobs:
         with:
           key: ${{ matrix.os }}-${{ matrix.arch }}-6
           path: cache/hdf5
+          lookup-only: true
```
takluyver, line 95: I think this has made the caching useless - Windows & Mac builds (e.g. [this one](https://github.com/h5py/h5py/actions/runs/17918521517/job/50947020448) are now building HDF5 from source every time, with the cache action saying:

```
Cache hit for: windows-latest-AMD64-6
Lookup only - skipping download
Cache found and can be restored from key: windows-latest-AMD64-6
```

takluyver: I've opened #2699 to hopefully restore the caching.

neutrinoceros: thanks !

PR #2697, opened 2025-09-22
CI: give the upload artifacts job more power

tacaswell:
Fixes a bug introduced in #2696 which over-restricted permission causing failures: 
https://github.com/h5py/h5py/actions/runs/17894803441/job/50880501354

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -161,6 +161,8 @@ jobs:
 
   upload_artifacts:
     name: Upload build artifacts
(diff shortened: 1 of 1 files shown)


tacaswell:
There is no easy way to test that this is fixed other than merging (I guess I could change the code owner checks and test it on my fork...)

neutrinoceros:
LGTM, thanks for catching it !

neutrinoceros:
it didn't fix it. I tried again in #2698

PR #2698, opened 2025-09-22
CI: fix permission settings for wheel upload again

neutrinoceros:
Follow up to #2696 and #2697
I hope this time we got it right
doc: https://github.com/softprops/action-gh-release#permissions

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -162,7 +162,7 @@ jobs:
   upload_artifacts:
     name: Upload build artifacts
     permissions:
-      actions: write
+      contents: write
     needs:
       - build_sdist
       - build_wheels


takluyver:
Thanks, this looks like it should work.

Clearly in this case it was useful for catching the issue, but do we think there's a good reason to (re)create a Github draft release on each push to master? The wheels produced from CI are available for download anyway as 'artifacts' from Github actions, and this also works for PRs. And then wheels are uploaded to anaconda.org every night for a more convenient way to install a pre-release. Maybe we should drop the bit publishing to a `dev` release on GH? CC @tacaswell

neutrinoceros:
IIRC @larsoner implemented it, so maybe he can provide a use case not covered by nighties ?

tacaswell:
It is also private and only people in the h5py org can see it (look in an incognito browser).

I am also not sure of the value of it, but I very firmly believe in not removing things I do not understand why exist!

larsoner:
> Clearly in this case it was useful for catching the issue, but do we think there's a good reason to (re)create a Github draft release on each push to master?

Probably not -- might be better to make it something you could trigger from the GHA interface instead (`worflow_dispatch` triggered). I think I've used it in other repos to notice when a release wasn't going to work before actually clicking the release button. But if you don't use it in your workflows you could get rid of it

takluyver:
Thanks! So far I don't think we've used it. I can see the advantage of having a way to test the release process, but it's also valuable to have the CI steps for a release be as concise and readable as practical, and cutting out the draft release part, as I've tried in #2699, does make it a fair bit simpler.

PR #2699, opened 2025-09-22
Restore caching of HDF5 builds in Github actions

takluyver:
See comment on #2696.

At present, this also brings back the warning from zizmor about the potential for cache poisoning (albeit with "audit confidence → Low"). If we agree that this is sufficiently protected, I'll add a [comment to ignore this warning](https://docs.zizmor.sh/usage/#ignoring-results).

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -163,6 +163,10 @@ jobs:
 
   upload_artifacts:
     name: Upload build artifacts
+    if: ${{ github.repository_owner == 'h5py' && (
+                (github.event_name == 'push' && github.ref_type == 'tag')
+                || (github.event_name == 'schedule'))
+      }}
     permissions:
       contents: write
     needs:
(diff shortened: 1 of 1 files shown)


tacaswell (approved):
👍🏻 on adding a comment to accept the mitigation

.github/workflows/build_wheels.yml
```diff
@@ -185,40 +191,22 @@ jobs:
...
         with:
-          name: ${{ steps.vars.outputs.release_name }}
           # Note that draft releases are not public
-          draft: ${{ github.ref_type != 'tag' }}
-          prerelease: ${{ github.ref_type != 'tag' }}
-          generate_release_notes: ${{ steps.vars.outputs.generate }}
+          generate_release_notes: false
           files: |
-            dist/*.whl
+            dist/*
```
neutrinoceros, line 211: nice catch !

PR #2701, opened 2025-09-24
Enable use of Github PAT for Python download on Azure

takluyver:
Follow up to #2650. Hopefully I've set up the token stuff correctly.

### DIFF
--- a/ci/azure-pipelines-steps.yml
+++ b/ci/azure-pipelines-steps.yml
@@ -21,7 +21,7 @@ steps:
     versionSpec: '$(python.version)'
     architecture: '$(python.architecture)'
     allowUnstable: true
-    # githubToken: '$(githubToken)'
+    githubToken: '$(githubToken)'
   displayName: 'Use Python $(python.version) $(python.architecture)'
   condition: and(succeeded(), ne(variables['python.version'], 'Pre'))
 


takluyver:
> Failed to download Python versions manifest from the Github Actions python registry (https://github.com/actions/python-versions). Invalid GitHub token could cause this.

:face_exhaling:

takluyver:
I've got this working now. The `githubToken` needed to be supplied as a [secret variable](https://learn.microsoft.com/en-gb/azure/devops/pipelines/process/set-secret-variables?view=azure-devops&tabs=yaml%2Cbash) in the pipeline config, as in [this SO answer](https://stackoverflow.com/a/79046161/434217), instead of a service connection in the project config.

I've created a token that's valid for 1 year (i.e. until September 2026), which is the maximum allowed. Hopefully by the time that expires the relevant Python builds will be readily available on Azure anyway, and we won't need another one. If we do need to replace it, I made a 'classic' PAT with scopes `public_repo, read:user, user:email`, i.e. it's read only and most of what it can read is public anyway, so we don't have to be especially paranoid about it.

neutrinoceros:
> I've created a token that's valid for 1 year (i.e. until September 2026), which is the maximum allowed. Hopefully by the time that expires the relevant Python builds will be readily available on Azure anyway, and we won't need another one. 

awesome. I'd suggest adding this information as a comment in the vicinity of the line you edited, which might make it easier to debug if it fails in a year and the first person to see it forgot or never knew what to expect !

takluyver:
Good point, thanks!

takluyver:
(I accidentally pushed the branch to the h5py repo instead of my fork, so all the CI is running twice and slowing everything down. Mea culpa.)

ci/azure-pipelines-steps.yml
```diff
@@ -21,7 +21,9 @@ steps:
     versionSpec: '$(python.version)'
     architecture: '$(python.architecture)'
     allowUnstable: true
-    # githubToken: '$(githubToken)'
+    # GH 'classic' PAT token, set as a secret variable in Azure pipeline config.
+    # Max validity 1 year. Used for Python builds not already installed on Azure.
```
neutrinoceros, line 25: it's true that `git blame` fills in the gaps for now, but I'd prefer to have the date-of-last-update included in the comment itself

takluyver: Fair enough

PR #2702, opened 2025-09-24
Enable cp314t wheels

crusaderky:
- Tracker: #2475
- Follow-up to #2650

### DIFF
--- a/setup_build.py
+++ b/setup_build.py
@@ -201,7 +201,10 @@ class h5py_build_ext(build_ext):
         }
         compiler_directives = {}
         if Version(cython_version) >= Version("3.1.0b1"):
-            compiler_directives["freethreading_compatible"] = True
+            # TODO it was decided to not mark the module as freethreading-compatible
+            # for the time being due to instability concerns.
+            # See https://github.com/h5py/h5py/pull/2650
+            compiler_directives["freethreading_compatible"] = False
 
         # Run Cython
         print("Executing cythonize()")
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -90,6 +90,11 @@ jobs:
         HDF5_DIR: $(HDF5_CACHE_DIR)/$(HDF5_VERSION)
         H5PY_ENFORCE_COVERAGE: yes
 
+        # TODO it was decided to not mark the module as freethreading-compatible
+        # for the time being due to instability concerns.
+        # See https://github.com/h5py/h5py/pull/2650
+        PYTHON_GIL: 0
+
     maxParallel: 4
 
   steps:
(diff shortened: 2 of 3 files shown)


neutrinoceros:
Thanks ! I'd suggest drafting this PR for now.
Genuine question: is there a known application for 3.13t, waiting for these wheels in the wild, that cannot sustain having *just* 3.14t wheels available ?

crusaderky:
> Thanks ! I'd suggest drafting this PR for now. Genuine question: is there a known application for 3.13t, waiting for these wheels in the wild, that cannot sustain having _just_ 3.14t wheels available ?

I can't answer with real examples from the top of my head, but if there is any package down the line that does not ship 3.14t for any reason (namely, because it's still in rc) then it's going to be a problem.

Also, any dependent package that DOES want to ship 3.13t will face CI pain if there are no wheels available.
I don't see any reason _not_ to build for 3.13t?

ngoldbaum:
I think if h5py only wants to build cp314t wheels that's fine at this point. Several important projects decided to only support free-threading on 3.14 and newer.

That said, there isn't much extra marginal cost to upload cp313t wheels, and doing so would reduce build times and setup pain for people who can't upgrade yet for whatever reason.

neutrinoceros:
> I don't see any reason not to build for 3.13t?

> there isn't much extra marginal cost to upload cp313t wheels

I'd like to keep the number of wheels we ship down to a reasonable amount (to be defined), because storage has a cost (albeit an externality). I see no issue with adding wheels for 3.13t when and *if* we get actual demand, but I also don't see much gain from adding them now.

pyproject.toml
```diff
@@ -125,9 +126,12 @@ wheels = [
 
 [tool.cibuildwheel]
 skip = [
-    "cp314t-*",  # Disabled due to instability concerns.
     "cp310-win_arm64", # Skipped due to unavailability of Python binary and NumPy for Win-ARM64
 ]
+enable = [
+    "cpython-freethreading",  # Build cp313t wheels
+]
```
ngoldbaum: Let's delete this, per https://github.com/h5py/h5py/pull/2702#issuecomment-3330232792.

setup_build.py
```diff
@@ -198,11 +199,16 @@ def run(self):
             "OBJECTS_DEBUG_ID": False,
             "FREE_THREADING": sysconfig.get_config_var("Py_GIL_DISABLED") == 1,
         }
+        compiler_directives = {}
+        if Version(cython_version) >= Version("3.1.0b1"):
+            compiler_directives["freethreading_compatible"] = True
```
ngoldbaum: If it would speed releasing cp314t wheels on pypi, maybe we can delay setting this for now and just let the GIL be re-enabled?

IMO shipping wheels that re-enable the GIL is preferable to build errors or for downstream testing free-threading needing to build h5py from source.

pyproject.toml
crusaderky:
```suggestion
```

setup_build.py
neutrinoceros: Sounds good to me.

crusaderky:
@ngoldbaum @neutrinoceros done as requested; this should be ready for merge.

neutrinoceros:
Planning to merge is in a couple days if nothing else comes up. Thanks !

PR #2706, opened 2025-10-04
MNT: add PEP 723 inline script metadata

neutrinoceros:
A quick quality of life improvement. This makes installing `requests` explicitly unnecessary if you're running the script via `uv` or `pipx` (and any other PEP 723-compliant script runners).

(note that this is the only such script I found with any third party dependency)


takluyver:
Neat, thanks

PR #2708, opened 2025-10-04
MNT: replace unused Travis CI status badge with a GHA one

neutrinoceros:


takluyver:
It looks like the token I set up in #2701 is not working. :confused: I've just generated a new token and put that in to Azure, but still nothing.

Can you figure anything out about it? If not, maybe we just have to scrap the free-threading test jobs until Python 3.14 is actually released - hopefully at that point it will be available on Azure without having to download it.

neutrinoceros:
I don't grok what's happening, but luckily, 3.14.0 is set for its final release tomorrow, so let's see if things improve by themselves over the course of the week

PR #2718, opened 2025-10-12
MNT/CLN: upgrade cibuildwheel, cleanup temporary override

### LINKED ISSUE #2711: WHL: incorrect default deployment target for cp314 macos wheels

neutrinoceros:
close #2711

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -157,12 +157,6 @@ delocate-wheel --require-archs {delocate_archs} -w {dest_dir} -v {wheel} \
 H5PY_ROS3 = "0"
 H5PY_DIRECT_VFD = "0"
 
-[[tool.cibuildwheel.overrides]]
-# https://github.com/pypa/cibuildwheel/issues/2612
-select = "cp314-macosx_x86_64"
-inherit.environment="append"
-environment = {MACOSX_DEPLOYMENT_TARGET="10.15"}
-
 [tool.uv]
 # never build numpy from source
 no-build-package = ["numpy"]


takluyver:
Thanks!

takluyver:
The Mac ARM build failed when I tagged a release. I'm not sure why it hasn't failed before.

```
  delocate.libsana.DelocationError: Library dependencies do not satisfy target MacOS version 11.0:
  /private/var/folders/y6/nj790rtn62lfktb1sh__79hc0000gn/T/tmpnxddtxxu/wheel/h5py/.dylibs/libhdf5_hl.310.dylib has a minimum target of 14.0
  /private/var/folders/y6/nj790rtn62lfktb1sh__79hc0000gn/T/tmpnxddtxxu/wheel/h5py/.dylibs/libz.1.3.1.dylib has a minimum target of 14.0
  /private/var/folders/y6/nj790rtn62lfktb1sh__79hc0000gn/T/tmpnxddtxxu/wheel/h5py/.dylibs/libhdf5.310.dylib has a minimum target of 14.0
  Set the environment variable 'MACOSX_DEPLOYMENT_TARGET=14.0' to update minimum supported macOS for this wheel.
```

https://github.com/h5py/h5py/actions/runs/18463305725/job/52599443858

I'm going to rerun this job, and see if it's something random, though this doesn't look random. :confused:

takluyver:
Ah, I bet the cache has been hiding an issue. We made it bypass the cache when building a release, to protect against cache poisoning attacks, so the release tag causes it to rebuild HDF5 from source.

So I guess we need to either set `MACOSX_DEPLOYMENT_TARGET` to a lower version when we build HDF5, and then bump the number in the cache key, or override it to a higher value for cibuildwheel?

neutrinoceros:
I think the correct solution is to actually set `MACOSX_DEPLOYMENT_VERSION` to match the runtime version of the image we build against, as suggested by the error message. It should probably invalidate the cache indeed, but the safest way to do that would be to include the contents `pyproject.toml` and CI configs in the hash computation.
I can work out a quick fix to unblock the release.

neutrinoceros:
there you go @takluyver : https://github.com/h5py/h5py/pull/2719

PR #2719, opened 2025-10-13
WHL: match runner images macosx versions in MACOSX_DEPLOYMENT_TARGET

neutrinoceros:
This isn't ideal; and I'd like to spend more time studying if it can be reverted for future releases, but this should unblock h5py 3.15

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -72,9 +72,9 @@ jobs:
             arch: AMD64
           - os: windows-11-arm
             arch: ARM64
-          - os: macos-14
+          - os: macos-14 # keep in sync with cibuildwheel.overrides
             arch: arm64
-          - os: macos-15-intel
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks for the quick fix, hopefully it will work for the release. Here's the build, still in progress: https://github.com/h5py/h5py/actions/runs/18465128941/job/52605204982

For the future, am I right that we can compile HDF5 itself with a lower MACOSX_DEPLOYMENT_TARGET to match what cibuildwheel uses by default, and thus have broader compatibility?

neutrinoceros:
Yes, that should absolutely be possible. I think we should update the cache invalidation rules first. I expect you know more about it than I do. Do you have time for a PR in that area ?

neutrinoceros:
I opened https://github.com/h5py/h5py/issues/2720 to follow up on this

PR #2721, opened 2025-10-13
Include HDF5 build scripts in key for HDF5 cache

takluyver:
This should improve cache invalidation for HDF5 builds.

I've split up the cache step between Windows & Mac, as there are different scripts involved in the build for each platform.


neutrinoceros:
excellent. I was actually working concurrently on my own solution, involving a python script doing essentially the same thing as `hashFiles` (which I just learned about !), but this is much better !

PR #2715, opened 2025-10-10
WHL: workaround broken defaut MACOSX_DEPLOYMENT_TARGET on cp314 wheels

neutrinoceros:
workaround issue #2711 
closing that issue will require reverting this change + upgrading cibuildwheel

PR #2722, opened 2025-10-13
WHL: build HDF5 with backward compatible system APIs on macOS

### LINKED ISSUE #2720: macos: restore support for versions older than 14 (arm64) and 15 (x86_64) respectively
During the release process for h5py 3.15, we encountered an issue with macos wheels (both archs) where `delocate-wheel` flagged dynamic libraries we were bundling (HDF5) as incompatible. The release was allowed to continue with a temporary patch (#2719), but the plan is to revert it for the next release.

We identified two problems that need addressing:
1) when building HDF5 in CI, we should explicitly set `MACOSX_DEPLOYMENT_TARGET` to match the lowest version supported required for our oldest supported CPython (see for reference https://cibuildwheel.pypa.io/en/stable/platforms/#macos-version-compatibility)
1) the cache used in regular CI (not when releasing) isn't being properly invalidated; hash computation should include contents from `pyproject.toml` and *every* CI config files (roughly `.github/workflows/*`) and scripts (`ci/*`).

I think it would be useful to tackle the second point first so we know that we're not hiding another problem when handling the first.

neutrinoceros:
Close #2720

revert #2719 and part of #2715, which I now understand is when I created the problem in the first place. Will need to carefully look at logs before undrafting this (I need to check that cibw isn't affected by `MACOSX_DEPLOYMENT_TARGET` set for `HDF5`).

### DIFF
--- a/ci/get_hdf5_if_needed.sh
+++ b/ci/get_hdf5_if_needed.sh
@@ -51,6 +51,15 @@ else
             ARCH=$(uname -m)
             ZLIB_VERSION="1.3.1"
 
+            # When compiling HDF5, we should use the minimum across all
+            # Python versions for a given arch.
+            # See cibuildwheel.pypa.io/en/stable/platforms#macos-version-compatibility
+            if [[ "$ARCH" == "arm64" ]]; then
+                export MACOSX_DEPLOYMENT_TARGET="11.0"
+            else
+                export MACOSX_DEPLOYMENT_TARGET="10.9"
+            fi
+
             pushd /tmp
             curl -sLO https://zlib.net/fossils/zlib-$ZLIB_VERSION.tar.gz
             tar xzf zlib-$ZLIB_VERSION.tar.gz
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -72,9 +72,9 @@ jobs:
             arch: AMD64
           - os: windows-11-arm
             arch: ARM64
-          - os: macos-14 # keep in sync with cibuildwheel.overrides
+          - os: macos-14
             arch: arm64
-          - os: macos-15-intel # keep in sync with cibuildwheel.overrides
+          - os: macos-15-intel
             arch: x86_64
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -157,16 +157,6 @@ delocate-wheel --require-archs {delocate_archs} -w {dest_dir} -v {wheel} \
 H5PY_ROS3 = "0"
 H5PY_DIRECT_VFD = "0"
 
-[[tool.cibuildwheel.overrides]]
-select = "cp3*-macosx_arm64"
-inherit.environment = "append"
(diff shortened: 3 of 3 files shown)


neutrinoceros:
I think I got it. The real issue was that `MACOSX_DEPLOYMENT_TARGET` was set far from were it's actually needed (building HDF5) and, crucially, never unset.

ci/get_hdf5_if_needed.sh
```diff
@@ -85,6 +94,13 @@ else
         popd
         popd
 
+        if [[ "$OSTYPE" == "darwin"* ]]; then
+            # restore cibuildwheel's defaults,
+            # which vary across Python versions and architectures
+            # and are in general higher than what is used above
+            unset MACOSX_DEPLOYMENT_TARGET
```
takluyver: I don't understand how unsetting this affects anything. This script exits just afterwards. It is sourced from `cibw_before_all_macos.sh`, so its variables could affect that script, but that also exits without building anything else. cibuildwheel runs in a separate step in Github actions, which won't pick up environment variables from this step, unless they're explicitly written to $GITHUB_ENV.

neutrinoceros: Hum, that seems to makes sense, although empirically this is the only setup that _didn't_ leak this variable...
Maybe shells leak exports differently on macOS ??

neutrinoceros:
I pushed a new commit to comment out the dubious part while I research "`export` on macOS"

neutrinoceros:
Screenshot 2025-10-14 at 13 22 34

well, that explains it; `unset` is indeed not needed. I guess I was confused and probably changed more than one thing at once at some point... any way, I can clean it up now. Thanks @takluyver for not trusting my word on this haha

takluyver:
Thanks! :+1:

bruAristimunha:
hey @takluyver and @neutrinoceros!

I was wondering, could you please release h5py? It is affecting mne_bids that have pymatreader as a dependency, which in turn have h5py as a dependency—a long list of links that is blocking the CI. 😭 

https://github.com/mne-tools/mne-bids/pull/1450

neutrinoceros:
@bruAristimunha it's failing to build on your CI but you have the option of disabling installing from an sdist with `pip install h5py --only-binary h5py`.

bruAristimunha:
Hey @neutrinoceros ,

Thank you so much for the answer🙏 

The tricky part is that it comes from an intermediate dependency, pymatreader. 

I think it would be much better to resolve it in the source, but I can ask them to do the release using this instructions.

Since hyp5 is heavily used in many libraries, I imagine I'm just one of the first to ask for a release here.

neutrinoceros:
There are other problems caused by h5py 3.15
We are currently debating how to best resolve them, and a new release is indeed one of the options.

takluyver:
Yeah, I think there's a good chance we do another release soon, but we want to understand the problems and the solution first - there's a discussion happening on #2726.

In the meantime, you can still restrict the version of a transitive dependency in a couple of ways:

- The simple way: install a different version of h5py before installing everything else - pip should see that it's already installed and use that version
- The fancy way: Create a [constraints file](https://pip.pypa.io/en/stable/user_guide/#constraints-files) specifying `h5py !=3.15.0` and point pip at that when installing your package, to tell it to pick a different version of h5py.

PR #2725, opened 2025-10-14
TST: File hygiene in test_lowlevel_vds

### LINKED ISSUE #2714: Can't open the same file in read-only multiple times from multiple threads
`TestPercivalLowLevel` creates a File in setUp, closes it, and then reopens the same file in the `test` function in read-only mode, assigning the File object to a local variable.
The test fails when it is run in pytest-run-parallel (#2655), which runs setUp in a single thread and the `test` functions in many threads. Copying the .h5 file on disk, so that each thread opens a separate copy, and then opening the copy works around the issue.
I cannot reproduce the issue when running on a single thread.

crusaderky:
- Closes #2714
- Blocks #2655
- Don't open File for read while the same file on disk is already opened for write
- Review File open/close hygiene and use `with h5.File` when sensible


takluyver:
Thanks!

PR #2728, opened 2025-10-19
RFC: avoid direct mutations to ndarray.shape attributes

neutrinoceros:
xref: https://github.com/numpy/numpy/pull/29536

this is the only block in the whole code base using this soon-to-be deprecated pattern


takluyver:
Thanks!

PR #2729, opened 2025-10-20
fix selecting full slices from zero-sized datasets and zero-size slices from sized datasets

### LINKED ISSUE #2549: Writing to zero-size resizable dataset doesn't raise
Over in https://github.com/h5netcdf/h5netcdf/issues/249 we are trying to get behind some unexpected behaviour.

Consider the following two examples. The first one does not raise and silently finishes:

```python
import h5py
import numpy as np
with h5py.File("test.h5", mode="w") as fh:
    dset = fh.create_dataset("resizable", (0,), maxshape=(None,))
    dset[:] = np.ones((4))
!h5dump test.h5
```
```
HDF5 "test.h5" {
GROUP "/" {
   DATASET "resizable" {
      DATATYPE  H5T_IEEE_F32LE
      DATASPACE  SIMPLE { ( 0 ) / ( H5S_UNLIMITED ) }
      DATA {
      }
   }
}
}
```

The second example where we set the dataset length to some number, raises instead:

```python
import h5py
import numpy as np
with h5py.File("test.h5", mode="w") as fh:
    dset = fh.create_dataset("resizable", (1,), maxshape=(None,))
    dset[:] = np.ones((4))
!h5dump test.h5
```
```python
Cell In[14], line 5
      3 with h5py.File("test.h5", mode="w") as fh:
      4     dset = fh.create_dataset("resizable", (1,), maxshape=(None,))
----> 5     dset[:] = np.ones((4))
      6 get_ipython().system('h5dump test.h5')

File h5py/_objects.pyx:54, in h5py._objects.with_phil.wrapper()

File h5py/_objects.pyx:55, in h5py._objects.with_phil.wrapper()

File [/home/kai/data/mambaforge/envs/xr_312_np2/lib/python3.12/site-packages/h5py/_hl/dataset.py:1020](http://localhost:8888/lab/tree/home/kai/python/gists/h5netcdf/home/kai/data/mambaforge/envs/xr_312_np2/lib/python3.12/site-packages/h5py/_hl/dataset.py#line=1019), in Dataset.__setitem__(self, args, val)
   1017     mshape = val.shape
   1019 # Perform the write, with broadcasting
-> 1020 mspace = h5s.create_simple(selection.expand_shape(mshape))
   1021 for fspace in selection.broadcast(mshape):
   1022     self.id.write(mspace, fspace, val, mtype, dxpl=self._dxpl)

File [/home/kai/data/mambaforge/envs/xr_312_np2/lib/python3.12/site-packages/h5py/_hl/selections.py:264](http://localhost:8888/lab/tree/home/kai/python/gists/h5netcdf/home/kai/data/mambaforge/envs/xr_312_np2/lib/python3.12/site-packages/h5py/_hl/selections.py#line=263), in SimpleSelection.expand_shape(self, source_shape)
    262             eshape.append(t)
    263         else:
--> 264             raise TypeError("Can't broadcast %s -> %s" % (source_shape, self.array_shape))  # array shape
    266 if any([n > 1 for n in remaining_src_dims]):
    267     # All dimensions from target_shape should either have been popped
    268     # to match the selection shape, or be 1.
    269     raise TypeError("Can't broadcast %s -> %s" % (source_shape, self.array_shape))  # array shape

TypeError: Can't broadcast (4,) -> (1,)
```

Is this intended? I would have expected for the first example to raise a similar error, "writing to a zero-size dimension".

The location where this is silently returned is here:

https://github.com/h5py/h5py/blob/c02a7e70c55733746948d73b1ccc66a093ef9e01/h5py/_hl/dataset.py#L987-L992

Maybe just checking for empty dimensions in `self.shape` and raise an informative error would help here?

kmuehlbauer:
- closes #2549

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -1061,11 +1061,11 @@ class Dataset(HLObject):
         # Perform the dataspace selection
         selection = sel.select(self.shape, args, dataset=self)
         # return early on zero-size selection
-        # but not if dataset shape and selection shape are equal
+        # but only if dataset shape and selection shape differ
         # this prevents returning here for zero size datasets
         # where proper errors should be raised later
         # see https://github.com/h5py/h5py/issues/2549
-        if selection.nselect == 0 and not (self.shape == selection.shape):
+        if selection.nselect == 0 and self.shape != selection.shape:
             return
 
         # Broadcast scalars if necessary.
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -1060,13 +1060,6 @@ class Dataset(HLObject):
 
         # Perform the dataspace selection
         selection = sel.select(self.shape, args, dataset=self)
-        # return early on zero-size selection
-        # but only if dataset shape and selection shape differ
-        # this prevents returning here for zero size datasets
-        # where proper errors should be raised later
-        # see https://github.com/h5py/h5py/issues/2549
-        if selection.nselect == 0 and self.shape != selection.shape:
-            return
 
         # Broadcast scalars if necessary.
         # In order to avoid slow broadcasting filling the destination by
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -1060,8 +1060,12 @@ class Dataset(HLObject):
 
         # Perform the dataspace selection
         selection = sel.select(self.shape, args, dataset=self)
-
-        if selection.nselect == 0:
+        # return early on zero-size selection
+        # but not if dataset shape and selection shape are equal
+        # this prevents returning here for zero size datasets
+        # where proper errors should be raised later
+        # see https://github.com/h5py/h5py/issues/2549
+        if selection.nselect == 0 and not (self.shape == selection.shape):
             return
 
         # Broadcast scalars if necessary.
--- a/h5py/tests/test_dataset_getitem.py
+++ b/h5py/tests/test_dataset_getitem.py
@@ -652,3 +652,36 @@ def test_bool_selection_1d(writable_file):
     str_dset = writable_file['strings']
     np.testing.assert_array_equal(str_dset[sel], str_arr[sel])
     np.testing.assert_array_equal(str_dset[sel, :], str_arr[sel, :])
+
+
+class TestZeroSizeSelectionResizableDataset(TestCase):
+    """
+    Tests for indexing of zero Resizable Datasets
+    see https://github.com/h5py/h5py/issues/2549
+    """
+    def setUp(self):
+        super().setUp()
+        self.dset0 = self.f.create_dataset('x', (0,), 'f4', maxshape=(None,))
+        self.dset1 = self.f.create_dataset('y', (1,), 'f4', maxshape=(None,))
+
+    def test_set_with_scalar(self):
+        sel = np.s_[:]
+        self.dset0[sel] = 1
+        self.arr = np.array([], dtype=np.float32)
+        self.assertNumpyBehavior(self.dset0, self.arr, sel)
+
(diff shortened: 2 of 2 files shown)


kmuehlbauer:
@takluyver I've finally came to this issue again. Please let me know, if this really solves the issue. At least it solves the reported problem and also keeps the behaviour according to your comment https://github.com/h5py/h5py/issues/2549#issuecomment-2637500041.

neutrinoceros:
just a quick note that failures appear to be unrelated network errors. We'll retry these jobs later.

kmuehlbauer:
@neutrinoceros Any chance to get a review here?

h5py/_hl/dataset.py
```diff
@@ -1045,8 +1045,12 @@ def __setitem__(self, args, val):
...
         # Perform the dataspace selection
         selection = sel.select(self.shape, args, dataset=self)
-
-        if selection.nselect == 0:
+        # return early on zero-size selection
+        # but not if dataset shape and selection shape are equal
+        # this prevents returning here for zero size datasets
+        # where proper errors should be raised later
+        # see https://github.com/h5py/h5py/issues/2549
+        if selection.nselect == 0 and not (self.shape == selection.shape):
```
neutrinoceros:
```suggestion
        # but only if dataset shape and selection shape differ
        # this prevents returning here for zero size datasets
        # where proper errors should be raised later
        # see https://github.com/h5py/h5py/issues/2549
        if selection.nselect == 0 and self.shape != selection.shape:
```

neutrinoceros (commented):
LGTM, with one minor suggestion to avoid negations in conditionals and comments

neutrinoceros (approved):
approving now, but I also want to give a couple more (buisiness) days for others to chime in, as I'm not very familiar with this part of the code. In essence, I just verified that your tests reveal the bug in question and that your patch fixes it, but I can't really judge the approach itself. Thanks !

kmuehlbauer:
No worries @neutrinoceros, your efforts are much appreciated.

neutrinoceros:
whoops, just stumbling upon this again. I just resolved a trivial merge conflict. Will merge once CI completes.

takluyver:
Closed/reopened to re-run CI, the errors before looked like network failures reaching anaconda.org.

takluyver:
The tests have passed now, but thinking about this some more, I realised a couple of issues.

1. `selection.shape` is the shape of the whole dataspace you're selecting in, not the selected part. So it's always the same as the dataset shape, and the effect is to remove the early exit entirely.
2. We could use `.mshape` (memory shape) or `.array_shape` instead, but this still isn't quite right. Say you have a dataset with shape `(0, 5)`, and you write to it with `ds[:, :4] = ...`. The selected shape is `(0, 4)`, different from the dataset shape, so it will take the early exit, but depending on the data shape you're writing, it may still be an error.

Maybe the simplest thing is to get rid of the early return entirely, so later checks can run. Or maybe checking the selection shape against the value shape (i.e. the data to be written, rather than the dataset we're writing into) still allows a useful shortcut. :thinking:

takluyver:
I think I prefer getting rid of the early exit entirely: it means doing a bit more work in some cases, but if writing into zero-length slices in a dataset is happening often enough to matter for performance, the caller can probably do better by checking for it themselves anyway.

kmuehlbauer:
Thanks @takluyver for taking the time. I agree, those checks can be done downstream if performance requires.

takluyver:
OK, that's passing the tests. Can you (@kmuehlbauer @neutrinoceros or anyone) see any possibilities where getting rid of this shortcut might do the wrong thing?

takluyver:
I'll merge this one in another day or two if there are no objections.

PR #2730, opened 2025-10-20
Build zlib from source for Windows wheels

### LINKED ISSUE #2261: Update zlib in Windows wheels
Follow up from #2254.

We are bundling zlib for Windows from [this NuGet package](https://www.nuget.org/packages/zlib-msvc-x64/), which hasn't been updated since 2017. It would be good to use a newer version.

I don't think any of the maintainers are very familiar with Windows development tools, so we're ideally looking for someone to either provide a newer, compatible version on NuGet, or figure out some other way of getting a compatible build of zlib in the CI environment.

NeoKish:
- Fixes #2261 
- Script for building zlib from source is added instead of nuget installation(which is at lower version)

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -95,8 +95,6 @@ jobs:
         # When preparing to upload a release, always build HDF5 to avoid potential
         # cache-poisoning attacks.
         if: ${{ runner.os == 'Windows' && github.ref_type != 'tag' && github.event_name != 'schedule' }}
-      - uses: nuget/setup-nuget@323ab0502cd38fdc493335025a96c8fdb0edc71f # v2.0.1
-        if: runner.os == 'Windows'
       - uses: actions/setup-python@e797f83bcb11b83ae66e0230d6156d7c80228e7c # v6.0.0
         with:
           python-version: 3.12


tacaswell:
Is the renaming of the dll wrong?

NeoKish:
> Is the renaming of the dll wrong?

no, that's the dll name from build outputs

ci/cibw_before_all_windows.sh
```diff
@@ -23,9 +24,25 @@ if [[ "$ARCH" == "ARM64" ]]; then
     export LINK="/LIBPATH:$ZLIB_ROOT/lib"
     export HDF5_VSVERSION="17-arm64"
 elif [[ "$ARCH" == "AMD64" ]]; then
-    # NuGet for Windows x64
-    nuget install zlib-msvc-x64 -ExcludeVersion -OutputDirectory "$PROJECT_PATH"
-    ZLIB_ROOT="$PROJECT_PATH/zlib-msvc-x64/build/native"
+    # Build zlib from source for Windows AMD64
+    mkdir zlib_win_x64
+    curl -sLO https://github.com/madler/zlib/archive/refs/tags/v$ZLIB_VERSION.tar.gz
```
takluyver: For Mac, we download a tarball from `https://zlib.net/fossils/zlib-$ZLIB_VERSION.tar.gz`. Does the archive from Github give us anything extra? If not, could we use the same URL on Windows.

ci/cibw_before_all_windows.sh
```diff
@@ -23,9 +24,25 @@ if [[ "$ARCH" == "ARM64" ]]; then
...
+    tar -xzf v$ZLIB_VERSION.tar.gz && rm v$ZLIB_VERSION.tar.gz
+    cmake -S zlib-$ZLIB_VERSION -B build -G "Visual Studio 17 2022" \
+    -DBUILD_SHARED_LIBS:BOOL=ON -DCMAKE_INSTALL_PREFIX:PATH="$PROJECT_PATH/zlib_win_x64"
```
takluyver:
```suggestion
        -DBUILD_SHARED_LIBS:BOOL=ON \
        -DCMAKE_INSTALL_PREFIX:PATH="$PROJECT_PATH/zlib_win_x64"
```

I think this would make the list of args clearer if we need to add more to it.

ci/cibw_before_all_windows.sh
```diff
@@ -23,9 +24,25 @@ if [[ "$ARCH" == "ARM64" ]]; then
...
+    # Release build
+    cmake --build build --config Release
+    cmake --install build --config Release
+    mv zlib_win_x64/bin zlib-win-x64/bin_release
+    mv zlib_win_x64/lib zlib-win-x64/lib_release
+    # Debug build
+    cmake --build build --config Debug 
+    cmake --install build --config Debug
+    mv zlib_win_x64/bin zlib-win-x64/bin_debug
+    mv zlib_win_x64/lib zlib-win-x64/lib_debug
```
takluyver: Is there any reason to produce a debug build, if we're only going to use the release build?

ci/fix_paths.py
```diff
@@ -29,7 +29,7 @@ def main():
             if arch in ("arm64", "aarch64"):
                 f = pjoin(zlib_root, 'bin', 'zlib1.dll')
             elif arch in ("amd64", "x86_64"):
-                f = pjoin(zlib_root, 'bin_release', 'zlib.dll')
+                f = pjoin(zlib_root, 'bin_release', 'zlib1.dll')
```
takluyver: We also have Windows builds on Azure pipelines, which still get the old nuget package:

https://github.com/h5py/h5py/blob/c32aae5e3f4c78a1a49542f6196863abef719097/ci/azure-pipelines-steps.yml#L40-L46

Those are now failing because of this name change. I think the best thing to do is to switch those over to building zlib from source as well, either by integrating it into the `get_hdf5_win.py` script, or splitting the commands you added out as a separate script which can be called from the Azure YAML as well.

ci/cibw_before_all_windows.sh
```diff
@@ -23,9 +24,25 @@ if [[ "$ARCH" == "ARM64" ]]; then
...
+    # Build zlib from source for Windows AMD64
+    mkdir zlib_win_x64
+    curl -sLO https://github.com/madler/zlib/archive/refs/tags/v$ZLIB_VERSION.tar.gz
+    tar -xzf v$ZLIB_VERSION.tar.gz && rm v$ZLIB_VERSION.tar.gz
+    cmake -S zlib-$ZLIB_VERSION -B build -G "Visual Studio 17 2022" \
+    -DBUILD_SHARED_LIBS:BOOL=ON -DCMAKE_INSTALL_PREFIX:PATH="$PROJECT_PATH/zlib_win_x64"
+    # Release build
+    cmake --build build --config Release
+    cmake --install build --config Release
+    mv zlib_win_x64/bin zlib-win-x64/bin_release
```
takluyver:
```
mv: cannot move 'zlib_win_x64/bin' to 'zlib-win-x64/bin_release': No such file or directory
```

On the Windows AMD64 job.

takluyver (commented):
Thanks, a few things to figure out.

NeoKish:
@takluyver , thanks for the suggestions. 

- I have added the separated zlib script for use in both cibw_before_all_windows.sh and azure-pipelines-steps.yml and updated the paths accordingly

ci/cibw_before_all_windows.sh
NeoKish: Have used the same tarball used for Mac. The dll name(zlib1.dll) is different for the tarball I was using earlier. I have reverted it back to the previous(zlib.dll) in the later commits pushed

ci/cibw_before_all_windows.sh
NeoKish: This suggestion has been taken into account in the new script for zlib source build

ci/cibw_before_all_windows.sh
NeoKish: This was because zlib library debug path was used in get_hdf5_win.py script. this has been removed for windows amd64 as hdf5 is built only for release based on the script

ci/fix_paths.py
NeoKish: Have modified this

ci/cibw_before_all_windows.sh
NeoKish: Fixed this

takluyver:
Thanks! One more thing: the Github workflow (`.github/workflows/build_wheels.yml`) still calls an action to setup nuget, although we're no longer using it anywhere, as far as I can see. Could you get rid of that?

takluyver:
Thank you @NeoKish !

PR #2731, opened 2025-10-21
Add specific error message when slicing dataset with None/newaxis

### LINKED ISSUE #2409: No Numpy newaxis slicing support
# Description:
H5py currently doesn't support adding np.newaxis when slicing a data array. 
```
with h5py.File(file, 'r') as F:
        return F[array_group][:,np.newaxis,:,:]
```

as following return is given:

' File "****", line 77, in read_hdf_frame_v1
    return F[array_group][:,np.newaxis,:,:]
  File "h5py\_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py\_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "****\lib\site-packages\h5py\_hl\dataset.py", line 831, in __getitem__
  File "****lib\site-packages\h5py\_hl\selections.py", line 82, in select
  File "h5py\_selector.pyx", line 282, in h5py._selector.Selector.make_selection
  File "h5py\_selector.pyx", line 181, in h5py._selector.Selector.apply_args
TypeError: Selection can't process None'

'Summary of the h5py configuration
h5py    3.10.0
HDF5    1.14.2
Python  3.10.13 | packaged by Anaconda, Inc. | (main, Sep 11 2023, 13:24:38) [MSC v.1916 64 bit (AMD64)]
sys.platform    win32
numpy   1.26.4
cython (built with) 0.29.36
numpy (built against) 1.21.6
HDF5 (built against) 1.14.2'

takluyver:
`np.newaxis` is an alias for None, but that might not be obvious (I wasn't sure), so it could be helpful to explicitly mention it in an error message.

Closes #2409

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -835,6 +835,9 @@ class Dataset(HLObject):
         """
         args = args if isinstance(args, tuple) else (args,)
 
+        if any(a is None for a in args):  # 'None in args' would fail on arrays
+            raise TypeError("Indexing with None (or np.newaxis) is not supported")
+
         if self._fast_read_ok and (new_dtype is None):
             try:
                 return self._fast_reader.read(args)
--- a/h5py/tests/test_dataset_getitem.py
+++ b/h5py/tests/test_dataset_getitem.py
@@ -44,8 +44,9 @@
 import sys
 
 import numpy as np
-import h5py
+import pytest
 
+import h5py
 from .common import ut, TestCase
 
 
@@ -616,3 +617,9 @@ class TestBoolIndex(TestCase):
         sel = np.s_[[False, True, False, False],:]
         with self.assertRaises(TypeError):
             self.dset[sel]
+
+
+def test_error_newaxis(writable_file):
+    ds = writable_file.create_dataset('a', data=np.arange(5))
+    with pytest.raises(TypeError, match="newaxis"):
+        ds[np.newaxis, :]


h5py/_hl/dataset.py
```diff
@@ -835,6 +835,9 @@ def __getitem__(self, args, new_dtype=None):
         """
         args = args if isinstance(args, tuple) else (args,)
 
+        if any(a is None for a in args):  # 'None in args' would fail on arrays
```
neutrinoceros, line 838: great comment. I was *about* to suggest the naive approach before my eyes reached the end of the line ;-)

h5py/tests/test_dataset_getitem.py
```diff
@@ -616,3 +617,9 @@ def test_wrong_size(self):
         sel = np.s_[[False, True, False, False],:]
         with self.assertRaises(TypeError):
             self.dset[sel]
+
+
+def test_error_newaxis(writable_file):
+    ds = writable_file.create_dataset('a', data=np.arange(5))
+    with pytest.raises(TypeError, match="newaxis"):
```
neutrinoceros, line 624: since there's nothing dynamic about this error message, we can afford a stricter check
```suggestion
    with pytest.raises(TypeError, match=r"^Indexing with None \(or np\.newaxis\) is not supported$"):
```

takluyver: I prefer not to assert on entire error messages, so if we reword the message we don't need to change the test as well. The idea is just to check that it mentions `newaxis`.

neutrinoceros: dealer's choice then

takluyver:
I'll merge this in a couple more days if there are no further comments.

PR #2734, opened 2025-11-03
New default chunk cache nslots value in HDF5 2.0

ajelenak:
Update tests and documentation for new default chunk cache nslots constant in HDF5 2.0.


takluyver:
Thanks!

PR #2738, opened 2025-11-09
TST/DEP: upgrade pytest to version 9

neutrinoceros:
This is arguably a *very* aggressive upgrade; pytest 9.0.0 was just released yesterday.
However, I wanted to give it a spin for demonstration, if nothing else. Most importantly, this version is the first one that has proper lower bounds on its own dependencies, which enables switching uv resolution from `lowest-direct` to `lowest` the difference being that, with `lowest`, transitive deps are *also* resolved to oldest, which is much more consistent and reproducible than just pinning *direct* ones.

While I'm at it, I also took advantage of new features:
- native toml support (`[tool.pytest]`)
- `strict` mode, which makes things like typos in markers raise exceptions instead of mere warnings

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -112,7 +112,8 @@ tables = [
     "tables>=3.7.0",
 ]
 test = [
-    "pytest>=8.2.2",
+    "coverage>=7.10.0",
+    "pytest>=9.0.1",
     "pytest-cov>=5.0.0",
     "pytest-mpi>=0.6",
 ]
@@ -121,6 +122,16 @@ wheels = [
     "coverage>=7.9.1",
     "tox>=4.26.0",
     "tox-uv>=1.27.0",
+]
+
+[tool.pytest]
+minversion = "9.0"
+strict = true
+addopts = ["-ra"]
+filterwarnings = ["error"]
+markers = [
+    "network: Tests requiring network access, skipped with --no-network",
+    "slow: Slow tests",
 ]
 
 [tool.cibuildwheel]
@@ -160,3 +171,7 @@ H5PY_DIRECT_VFD = "0"
 [tool.uv]
 # never build numpy from source
 no-build-package = ["numpy"]
+constraint-dependencies = [
+    # via check-manifest
+    "setuptools>=61.0",
+]
--- a/tox.ini
+++ b/tox.ini
@@ -47,8 +47,7 @@ setenv =
     pre: UV_INDEX=https://pypi.anaconda.org/scientific-python-nightly-wheels/simple
     pre: UV_INDEX_STRATEGY=unsafe-best-match  # match pip's behavior
 uv_resolution =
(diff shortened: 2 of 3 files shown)


h5py/tests/test_filters.py
```diff
@@ -85,7 +91,10 @@ def test_filter_ref_obj_eq():
     assert gzip8 != h5py.filters.Gzip(level=7)
 
 
-@ut.skipIf(not os.getenv('H5PY_TEST_CHECK_FILTERS'),  "H5PY_TEST_CHECK_FILTERS not set")
```
neutrinoceros, line 88: I don't really get why, but this test started failing with pytest 9.0.0. Using pytest's own `skipif` decorator resolved the problem, so I applied this change throughout the module (and I plan to do it over the entire suite, but it feels out of scope here).
This is *not* caused by `strict` mode.

tox.ini
```diff
@@ -47,8 +47,7 @@ setenv =
     pre: UV_INDEX=https://pypi.anaconda.org/scientific-python-nightly-wheels/simple
     pre: UV_INDEX_STRATEGY=unsafe-best-match  # match pip's behavior
 uv_resolution =
-    mindeps: lowest-direct
-    !mindeps: highest
```
neutrinoceros, line 51: this line doesn't do anything because that's the default behavior, so I removed it.

h5py/tests/test_filters.py
neutrinoceros: looks like I *did* need to apply this everywhere in the end... I'll probably isolate the commit for this change so it can be merged in isolation if needed.

setup.py
```diff
@@ -30,6 +30,7 @@
     # works (assuming you build it from source too).
     # NumPy 1.21.2 is the first release with any wheels for Python 3.10, our minimum Python.
     "numpy >=1.21.2",
+    "numpy >=1.21.3 ; platform_machine=='arm64' and platform_system=='Darwin'",
```
neutrinoceros: this was useful for local testing and as an intermediate step but it *shouldn't* be needed so I'll try reverting it once everything else looks stable.

neutrinoceros:
rebased on top of #2739

pyproject.toml
```diff
@@ -123,6 +124,16 @@ wheels = [
     "tox-uv>=1.27.0",
 ]
 
+[tool.pytest]
+minversion = "9.0"
+#strict = true
```
neutrinoceros:
```suggestion
strict = true
```

neutrinoceros:
this is stable with pytest 9.0.1

tacaswell:
definitely too aggressive, I would be inclined to give it 12mo before we adjust the floor.

neutrinoceros:
sounds fair. In that case, I'll close this in favor of #2747

PR #2739, opened 2025-11-09
TST/RFC: replace unittest.skipIf markers with pytest.mark.skipif

neutrinoceros:
Extracted from #2738
This should be fully backward+forward compatible, but for some reason pytest 9.0.0 seems to choke on unittest decorators.
xref: https://github.com/pytest-dev/pytest/issues/13895

Independently of this problem, I think we should aim to use `pytest` API as much as possible, if nothing else, for the sake of consistency.

### DIFF
--- a/h5py/tests/test_base.py
+++ b/h5py/tests/test_base.py
@@ -15,11 +15,13 @@
 
 from h5py import File
 from h5py._hl.base import is_hdf5, Empty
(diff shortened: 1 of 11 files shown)


neutrinoceros:
In hindsight, this will likely be fixed in a patch version of pytest, and the patch here may create unnecessary conflicts with #2655, so I'd rather not merge it right now. Drafting for now.

PR #2745, opened 2025-11-21
Enable freethreading_compatible flag

crusaderky:
- Part of #2475
- Extracted from #2702

As discussed in https://github.com/h5py/h5py/pull/2650, we should not enable this flag until the unit tests suite gives us confidence that there are no race conditions. See parent issue for checklist.


tacaswell:
I am happy to flip this switch, the fix we added as part of https://github.com/h5py/h5py/pull/2658 along with all of the libhdf5 work being serialized by `phil` I am satisfied that we are safe-as-a-library for freethreading.

crusaderky:
@tacaswell this is ready then

PR #2746, opened 2025-11-21
Fix test failure in conda-forge CI

crusaderky:
This test fails when the test suite is triggered by the conda-forge recipe for 3.14t:
https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1386909&view=logs&jobId=aa85f627-05cc-5265-2465-79d7168d3859&j=f2822e16-0094-5d6c-29fd-a9f5a365b7fd&t=c158e018-8e8e-5c27-0ba4-2482c0a9e77e

```python
______________________ TestFileLocking.test_multiprocess _______________________

self = <h5py.tests.test_file.TestFileLocking object at 0x22f0e895650>
tmp_path = PosixPath('/tmp/pytest-of-conda/pytest-0/test_multiprocess0')

        def test_multiprocess(self, tmp_path):
            """Test file locking option from different concurrent processes"""
            fname = tmp_path / "test.h5"
    
            def open_in_subprocess(filename, mode, locking):
                """Open HDF5 file in a subprocess and return True on success"""
                h5py_import_dir = str(pathlib.Path(h5py.__file__).parent.parent)
    
                process = subprocess.run(
                    [
                        sys.executable,
                        "-c",
                        f"""
    import sys
    sys.path.insert(0, {h5py_import_dir!r})
    import h5py
    f = h5py.File({str(filename)!r}, mode={mode!r}, locking={locking})
                        """,
                    ],
                    capture_output=True)
                return process.returncode == 0 and not process.stderr
    
            # Create test file
            with h5py.File(fname, mode="w", locking=True) as f:
                f["data"] = 1
    
            with h5py.File(fname, mode="r", locking=False) as f:
                # Opening in write mode with locking is expected to work
>               assert open_in_subprocess(fname, mode="w", locking=True)
E               AssertionError: assert False
E                +  where False = <function TestFileLocking.test_multiprocess.<locals>.open_in_subprocess at 0x22f1120b4c0>(PosixPath('/tmp/pytest-of-conda/pytest-0/test_multiprocess0/test.h5'), mode='w', locking=True)
```

Unsure why, but the test uses an anti-pattern that encroaches on h5py's bespoke import design.
This PR fixes the issue, as demonstrated by https://github.com/conda-forge/h5py-feedstock/pull/160

CC @zklaus

### DIFF
--- a/h5py/tests/test_file.py
+++ b/h5py/tests/test_file.py
@@ -1020,7 +1020,8 @@ class TestFileLocking:
             File(fname, mode="r", locking=False) as f,
         ):
             # Opening in write mode with locking is expected to work
-            ex.submit(open_and_close, fname, mode="w", locking=True).result()
+            future = ex.submit(open_and_close, fname, mode="w", locking=True)
+            future.result(timeout=10)
 
 
 def open_and_close(*args, **kwargs):


tacaswell:
> Unsure why, but the test uses an anti-pattern that encroaches on h5py's bespoke import design.

What does this mean?

tacaswell:
Do we understand why this test is failing on conda-forge?  Is the critical change removing the path-hacking or the move to the `ProcessPool`?  What is the minimal change we can make that will fix CF?

Carreau:
from a DM chat with @tacaswell, i think also patching this test out on CF for now until we get in agreement on this and then have a release out would be ok.

crusaderky:
> > Unsure why, but the test uses an anti-pattern that encroaches on h5py's bespoke import design.
> 
> What does this mean?

h5py includes guards that prevent it from running unit tests in its install directory.
e.g. locally, in order to run tests I need to
```bash
pip install .
mkdir /tmp/h5py-tests
cp pyproject.coml /tmp/h5py-tests/
pushd /tmp/h5py-tests
python -m pytest --pyargs h5py
```
I did not audit the conda-forge script, but I expect some sort of subtle difference going on there compared to the h5py CI. Ping @zklaus for more info.

> Do we understand why this test is failing on conda-forge? Is the critical change removing the path-hacking or the move to the ProcessPool? 

The critical change is removing the path-hacking.

> What is the minimal change we can make that will fix CF?

We could _try_ removing the `sys.path.insert` line and see what happens. I did not test that. This PR goes for the IMHO easier and cleaner route of letting the spawn multiprocessing context figure everything out.

zklaus:
The tests on conda-forge are run in a dedicated test environment and in the case of the h5py package via the [`run_test.py` script](https://github.com/zklaus/h5py-feedstock/blob/main/recipe/run_test.py), which does a couple of import tests first, then checks that `h5py.get_config()` reflects the expected mpi configuration, and finally uses `h5py.run_tests()` to run the actual tests.

Carreau:
https://github.com/conda-forge/h5py-feedstock/pull/161 skip this test on the feedstock and with it the build passes. Tf there are disagreement on how to fix this for the time beeing, maybe skipping the test on CF for now will be ok, and unblock dependant packages for 3.14t, that woudl be great.

Carreau:
I think this was autoclosed by github on mistake, the CF patch references this, but a fix is still necessary.

neutrinoceros:
reopened

h5py/tests/test_file.py
```diff
@@ -13,13 +13,14 @@
...
 
+import multiprocessing
 import os
 import stat
 import pickle
 import tempfile
 import time
-import subprocess
 import sys
+from concurrent.futures import ProcessPoolExecutor
```
takluyver, line 23: Is there a particular reason to use `concurrent.futures` rather than multiprocessing's `Pool` class, since we have to import multiprocessing anyway? I never really know which one is preferable.

h5py/tests/test_file.py
```diff
@@ -1009,31 +1010,23 @@ def test_multiprocess(self, tmp_path):
...
+            ProcessPoolExecutor(mp_context=ctx, max_workers=1) as ex,
+            File(fname, mode="r", locking=False) as f,
+        ):
             # Opening in write mode with locking is expected to work
-            assert open_in_subprocess(fname, mode="w", locking=True)
+            ex.submit(open_in_subprocess, fname, mode="w", locking=True).result()
+
+
+def open_in_subprocess(*args, **kwargs):
+    """Open and close HDF5 file in a subprocess"""
```
takluyver:
```suggestion
def open_and_close(*args, **kwargs):
    """Open and close HDF5 file, for use in a subprocess"""
```

Since it's not actually doing anything with a subprocess inside the function, now.

h5py/tests/test_file.py
```diff
@@ -1009,31 +1010,23 @@ def test_multiprocess(self, tmp_path):
...
 
-        with h5py.File(fname, mode="r", locking=False) as f:
+        ctx = multiprocessing.get_context("spawn")
+        with (
+            ProcessPoolExecutor(mp_context=ctx, max_workers=1) as ex,
+            File(fname, mode="r", locking=False) as f,
+        ):
             # Opening in write mode with locking is expected to work
-            assert open_in_subprocess(fname, mode="w", locking=True)
+            ex.submit(open_in_subprocess, fname, mode="w", locking=True).result()
```
takluyver:
```suggestion
            ex.submit(open_and_close, fname, mode="w", locking=True).result()
```

takluyver: Also, let's pass a timeout to `.result()`, something generous like 60 seconds, just to make sure it can't get stuck forever.

h5py/tests/test_file.py
crusaderky: I've always used `concurrent.futures` without much thought. It's just easier to remember how to use one that has identical API for the threaded version.

takluyver: Fair enough. In fact multiprocessing also has a `ThreadPool` class, although the docs for that say one should 'generally prefer' the version in concurrent.futures.

crusaderky:
@takluyver all comments have been addressed

takluyver:
Thanks!

crusaderky:
@zklaus FYI the hack you put into the conda-forge recipe can be reverted when 3.16 comes out.

PR #2747, opened 2025-11-21
TST: more stable backward compatibility testing with lowest resolution strategy

neutrinoceros:
A less aggressive alternative to #2738


neutrinoceros:
I suspect the only reason this is still in draft is that I forgot to undraft it. I'll refresh CI and hopefully open it for review today.

neutrinoceros:
@tacaswell

PR #2750, opened 2025-12-02
Test building with HDF5 2.0

takluyver:
This is working on Azure, opening a PR to test with Github actions too.

### DIFF
--- a/ci/get_hdf5_win.py
+++ b/ci/get_hdf5_win.py
@@ -31,15 +31,22 @@ arch = platform.machine().lower()
 CI_DIR = dirname(abspath(__file__))
 
 CMAKE_CONFIGURE_CMD = [
-    "cmake", "-DBUILD_SHARED_LIBS:BOOL=ON", "-DCMAKE_BUILD_TYPE:STRING=RELEASE",
-    "-DHDF5_BUILD_CPP_LIB=OFF", "-DHDF5_BUILD_HL_LIB=ON",
-    "-DHDF5_BUILD_TOOLS:BOOL=OFF", "-DBUILD_TESTING:BOOL=OFF",
+    "cmake",
+    "-DBUILD_SHARED_LIBS:BOOL=ON",
+    "-DBUILD_STATIC_LIBS:BOOL=OFF",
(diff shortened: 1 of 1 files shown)


neutrinoceros:
I'd suggest adding a `brew install zlib` somewhere on macos and see if that's enough

takluyver:
The low-level API docs failure is unrelated - Sphinx 9 was released in the last few days, and the release notes mention:

> [sphinx.ext.autodoc](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#module-sphinx.ext.autodoc) has been substantially rewritten, and there may be some incompatible changes in edge cases, especially when extensions interact with autodoc internals. The [autodoc_use_legacy_class_based](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_use_legacy_class_based) option has been added to use the legacy (pre-9.0) implementation of autodoc.

So I'd guess that's biting us.

neutrinoceros:
clearly. See #2752 for a quick fix

ajelenak:
Why not switch to zlib-ng instead of zlib. It is faster. Can be built at the same time as libhdf5 which then includes in its dynamic library artifact.

takluyver:
For Linux, which I suspect is a large chunk of where h5py is used, zlib is ubiquitous enough that we can assume it's on the system and don't have to bundle it. We should keep testing somewhere with stock zlib so we can ensure that works for people building it themselves, even if we bundle zlib-ng.

I'm not against switching to zlib-ng for our pre-built packages - performance is always nice. But I'd like to have the process of downloading & building it under our control in our CI, not something that HDF5's build system tries to do for us. To my mind it's a separate question from building with HDF5 2.0, though.

takluyver:
I've prepared https://github.com/h5py/hdf5-manylinux/pull/24 to build HDF5 2.0 for use in the Linux wheels as well.

takluyver:
@tacaswell any thoughts on this or the corresponding update in the Docker images for building Linux wheels (h5py/hdf5-manylinux#24)? I'll merge both in a couple more days if there are no objections.

Issue #2690, opened 2025-09-12
Feature request: allow specifying access options for external links and virtual datasets


woutdenolf:
Currently, when following an `ExternalLink` or virtual dataset, `h5py` opens the target file with the same access options as the parent file. There is no way to override these options (e.g. mode, locking, swmr).

This can make it impossible to read data from external links if the external file requires different options than the parent. For example, when opening the parent file in append mode and the external file is read-only (a common situation for data processing result files linking to raw data from an experiment owned by another user):

```python
import h5py
import os, stat, tempfile

tmpdir = tempfile.mkdtemp()
ext_path = os.path.join(tmpdir, "ext.h5")
root_path = os.path.join(tmpdir, "root.h5")

with h5py.File(ext_path, "w") as ext:
    ext.create_dataset("data", data=[1, 2, 3])

os.chmod(ext_path, stat.S_IREAD)

with h5py.File(root_path, "w") as root:
    root["ext_data"] = h5py.ExternalLink(ext_path, "/data")

with h5py.File(root_path, "a") as f:
    dset = f["ext_data"]  # raises KeyError: "Unable to synchronously open object (can't open file)"
    print(dset[()])
```

If we can support this for virtual datasets as well that would be even better.

Ping @t20100

neutrinoceros:
Interesting. What kind of API do you envision to support this ?

woutdenolf:
Something like this:

```python
class File(Group):

    def __init__(self, name, mode='r', driver=None, libver=None, userblock_size=None, swmr=False,
                 rdcc_nslots=None, rdcc_nbytes=None, rdcc_w0=None, track_order=None,
                 fs_strategy=None, fs_persist=False, fs_threshold=1, fs_page_size=None,
                 page_buf_size=None, min_meta_keep=0, min_raw_keep=0, locking=None,
                 alignment_threshold=1, alignment_interval=1, meta_block_size=None,
                 *, track_times=False, **kwds):

    def update_external_access(self, **kwds):
        """File access properties for external files. The properties of the parent file are used by default.
        """
```

neutrinoceros:
I meant, how would a user script look ?

woutdenolf:
[...]

Issue #2257, opened 2023-05-01
Possibility to control the link access/creation property lists from the high level interface


cderemble:
Currently, link access/creation property lists are global instances created when loading the library via `default_lapl` / `default_lcpl`.
It would be cool if individual groups could have the possibility to override these properties. Ideally, these properties would be passed on to groups returned by `__getitem__`

One possible implementation would be to be to override the properties `_lapl` and `_lcpl` in `Group` : they would default to the global property lists and use the ones optionally provided in the constructor.
The `__init__` function of `Group` and `File` would then need to be modified by adding 2 optional arguments: `lapl` and `lcpl`. The rest of the library should not be affected. By default, the behavior of the library should be unchanged.

cderemble:
possible implementation in #2258

aragilar:
It would seem reasonable to me to have externally referenced files default to whatever mode closest matches the the mode of the main one (with some option to override the mode separately at the high level).

takluyver:
I agree that that makes sense, but I'm also confused - the [HDF5 docs](https://portal.hdfgroup.org/display/HDF5/H5P_SET_ELINK_ACC_FLAGS) suggest that this should already be the default:

> The library will normally use the file access flag used to open the parent file as the file access flag for the target file. 

And h5py doesn't seem to override that:

https://github.com/h5py/h5py/blob/06d9d0d095c1e972f938d874a78598ae42b1df8e/h5py/_hl/base.py#L129-L135

Am I missing somewhere that we're setting that to something different? Or is this a bug?

(I'm more and more convinced that making references to separate files transparent, as external links, external datasets or virtual datasets, is a mistake. :thinking: )

takluyver:
At least for me, following an external link from a read-only file seems to do the right thing (the second file is also read-only):

```
In [3]: f1 = h5py.File('foo.h5', 'r')

In [4]: ds = f1['extlink']

In [5]: ds
Out[5]: <HDF5 dataset "a": shape (10, 10), type "<f4">

In [6]: ds.file
Out[6]: <HDF5 file "foo2.h5" (mode r)>
```

cderemble:
My bad: I got it wrong, let me reformulate.

The problem is when you want to modify a file that contains external links to other files that you know wont be modified. For example, let's copy a dataset from an external file via a link: 
```
f = h5py/File('foo.h5', mode='w')
f['ext_data_set'] = h5py.ExternalLink('foo2.h5', 'data_set')
[...]

PR #2751, opened 2025-12-02
Add arguments to Group.get to configure external link access

t20100:
This PR proposes to add some arguments to `Group.get` in order to configure external links access behavior as suggested in https://github.com/h5py/h5py/issues/2690#issuecomment-3398115742.

It adds 3 arguments: `elink_mode`, `elink_swmr` and `elink_locking` (using a naming similar to `h5py.File` arguments + `elink_` prefix).

It turned out not all combinations of current file access mode/external link access mode are supported: In those cases an exception is raised while creating the file access property list, otherwise `Group.get` would silent it and returns its `default` value.

Related to #2690 and #2257

### DIFF
--- a/h5py/_hl/group.py
+++ b/h5py/_hl/group.py
@@ -410,8 +410,8 @@ class Group(HLObject, MutableMappingHDF5):
     def _get(self, name, lapl=None):
         if isinstance(name, h5r.Reference):
             if lapl is not None:
-                raise NotImplementedError(
-                    "Accessing a reference with link access properties is not supported")
+                raise TypeError(
+                    "Access with link access properties is done with bytes or str, not Reference")
             oid = h5r.dereference(name, self.id)
             if oid is None:
                 raise ValueError("Invalid HDF5 object reference")


h5py/_hl/group.py
```diff
@@ -12,20 +12,64 @@
...
+    elink_mode: Literal["r", "r+"] | None = None,
+    elink_swmr: bool | None = None,
+    elink_locking: Literal["true", "false", "best-effort"] | bool | None = None,
+) -> h5p.PropLAID | None:
+    """Set up a link access property list"""
+    if elink_mode is None and elink_swmr is None and elink_locking is None:
+        return None
+
+    if file.mode == "r" and elink_mode == "r+":
+        raise ValueError("Opening external links in write mode from a file opened in read-only mode is not supported")
```
t20100, line 45: This would otherwise raise an exception when getting the item which is catched in `Group.get`

h5py/_hl/group.py
```diff
@@ -12,20 +12,64 @@
...
+
+    lapl = h5p.create(h5p.LINK_ACCESS)
+
+    if elink_mode is not None or elink_swmr is not None:
+        mode = file.mode if elink_mode is None else elink_mode
+        swmr_mode = file.swmr_mode if elink_swmr is None else elink_swmr
+
+        if file.mode == "r+" and file.swmr_mode and (mode != "r+" or not swmr_mode):
+            raise ValueError("Changing external links access mode from a file opened in SWMR write mode is not supported")
```
t20100, line 54: From what I tested, changing external link access mode from a file opened in SWMR write mode is ignored.

h5py/_hl/plist.py
```diff
@@ -0,0 +1,12 @@
+from typing import Literal
+
+
+def set_file_locking(fapl, locking: Literal["true", "false", "best-effort"] | bool):
```
t20100: This conversion needs to be shared between `Group` and `File`.
It could go in `group.py` if you prefer to avoid creating new modules.

t20100:
I made PR #2760 to propose an alternative/complete this one with extra arguments in `File.__init__`.

h5py/_hl/plist.py
takluyver: I maybe lean towards putting in `group.py` - I think my perceptions are skewed by working with a filesystem which doesn't really like the thousands of tiny files that make up a Python environment. 🙃

h5py/tests/test_group.py
```diff
@@ -1046,6 +1046,125 @@ def test_unicode_hdf5_path(self):
...
+    with File(main_filepath, "r", locking=file_locking) as main_file:
+        cls = main_file.get("/external/data", getclass=True, elink_locking=elink_locking)
+        assert cls is Dataset
+
+        link = main_file.get("/external/data", getlink=True, elink_locking=elink_locking)
+        assert isinstance(link, HardLink)
+
+        ext_group = main_file.get("external", elink_locking=elink_locking)
+        if h5py.version.hdf5_version_tuple >= (1, 14, 4):
+            # Retrieving file locking is broken for older versions of HDF5
```
t20100, line 1163: From 
https://raw.githubusercontent.com/HDFGroup/hdf5/hdf5_1_14_4/release_docs/RELEASE.txt:

>     - Fixed H5Fget_access_plist so that it returns the file locking
>       settings for a file
> 
>       When H5Fget_access_plist (and the internal H5F_get_access_plist)
>       is called on a file, the returned File Access Property List has
>       the library's default file locking settings rather than any
>       settings set for the file.

h5py/_hl/plist.py
t20100: Done

h5py/_hl/group.py
```diff
@@ -12,20 +12,73 @@
 """
 
 from contextlib import contextmanager
+from typing import Literal
```
tacaswell: I don't think we have any other inline typing.  I'm wary of adding it in a non-systematic way.

h5py/_hl/group.py
```diff
@@ -355,16 +408,20 @@ def require_group(self, name):
                 raise TypeError("Incompatible object (%s) already exists" % grp.__class__.__name__)
             return grp
 
-    @with_phil
     def __getitem__(self, name):
         """ Open an object in the file """
+        return self._get(name)
 
+    @with_phil
+    def _get(self, name, lapl=None):
```
tacaswell, line 410: This now leads to `lapl` getting ignored in the `h5r.Reference` case.

As this is new functionality, I'm inclined to fail in the `h5r.Reference` case if `lapl` is passed.  No one should be doing in now (so no user breakage) and we can re-address if we should check that they are consistent or actively ignore when someone has a use case to justify either way.

h5py/_hl/group.py
```diff
@@ -405,17 +497,22 @@ def get(self, name, default=None, getclass=False, getlink=False):
         # pylint: disable=arguments-differ
 
         with phil:
+            if elink_mode is None and elink_swmr is None and elink_locking is None:
+                lapl = self._lapl
+            else:
+                lapl = make_lapl(self.file, elink_mode, elink_swmr, elink_locking)
```
tacaswell, line 493: Should we try to merge if there is any `self._lapl`?  I tried to track through where this is set, but it appears to only ever be `None` as of 3c7fc2e33e49c0f0f1bf2bcbca3336512d090bc4 .

We can probably drop `_lapl` all together at this point.

tacaswell:
overall 👍🏻 , just a few nit picks.

Sorry this took so long to get looked at.

h5py/_hl/group.py
takluyver: Yes, I think it's always None, so no need to fiddle with merging. I've made a note in #2756 to remove it for v4 - which is probably unnecessarily cautious, but as we're planning v4 anyway...

h5py/_hl/group.py
```diff
@@ -397,6 +463,32 @@ def get(self, name, default=None, getclass=False, getlink=False):
...
+
+        "elink_locking":
+            External links file locking behavior:
+
+            - None (default)     --  Use the current file locking
+            - False (or "false") --  Disable file locking
+            - True (or "true")   --  Enable file locking
+            - "best-effort"      --  Enable file locking but ignore some errors
+
+            .. warning::
```
takluyver: The docstrings in the high-level API (`h5py/_hl/`) don't get built into Sphinx docs, so AFAIK there's no need to use rst syntax.

h5py/_hl/group.py
t20100: Raising `NotImplementedError`. I hesitate with `ValueError`.

h5py/_hl/group.py
t20100: Yes, `self._lapl`, `base.dlapl` and `base.default_lapl()` are no longer needed, but I was worried to break code that patch this.

h5py/_hl/group.py
takluyver: Sorry to nitpick, but I don't think NotImplementedError is quite right - that implies that something *could* be implemented, it's just not there yet. If I've understood references correctly, they point directly to an object, not a link, so it doesn't make sense to combine them with a LAPL.

So I would use ValueError, or maybe even TypeError (as Python raises that for unexpected arguments to a function).

t20100: Indeed.

I propose `TypeError`, but it's not clear to me what's best between `ValueError` & `TypeError` when raised for a combination of arguments.

takluyver: Thanks! I suspect TypeError vs. ValueError doesn't matter too much - this is mainly there to give a 'this is wrong' message to a developer, not to be handled. :+1:

takluyver:
Thanks @t20100 !

Issue #2754, opened 2025-12-02
MNT/DEP: reminder to lift upper bound on Sphinx


neutrinoceros:
Follow up to #2752

We should lift the upper bound (allow Sphinx 9), but it's not clear yet how to do that. Maybe we won't be able to support using both Sphinx 9 *or* older versions, and maybe we need so patches upstream before we can make it work at all. In any case this is affecting a lot of projects so I expect Sphinx to adapt over the following days or weeks.

xref: https://github.com/sphinx-doc/sphinx/issues/14089

AA-Turner:
@neutrinoceros @takluyver what errors did you run into? What fixes are needed in Sphinx?

A

neutrinoceros:
Thanks for reaching out @AA-Turner 
Here's a log with the relevant failure mode : https://github.com/h5py/h5py/actions/runs/19864317402/job/56922610671


I haven't attempted to dig deeper than this yet, but I should be able to do so in the morning if needed.

AA-Turner:
```
WARNING: error while formatting signature for h5py.h5a.create: list assignment index out of range [autodoc]
WARNING: error while formatting signature for h5py.h5a.open: list assignment index out of range [autodoc]
WARNING: error while formatting signature for h5py.h5a.delete: list assignment index out of range [autodoc]
WARNING: error while formatting signature for h5py.h5a.get_info: list assignment index out of range [autodoc]
WARNING: error while formatting signature for h5py.h5o.get_info: list assignment index out of range [autodoc]
```

This looks suspect indeed.

AA-Turner:
Could you run with `-Werror` / `PYTHONWARNINGS=error` & `--show-traceback --exception-on-warning -vv`? It's being helpful by swallowing the errors, but hopefully we can get it to actually quit & show a traceback.

A

neutrinoceros:
you bet. See https://github.com/h5py/h5py/pull/2757

PR #2757, opened 2025-12-04
EXP: debug API docs build with Sphinx 9

neutrinoceros:
This is an experiment to inquire on #2754


AA-Turner:
```pytb
      File "/home/runner/work/h5py/h5py/docs_api/automod.py", line 76
        const_exclude = ["%s(?:\W|$)" % x for x in const_exclude]
                               ^^
    SyntaxError: invalid escape sequence '\W'
```

first error: could you replace it with a raw string (`const_exclude = [r"%s(?:\W|$)" % x for x in const_exclude]`)?

A

neutrinoceros:
ok _now_ I think we're hitting the actual failure mode from Sphinx 9

AA-Turner:
```pytb
  File "/home/runner/work/h5py/h5py/.tox/apidocs/lib/python3.12/site-packages/sphinx/ext/autodoc/_dynamic/_loader.py", line 131, in _load_object_by_name
    signatures = _format_signatures(
                 ^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/h5py/h5py/.tox/apidocs/lib/python3.12/site-packages/sphinx/ext/autodoc/_dynamic/_signatures.py", line 128, in _format_signatures
    signatures[0] = (args, retann if isinstance(retann, str) else '')
    ~~~~~~~~~~^^^
IndexError: list assignment index out of range

...

  File "/home/runner/work/h5py/h5py/.tox/apidocs/lib/python3.12/site-packages/sphinx/util/logging.py", line 412, in filter
    raise SphinxWarning(message)
sphinx.errors.SphinxWarning: error while formatting signature for h5py.h5a.create: list assignment index out of range
```

The relevant lines in Sphinx are https://github.com/sphinx-doc/sphinx/blob/69596331b3680ebfc3b6dfbc60ce86c885423316/sphinx/ext/autodoc/_dynamic/_signatures.py#L117-L128

Do you have code that acts on `'autodoc-process-signature'`?

A

AA-Turner:
It's probably erroring as the signatures list is empty but some result was returned from the event handler. The fix might be as simple as `signatures[:1] = [(args, retann if isinstance(retann, str) else '')]`.

A

neutrinoceros:
I don't know our docs infrastructure well, but I just learned we had a very old patched version of autodoc, so, before I try anything else, I'm going to take a shot in the dark by simply removing it and see what happens.

takluyver:
I think that was a misreading, the `automod.py` file is not itself a patched version of autodoc, rather it's saying that it *requires* a patched version of autodoc to work properly. Though the patch it refers to was applied in Sphinx in 2008, so it's not really 'a patched version' now.

`automod` is doing a bunch of h5py-specific things to work with the peculiarly structured docstrings in h5py's low-level API and cross-link references to h5py specific objects. We definitely want to keep that. Honestly, I'm impressed that a little custom extension like this has worked for over a decade with only a few minor changes (and none at all in the last 5 years).

neutrinoceros:
thanks for clarifying this. I'll update the comment to avoid shoting myself in the foot again !

PR #2758, opened 2025-12-04
BLD: make version parsing robust against NumPy pre-releases

neutrinoceros:
Fix building against a pre-release, [as seen on the latest nightly run](https://github.com/h5py/h5py/actions/runs/19916089296/job/57095022349).
The second commit is an unrelated -strictly speaking- cleanup, and could be separated.

### DIFF
--- a/setup_build.py
+++ b/setup_build.py
@@ -108,14 +108,7 @@ class h5py_build_ext(build_ext):
             settings['library_dirs'].extend(config.msmpi_lib_dirs)
             settings['libraries'].append('msmpi')
 
-        try:
-            numpy_includes = numpy.get_include()
-        except AttributeError:
-            # if numpy is not installed get the headers from the .egg directory
-            import numpy.core
-            numpy_includes = os.path.join(os.path.dirname(numpy.core.__file__), 'include')
-
-        settings['include_dirs'] += [numpy_includes]
+        settings['include_dirs'] += [numpy.get_include()]
         if config.mpi:
             import mpi4py
             settings['include_dirs'] += [mpi4py.get_include()]


neutrinoceros:
(added a third, yet separable, cleanup commit)

tacaswell:
I trust `numpy.__version__` but do not really trust the packaging metadata. `__version__` will always match the code we are actually running whereas the metadata is implementation details of the pip/wheel packaging ecosystem and it should not leak into our runtime or buildtime code paths (any more than we would let details of debian or conda leak in).

neutrinoceros:
the `__version__` attribute is generally an implementation detail too and could be out of sync (granted, that's pretty unlikely). Why do you trust it more than the package metadata ?

neutrinoceros:
Ah, I think I see where you're coming from; the package metadata can indeed easily go out of sync when working with an editable, source installation of numpy. Though, I don't think I would trust `numpy.__version__` either in this context, so I'm still left wondering why you have a preference. (Arguably I do too; I prefer the method that always works, though it's not a strong opinion in this context)

tacaswell:
See https://discuss.python.org/t/please-make-package-version-go-away/58501/90 and links for way more detail than any reasonable person wants ;)

The very short version is `__version__` is controlled by the library and is (in my view) the long standing defacto standard for how a library should report its version at run time.  If it is wrong that is a bug in the library and should be treated as such.

The values in the pip/wheel metadata are strictly derived values there for things that install wheels (e.g. pip and uv) to do their jobs.  We should stay away from picking up any sort of dependency on a particular packaging ecosystem.

I also prefer the method that always works which in my view is `__version__` .

neutrinoceros:
> I also prefer the method that always works which in my view is __version__

this is what I thought before I wrote [`wxc`](https://github.com/neutrinoceros/wxc).

Anyway, thanks for the link and your reply ! happy to go back to `__version__` in the build process and for these two packages.

neutrinoceros:
Coming back to this I don't get why scheduled jobs are still failing; granted, the bug fixed here isn't addressed on `master`, but what surprises me is that numpy `2.4.0rc1` is still being picked of PyPI *over* `2.5.0.dev0` nightlies, which are now available.
I couldn't find a relevant entry on pip's recent changelog so I really don't have an idea what could be causing this. I do hope that `2.5.0.dev0` starts being picked once `2.4.0` (final) is available, and I'll keep an eye on that. 
@tacaswell do you have any reservations regarding merging this now ?

tacaswell (approved):

PR #2760, opened 2025-12-04
Add arguments to File to configure external link access

t20100:
This PR proposes to add 3 arguments (`elink_mode`, `elink_swmr` and `elink_locking`) to `h5py.File` in order to configure the behavior when accessing external links.
This is similar to suggestion: https://github.com/h5py/h5py/issues/2257#issuecomment-1581044576
This is a rework of PR #2751 to propose an alternative (or to complete it) since IMO this proposes an API closer than what would be needed for #2690 and #2257.

As it is, it requires more testing and there is still some issues to tackle:
- SWMR write mode seems to override lapl.
- When `elink_mode="r"`, `create_dataset("external_link/new_data", ...)` fails as expected, but `create_group("external_link/new_group") works. I would expect other cases like this one.

### DIFF
--- a/h5py/tests/test_file.py
+++ b/h5py/tests/test_file.py
@@ -1073,6 +1073,87 @@ def test_file_locking_external_link(tmp_path, locking_arg, file_locking_props):
         assert elink_locking_info == file_locking_props
 
 
+class TestExternalLinkArguments:
+
+    def _create_files(self, tmp_path):
+        external_filepath = tmp_path / make_name("external{}.h5")
+        main_filepath = tmp_path / make_name("main{}.h5")
+
+        with h5py.File(external_filepath, "w", libver="latest") as external_file:
+            external_file.create_group("/group")
+            external_file["/group/data"] = 1, 2, 3
+
+        with h5py.File(main_filepath, "w", libver="latest") as main_file:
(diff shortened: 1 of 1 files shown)


takluyver:
I find the stateless alternative in #2751 more intellectually appealing. The version in this PR, setting parameters when opening a file, is probably more convenient in a lot of cases, though - I imagine there aren't many times that you want to do different things with external links in the same file.

IMO, h5py works well as a relatively low-level Python interface to HDF5, allowing a lot of precise control, and providing a good solid platform for building higher levels of abstraction on top. This makes me lean towards #2751 even if it's not quite as convenient for direct use. But my view might be influenced by my work on [one such higher level abstraction](https://extra-data.readthedocs.io/en/latest/).

t20100:
Here we tend to use `h5py` directly as the API to access data, so I'm influenced the other way around.

> setting parameters when opening a file, is probably more convenient in a lot of cases

Yes that was the idea. At the same time, to be really simple, a single option to open any external files (elink, VDS) in read-only mode would be best.

Let's see how #2751 gets used: this should already save parsing hdf5 entity path to check whether each part is an external link or not.

I'm closing this for now.

Issue #2756, opened 2025-12-04
h5py v4 planning


takluyver:
What should we do in h5py 4.0? These aren't set in stone, but:

- [ ] Read variable-length string data to Numpy variable-width strings by default (from #2557)
  - [ ] Drop support for Numpy 1.x (according to [SPEC-0](https://scientific-python.org/specs/spec-0000/), this is OK from September 2025)
- [ ] Convert numpy complex dtype to HDF5 2.0 native complex datatype by default (from #2694)
  - Though this means that earlier versions of HDF5 won't be able to read the complex data 🤔 
- [ ] Any other disruptive changes to properly support HDF5 2.x
- [ ] Any disruptive changes required for free-threading support

To be clear: this doesn't mean we're doing v4 particularly soon, it's just a place to track & discuss the major changes. More ideas are welcome, but I'm still aiming for a fairly boring release; I don't want to redesign everything. I expect lots of code using h5py will keep working unmodified.

neutrinoceros:
> Drop support for Numpy 1.x (according to [SPEC-0](https://scientific-python.org/specs/spec-0000/), this is OK from September 2025)

two notes here:
- I think this particular change could land on a minor release
- I've been less and less inclined to drop backwards compatibility layer on the sole argument that SPEC 0 allows us to (and we've been supporting Python versions for much longer than SPEC 0 mandates)

At the moment, and as far as I know, our technical debt to support numpy 1 is very manageable, so I would just put it off for now.
At the latest, it should shed itself off when we drop support for Python 3.12, which was the latest version supported by numpy 1, and which will reach EOL about 3 years from now. Realistically, I expect the incentive to preserve backward compatibility will be marginal by then, and we'll probably end up dropping it as it becomes a burden.

neutrinoceros:
(this is a sensitive topic for me at the moment, sorry I come out as needlessly defensive. I understand you didn't suggest we drop it *now*, or even soon, I just wanted to clarify my own goals in that area)

aragilar:
Not that I'm suggesting we do this (or don't do this), but I can see changing build system might be one (depending on if we can continue using the same way of configuring the build)?

I don't think we need to drop numpy 1.x asap, and I think as long as supporting it remains uneventful (e.g. we can build against 2.x and 1.x still is supported), I don't think we need to go out of our way to drop it.

neutrinoceros:
[...]

PR #2766, opened 2026-01-09
Warn on use of h5py.config().read_byte_strings (now a no-op)

takluyver:
The `read_byte_strings` context manage is a no-op now. This adds a warning, prior to getting rid of it in h5py v4 (#2756).

### DIFF
--- a/h5py/h5.templ.pyx
+++ b/h5py/h5.templ.pyx
@@ -120,6 +120,11 @@ cdef class H5PYConfig:
     def read_byte_strings(self):
         """ Returns a context manager which forces all strings to be returned
         as byte strings. """
+        warn(
+            "h5py.get_config().read_byte_strings is deprecated, "
+            "and no longer has any effect.",
+            category=H5pyDeprecationWarning,
+        )
         with phil:
             return self._bytestrings
 


takluyver:
I think this one is straightforward, so if there are no comments I'll merge it next week.

h5py/h5.templ.pyx
```diff
@@ -120,6 +120,11 @@ cdef class H5PYConfig:
     def read_byte_strings(self):
         """ Returns a context manager which forces all strings to be returned
         as byte strings. """
+        warn(
+            "h5py.get_config().read_byte_strings is deprecated, "
+            "and no longer has any effect.",
+            category=H5pyDeprecationWarning,
+        )
```
neutrinoceros, line 127:
```suggestion
            category=H5pyDeprecationWarning,
            stacklevel=2,
        )
```

takluyver: This is generally what we want, but in this case it points to the right line without setting this, and the wrong one if I do set it. I presume it's because this is Cython code - either it doesn't count as a proper stack frame, or Cython is doing something behind the scenes that affects warnings.

neutrinoceros: oh wow I completely missed that. TIL

PR #2770, opened 2026-01-13
Fix passing a low-level TypeID object for dtype in g.create_dataset()

### LINKED ISSUE #2576: Saving a byte array in h5py is not the same as saving a byte array in HDF.PInvoke
I am saving a byte array in C# using HDF.PInvoke:

```
status = H5D.write(arrID, H5T.NATIVE_B8, H5S.ALL, H5S.ALL, H5P.DEFAULT, ptr);
```

and saving a byte array using h5py:
```
buffer = np.array(..., dtype=np.uint8)
f.create_dataset("name", data=buffer)
```

Loading either the C# or python dataset is ok from h5py.

Loading the python generated dataset in C# gives this problem:

```

HDF5-DIAG: Error detected in HDF5 (1.10.6) thread 972:
  #000: C:\projects\hdf-pinvoke-1-10\CMake-hdf5-1.10.6\hdf5-1.10.6\src\H5Dio.c line 199 in H5Dread(): can't read data
    major: Dataset
    minor: Read failed
  #001: C:\projects\hdf-pinvoke-1-10\CMake-hdf5-1.10.6\hdf5-1.10.6\src\H5Dio.c line 467 in H5D__read(): unable to set up type info
    major: Dataset
    minor: Unable to initialize object
  #002: C:\projects\hdf-pinvoke-1-10\CMake-hdf5-1.10.6\hdf5-1.10.6\src\H5Dio.c line 985 in H5D__typeinfo_init(): unable to convert between src and dest datatype
    major: Dataset
    minor: Feature is unsupported
  #003: C:\projects\hdf-pinvoke-1-10\CMake-hdf5-1.10.6\hdf5-1.10.6\src\H5T.c line 4546 in H5T_path_find(): can't find datatype conversion path
    major: Datatype
    minor: Can't get value
  #004: C:\projects\hdf-pinvoke-1-10\CMake-hdf5-1.10.6\hdf5-1.10.6\src\H5T.c line 4762 in H5T__path_find_real(): no appropriate function for conversion path
    major: Datatype
    minor: Unable to initialize object

```

I imagine it has something so do with my choice of NATIVE_B8 as a type on the C# side.

Checking the type in python is not particularly illuminating both the C# and python generated datasets are '|u1'.

I can't find any documentation on what hdf5 datatypes the h5py datatypes correspond to.

Any information on what's causing this would be most appreciated!

takluyver:
Passing a low-level TypeID object when creating a dataset (e.g. `f.create_dataset('b', dtype=h5py.h5t.STD_B16LE, shape=(12,))`) works, but it's the same as passing e.g. `h5py.h5t.STD_B16LE.dtype`, i.e. the equivalent Numpy dtype. For most types this will just roundtrip back to the same HDF5 datatype, but not bitfields, since they translate to numpy uints.

This would also fail for non-standard int and float formats which don't have a numpy equivalent; either with an outright error, or using a different format to what was specified.

Is this a bugfix for the next 3.x release? Or is it a breaking change for 4.0? I don't think anyone would intentionally rely on this translation of TypeID objects to dtype and back, but I could imagine some code that happens to work because of this quirk. :thinking: 

Closes #2576.

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -17,6 +17,7 @@ from abc import ABC, abstractmethod
 
 import numpy
 
+import h5py.h5t
 from .. import h5, h5s, h5t, h5r, h5d, h5p, h5fd, h5ds, _selector
 from .base import (
     array_for_new_object, cached_property, Empty, find_item_type, HLObject,
@@ -79,6 +80,9 @@ def make_new_dset(parent, shape=None, dtype=None, data=None, name=None,
         # Named types are used as-is
         tid = dtype.id
         dtype = tid.dtype  # Following code needs this
+    elif isinstance(dtype, h5py.h5t.TypeID):  # Low-level HDF5 data type
+        tid = dtype
+        dtype = tid.dtype
     else:
         # Validate dtype
         if dtype is None and data is None:
--- a/h5py/tests/test_dtype.py
+++ b/h5py/tests/test_dtype.py
@@ -530,3 +530,22 @@ def test_opaque(writable_file):
     assert isinstance(ds.id.get_type(), h5py.h5t.TypeOpaqueID)
     assert ds.id.get_type().get_size() == 2
     np.testing.assert_array_equal(ds[:], arr)
+
+
+def test_create_bitfield(writable_file):
+    data = np.arange(12, dtype=np.uint16)
+    ds = writable_file.create_dataset(
+        'b1', dtype=h5py.Datatype(h5py.h5t.STD_B16LE), shape=(12,)
+    )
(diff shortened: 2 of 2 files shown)


takluyver:
I think this can go in 3.16 - it's clearly a bug that passing in a specific HDF5 datatype to make a new dataset doesn't always use that datatype, the cases that change are pretty niche, and we're not generally dogmatic about avoiding any possible breaks in 3.x releases, not least because [every change is potentially breaking](https://xkcd.com/1172/). I only thought about this because I'm planning 4.0, but that may still take some time.

tacaswell (approved):

PR #2772, opened 2026-01-14
Add native ppc64le wheel builds alongside existing platforms.

sandeepgupta12:
This PR adds native ppc64le wheel builds to the CI matrix, bringing Power (ppc64le) to parity with existing x86_64, aarch64, macOS, and Windows platforms.

Because official NumPy wheels are not available for ppc64le, the workflow builds NumPy from source in a controlled and policy-aligned manner and reuses those wheels when building h5py.

All existing platforms continue to use the current cibuildwheel flow unchanged.

Happy to adjust the workflow based on CI or community feedback.


.github/workflows/build_wheels.yml
```diff
@@ -121,6 +123,83 @@ jobs:
...
+      
+      - name: Setup Python and system dependencies (ppc64le)
+        if: matrix.arch == 'ppc64le'
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y software-properties-common pkg-config
+          sudo add-apt-repository ppa:deadsnakes/ppa -y
+          sudo apt-get update
+          sudo apt-get install -y \
+            python3.11 python3.11-dev python3.11-venv \
```
neutrinoceros: what's the reason not to include Python 3.10 ?

.github/workflows/build_wheels.yml
```diff
@@ -121,6 +123,83 @@ jobs:
...
+            pip install -U pip setuptools wheel meson ninja meson-python Cython build
+            
+            python -m build --wheel --no-isolation -v
+            cp dist/*.whl ../numpy_wheels/
+            deactivate
+            rm -rf build dist .mesonpy-* _skbuild venv$PY
+            cd ..
+          }
+          # NumPy 2.4.x for Python 3.11
+          git clone --depth 1 --branch v2.4.1 --recurse-submodules https://github.com/numpy/numpy.git numpy-2.4.1
```
neutrinoceros: I would prefer using a commit sha rather than a branch name here

.github/workflows/build_wheels.yml
```diff
@@ -121,6 +123,83 @@ jobs:
...
+          build_numpy 3.11 numpy-2.4.1
+          # NumPy main for Python >= 3.12
+          git clone --depth 1 --recurse-submodules https://github.com/numpy/numpy.git numpy-main
```
neutrinoceros: Is there a reason not to use numpy 2.4.1 for all versions ? I would strongly prefer the build process to be reproducible

.github/workflows/build_wheels.yml
```diff
@@ -121,6 +123,83 @@ jobs:
...
+          for PY in 3.12 3.13 3.14; do
+            build_numpy $PY numpy-main
+          done
+      
+      - name: Setup venv for h5py build
+        if: matrix.arch == 'ppc64le'
+        run: |
+          python3.12 -m venv venv
+          source venv/bin/activate
+          pip install -U pip setuptools wheel cibuildwheel
```
neutrinoceros: do we actually need pip, setuptools and wheel to be upgraded ?
I would expect cibuildwheel to resolve our build dependencies by itself, especially seeing that Cython didn't make it here and the job kept running.

neutrinoceros (commented):
Thank you ! An overarching question is whether you're also working with numpy to get wheels out there ?

sandeepgupta12:
> Thank you ! An overarching question is whether you're also working with numpy to get wheels out there ?

Yes we are working on it.

.github/workflows/build_wheels.yml
sandeepgupta12: Thanks @neutrinoceros for reviewing this PR. 
I tested Python 3.10, but NumPy 2.4.x fails to build because meson-python now requires Python ≥3.11. Supporting Python 3.10 would require pinning an older NumPy version and adding a separate build path, so it’s intentionally excluded here to keep the workflow aligned and simple.

Happy to explore Python 3.10 with an older NumPy version if the project considers that useful.

takluyver:
Oh, I assumed your merged PR on Numpy meant they were already publishing ppc64le wheels, but it's actually only testing, not releasing wheels.

I think doing this in h5py is essentially dependent on Numpy publishing wheels first. There's not a lot of point h5py having ppc64le wheels if Numpy doesn't have wheels, and I don't want the extra complexity and slowness of building Numpy from source for our CI. The conda packages of h5py on conda-forge have ppc64le builds which people can use in the meantime.

sandeepgupta12:
Hi @takluyver @neutrinoceros ,

Thanks for the clarification — I agree that publishing h5py ppc64le wheels is effectively dependent on NumPy wheels being available, and that building NumPy from source in the main h5py CI would add undesirable complexity and latency.

To address that, one possible approach is to leverage the existing hdf5-manylinux container image setup. Since that repository already builds the manylinux container images used for h5py wheel builds (with HDF5 layered on top of the standard manylinux images), we could extend those images to also build NumPy ppc64le wheels inside the container.

Those prebuilt NumPy wheels could then be reused during the h5py wheel build jobs, so the h5py CI itself would not need to build NumPy from source. This keeps the additional complexity isolated to the container image build, avoids slowing down the main CI, and provides a practical way to enable h5py ppc64le wheels until NumPy publishes official wheels upstream.

neutrinoceros:
This seems like a reasonable approach to me.

takluyver:
To be honest, my preference is just to leave it alone for now, and not attempt to test on ppc64le until there are Numpy wheels. The point about adding complexity still stands if we add it in the hdf5-manylinux repo - it's currently very simple, basically only needing changes when a new HDF5 version comes out. If it also has to build Numpy for several Python versions, it's suddenly a rather more involved project.

I don't think we're very likely to accidentally break ppc64le compatibility. It didn't really come up when we had ppc64le testing on Travis CI - that's why we didn't worry too much about removing it when it became too flaky - and we seemingly haven't broken it since then. conda-forge is already building on ppc64le, so there's some means to catch bugs if they do arise.

To my mind, including ppc64le in our own CI is only worthwhile if we're going to release wheels from it, and that only makes sense if Numpy does so first.

sandeepgupta12:
Hi @takluyver, thanks for the explanation — I understand the motivation to keep CI and the manylinux setup simple and low‑maintenance.
From an implementation standpoint, one possible way to enable ppc64le wheel builds without introducing NumPy source builds into h5py CI is to consume prebuilt NumPy wheels from a devpi index. This is a workflow already used in practice by teams on IBM Power systems: NumPy wheels are built reproducibly from upstream sources once and then reused downstream, avoiding repeated source builds in each project’s CI. IBM documents this approach here:
https://community.ibm.com/community/user/blogs/janani-janakiraman/2025/09/10/developing-apps-using-python-packages-on-ibm-power

Concretely, this would mean:
- no NumPy source builds in h5py CI
- CI changes scoped only to ppc64le (e.g. configuring an alternate index)
- reuse of the existing h5py wheel build flow
- minimal ongoing maintenance, similar to consuming any other external wheel source

If this approach sounds acceptable, I can propose a small, concrete change showing how h5py CI could install NumPy from a devpi index for ppc64le and proceed with wheel builds.

tacaswell:
@sandeepgupta12 That would be acceptable to pull wheels from a non-pypi source.  Would you want to use our CI to push h5py wheels back or are those built through some other mechanism?

I would also like an agreement that IBM will support these CI jobs if we have problems going forward (with the backup plan of "turn it off again").

sandeepgupta12:
Hi @tacaswell , thanks for the confirmation.

- For ppc64le, h5py CI will build h5py wheels as usual, but it will consume prebuilt NumPy wheels from a devpi index instead of building NumPy from source for ppc64le.

- Devpi publishing is handled via a separate IBM‑managed mechanism; h5py CI will not push wheels to devpi. From the CI perspective, the only change is the NumPy install source for ppc64le.

- IBM will support any Power/ppc64le‑related issues, and we’re aligned on having a clear fallback plan to disable these jobs if needed.

- I will update this PR with the devpi configuration details and the backup/disable plan.

More details on the devpi workflow are documented here:
https://community.ibm.com/community/user/blogs/janani-janakiraman/2025/09/10/developing-apps-using-python-packages-on-ibm-power

Thanks

sandeepgupta12:
HI @takluyver @neutrinoceros @tacaswell 

I’ve pushed the updates to the PR now.
Concretely, the changes configure h5py CI to consume prebuilt NumPy wheels from an IBM‑managed devpi index for ppc64le only, while keeping the existing wheel build flow unchanged for all other architectures.

Regarding CI robustness and support: IBM will own support for any Power/ppc64le‑specific issues that arise from this setup. The changes are deliberately scoped so that disabling the ppc64le jobs remains straightforward should we need to fall back in the future.

Please review the changes and let me know if you have any feedback or would like adjustments.

.github/workflows/build_wheels.yml
```diff
@@ -122,6 +128,33 @@ jobs:
...
+        run: |
+          # Install cibuildwheel
+          python3 -m pip install --upgrade pip setuptools wheel
+          python3 -m pip install cibuildwheel
+          
+          # Build wheels
+          python3 -m cibuildwheel --output-dir wheelhouse
+        env:
+          CIBW_ARCHS_LINUX: ppc64le
+          CIBW_BUILD: "cp311-manylinux_ppc64le cp312-manylinux_ppc64le cp313-manylinux_ppc64le cp314-manylinux_ppc64le"
```
neutrinoceros: Why is Python 3.10 not included ? In any case, there are much more compact ways to write this
```suggestion
          CIBW_BUILD: "cp3{11,12,13,14}-manylinux_ppc64le"
```

neutrinoceros (commented):
I guess a general question I have is: do we have a good reason to use `CIBW_*` environment variables instead of static configuration that'd live in `pyproject.toml` with the rest of our cibw config ? I'm 99% confident we could do all this with [overrides](https://cibuildwheel.pypa.io/en/stable/configuration/#overrides)

sandeepgupta12:
> I guess a general question I have is: do we have a good reason to use `CIBW_*` environment variables instead of static configuration that'd live in `pyproject.toml` with the rest of our cibw config ? I'm 99% confident we could do all this with [overrides](https://cibuildwheel.pypa.io/en/stable/configuration/#overrides)

Thanks for the review. Agreed — there’s no strong reason to keep using CIBW_* env vars here; this can (and should) be expressed cleanly via cibuildwheel overrides in pyproject.toml for better clarity and maintainability. The only remaining item is a NumPy wheel issue on ppc64le, which depends on an upcoming DevPi release; I’m working with the DevPi team on this and will follow up once it’s resolved.

.github/workflows/build_wheels.yml
sandeepgupta12: @neutrinoceros, I have applied your suggestions. Currently I have skipped python 3.10 as I notice one PR is opened to drop python 3.10 support.

neutrinoceros: sounds good

sandeepgupta12:
@neutrinoceros 
Regarding the static check failure: The zizmor security audit is flagging cache poisoning errors on lines 98 and 116 of .github/workflows/build_wheels.yml for the existing HDF5 cache configuration (Windows and macOS). These lines are not modified by this PR and predate the PPC64LE changes.

https://dev.azure.com/h5pyappveyor/h5py/_build/results?buildId=3557&view=logs&j=c4851542-858c-5aa0-9714-b02cde6c1394&t=230ae624-38df-5377-5338-7a322a9283fe&l=444

sandeepgupta12:
@neutrinoceros, Also can you please enable these org permission to run the ppc64le jobs? https://github.com.mcas.ms/IBM/actionspz/blob/main/docs/onboarding.md#for-organization-admins-update-runner-groups

I see that following job waiting for runner due to this- https://github.com/h5py/hdf5-manylinux/actions/runs/27741828771/job/82070623823

neutrinoceros:
> @neutrinoceros, Also can you please enable these org permission to run the ppc64le jobs?

Sorry I still don't have org-level permissions.

> Regarding the static check failure: The zizmor security audit is flagging cache poisoning errors on lines 98 and 116 of .github/workflows/build_wheels.yml for the existing HDF5 cache configuration (Windows and macOS). These lines are not modified by this PR and predate the PPC64LE changes.

right, but these diagnostics are ignored in kind of a flimsy way. You'll need to update `.github/zizmor.yaml` with the new line numbers.
See https://github.com/h5py/h5py/pull/2863#discussion_r3105656761 for why I chose this workflow (I had to search for it to refresh my own memory).

sandeepgupta12:
@neutrinoceros , can you please review the changes and merge this PR.

sandeepgupta12:
@takluyver, can you please enable these org permission to run the ppc64le jobs? https://github.com.mcas.ms/IBM/actionspz/blob/main/docs/onboarding.md#for-organization-admins-update-runner-groups

sandeepgupta12:
@neutrinoceros , can you please review this PR once again?

.github/workflows/build_wheels.yml
```diff
@@ -4,6 +4,10 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.number }}-${{ github.event.ref }}
   cancel-in-progress: true
 
+env:
+  # Feature flags for platform builds
+  ENABLE_PPC64LE: true  # Set to false to skip ppc64le builds
```
neutrinoceros: is this used anywhere else than this workflow ? If not, maybe we could simply remove it as it is logically inert at the moment.

.github/workflows/build_wheels.yml
```diff
@@ -140,6 +149,14 @@ jobs:
 
       # Now actually build the wheels
       - uses: pypa/cibuildwheel@294735312765b09d24a2fbec22660ce817587d55 # v4.1.0
+        if: matrix.arch != 'ppc64le'
+
+      # Build wheels for ppc64le (config in pyproject.toml overrides)
```
neutrinoceros: I'm not sure I understand why this is needed, so let me ask a couple mundane questions. Is this the only way to run cibuildwheel without a container so we control the image/arch ? Is it the case that cibuildwheel won't auto-install build dependencies when run in this fashion ?
Whatever the answers, I think it'd make sense to add inline comments here to explain your reasoning.

pyproject.toml
```diff
@@ -131,17 +131,51 @@ wheels = [
 [tool.cibuildwheel]
 skip = [
     "cp310-win_arm64", # Skipped due to unavailability of Python binary and NumPy for Win-ARM64
+    "cp314t-manylinux_ppc64le", # Skipped due to unavailability of Python binary and NumPy for manylinux_ppc64le
+    "cp310-*linux_ppc64le", # Build only Python 3.11+ for ppc64le
+    "*-musllinux_ppc64le", # Skip all musllinux builds for ppc64le
```
neutrinoceros: ok, but why ?

neutrinoceros (commented):
a couple superficial questions but overall LGTM now, sorry for the long delay

.github/workflows/build_wheels.yml
sandeepgupta12: Good point, it's not used anywhere else in the repo. Removed in the latest commit.

.github/workflows/build_wheels.yml
sandeepgupta12: The pypa/cibuildwheel action hardcodes an actions/setup-python step internally that can't be skipped, and actions/setup-python has no prebuilt Python binaries for ppc64, so it always fails. Calling python3 -m cibuildwheel directly bypasses that and uses the Python already on the runner. All config still comes from pyproject.toml as normal. Added inline comments to explain this.

pyproject.toml
sandeepgupta12: I intentionally skipped the following targets:

- `cp310-*manylinux_ppc64le`
  - Python 3.10 was skipped because there is already an PR to drop Python 3.10 support for h5py.

- `cp314t-manylinux_ppc64le`
  - Skipped due to the unavailability of Python 3.14t binaries and compatible NumPy packages for `manylinux_ppc64le`.

- `*-musllinux_ppc64le`
  - Skipped all `musllinux` builds for `ppc64le` as the required toolchain and dependency ecosystem are currently not available/supported for this architecture.

sandeepgupta12:
> a couple superficial questions but overall LGTM now, sorry for the long delay

Thanks. I have updated the PR with your suggestion. Please have a look.

pyproject.toml
neutrinoceros: sure, it's all fine. What I meant to suggest is to make sure these reasons are recorded as inline comments, but that last one currently isn't.

sandeepgupta12:
> @takluyver, can you please enable these org permission to run the ppc64le jobs? https://github.com.mcas.ms/IBM/actionspz/blob/main/docs/onboarding.md#for-organization-admins-update-runner-groups

Hi @takluyver, can you please perform the steps so that ppc64le CI jobs are run?

pyproject.toml
sandeepgupta12: done.

neutrinoceros (approved):

sandeepgupta12:
Hi @neutrinoceros @takluyver, 
The ppc64le GitHub Actions job has been received by GitHub, but the IBM Power Runner is not picking it up due to runner group permission restrictions.

Could you please verify and perform the runner group onboarding steps described in the IBM Actions documentation? Specifically:

For Organization Admins: Update Runner Groups

1. Navigate to Organization Settings > Actions > Runner groups.
2. Select the Default group.
3. Update the following settings:

- [ ] Check "Allow public repositories" to enable it. (This is necessary to allow forks of repositories to execute CI workflows via Pull Requests.)
- [ ] Ensure "Repository access" is set to "All repositories" (or verify the specific repositories are included in the "Selected repositories" list).

At present, the workflow is queued successfully, but the runner assignment appears to be blocked by the organization-level runner group configuration rather than any issue in the workflow itself.

**Documentation**: https://github.com/IBM/actionspz/blob/main/docs/onboarding.md#for-organization-admins-update-runner-groups

PR #2775, opened 2026-01-16
Warn on creating a dataset with neither dtype nor data

takluyver:
We plan to require one of these in h5py 4.0 (see #2756).

Of course, one application which does create many datasets without really caring about the dtype is h5py's own test suite, so most of the diff is fixing tests.


tacaswell:
It is probably worth adding one test that exercises the warning.

PR #2777, opened 2026-01-20
Threads+fork test fails on 3.15t

crusaderky:
I'm observing this test failing frequently on 3.15a4t (without pytest-run-parallel) with

> E           DeprecationWarning: This process (pid=51457) is multi-threaded, use of fork() may lead to deadlocks in the child.

e.g. https://github.com/crusaderky/hdf5-pixi/actions/runs/21150734896/job/60826052716

I have not tested it on 3.14t extensively enough to figure out if this is a change in 3.15 or not. What we know is that it never happens on 3.14 with GIL, according to the h5py CI.

The warning is from 2022: https://github.com/python/cpython/pull/100229
Reading the code suggests that the warning should be deterministic in this test (assuming that fork takes less than 1 second to complete, which should hold almost always true even on the slow CI boxes). I can't figure out why it only happens occasionally, and why only on free-threading builds.

### DIFF
--- a/h5py/tests/test_objects.py
+++ b/h5py/tests/test_objects.py
@@ -7,8 +7,6 @@
 # License:  Standard 3-clause BSD; see "license.txt" for full license terms
 #           and contributor agreement.
 import os
-import sys
-import threading
 
 import pytest
 
@@ -42,47 +40,21 @@ class TestObjects(TestCase):
 
     @pytest.mark.thread_unsafe(reason="fork() from a thread may deadlock")
     @pytest.mark.skipif(not hasattr(os, "fork"), reason="os.fork() not available")
-    @pytest.mark.skipif(sys.version_info >= (3, 15), reason="os.fork() deadlocks")
-    def test_phil_fork_with_threads(self):
+    def test_phil_fork(self):
         # Test that handling of the phil Lock after fork is correct.
-        # We simulate a deadlock in the forked process by explicitly
-        # waiting for the phil Lock to be acquired in a different thread
-        # before forking.
-
-        thread_acquired_phil_event = threading.Event()
-        thread_stop_event = threading.Event()
-
-        def f():
-            o.phil.acquire()
-            try:
-                thread_acquired_phil_event.set()
-                thread_stop_event.wait()
-            finally:
+        # Note that threading and fork() are mutually exclusive, so
+        # the use case of a locked phil Lock during a fork() is unsupported.
+        pid = os.fork()
+        if pid == 0:
+            # child process
+            if o.phil.acquire(blocking=False):
                 o.phil.release()
-
-        thread = threading.Thread(target=f)
-        thread.start()
-        try:
-            # wait for the thread running "f" to have acquired the phil lock
-            thread_acquired_phil_event.wait()
-
-            # now fork the current (main) thread while the other thread holds the lock
-            pid = os.fork()
-            if pid == 0:
-                # child process
-                # If we handle the phil lock correctly, this should not deadlock,
-                # and we should be able to acquire the lock here.
-                if o.phil.acquire(blocking=False):
-                    o.phil.release()
-                    os._exit(0)
-                else:
-                    os._exit(1)
+                os._exit(0)
             else:
-                # parent process
-                # wait for the child process to finish
-                _, status = os.waitpid(pid, 0)
-                assert os.WIFEXITED(status)
-                assert os.WEXITSTATUS(status) == 0
-        finally:
-            thread_stop_event.set()
-            thread.join()
+                os._exit(1)
+        else:
+            # parent process
+            # wait for the child process to finish
+            _, status = os.waitpid(pid, 0)
+            assert os.WIFEXITED(status)
+            assert os.WEXITSTATUS(status) == 0
--- a/h5py/tests/test_objects.py
+++ b/h5py/tests/test_objects.py
@@ -43,8 +43,6 @@ class TestObjects(TestCase):
     @pytest.mark.thread_unsafe(reason="fork() from a thread may deadlock")
     @pytest.mark.filterwarnings(
         # https://github.com/python/cpython/pull/100229
-        # The warning was introduced in Python 3.12 but for some reason it only starts
-        # appearing in Python 3.15
         r"ignore:.*use of fork\(\) may lead to deadlocks:DeprecationWarning"
     )
     @pytest.mark.skipif(not hasattr(os, "fork"), reason="fork() not available")
@@ -85,6 +83,7 @@ class TestObjects(TestCase):
                     os._exit(1)
             else:
                 # Parent process
+                assert not o.phil.is_locked()
                 # Wait for the child process to finish
                 _, status = os.waitpid(pid, 0)
                 assert os.WIFEXITED(status)
--- a/h5py/tests/test_objects.py
+++ b/h5py/tests/test_objects.py
@@ -7,6 +7,8 @@
 # License:  Standard 3-clause BSD; see "license.txt" for full license terms
 #           and contributor agreement.
 import os
+import sys
+import threading
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


neutrinoceros (approved):
It took me a couple minutes to understand what was happening but when it clicked, it was extremely satisfying. This is a clear improvement, and I would go as far as calling the original test's implementation as buggy and unreliable.

>  I can't figure out why it only happens occasionally, and why only on free-threading builds.

too bad, but no big deal. Your fix is clearly correct. I'll merge once CI goes green. Thank you !

crusaderky:
Once you suppress the warning, 3.15a4 deadlocks very very fast.
I've changed the PR to skip Python 3.15 and later instead.

neutrinoceros (approved):
still looks goods to me. We never meant to support forking in the first place, so if it breaks with future Python versions, I don't think we can do anything about it any way.

tacaswell (dismissed):
This seems like there is a bug in the way we are handling phil with freethreading that is being identified by this test?

crusaderky:
> This seems like there is a bug in the way we are handling phil with freethreading that is being identified by this test?

I'm seeing an issue which causes a `threading.RLock` ([which is how phil is implemented on free-threading](https://github.com/h5py/h5py/blob/52e35798cfd0cfa38c15b60fc38da39c0653972c/h5py/_objects.templ.pyx#L40-L45)), which was defined as a Cython 3.1 module attribute, to deadlock on Python 3.15a5t if you spawn a thread, acquire the lock inside the thread, and then fork from the main thread.

Changing to a `threading.RLock` defined inside a regular python module changes the behaviour to a failure to acquire the RLock in the forked process.

Mixing fork() and threads is explicitly unsupported by CPython, so I don't think this warrants opening an upstream issue (I guess in Cython?)

## Minimal reproducer
```python
import os
import threading

from h5py._objects import phil
# Uncommenting this line changes behaviour from deadlock on fork()
# to failure to acquire phil in the subprocess
# phil = threading.RLock() 
assert type(phil) is type(threading.RLock())


def main():
    thread_acquired_phil_event = threading.Event()
    thread_stop_event = threading.Event()

    def f():
        phil.acquire()
        try:
            thread_acquired_phil_event.set()
            thread_stop_event.wait()
        finally:
            phil.release()

    thread = threading.Thread(target=f)
    thread.start()
    # wait for the thread running "f" to have acquired the phil lock
    thread_acquired_phil_event.wait()
    assert not phil.acquire(blocking=False)

    try:

        print("forking")
        # now fork the current (main) thread while the other thread holds the lock
        pid = os.fork()
        if pid == 0:
            print("forked to child process")
            # child process
            # If we handle the phil lock correctly, this should not deadlock,
            # and we should be able to acquire the lock here.
            if phil.acquire(blocking=False):
                print("acquired phil in child process")
                phil.release()
                print("released phil in child process")
                os._exit(0)
            else:
                print("failed to acquire lock in child process")
                os._exit(1)
        else:
            print("parent process after fork; child pid =", pid)
            # parent process
            # wait for the child process to finish
            _, status = os.waitpid(pid, 0)
            assert os.WIFEXITED(status)
            assert os.WEXITSTATUS(status) == 0
    finally:
        thread_stop_event.set()
        thread.join()


if __name__ == "__main__":
    main()
```

crusaderky:
CPython limitations aside, The feature that this test is verifying is fundamentally wrong.

The whole point of the phil lock is to maintain a coherent global state in libhdf5.
This test says that, while thread A is busy updating the hdf5 global state (holding the phil lock), thread B can call fork, which will snapshot the transient, half-updated global state and feed it to another process, where the cloned phil lock must release itself, leaving the user open to a broken state.

I think this test should be heavily reworked. The only feature that IMHO is sane to test is that it's ok to fork an _unlocked_ global phil lock, e.g. the typical single-threaded use case.

neutrinoceros:
thanks for the detailed explanation. I'm on board with changing the test as proposed.

crusaderky:
> thanks for the detailed explanation. I'm on board with changing the test as proposed.

done

takluyver:
The test was added just last year in #2585 - I'd like to ask @ArvidJB to have a look at this PR and comment, so that the h5py maintainers don't end up playing telephone between different viewpoints.

> This test says that, while thread A is busy updating the hdf5 global state (holding the phil lock), thread B can call fork, which will snapshot the transient, half-updated global state and feed it to another process, where the cloned phil lock must release itself, leaving the user open to a broken state.

I believe you've missed something here. In this scanario, thread B is meant to acquire the phil lock in an atfork handler, so there are no HDF5 library calls happening at the moment of forking. This stuff is admittedly easy to get wrong, and we may still be getting it wrong, but I don't think it's wrong in the way you describe.

IDK what to do about the larger issue of threads+fork not being supported from Python. I understand that it's officially never been supported, but for a long time it has worked well enough the people (including me) could and did use it, although it seems to be becoming a bigger issue in more recent Python versions, for reasons I don't understand.

takluyver:
> thread B is meant to acquire the phil lock in an atfork handler,

I think this also makes sense of why the test doesn't always trigger Python's warning about fork() - forking is delayed until the thread started in the test releases the lock, so the fork is only racing against the thread terminating.

crusaderky:
...right. I now understand the original intent. I'll revert to the second commit and add an explanation.
I'll investigate if this is a genuine regression that impacts _all_ `os.register_at_fork` calls in Python 3.15a5t, or is for some reason specific to when threads are involved.

ArvidJB:
> thread B is meant to acquire the phil lock in an atfork handler, so there are no HDF5 library calls happening at the moment of forking.

Yes, that's exactly the intent of the change. This test should not deadlock even without the GIL? I know that the `FastRLock` relies on the GIL, but the free-threading implementation is not using that, no?

tacaswell:
We went with a `threading.RLock()` out of expediency.  Maybe we do need to implement our own after all that is robust to this case (with the added benefit of possibly being faster)?

crusaderky:
> Once you suppress the warning, 3.15a4 deadlocks very very fast. I've changed the PR to skip Python 3.15 and later instead.

This was because I erroneously added a double-event system and removed the sleep. You can't do that because of fork internally acquiring the phil lock.

I think this PR is now ready for a final review.
it is still unknown why the warning from python 3.12 only appears in 3.15.

h5py/tests/test_objects.py
```diff
@@ -42,37 +41,38 @@ def test_hash(self):
             hash(oid)
 
     @pytest.mark.thread_unsafe(reason="fork() from a thread may deadlock")
+    @pytest.mark.filterwarnings(
+        # https://github.com/python/cpython/pull/100229
+        r"ignore:.*use of fork\(\) may lead to deadlocks:DeprecationWarning"
+    )
```
crusaderky, line 47: this is the only material change in this PR now. Everything else is just minor quality QOL and documentation.

ArvidJB:
Changes look good to me, thanks @crusaderky!

> it is still unknown why the warning from python 3.12 only appears in 3.15.

Probably this change? https://github.com/python/cpython/commit/fd8f42d3d1038a812340c3ec3cbfc995a80c4e13

neutrinoceros (approved):
In retrospect, I've approved previous states prematurely, so I'll abstain from merging here, but FWIW I'm on board with the current implementation too

h5py/tests/test_objects.py
```diff
@@ -42,37 +41,38 @@ def test_hash(self):
...
-        # waiting for the phil Lock to be acquired in a different thread
-        # before forking.
-
-        # On Windows forking (and the register_at_fork handler)
-        # are not available, skip this test.
-        if not hasattr(os, "fork"):
-            raise SkipTest("os.fork not available")
+        """Test that handling of the phil Lock after fork is correct.
 
+        h5py changes os.fork() to internally acquire the phil lock before forking
```
tacaswell: We are using public API https://docs.python.org/3/library/os.html#os.register_at_fork to do this.

This documentation is deeply misleading and makes it sound like we are monkey patching the std modules 🤣 .

takluyver:
Thanks everyone. :slightly_smiling_face: I agree with @tacaswell that the docstring is misleading - we're not modifying `os.fork()` - but once that's rephrased I think this looks good.

h5py/tests/test_objects.py
crusaderky: I've updated the wording.

takluyver:
Thanks @crusaderky !

PR #2778, opened 2026-01-20
Make `write_direct_chunk` thread-safe

crusaderky:
This race condition was detected by `pytest-run-parallel` segfaulting on  `test_write_direct_chunk`.

### DIFF
--- a/h5py/h5d.templ.pyx
+++ b/h5py/h5d.templ.pyx
@@ -534,6 +534,7 @@ cdef class DatasetID(ObjectID):
     @with_phil
     @cython.boundscheck(False)
     @cython.wraparound(False)
+    @with_phil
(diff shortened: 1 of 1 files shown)


h5py/h5d.templ.pyx
```diff
@@ -534,6 +534,7 @@ cdef class DatasetID(ObjectID):
     @with_phil
     @cython.boundscheck(False)
     @cython.wraparound(False)
+    @with_phil
```
neutrinoceros, line 537: the decorator is already present on line 534. Does it not compound correctly with cython decorators ?

crusaderky: Whoops. Merge conflict. This PR duplicates #2764.

PR #2779, opened 2026-01-20
MNT: customize RTD build process to utilize uv

neutrinoceros:
Rationale: I'm seeing traces of an attempt to pin doc dependencies to exact versions on RTD (`docs/requirements-std.txt`). However, this approach only includes *direct* dependencies, leaving out any transitive ones, which means we can still get stealth upgrades there, and are generally at risk of builds suddenly breaking.
The general solution is to use a lock file, and we could pretty easily do that with uv (which we already use in CI via tox-uv), so I'm proposing, as a first step towards this goal, to configure RTD to use uv instead of pip. This configuration is directly taken from [RTD docs](https://docs.readthedocs.com/platform/stable/build-customization.html#install-dependencies-with-uv)

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,16 +1,22 @@
 version: 2  # Required
 
 build:
-  os: ubuntu-20.04
+  os: ubuntu-24.04
   tools:
-    python: "3.10"
+    python: '3'
+  jobs:
+   pre_create_environment:
+     - asdf plugin add uv
+     - asdf install uv latest
+     - asdf global uv latest
+   create_environment:
+     - uv venv "${READTHEDOCS_VIRTUALENV_PATH}"
+   install:
+     # note that uv sync should be run as `--frozen`, but this requires uv.lock to be committed to the repo
+     - UV_PROJECT_ENVIRONMENT="${READTHEDOCS_VIRTUALENV_PATH}" uv sync --only-group docs
 
 sphinx:
    configuration: docs/conf.py
    fail_on_warning: true
-
-python:
-   install:
-   - requirements: docs/requirements-rtd.txt
 
 formats: all
--- a/docs/requirements-rtd.txt
+++ b/docs/requirements-rtd.txt
@@ -1,2 +1,0 @@
-sphinx==7.2.6
-sphinx_rtd_theme==1.3.0
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -101,7 +101,8 @@ check-readme = [
     "twine>=6.1.0",
 ]
 docs = [
-    "sphinx>=7.4.7, <9",
+    "sphinx>=7.4.7,<9",
+    "sphinx-rtd-theme>=1.3.0",
 ]
 lint = [
     "pre-commit>=4.2.0",


takluyver:
I'd rather not customise what RTD does like this - it's one of the big advantages of RTD that you don't have to write yet another set of instructions in YAML, you can describe what the docs are in mostly declarative terms and it knows how to build them.

I'm not against using a lock file style approach to doc dependencies, though. There's a well established pattern for doing something similar using pip by generating a `requirements.txt` file using [pip-tools](https://pypi.org/project/pip-tools/) - it's not a full lock file, but it gives us ~95% of the benefit, and this is [what RTD recommends](https://docs.readthedocs.com/platform/stable/guides/reproducible-builds.html#pin-your-transitive-dependencies). Lock files were officially standardised last year ([PEP 751](https://peps.python.org/pep-0751/)), so pip and/or RTD might gain more support for them some time soon.

neutrinoceros:
Pip already knows how to read pylock.toml files (uv does too). A disadvantage of this standard format is that it's not universal yet (as of v1), meaning you can lock for a single deployment platform but not have one lockfile work on macOS and windows, for instance. Another issue is that I don't know of an existing tool that would handle upgrades automatically.

requirements files have more support but they have none of the security properties of a lockfile (I guess that falls in the 5 other %). If we're going to use a non standard format I'd rather go with uv.lock, and hope that the gap in feature set with pylock.toml is bridged in the future.

neutrinoceros:
Btw if we were to use a pylock.toml, we'd still need to customize the build very similarly to what I'm doing here. RTD maintainers added hooks precisely because they don't have the capacity to add support for every setup flavor while keeping the configuration minimal, so I doubt that more specific support will be added beyond this.

takluyver:
Oh, I didn't know pip had already adopted the lockfile format. Given what you say about cross platform support, I think I still prefer `requirements.txt` for this, though.

>  If we're going to use a non standard format I'd rather go with uv.lock,

Sorry, but I'd still rather not. I want to stick with the tools that RTD offers already if that's remotely practical, and it clearly is here. The simple list of `pkg==version` requirements is also a de-facto standard format, even if it's not formally specified, so long as you don't use the ability to pass pip command line options in the file.

neutrinoceros:
what about the broader picture ? I was actually also thinking of locking *most* CI jobs by re-using the same uv.lock (combined with automated upgrade PRs), so that only the bleeding-edge job is responsible for finding out about incoming incompatibilities in our stack.

takluyver:
Separate question, IMO. The docs requirements are distinct from the general CI dependencies (building the main docs doesn't actually require building h5py).

More generally... I dislike the current trend for Python tools written in Rust. I understand why people are doing it - "blazing fast :rocket:" - and I've [played with](https://github.com/takluyver/sqlite-glance) Rust myself - but I think we're ignoring the longer term value of having open source Python tools written in Python, and thus relatively easy for Python developers to modify. We can probably wring a fair bit more speed out of Python based tools where people focus on that (e.g. [this recent post](https://iscinumpy.dev/post/packaging-faster/)), especially if we sprinkle in a bit of compiled code for the hottest parts. But I recognise that I'm an old fogey, and my opinions are now cutting against the grain of how everyone else does things. :shrug:

neutrinoceros:
>  I understand why people are doing it - "blazing fast 🚀" 

Since we're on the topic, I should note that while enjoyable, I believe performance isn't actually the main reason to use another language for (some) Python tools: uv, amongst many other functionalities, can manage python interpreters installation, upgrade, and isolation. This would not be practical if the tool itself required a working interpreter.
But in other cases, sufficient performance actually seems impossible to achieve in Python; I'm thinking of Python LSPs. I think VSCode's Pylance used to have Python components, and as a result was putting enormous amounts of pressure on hardware resources. Meanwhile, `ty` and `pyrefly` (both written in rust) run with a fraction of the requirement.

In any case, it's not why I'm proposing uv.lock here (or elsewhere); I'm just trying to reduce the amount of pressure I personally feel with the many packages I maintain where CI is always one external change away from collapsing. I'll gladly admit that h5py hasn't been noticeably unstable in that regard (thank `<deity>`), though I am trying to aim at a cultural shift.

Anyway. I suppose settling for `requirements.txt` would still be better than not locking at all.

would the following work for you ?
```
uv pip compile --group docs -o docs/requirements.txt
```

<details><summary> output</summary>

```
# This file was autogenerated by uv via the following command:
#    uv pip compile --group docs -o docs/requirements.txt
alabaster==1.0.0
    # via sphinx
babel==2.17.0
    # via sphinx
certifi==2026.1.4
    # via requests
charset-normalizer==3.4.4
    # via requests
docutils==0.21.2
    # via sphinx
idna==3.11
    # via requests
imagesize==1.4.1
    # via sphinx
jinja2==3.1.6
    # via sphinx
markupsafe==3.0.3
    # via jinja2
packaging==25.0
    # via sphinx
pygments==2.19.2
    # via sphinx
requests==2.32.5
    # via sphinx
roman-numerals==4.1.0
    # via roman-numerals-py
roman-numerals-py==4.1.0
    # via sphinx
snowballstemmer==3.0.1
    # via sphinx
sphinx==8.2.3
    # via h5py (pyproject.toml:docs)
sphinxcontrib-applehelp==2.0.0
    # via sphinx
sphinxcontrib-devhelp==2.0.0
    # via sphinx
sphinxcontrib-htmlhelp==2.1.0
    # via sphinx
sphinxcontrib-jsmath==1.0.1
    # via sphinx
sphinxcontrib-qthelp==2.0.0
    # via sphinx
sphinxcontrib-serializinghtml==2.0.0
    # via sphinx
urllib3==2.6.3
    # via requests
```

</details>

neutrinoceros:
> We can probably wring a fair bit more speed out of Python based tools where people focus on that 

(FTR, I'm hugely favorable to this approach too)

takluyver:
> uv, amongst many other functionalities, can manage python interpreters installation, upgrade, and isolation

Thanks, that is a good point. I realise I sounded rather negative about uv & friends yesterday. I'm not that against them, and if they can simplify our CI, that would be great. But I've been around packaging stuff long enough that the appearance of a new wave of tooling ready to solve all our problems tends to make me feel more weary than excited.

> `uv pip compile --group docs -o docs/requirements.txt`

Where would the input for this live? If it's a standard location like the `[optional-dependencies]` or `[dependency-groups]` table in pyproject.toml, great. If it's something uv-specific, I'd prefer to go with a familiar `requirements.in` input file for now; it could be changed later if we do adopt uv more broadly.

neutrinoceros:
> I've been around packaging stuff long enough that the appearance of a new wave of tooling ready to solve all our problems tends to make me feel more weary than excited.

I understand your position. To be honest, I'll still be a bit worried about the future of `uv` as long as astral doesn't publicly announce they're making a profit and can cut VC investments, but not *that* worried, considering the licensing of their tools is permissive enough that the community could potentially take over if needed.

> Where would the input for this live?

Oh, sorry I wasn't clear about that: the command I proposed, together with its output, are not proof of concept; it already works exactly as proposed because `dependency-groups.docs` is defined since #2604. All I'd need to do is to commit the output to version control (and set up dependabot for auto upgrades).

takluyver:
Aha, thank-you! Yes, in that case I'm fine with compiling the docs requirements from the dependency group, and checking it in to git. :+1:

neutrinoceros:
cool. Closing in favor of #2786

PR #2782, opened 2026-01-21
Use PyDict_DelItem to remove objects from registry

### LINKED ISSUE #2771: Memory leak in h5py 3.15 (for attrs and dataset)
While investigating a memory leak in a larger app, I realized that h5py 3.15 leaks memory when overwriting an attribute and when writing to a dataset. Actually there may be other occurrences but those are the one tested by the included example.

Using this example on 3.15, I observe leakage of the order of 10 MB for a 100 iteration when modifying an attribute. On 3.14, the memory usage very quickly stabilize after an initial increase of less than 1 MB.

I am willing to help but I would need some pointer as to where to start.

I ran the tests on a Windows machine under Python 3.12.

```python
import gc
import pathlib
import tempfile
import tracemalloc
import uuid

import h5py
import psutil

ATTR_SET_ITEM = True
ATTR_MODIFY = False
DATASET_WRITING = False

counter = 0
d = pathlib.Path(tempfile.gettempdir())
tracemalloc.start()
snapshot = tracemalloc.take_snapshot()
while True:
    if counter % 100 == 0:
        print(f"{counter}: Memory used {psutil.Process().memory_full_info().uss / 1e6}")
        new = tracemalloc.take_snapshot()
        if counter != 0:
            for s in new.compare_to(snapshot, "lineno")[:5]:
                print(f"  {s}")
        snapshot = new

    counter += 1
    p = d / f"{uuid.uuid4()}.h5"
    with h5py.File(p, mode="a") as f:
        if ATTR_MODIFY:
            f.attrs["a"] = 0
        if DATASET_WRITING:
            f.create_dataset("d", (200,), int)
        for i in range(200):
            if ATTR_SET_ITEM:
                f.attrs["a"] = i
            if ATTR_MODIFY:
                f.attrs.modify("a", i)
            if DATASET_WRITING:
                f["d"][i] = i
        pass

    gc.collect()

```

h5py version info
```
h5py    3.15.1
HDF5    1.14.6
Python  3.12.11 (main, Jun 12 2025, 12:44:17) [MSC v.1943 64 bit (AMD64)]
sys.platform    win32
numpy   2.4.1
cython (built with) 3.1.4
numpy (built against) 2.3.4
HDF5 (built against) 1.14.6
```

takluyver:
This is option 1 for fixing the memory leak describe in #2771. I don't understand exactly why it fixes it - I guess there must be a bug somewhere in Cython, or even CPython, on the generic code path for deleting an item from a collection.

I'm hoping to prepare a second option as well, in a few minutes, so hold off merging for now.

Closes #2771.

### DIFF
--- a/h5py/_objects.templ.pyx
+++ b/h5py/_objects.templ.pyx
@@ -12,6 +12,7 @@
 """
 
 include "_locks.pxi"
+from cpython.dict cimport PyDict_DelItem
 from .defs cimport *
 
 import os
@@ -239,7 +240,7 @@ cdef class ObjectID:
                         )
                     )
             if self._pyid is not None:
-                del registry[self._pyid]
+                PyDict_DelItem(registry, self._pyid)
 
     def _close(self):
         """ Manually close this object. """


takluyver:
#2783 is the alternative PR.

The version here is a smaller diff, but I think we can get rid of some complexity - or rather, shift it onto the Python standard library - with the other option.

takluyver:
We went with #2783 instead of this.

Issue #775, opened 2016-11-14
"dictionary changed size during iteration" with Python 3


pitrou:
This is happening randomly to us on some Travis CI builds.
See full traceback here:
https://travis-ci.org/dask/distributed/jobs/175780263#L1763

It happens when h5py tries to remove a "murdered identifier" when closing. `registry.items()` returns an iterator on Python 3, it should be wrapperd in a `list()`.

(what is a "murdered identifier"?)

See also https://github.com/h5py/h5py/issues/534#issuecomment-231385966

pitrou:
What happens is certainly that a GC run is triggered during `nonlocal_close()`. Some ObjectID is then reaped (probably it was dangling on a reference cycle) and removed from the registry. I don't think there is any point in issueing a warning because of that.

shoyer:
I've been getting occasional errors like this when running the xarray test suite locally using Python 3 / OS X. Thanks @pitrou for tracking this down!

PR #2783, opened 2026-01-21
Use a WeakValueDictionary for the object registry

### LINKED ISSUE #2771: Memory leak in h5py 3.15 (for attrs and dataset)

takluyver:
This is an alternative to #2782, fixing the same memory leak.

This one is a bigger change, using the [WeakValueDictionary](https://docs.python.org/3/library/weakref.html#weakref.WeakValueDictionary) class, which automatically discards its own entries when they're garbage collected, rather than managing a regular dict of weakref objects ourselves. That simplifies the code in h5py, albeit relying on more complexity in Python. I think this trade-off is worth it - weakrefs are fiddly to work with, and the Python developers probably have the best chance of getting it right.

Closes #2771.

### DIFF
--- a/h5py/_objects.templ.pyx
+++ b/h5py/_objects.templ.pyx
@@ -118,21 +118,18 @@ import weakref
 import warnings
 
 # Will map id(obj) -> weakref(obj), where obj is an ObjectID instance.
-# Objects are added only via ObjectID.__cinit__, and removed only by
-# ObjectID.__dealloc__.
-cdef dict registry = {}
+# Objects are added only via ObjectID.__cinit__, and dropped by the weakref system
+cdef object registry = weakref.WeakValueDictionary()
 
 @with_phil
 def print_reg():
     import h5py
-    refs = registry.values()
-    objs = [r() for r in refs]
+    objs = list(registry.values())
 
-    none = len([x for x in objs if x is None])
     files = len([x for x in objs if isinstance(x, h5py.h5f.FileID)])
     groups = len([x for x in objs if isinstance(x, h5py.h5g.GroupID)])
 
-    print("REGISTRY: %d | %d None | %d FileID | %d GroupID" % (len(objs), none, files, groups))
+    print("REGISTRY: %d | %d FileID | %d GroupID" % (len(objs), files, groups))
 
 
 @with_phil
@@ -142,30 +139,7 @@ def nonlocal_close():
     cdef ObjectID obj
     cdef list reg_ids
 
-    # create a cached list of ids whilst the gc is disabled to avoid hitting
-    # the cyclic gc while iterating through the registry dict
-    gc_was_enabled = gc.isenabled()
-    gc.disable()
-    try:
-        reg_ids = list(registry)
-    finally:
-        if gc_was_enabled:
-            gc.enable()
-
-    for python_id in reg_ids:
-        ref = registry.get(python_id)
-
-        # registry dict has changed underneath us, skip to next item
-        if ref is None:
-            continue
-
-        obj = ref()
-
-        # Object died while walking the registry list, presumably because
-        # the cyclic GC kicked in.
-        if obj is None:
-            continue
-
+    for python_id, obj in registry.items():
         # Locked objects are immortal, as they generally are provided by
         # the HDF5 library itself (property list classes, etc.).
         if obj.locked:
@@ -218,7 +192,7 @@ cdef class ObjectID:
             ### {{if OBJECTS_DEBUG_ID}}
             print("CINIT - registering %d of kind %s HDF5 id %d" % (self._pyid, type(self), self.id))
             ### {{endif}}
-            registry[self._pyid] = weakref.ref(self)
+            registry[self._pyid] = self
 
 
     def __dealloc__(self):
@@ -226,7 +200,7 @@ cdef class ObjectID:
 
     # During interpreter shutdown, module attributes are set to None
     # before __dealloc__ and __del__ methods are executed.
-    def _dealloc(self, _phil=_phil, warn=warnings.warn, registry=registry):
+    def _dealloc(self, _phil=_phil, warn=warnings.warn):
         with _phil:
             ### {{if OBJECTS_DEBUG_ID}}
             print("DEALLOC - unregistering %d HDF5 id %d" % (self._pyid, self.id))
@@ -238,8 +212,6 @@ cdef class ObjectID:
                             self.id
                         )
                     )
-            if self._pyid is not None:
-                del registry[self._pyid]
 
     def _close(self):
         """ Manually close this object. """
--- a/h5py/_objects.templ.pyx
+++ b/h5py/_objects.templ.pyx
@@ -120,10 +120,28 @@ import warnings
 # Objects are added only via ObjectID.__cinit__, and dropped by the weakref system
 cdef object registry = weakref.WeakValueDictionary()
 
+# These 2 functions are part of a bugfix from Python 3.14. Before this,
+# a race condition meant that an entry could be removed from wvd.data while
+# another thread was iterating over it. We can remove these and use the
+# WeakValueDict methods once 3.14 is the minimum supported version.
+# https://github.com/python/cpython/issues/89967
+def wvd_values(self):
+    for wr in self.data.copy().values():
+        obj = wr()
+        if obj is not None:
+            yield obj
+
+def wvd_items(self):
+    for k, wr in self.data.copy().items():
+        v = wr()
+        if v is not None:
+            yield k, v
+
+
 @with_phil
 def print_reg():
     import h5py
-    objs = list(registry.values())
+    objs = list(wvd_values(registry))
 
     files = len([x for x in objs if isinstance(x, h5py.h5f.FileID)])
     groups = len([x for x in objs if isinstance(x, h5py.h5g.GroupID)])
@@ -138,7 +156,7 @@ def nonlocal_close():
     cdef ObjectID obj
     cdef list reg_ids
 
-    for python_id, obj in registry.items():
+    for python_id, obj in wvd_items(registry):
         # Locked objects are immortal, as they generally are provided by
         # the HDF5 library itself (property list classes, etc.).
         if obj.locked:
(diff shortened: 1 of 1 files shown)


neutrinoceros:
At a surface level, this does seem like the better approach to me. I can review in greater depth tomorrow to make sure I understand the details enough to form an informed opinion.

MatthieuDartiailh:
Anything that could be done to help merge this and get a new release ?

h5py/_objects.templ.pyx
```diff
@@ -113,26 +113,22 @@ if hasattr(os, "register_at_fork"):
...
+cdef object registry = weakref.WeakValueDictionary()
 
 @with_phil
 def print_reg():
     import h5py
-    refs = registry.values()
-    objs = [r() for r in refs]
+    objs = list(registry.values())
 
-    none = len([x for x in objs if x is None])
```
neutrinoceros, line 131: I don't understand this part. Is this some unrelated, fly-by clean up , or is it actually part of your fix ? either is fine, of course

h5py/_objects.templ.pyx
```diff
@@ -142,30 +138,7 @@ def nonlocal_close():
...
-            continue
-
-        obj = ref()
-
-        # Object died while walking the registry list, presumably because
-        # the cyclic GC kicked in.
-        if obj is None:
-            continue
-
+    for python_id, obj in registry.items():
```
neutrinoceros: lovely

neutrinoceros (approved):
@MatthieuDartiailh you just did ! I'm sorry I forgot about this, my system for reminding myself of things I promised to do appears to be broken at the moment. I confirm that I'm happy to get behind this patch. I just have a (minor) question for @takluyver

h5py/_objects.templ.pyx
takluyver: It's related. 🙂 

When we were managing a dict of weakrefs directly, resolving `ref()` gives None if the object it's pointing to has already been destroyed. So `print_reg()` included a count of objects which had been destroyed but not removed from the registry. That ought to be impossible, so I guess it was there to catch bugs in this mechanism.

With `WeakValueDictionary()`, entries conceptually disappear as soon as the value has destroyed. The actual implementation is more complex, but every method to get objects out (e.g. [WeakValueDict.values()](https://github.com/python/cpython/blob/6543720b63a62363de540deb5a6701f1ae431bce/Lib/weakref.py#L198-L202)) has `is not None` checks to skip over destroyed-but-not-yet-cleaned-up entries.

takluyver: > That ought to be impossible

Actually, I was wrong about this. If gc runs between retrieving the references from the dict and resolving them to objects, you could see None here without it being a bug.

aragilar:
I think this is preferable over #2782, though I'm a little concerned we may be missing something given from what I could see weakref has been around since Python 2.1.

takluyver:
This PR is in a tricky, old and crucial bit of h5py, so I'll wait a couple more days to give @tacaswell a chance to look at it as well. That's not to disregard your review, @neutrinoceros (thanks!) but it's the kind of thing where Tom might notice or remember something that I've overlooked. And bugs in the reference logic can have baffling effects and be difficult to debug, so it's worth taking it slow and getting another pair of eyes on it, IMO.

I'm feeling fairly positive about this, change, though - I think the likelihood is that it gets merged next week. I'm hoping to make a 3.16 release in the fairly near future.

h5py/_objects.templ.pyx
neutrinoceros: Awesome, thanks for the detailed explanation!

h5py/_objects.templ.pyx
```diff
@@ -238,8 +211,6 @@ cdef class ObjectID:
                             self.id
                         )
                     )
-            if self._pyid is not None:
-                del registry[self._pyid]
```
aragilar, line 242: I can't think of any reason why this might make a difference (even though it does seem to fix the leak), but I think we've implicitly switched from before `__dealloc__` is finalised to after `__dealloc__` is, in case that has some implications.

aragilar:
I can't see any things in https://github.com/python/cpython/blob/3.14/Lib/weakref.py#L92 that would concern me, and it does look like that code is smarter around the GC than the gc disable/enable logic we have (https://github.com/python/cpython/blob/3.14/Lib/weakref.py#L105-L110). I also can't see a issue where either Python or HDF5 drop a reference and it's not handled by the other as it is currently.

I'm somewhat surprised we didn't make this change previously (the current registry only dates from 2014, so `WeakValueDictionary` could have been used then), as having looked at the history, we'd have avoided a few bugs in the logic over the years. Hopefully this means any bugs related to free-threading (or any other major changes) that would result from our custom code get fixed before we even see them.

I'm happy to wait for @tacaswell to do a final review though.

takluyver:
@aragilar Good point about the history, thanks! This has sent me down a rabbit hole...

h5py previously had a registry based on WeakValueDictionary way back when, and it was changed away from that in 2011 (b8ab902d4b79bfa28736a9fceb7cf80887c3a20f) because of thread safety concerns.

In 2014 (#454), Andrew totally rewrote the registry. Before that, each HDF5 identifier was meant to have exactly 1 low-level h5py wrapper, and the registry was there to find them and avoid creating duplicates. The new version - the basis of what we still use now - allows multiple low-level wrappers of each HDF5 ID, using HDF5's reference counting. The remaining need for a global registry is that when you close a file, we need to find all the h5py objects which were opened from that file and clear their `.id` field before the HDF5 ID can be reused for a new object.

The newer version of the registry avoids the original threading issue, because the removal of out-of-scope objects is no longer critical. They still need to be removed to prevent a memory leak, but exactly when isn't important, because invalidated entries are ignored.

However, while I was looking at the history of the WeakValueDictionary class, I spotted there was a fix for [another threading issue](https://github.com/python/cpython/issues/89967) in Python 3.14. I think that bug could result in a RuntimeError if you call `file.close()` on one thread (triggering `nonlocal_close()`), while GC is running in another thread. You'd have to be pretty unlucky, but...

So I think we either need to go with #2782 instead, or backport part of the fix from Python 3.14 to avoid this issue. I'll do the latter in this PR for comparison.

aragilar:
Oh interesting, that was the same bug we saw as #775, I hadn't dug into old versions of `weakref.py`. So I guess we'd have hit #775 either way.

takluyver:
Yup! I think it would be much rarer to hit this variant - WeakValueDictionary has tried to protect against GC-while-iterating since Python 2.7. But it checked 'is someone iterating' before doing the removal, so it's possible for an iterator to start between the check and the operation.

The new fix relies on `dict.copy()` being atomic. I think that's guaranteed by the GIL in all Python versions before 3.14, and it [looks like it's ensured](https://github.com/python/cpython/blob/6543720b63a62363de540deb5a6701f1ae431bce/Objects/dictobject.c#L4181-L4185) for free-threading builds too.

takluyver:
@tacaswell asked me via a separate channel:

> a quick concern, does this now force us back up to the interpreter more often?

I think it might. WeakValueDictionary hooks up [a Python callback](https://github.com/python/cpython/blob/cb1dc91dcb1eb6637d1c79b34a0bab728939f717/Lib/weakref.py#L105-L110) to remove entries when the object is finalised. In the current code (and in #2782) this is done by Cython code in the `__dealloc__` method.

I'm not sure how much this matters, though. The `__dealloc__` method needs to be called with the GIL (if Python is using the GIL), so we don't avoid that either way. There is probably some overhead in having a callback and executing bytecode, but hopefully not much.

To get a sense of it, I timed these functions:

```python
def f3():  # ~ 0.55 µs
    obj = MyObject()
    d[id(obj)] = weakref.ref(obj)
    del d[id(obj)]

def f4():  # ~ 1.2 µs
    obj = MyObject()
    wvd[id(obj)] = obj

def f_h():  # ~ 2.6 µs (on h5py 3.15.1)
    space = h5py.h5s.SpaceID(sid)
```

So if you do nothing but create and destroy h5py wrappers of HDF5 IDs, this PR might make it around 25% slower. ((1.2 - 0.55) / 2.6). I don't think any realistic use case comes close to that, though.

tacaswell (approved):

PR #2785, opened 2026-01-22
Get rid of obsolete swmr_support variable

takluyver:
We no longer support pre-SWMR versions of HDF5, so this variable was set to True in all cases. It's not part of the public API, so we can get rid of it.


neutrinoceros (approved):
:shipit:

PR #2786, opened 2026-01-22
DEP: pin all docs dependencies exactly in RTD

neutrinoceros:
replace #2779


docs/requirements-rtd.txt
```diff
@@ -1,2 +1,66 @@
...
+    # via sphinx
+requests==2.32.5
+    # via sphinx
+roman-numerals==4.1.0 ; python_full_version >= '3.11'
+    # via roman-numerals-py
+roman-numerals-py==4.1.0 ; python_full_version >= '3.11'
+    # via sphinx
+snowballstemmer==3.0.1
+    # via sphinx
+sphinx==8.1.3 ; python_full_version < '3.11'
```
neutrinoceros, line 37: for reference, I do not know how well dependabot integrates with such a fork. I *do* know that the equivalent in `uv.lock` is properly handled.

docs/requirements-rtd.txt
```diff
@@ -1,2 +1,66 @@
-sphinx==7.2.6
-sphinx_rtd_theme==1.3.0
+# This file was autogenerated by uv via the following command:
+#    uv pip compile -p 3.10 --universal --group docs -o docs/requirements-rtd.txt
```
neutrinoceros, line 2: `-p 3.10` is a shorthand for "make sure the solution includes Python 3.10", otherwise we get a version of Sphinx that requires 3.11 or later, breaking the build. I suppose we could also consider using a more recent version to build docs, but there might be good reasons to stick to an older one that I'm just missing.

takluyver: I suspect we just set a version at some point and didn't update it (except for you bumping it from 3.9 a few months ago). It can probably use a newer version OK, if you want to do another PR for that.

docs/requirements-rtd.txt
takluyver: We'll find out :slightly_smiling_face:

PR #2792, opened 2026-01-27
Better access to compression/filter information (simplified)

### LINKED ISSUE #2161: Dataset.compression is None when using hdf5plugin compressors

takluyver:
This provides `ds.filter_ids` and `ds.filter_names` properties to inspect all of the filters which a dataset uses, and makes `dset.compression` 'unknown' if h5py's 3 built-in compression options are not used, but an unknown filter is.

This is a simplified version of #2180. In that PR, I tried to improve `create_dataset_like()` to handle combining an arbitrary filter pipeline on the source dataset and h5py keyword arguments which could add/replace/remove filters. That added a bunch of complexity and API surface for something that just doesn't come up very much. You can create a dataset with exactly the same settings as another, including third-party filters, by dipping slightly into the low-level API: `src_ds.id.get_create_plist()`.

Closes #2161.

### DIFF
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -703,7 +703,7 @@ Reference
         module, while filter IDs from plugins are listed in `HDF Group's registry
         <https://portal.hdfgroup.org/display/support/Registered+Filter+Plugins>`_.
 
-        .. versionadded:: 3.8
+        .. versionadded:: 3.16
 
     .. attribute:: fillvalue
 
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -685,6 +685,21 @@ Reference
 
         Whether Fletcher32 checksumming is enabled (T/F).  See :ref:`dataset_fletcher32`.
 
+    .. attribute:: filter_ids
+                   filter_names
+
+        The numeric filter IDs and the string names (as stored in the file) of
+        the filters in use. Each attribute is a tuple.
+
+        Filters are mostly used to compress data, but can also do things like
+        checksumming (see :ref:`dataset_compression`). Other attributes listed
+        above provide convenient shortcuts to check on common filters.
+        IDs for filters built into h5py can be found in the :mod:`h5py.h5z`
+        module, while filter IDs from plugins are listed in `HDF Group's registry
+        <https://portal.hdfgroup.org/display/support/Registered+Filter+Plugins>`_.
+
+        .. versionadded:: 3.8
+
     .. attribute:: fillvalue
 
         Value used when reading uninitialized portions of the dataset, or None
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -667,6 +667,11 @@ Reference
         String with the currently applied compression filter, or None if
         compression is not enabled for this dataset.  See :ref:`dataset_compression`.
 
+        This only recognises the built-in compression options ``'gzip'``,
+        ``'lzf'`` and ``'szip'``. Other compression mechanisms will show as
+        ``'unknown'`` from h5py 3.8. Use :attr:`filter_ids` and
+        :attr:`filter_names` to get more complete information.
+
     .. attribute:: compression_opts
 
         Options for the compression filter.  See :ref:`dataset_compression`.
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -669,7 +669,7 @@ Reference
 
         This only recognises the built-in compression options ``'gzip'``,
         ``'lzf'`` and ``'szip'``. Other compression mechanisms will show as
-        ``'unknown'`` from h5py 3.8. Use :attr:`filter_ids` and
+        ``'unknown'`` from h5py 3.16. Use :attr:`filter_ids` and
         :attr:`filter_names` to get more complete information.
 
     .. attribute:: compression_opts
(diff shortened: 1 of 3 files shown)


takluyver:
To be clear, code like this:

```python
ds1 = f.create_dataset('a', data=..., compression=hdf5plugin.Zstd())
ds2 = f.create_dataset_like('b', ds1)
```

currently runs without an error, but ds2 is uncompressed (because `ds1.compression is None`). With this PR, it will instead fail with an error, which IMO is preferable. This was what the extra complexity in #2180 was trying to deal with.

The error can be worked around by specifying `compression=` again, or specifying `compression=None, compression_opts=None` if removing the compression is intentional.

docs/high/dataset.rst
```diff
@@ -685,6 +690,21 @@ Reference
...
+        the filters in use. Each attribute is a tuple.
+
+        Filters are mostly used to compress data, but can also do things like
+        checksumming (see :ref:`dataset_compression`). Other attributes listed
+        above provide convenient shortcuts to check on common filters.
+        IDs for filters built into h5py can be found in the :mod:`h5py.h5z`
+        module, while filter IDs from plugins are listed in `HDF Group's registry
+        <https://portal.hdfgroup.org/display/support/Registered+Filter+Plugins>`_.
+
+        .. versionadded:: 3.8
```
tacaswell:
```suggestion
        .. versionadded:: 3.16
```

docs/high/dataset.rst
```diff
@@ -685,6 +690,21 @@ Reference
 
         Whether Fletcher32 checksumming is enabled (T/F).  See :ref:`dataset_fletcher32`.
 
+    .. attribute:: filter_ids
+                   filter_names
```
tacaswell, line 694: Do stacked attributes like this work?

takluyver: Yup 🙂 :

Image


https://h5py--2792.org.readthedocs.build/en/2792/high/dataset.html

docs/high/dataset.rst
```diff
@@ -685,6 +690,21 @@ Reference
...
+
+        The numeric filter IDs and the string names (as stored in the file) of
+        the filters in use. Each attribute is a tuple.
+
+        Filters are mostly used to compress data, but can also do things like
+        checksumming (see :ref:`dataset_compression`). Other attributes listed
+        above provide convenient shortcuts to check on common filters.
+        IDs for filters built into h5py can be found in the :mod:`h5py.h5z`
+        module, while filter IDs from plugins are listed in `HDF Group's registry
+        <https://portal.hdfgroup.org/display/support/Registered+Filter+Plugins>`_.
```
ajelenak: Should be https://github.com/HDFGroup/hdf5_plugins/blob/master/docs/RegisteredFilterPlugins.md

tacaswell: Don't we want to point to the rendered version?

tacaswell (approved):
but will hold off merging due to the discussion about where the link should point.

docs/high/dataset.rst
takluyver: HDF group's own pages (e.g. https://support.hdfgroup.org/documentation/ ) point to the Github link, so it looks like that's canonical. Thanks @ajelenak !

PR #2796, opened 2026-02-02
Clean up spurious print in tests

crusaderky:
Make sure that `pytest -s` doesn't produce any spurious output.
The `-s` flag, notably, is needed when running with TSAN or ASAN (https://github.com/crusaderky/hdf5-pixi for automation) to dump out their error message to the console.


takluyver:
I suspect that running pytest with `--capture=sys` would capture Python-level prints but not the output from tools like these. I don't necessarily see it as a problem to leave some `print()` calls in tests for things like the filename handling, to provide extra info if it fails. But it's not a big diff, and I'm not going to lose sleep over it.

PR #2799, opened 2026-02-10
Require mpi4py 4.0.0 at build time on Python <= 3.12

### LINKED ISSUE #2797: Parallel HDF5 installation broken for Python 3.12 and less
I'm running into issues doing a source install of h5py with support for parallel HDF5 wherein it fails to install mpi4py. The root cause seems to be the following:
- setuptools 81 (released two days ago) removed the `dry-run` parameter
- Installation of mpi4py versions <4 still uses the `dry-run` parameter and fails
- When installing h5py with Python versions 3.10-12, the mpi4py version is pinned to 3.1.x: https://github.com/h5py/h5py/blob/65fe5c9c7d500785bff09db42828473bb561aaee/_custom_build/backend.py#L24-L26
- Thus, the source install of h5py on these versions fails when trying to build mpi4py (see [this example](https://github.com/openmc-dev/openmc/actions/runs/21802905601/job/62901345650))

I'm not sure what the best solution or workaround would be, but would appreciate any advice!

takluyver:
I think this should be a simple solution. It possibly makes it slightly harder to build h5py with mpi4py 3.x, but I think it's common to skip build isolation for building with MPI support anyway, so this shouldn't be a big issue.

Closes #2797

### DIFF
--- a/_custom_build/backend.py
+++ b/_custom_build/backend.py
@@ -21,9 +21,10 @@ def get_requires_for_build_wheel(config_settings=None):
     # these build requirements for any reason.
     if os.getenv('HDF5_MPI') == 'ON' and os.getenv('H5PY_SETUP_REQUIRES') != '0':
         requires.extend([
-            "mpi4py ==3.1.2; python_version=='3.10.*'",
-            "mpi4py ==3.1.4; python_version=='3.11.*'",
-            "mpi4py ==3.1.6; python_version=='3.12.*'",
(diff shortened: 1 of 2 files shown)


aragilar:
This seems like the best we can do for the setuptools issue (and hopefully we don't need to do another bump due to setuptools any time soon), and I suspect those who need an older version of mpi4py either have it pre-installed, or are on an old system where newer packages in general will be a challenge.

I'm happy to merge this (assuming it passes).

neutrinoceros (approved):
Works for me

PR #2800, opened 2026-02-10
MNT: restore testing VS mpi4py<4

neutrinoceros:
Reverts h5py/h5py#2799
(I'll push another commit in a couple minutes)

I failed to remember earlier that we were *already* using `uv` as a build frontend in CI via `tox-uv`, so it should be possible to override mpi4py's build time dependencies instead of moving the target

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -169,4 +169,4 @@ no-build-package = ["numpy"]
 extra-build-dependencies = { mpi4py = ["setuptools<81"] }
 
 [tool.uv.sources]
-cython = { git = "https://github.com/neutrinoceros/cython" }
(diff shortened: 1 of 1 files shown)


neutrinoceros:
I *also* forgot about this (apparently still present) issue with this setup on Azure (which is also where the patch is needed)
```
fatal error C1083: Cannot open compiler generated file: '': Invalid
```
I suspect this might be a bug in Cython. I'm inquiring.

neutrinoceros:
I'm getting a *different* error with my patched Cython

```
LINK : fatal error LNK1104: cannot open file
```
I think I'm on a correct track, hunting for ways empty strings can end up in file lists within Cython.

PR #2808, opened 2026-03-06
Bundle zlib 1.3.2 on Windows & Mac

takluyver:
I'm also trying to build zlib the same way on Windows ARM64 that we do on x64, to simplify things a bit.


ci/get_zlib_windows.sh
neutrinoceros, line 1: if this file isn't used just for amd64 now, it'd help to also rename it

takluyver: Yup, I've got an un-pushed commit to do that. Just trying to understand the various CI failures.

NeoKish:
@takluyver don't know if it helps with ci issues. but in 1.3.2, for windows they have changed the library name form zlib.lib to z.lib

takluyver:
Thanks, that does help. :+1:

neutrinoceros (approved):
nice !

takluyver:
:face_exhaling: That took rather more effort than I expected when I started.

PR #2843, opened 2026-03-25
BUG: add missing nullptr checks following malloc calls

neutrinoceros:
close #2833
close #2832
partially address #2831

### DIFF
--- a/h5py/h5fd.templ.pyx
+++ b/h5py/h5fd.templ.pyx
@@ -110,10 +110,11 @@ ctypedef struct H5FD_fileobj_t:
 # parameters (dxpl, type) are ignored.
 
 from cpython cimport Py_INCREF, Py_DECREF
-from libc.stdlib cimport malloc as stdlib_malloc
 from libc.stdlib cimport free as stdlib_free
 cimport libc.stdio
 cimport libc.stdint
+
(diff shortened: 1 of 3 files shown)


neutrinoceros:
ping @devdanzin for review

devdanzin:
LGTM

takluyver:
Can we use our `emalloc` wrapper function for these?

neutrinoceros:
I tried, but hit compilation errors. Since the pattern of checking explicitly for `NULL` was already used elsewhere I figured I would take the path of least resistance.

takluyver:
I'd be curious about what exactly those errors were. I'd guess there are other places where it's not used just because they predate the utility function & were never cleaned up, or were added by people unaware of it.

neutrinoceros:
Thanks for asking. Trying to document my answer lead me to untangle my own confusion. I think the mistake I made in my previous attempt was to use a Python `import` instead of `cimport` for the wrapper, for which my compiler produced less than helpful error messages, but trying again it looks like it actually works.

takluyver:
Looks good, thanks. Yup, we'll need `cimport` to use `cdef` Cython functions - these are compiled to C functions without a Python wrapper.

PR #2844, opened 2026-03-25
BUG: avoid zombie identifiers arising in `ObjectID.__dealloc__`

neutrinoceros:
close #2837
The suggested fix would have made `ObjectID._dealloc` almost identical to `ObjectID._close` with only 2 unimportant (I think) differences:
- a slightly different debug message (unused in prod)
- extensibility beyond the builtin warnings.warn function (unused, as far as I can tell)

### DIFF
--- a/h5py/_objects.templ.pyx
+++ b/h5py/_objects.templ.pyx
@@ -213,36 +213,22 @@ cdef class ObjectID:
 
 
     def __dealloc__(self):
-        self._dealloc()
+        self._close(debug_message="DEALLOC - unregistering")
 
     # During interpreter shutdown, module attributes are set to None
     # before __dealloc__ and __del__ methods are executed.
-    def _dealloc(self, _phil=_phil, warn=warnings.warn):
+    # Here we attach _phil and warnings.warn to the function object as default values
+    # so they remain available during shutdown.
+    def _close(self, _phil=_phil, warn=warnings.warn, debug_message="CLOSE -"):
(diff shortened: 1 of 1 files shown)


takluyver:
The `warn=` parameter isn't about extensibility - the comment just above explains why it's there :wink: :

> During interpreter shutdown, module attributes are set to None before `__dealloc__` and `__del__` methods are executed.

Attaching globals to parameter defaults is 'one weird trick' - their values are attached to the function object, so you can still use them at interpreter shutdown, whereas using globals breaks. People also occasionally do it as a micro-optimisation - accessing local variables is (or was?) faster than globals.

I'd also kind of like to keep the debug message difference. You don't need it often, but when you are debugging something involving references & h5py IDs, details like explicitly closed vs. deallocated can be useful.

I think you're right that the `_close` and `_dealloc` methods can be combined, though.

neutrinoceros:
> the comment just above explains why it's there

you know, I'm sure I read it, but it doesn't have enough context that I could understand it without your additional explanation (thank you !)

> accessing local variables is (or was?) faster than globals.

I think it's still true.

takluyver:
No worries, it does assume a fair bit of knowledge. Feel free to expand it a bit.

neutrinoceros:
Thanks. I rewrote this and expanded the comment.

takluyver:
Thanks, I'm happy with this assuming the CI passes.

PR #2847, opened 2026-03-26
BUG: correctly release resources in `DatasetID.write_direct_chunk` before raising

neutrinoceros:
close #2829

### DIFF
--- a/h5py/h5d.templ.pyx
+++ b/h5py/h5d.templ.pyx
@@ -516,10 +516,10 @@ cdef class DatasetID(ObjectID):
         space_id = H5Dget_space(self.id)
         rank = H5Sget_simple_extent_ndims(space_id)
 
-        if len(offsets) != rank:
-            raise TypeError("offset length (%d) must match dataset rank (%d)" % (len(offsets), rank))
-
         try:
+            if len(offsets) != rank:
+                raise TypeError("offset length (%d) must match dataset rank (%d)" % (len(offsets), rank))
+
             offset = <hsize_t*>emalloc(sizeof(hsize_t)*rank)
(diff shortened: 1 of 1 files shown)


h5py/h5d.templ.pyx
```diff
@@ -517,6 +517,8 @@ cdef class DatasetID(ObjectID):
         rank = H5Sget_simple_extent_ndims(space_id)
 
         if len(offsets) != rank:
+            if space_id:
+                H5Sclose(space_id)
             raise TypeError("offset length (%d) must match dataset rank (%d)" % (len(offsets), rank))
 
         try:
```
takluyver, line 519: Would it work to just move the check inside this `try` block, so the cleanup is done by the `finally`?

neutrinoceros: I think it would, though the idea of using a `raise` within a `try` clause makes me slightly unconfortable, so I'd prefer not to.

takluyver: I don't think there's anything weird about using `raise` inside try/finally? Finally is there to execute some code whether or not an exception happens, that works just as well if the `raise` is directly inside the `try` block as if it's in a function called from that.

neutrinoceros: > I don't think there's anything weird about using raise inside try/finally?

I'm just used to seeing as an anti-pattern because I've seen it abused much more often than used correctly, but really the *whole* pattern is
```py
try:
   ...
   if ...:
      raise MyException
except MyException:
    raise
```
which is just obfuscation...
Any way, enough said about my PTSD, let's roll it

takluyver:
Thanks!

PR #2848, opened 2026-03-26
BUG: always release resources in `DatasetID.get_chunk_info`

neutrinoceros:
close #2830

### DIFF
--- a/h5py/h5d.templ.pyx
+++ b/h5py/h5d.templ.pyx
@@ -631,21 +631,25 @@ cdef class DatasetID(ObjectID):
         cdef unsigned filter_mask
         cdef hid_t space_id = 0
         cdef int rank
+        cdef tuple cot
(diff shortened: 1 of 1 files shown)


h5py/h5d.templ.pyx
```diff
@@ -639,13 +640,16 @@ cdef class DatasetID(ObjectID):
 
         rank = H5Sget_simple_extent_ndims(space_id)
         chunk_offset = <hsize_t*>emalloc(sizeof(hsize_t) * rank)
```
takluyver: I think technically both of these can raise an error as well, so they should be inside the `try`

neutrinoceros: gotcha

takluyver:
Thanks, looks good once the CI passes.

Issue #2816, opened 2026-03-11
CI failure on Windows ARM builds


takluyver:
The Windows ARM builds on Github actions have started failing as soon as they try to import h5py after building & installing it, on Python 3.11:

```
  py311-test-deps-pre: commands[2] C:\a\h5py\h5py\.tox> python -c "from h5py.version import info; print(info)"
  Traceback (most recent call last):
    File "<string>", line 1, in <module>
    File "C:\a\h5py\h5py\.tox\py311-test-deps-pre\Lib\site-packages\h5py\__init__.py", line 25, in <module>
      from . import _errors
  ImportError: cannot import name '_errors' from partially initialized module 'h5py' (most likely due to a circular import) (C:\a\h5py\h5py\.tox\py311-test-deps-pre\Lib\site-packages\h5py\__init__.py)
```

We initially discussed this on #2814. I tried downgrading cibuildwheel in #2815 to see if the recent upgrade was related, but @neutrinoceros was correct that it's not that.

neutrinoceros:
[first bad](https://github.com/h5py/h5py/actions/runs/22885840749/job/66398260713)
[last good](https://github.com/h5py/h5py/actions/runs/22837376087/job/66236395317)

I just checked for the VM image version number, but that's not it either: both runs used `20260209.46.1`

neutrinoceros:
note that even if we pinned `pypa/cibuildwheel` to an exact sha, *its* dependencies are not exactly pinned. Incidentally I just wanted to deploy zizmor scanning there, and now I have an even better motivation to.

neutrinoceros:
👁️ 
- https://github.com/pypa/cibuildwheel/pull/2767

neutrinoceros:
(actually, none of the workflows from the cibw repo is *used* by their action, so my suspicion is moot)

takluyver:
Hmmm, the first bad run you linked fails on Python 3.14t - the free-threaded build - whereas the subsequent failures, e.g. [this run](https://github.com/h5py/h5py/actions/runs/22947251259/job/66602638012) all fail on Python 3.11 (the first version tested).

takluyver:
Numpy 2.4.3 was released between the last good & first bad runs. That doesn't quite fit with why the first bad run only failed on cp314t, but...

neutrinoceros:
if numpy 2.4.3 ends up being the culprit (and I have no idea if it is), that would be an argument in favor of properly locking our CI env.

[...]

PR #2851, opened 2026-03-27
Disable Windows ARM builds

takluyver:
These are failing - see #2816 - and we have so far not identified why. I propose that we disable them until someone can figure it out, making it easier to see when there are other CI failures which we should pay attention to, and saving a little bit of energy.


MGN-GIT:
@khmyznikov fyi

PR #2852, opened 2026-03-28
Fix type handles leaking in check_compound_complex

takluyver:
Check the information from the main type ID first, and if that all matches, get the member types and check their details in turn.

My mistake, I'm too used to writing in a language with automatic memory management.

Closes #2826


neutrinoceros (approved):
nice. Thank you !

PR #2860, opened 2026-04-13
MNT: forbid uv to install mixed-language packages from sources

neutrinoceros:
From a quick scan, I think this is our only other (dev only) dependency that takes significant resources to compile, so it's worth ignoring versions that don't have compatible wheels when resolving.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -163,5 +163,18 @@ H5PY_ROS3 = "0"
 H5PY_DIRECT_VFD = "0"
 
 [tool.uv]
-# never build numpy from source
-no-build-package = ["numpy"]
+# never build the following packages from source
+no-build-package = [
+    "blosc2",
+    "cffi",
+    "charset-normalizer",
+    "cryptography",
+    "markupsafe",
+    "msgpack",
+    "ndindex",
+    "nh3",
+    "numexpr",
+    "numpy",
+    "pyyaml",
+    "tables",
+]


neutrinoceros:
I took a closer look at the dependency tree, and found many more packages that don't just a pure Python wheel. I added all of them too (with the exception of uv itself, which I don't think makes much sense)

neutrinoceros:
also note that it could also be argued that, for security reasons, we don't want to allow arbitrary code to run at install time for *any* third party package (by default), so really source installs should be white listed instead of black listed, but there currently is no good way to achieve this setup in uv.
xref: https://github.com/astral-sh/uv/issues/17416

aragilar:
`mpi4py` is always going to be a bit of a weird case for wheels ([as noted in their install docs](https://mpi4py.readthedocs.io/en/stable/install.html#wheel-packages)). I think for testing using the pre-built wheels makes sense, but I suspect outside demos a source build is required (because all of HDF5, mpi4py and h5py need to be linked with the correct MPI implementation, and those come from the platform and can't be provided by wheels). I guess we can point users to pip (rather than uv) if they need source builds (or does uv provide an easy environment variable to override the config file)?

neutrinoceros:
good point, I should probably remove mpi4py from this list.
But I do need to emphasize that this table only affects dev envs made with uv. In particular
- this is invisible to users (unless they install from the github repo and using uv...)
- no other tool will read this table

This is only useful for CI and local dev envs.

neutrinoceros:
(removed mpi4py)

PR #2863, opened 2026-04-18
MNT: upgrade zizmor pre-commit hook, address new results and adjust settings

neutrinoceros:
- **MNT: move zizmor ignores to a config file to allow dependabot to update version comments**
- **MNT: replace superflous action with gh**
- **MNT: upgrade zizmor pre-commit hook**
- **MNT: add missing cooldown setting**

### DIFF
--- a/.github/zizmor.yaml
+++ b/.github/zizmor.yaml
@@ -1,0 +1,6 @@
+rules:
+  cache-poisoning:
+    ignore:
+      # actions/cache is already disabled for releases
+      - build_wheels.yml:92
+      - build_wheels.yml:110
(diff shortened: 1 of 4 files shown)


.github/zizmor.yaml
```diff
@@ -0,0 +1,6 @@
+rules:
+  cache-poisoning:
+    ignore:
+      # actions/cache is already disabled for releases
+      - build_wheels.yml:92
```
neutrinoceros, line 5: arguably, moving these settings to a different file adds some maintenance burden (they need to be updated when relevant lines change numbers), but I'm taking a bet that it'll be worth the tradeoff of not having to manually edit dependabot's outputs.

neutrinoceros:
I'm planning to merge this one in a couple work days if there are no objections

aragilar:
Is dependabot just changing the github actions hashes?

neutrinoceros:
Currently yes, but when comments only contains version numbers it updates them too.

PR #2865, opened 2026-04-21
Automatically pick ros3 driver for supported URLs

### LINKED ISSUE #2864: Automatic driver selection for S3 URLs
Hello,

Would you be open to a PR introducing automatic `ros3` driver selection when the file is an `s3://` URL ?

The implementation would be something like

```python
class File(Group):
    def __init__(self, name, driver=None):
        if isinstance(name, (str, bytes)) and str(name).startswith("s3://") and driver is None:
            driver = "ros3"
        ...
```

which would run before the config check for S3 support.

The motivation behind this feature is to offer transparent handling of local/remote filenames in S3-enabled builds.

teobouvard:
Closes #2864

### DIFF
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -498,7 +498,7 @@ class File(Group):
         if (
             driver is None
             and isinstance(name, str)
-            and name.startswith(r("http://", "https://", "s3://"))
+            and name.startswith(("http://", "https://", "s3://"))
         ):
             driver = 'ros3'
 
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -495,11 +495,10 @@ class File(Group):
         Additional keywords
             Passed on to the selected file driver.
         """
-        ros3_schemes = ("http://", "https://", "s3://")
         if (
-            isinstance(name, str)
-            and str(name).startswith(ros3_schemes)
-            and driver is None
+            driver is None
+            and isinstance(name, str)
+            and name.startswith(r("http://", "https://", "s3://"))
         ):
             driver = "ros3"
 
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -500,7 +500,7 @@ class File(Group):
             and isinstance(name, str)
             and name.startswith(r("http://", "https://", "s3://"))
         ):
-            driver = "ros3"
+            driver = 'ros3'
 
         if driver == 'ros3':
             if not ros3:
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -495,6 +495,14 @@ class File(Group):
         Additional keywords
             Passed on to the selected file driver.
         """
+        ros3_schemes = ("http://", "https://", "s3://")
+        if (
+            isinstance(name, str)
+            and str(name).startswith(ros3_schemes)
+            and driver is None
+        ):
+            driver = "ros3"
+
         if driver == 'ros3':
             if not ros3:
                 raise ValueError("h5py was built without ROS3 support, can't use ros3 driver")
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -392,6 +392,7 @@ class File(Group):
         driver
             Name of the driver to use.  Legal values are None (default,
             recommended), 'core', 'sec2', 'direct', 'stdio', 'mpio', 'ros3'.
+            If None, the ros3 driver is automatically selected for supported URLs.
         libver
             Library version bounds.  Supported values: 'earliest', 'v108',
             'v110', 'v112', 'v114', 'v200' and 'latest' depending on the
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -496,6 +496,9 @@ class File(Group):
         Additional keywords
             Passed on to the selected file driver.
         """
+        # Automatically select the ros3 driver if it's an S3 URL.
+        # As the other drivers currently do not support HTTP(S), the ros3 driver is
+        # also automatically selected for these schemes.
         if (
             driver is None
             and isinstance(name, str)
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -502,7 +502,7 @@ class File(Group):
         if (
             driver is None
             and isinstance(name, str)
-            and urlsplit(name).scheme in ("http://", "https://", "s3://")
+            and urlsplit(name).scheme in ("http", "https", "s3")
         ):
             driver = 'ros3'
 
--- a/h5py/_hl/files.py
+++ b/h5py/_hl/files.py
@@ -14,6 +14,7 @@
 import inspect
 import os
 import sys
+from urllib.parse import urlsplit
 from warnings import warn
 
 from .compat import filename_decode, filename_encode
@@ -498,7 +499,7 @@ class File(Group):
         if (
             driver is None
             and isinstance(name, str)
-            and name.startswith(("http://", "https://", "s3://"))
+            and urlsplit(name).scheme in ("http://", "https://", "s3://")
         ):
             driver = 'ros3'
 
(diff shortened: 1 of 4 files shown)


h5py/_hl/files.py
```diff
@@ -495,6 +495,14 @@ def __init__(self, name, mode='r', driver=None, libver=None, userblock_size=None
         Additional keywords
             Passed on to the selected file driver.
         """
+        ros3_schemes = ("http://", "https://", "s3://")
+        if (
+            isinstance(name, str)
+            and str(name).startswith(ros3_schemes)
+            and driver is None
+        ):
```
neutrinoceros:
```suggestion
        if (
            driver is None
            and isinstance(name, str)
            and name.startswith(r("http://", "https://", "s3://"))
        ):
```

h5py/_hl/files.py
```diff
@@ -495,6 +495,14 @@ def __init__(self, name, mode='r', driver=None, libver=None, userblock_size=None
...
+            driver = "ros3"
```
neutrinoceros: for symmetry with the next line
```suggestion
            driver = 'ros3'
```

h5py/_hl/files.py
```diff
@@ -495,6 +495,13 @@ def __init__(self, name, mode='r', driver=None, libver=None, userblock_size=None
         Additional keywords
             Passed on to the selected file driver.
         """
+        if (
+            driver is None
+            and isinstance(name, str)
+            and name.startswith(r("http://", "https://", "s3://"))
```
neutrinoceros: Whoops, looks like my suggestion contained a typo

```suggestion
            and name.startswith(("http://", "https://", "s3://"))
```

teobouvard:
@neutrinoceros any blockers on merging this ?

neutrinoceros:
not that I'm aware of, though this isn't my area of expertise and I was hoping someone else would pick this up. Let me ping @takluyver just in case.

h5py/_hl/files.py
```diff
@@ -495,6 +495,13 @@ def __init__(self, name, mode='r', driver=None, libver=None, userblock_size=None
         Additional keywords
             Passed on to the selected file driver.
         """
+        if (
+            driver is None
+            and isinstance(name, str)
+            and name.startswith(("http://", "https://", "s3://"))
```
tacaswell: Are we sure that http and https will only every be used with the ros3 driver?

h5py/tests/test_ros3.py
```diff
@@ -37,6 +45,14 @@
         ),
         pytest.param(
             {"aws_region": b"us-east-2"},
+            id="HDF5-v2-auto-driver",
+            marks=pytest.mark.skipif(
```
tacaswell: I don't understand the logic of what these tests were before or of why these new tests were added.

tacaswell: mostly confused by the version gating.

h5py/_hl/files.py
teobouvard: I can't tell the future but automatically selecting the `ros3` driver when no driver is explicitly chosen seems like a sensible default in this case, because the other drivers do not support http and https schemes.

h5py/tests/test_ros3.py
jhendersonHDF: I'm not entirely familiar with pytest, but inferring from what's being passed I might be able to provide some information for a starting point. When planning for the HDF5 2.0.0 release, we decided that the ROS3 VFD needed some work because it was too simplistic. For example, it only performed minimal error handling and would fail as soon as it received a 429 error from S3. For the 2.0.0 release, we rewrote the ROS3 VFD to use https://github.com/awslabs/aws-c-s3 and as part of that rewrite the VFD was changed so that the AWS region needs to be explicitly specified at all times, either in the FAPL, the `AWS_REGION`/`AWS_DEFAULT_REGION` environment variables or an AWS configuration file. Not setting the AWS region will return an error during file open.

teobouvard: I did not add the version gating, it was already there. The modifications only adds `driver=None` to the test parametrization matrix. The `driver` keyword was hardcoded in the previous version of test, and it was lifted to pytest parametrization kwargs.

tacaswell: Can we collapse these test case with test_ros3_s3uri below?  It looks like we can always pass the region with both versions (or the s3 test would fail).  I Don't think we need to test that v1 of libhdf5 does not require the region, but if we do want to cover that we should split this into two `pytest.params` as this is currently pretty confusing.

tacaswell: https://github.com/h5py/h5py/blob/e134c37a5565dd25d16c681e0f9a7347f5f80b79/h5py/h5p.templ.pyx#L1082-L1143

confirms that we do not version-gate the support of aws_region in the FAPL so we should not test it.

teobouvard: I merged both tests and removed conditional skips.

teobouvard:
@tacaswell ping

teobouvard:
@neutrinoceros can I do anything to get this PR moving forward ?

neutrinoceros:
Thanks for the ping. I will try to fill in on behalf of the team today.

h5py/_hl/files.py
neutrinoceros: with `from urllib.parse import urlsplit`
```suggestion
            and urlsplit(name).scheme in ["http", "https", "s3"]
```

neutrinoceros: Sure we can't predict the future, but this is the kind of knowledge I'd hate having to unearth from a github thread. Can you please add this explanation as an inline comment ?

h5py/tests/test_ros3.py
```diff
@@ -24,32 +24,18 @@
...
-        pytest.param(
-            {"aws_region": b"us-east-2"},
-            id="HDF5-v2",
-            marks=pytest.mark.skipif(
-                h5py.version.hdf5_version_tuple < (2, 0, 0),
-                reason="Requires HDF5 >= 2.0",
-            ),
-        ),
-    ],
+        "s3://dandiarchive/ros3test.hdf5",
```
neutrinoceros, line 31: This new test isn't covering the same branches as the previous ones used to. I would feel more confortable merging this if we didn't also remove existing test cases

neutrinoceros (changes requested):
Got a couple nits I'd like addressed first, but otherwise LGTM

h5py/_hl/files.py
teobouvard: Done

h5py/tests/test_ros3.py
teobouvard: Collapsing the test cases by unconditionnally passing `aws_region`  was suggested by @tacaswell in https://github.com/h5py/h5py/pull/2865#discussion_r3312231627, can you confirm you want me to revert this change ?

neutrinoceros: nah, I missed that it was previously suggested. Thanks for reminding me !

h5py/_hl/files.py
```diff
@@ -495,6 +496,16 @@ def __init__(self, name, mode='r', driver=None, libver=None, userblock_size=None
         Additional keywords
             Passed on to the selected file driver.
         """
+        # Automatically select the ros3 driver if it's an S3 URL.
+        # As the other drivers currently do not support HTTP(S), the ros3 driver is
+        # also automatically selected for these schemes.
+        if (
+            driver is None
+            and isinstance(name, str)
+            and urlsplit(name).scheme in ("http://", "https://", "s3://")
```
neutrinoceros: the scheme doesn't include `://`
```suggestion
            and urlsplit(name).scheme in ("http", "https", "s3")
```

teobouvard: Whoops, my bad.

neutrinoceros (approved):

neutrinoceros:
Will merge once CI is green. Thank you for your patience !

teobouvard:
Thanks a lot!

PR #2871, opened 2026-04-25
Support user tags for opaque datatypes

ajelenak:
The changes here add ability to set an arbitrary tag on an opaque (`H5T_OPAQUE`) datatype. Care is taken not to change the current behavior with special `NUMPY:...` and `PYTHON:...` opaque datatype tags.

```python
>>> dt = h5py.opaque_dtype(np.dtype('V16'), tag=b'org.example.uuid')
>>> f.create_dataset('ids', shape=(10,), dtype=dt)
>>> h5py.get_opaque_tag(f['ids'].dtype)
b'org.example.uuid'
```

High-level helper functions for this datatype are updated accordingly. Tests for this feature included.

### DIFF
--- a/docs/special.rst
+++ b/docs/special.rst
@@ -225,13 +225,41 @@ Here's an example of storing and reading a datetime array::
     >>> print(f['data'][:])
     ['2019-09-22T17:38:30']
 
-.. function:: opaque_dtype(dt)
+.. function:: opaque_dtype(dt, *, tag=None)
 
    Return a dtype like the input, tagged to be stored as HDF5 opaque type.
+
+   :param dt: Any numpy dtype whose itemsize is non-zero and which is not an
+              object, structured, or sub-array dtype.
+   :param tag: Optional HDF5 opaque tag (``str`` or ``bytes``). Strings are
+               encoded as UTF-8. When ``None`` (default), h5py auto-generates
+               a ``b"NUMPY:..."`` tag that lets the original numpy dtype be
+               restored on read.
+
+   A user-supplied tag must be non-empty, at most 255 bytes, and must not
+   contain a NUL byte. The prefix ``b"NUMPY:"`` is reserved for h5py's
+   automatic dtype round-trip, and ``b"PYTHON:OBJECT"`` is reserved for
+   h5py internals; both raise :class:`ValueError`.
+
+   Setting a custom tag disables the numpy-dtype round-trip: on read, the
+   dataset dtype is ``|V<n>`` and the tag is exposed via
+   :func:`get_opaque_tag`. Example::
+
+       >>> dt = h5py.opaque_dtype(np.dtype('V16'), tag=b'org.example.uuid')
+       >>> f.create_dataset('ids', shape=(10,), dtype=dt)
+       >>> h5py.get_opaque_tag(f['ids'].dtype)
+       b'org.example.uuid'
 
 .. function:: check_opaque_dtype(dt)
 
    Return True if the dtype given is tagged to be stored as HDF5 opaque data.
+
+.. function:: get_opaque_tag(dt)
+
+   Return the custom HDF5 opaque tag for ``dt`` as ``bytes``, or ``None`` if
+   ``dt`` is not opaque or carries h5py's automatic ``b"NUMPY:..."`` tag.
+   The tag can also be read from a committed opaque datatype via
+   :attr:`Datatype.opaque_tag <h5py.Datatype.opaque_tag>`.
 
 .. note::
 
--- a/docs/special.rst
+++ b/docs/special.rst
@@ -237,7 +237,7 @@ Here's an example of storing and reading a datetime array::
                restored on read.
 
    A user-supplied tag must be non-empty, at most 255 bytes, and must not
-   contain a NUL byte. The prefix ``b"NUMPY:"`` is reserved for h5py's
+   contain a NULL byte. The prefix ``b"NUMPY:"`` is reserved for h5py's
    automatic dtype round-trip, and ``b"PYTHON:OBJECT"`` is reserved for
    h5py internals; both raise :class:`ValueError`.
 
--- a/h5py/h5t.templ.pyx
+++ b/h5py/h5t.templ.pyx
@@ -739,6 +739,15 @@ cdef class TypeOpaqueID(TypeID):
             # 6 = len("NUMPY:")
             return np.dtype(tag[6:], metadata={'h5py_opaque': True})
 
+        # A non-empty, non-reserved tag is user-supplied (set via
+        # opaque_dtype(..., tag=...) or by another tool). Surface it in the
+        # dtype metadata so copies and inspection can see it.
+        if tag and tag != b"PYTHON:OBJECT":
+            return np.dtype(
+                "|V" + str(self.get_size()),
+                metadata={'h5py_opaque': True, 'h5py_opaque_tag': tag},
+            )
+
         # Numpy translation function for opaque types
         return np.dtype("|V" + str(self.get_size()))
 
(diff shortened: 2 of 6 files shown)


ajelenak:
This PR has been open for two months now. If there aren't any requests for changes, I can merge at the end of next week.

ajelenak:
@tacaswell @takluyver @neutrinoceros Any comments? I'd like to merge this soon.

docs/special.rst
```diff
@@ -225,14 +225,42 @@ Here's an example of storing and reading a datetime array::
...
 
+   :param dt: Any numpy dtype whose itemsize is non-zero and which is not an
+              object, structured, or sub-array dtype.
+   :param tag: Optional HDF5 opaque tag (``str`` or ``bytes``). Strings are
+               encoded as UTF-8. When ``None`` (default), h5py auto-generates
+               a ``b"NUMPY:..."`` tag that lets the original numpy dtype be
+               restored on read.
+
+   A user-supplied tag must be non-empty, at most 255 bytes, and must not
+   contain a NUL byte. The prefix ``b"NUMPY:"`` is reserved for h5py's
```
neutrinoceros:
```suggestion
   contain a NULL byte. The prefix ``b"NUMPY:"`` is reserved for h5py's
```

h5py/h5t.templ.pyx
```diff
@@ -1935,7 +1967,39 @@ def opaque_dtype(np_dtype):
...
+        )
+
+    if len(tag_bytes) == 0:
+        raise ValueError("opaque tag must be non-empty")
+    if b'\x00' in tag_bytes:
+        raise ValueError("opaque tag must not contain a NUL byte")
+    if len(tag_bytes) > 255:
+        raise ValueError(
+            "opaque tag is limited to 255 bytes (got %d)" % len(tag_bytes)
+        )
```
neutrinoceros:
```suggestion
    if (tag_size := len(tag_bytes)) == 0:
        raise ValueError("opaque tag must be non-empty")
    if tag_size > 255:
        raise ValueError(f"opaque tag is limited to 255 bytes (got {tag_size})")
    if b'\x00' in tag_bytes:
        raise ValueError("opaque tag must not contain a NULL byte")
```

h5py/h5t.templ.pyx
```diff
@@ -1935,7 +1967,39 @@ def opaque_dtype(np_dtype):
...
+    if tag_bytes.startswith(b"NUMPY:"):
+        raise ValueError(
+            "tag prefix b'NUMPY:' is reserved for h5py's automatic numpy "
+            "dtype round-trip"
+        )
+    if tag_bytes == b"PYTHON:OBJECT":
+        raise ValueError(
+            "tag b'PYTHON:OBJECT' is reserved for h5py internals"
+        )
```
neutrinoceros: I love this kind of strong runtime validation pipeline, though I think we also have an opportunity to slightly improve the UX by reporting *all* problems we scan for instead of just the first one we encounter, using an `ExceptionGroup`; this does require dropping Python 3.10 though (#2890).
in case this pattern isn't familiar, here's the outline
```py
exceptions: list[ValueError] = []

if ...:
    exceptions.append(ValueError(...))
if ...:
    exceptions.append(ValueError(...))
if ...:
    exceptions.append(ValueError(...))

if len(exceptions) == 1:
    raise exceptions[0]
elif exceptions:
    raise ExceptionGroup("multiple errors with tag", exceptions)
```

h5py/h5t.templ.pyx
```diff
@@ -2069,6 +2133,21 @@ def check_opaque_dtype(dt):
         return False
 
 
+def get_opaque_tag(dt):
```
neutrinoceros, line 2134: I'd really like type annotations here because even though I just read the whole patch I'm still not sure what type `dt` is supposed to be here.

neutrinoceros (commented):
Reviewing this feels like an outreach to me; it's quite far from my domain of expertise, but here are a couple suggestions anyway !

docs/special.rst
ajelenak: Actually, "NUL" is the correct name for the zero byte. "NULL" is a pointer constant.

h5py/h5t.templ.pyx
ajelenak: The payoff here is smaller than it seems. The checks aren't independent. An empty tag can't also exceed 255 bytes, and a `NUMPY:` tag is a fixed prefix so the only combinations that could actually report together are oversized-plus-NUL and oversized-plus-reserved-prefix. Also it depends on removing Python 3.10 which further delays this PR being merged. It's been sitting for a while now already.

neutrinoceros: Definitely not blocking. And you're right that it may not even be worth implementing as a follow up

neutrinoceros: (I do note it's possible in Python 3.10 too using the `exceptiongroup` third party backport library)

h5py/h5t.templ.pyx
ajelenak: Are we doing type annotations now in h5py? Why should only this function get it?

docs/special.rst
neutrinoceros: TIL

h5py/h5t.templ.pyx
neutrinoceros: I've been trying to push for them, and even though we haven't added them to existing functions yet, I would like to at least get them from the get go for new functions, especially where they aren't obvious from a reviewer's standpoint.

ajelenak: Fair enough, willing to make history here. 😄

Should I annotate in this PR `get_opaque_tag`, the new
`tag` parameter on `opaque_dtype`, and the new `Datatype.opaque_tag` property?
Or just the first two so my annotations stay limited to Cython? Also, what to do with the pre-existing `np_dtype` parameter? Doing it properly there
means `npt.DTypeLike` and an import of `numpy.typing` into `h5t.templ.pyx`.

One thing I want to flag because this is my first time to type annotate Cython
code. Currently the Cython `annotation_typing` option is not set in h5py build
configuration. The default is `annotation_typing=True`. This will turn my
type annotations into runtime type checks, which will produce different
exception messages because they will not come from the code.

I have Cython 3.1.2 and this is how different type annotations were handled:

| annotation | effect |
| --- | --- |
| `dt: np.dtype` | inert, no check emitted |
| `tag: str \| bytes \| None` | inert (Cython won't reduce a union) |
| `-> bytes \| None` | inert |
| `tag: bytes` | **enforced**, a `str` argument raises `TypeError: Argument 'tag' has incorrect type (expected bytes, got str)` |

That last row is the problem. `opaque_dtype`'s `tag` accepts `str` and encodes
it as UTF-8, so annotating `tag: bytes` will break a supported path and pre-empt
the code's own validation with Cython's message. In this case
`test_opaque_tag_str_encoded_as_utf8` happens to cover it, but that's luck of
coverage from using bare `pytest.raises(...)` with no `match=`. A type
annotation basically quietly moves validation and error messaging from the code
to Cython. Is this the eventual intent?

I suggest setting`annotation_typing=False` in `compiler_directives` in `setup_build.py` but this is a decision that needs to be made not in this PR. This way annotations in the Cython sources mean the same thing they mean in `_hl`.

ajelenak: @neutrinoceros I need your input how to proceed here. Thanks!

neutrinoceros: > Should I annotate in this PR get_opaque_tag, the new
tag parameter on opaque_dtype, and the new Datatype.opaque_tag property?
Or just the first two so my annotations stay limited to Cython?

I started with a minimalist request, but any annotation you *can* add is a plus in my book.

> Doing it properly there
means npt.DTypeLike and an import of numpy.typing into h5t.templ.pyx.

right, maybe it's a bit inelegant since `numpy.typing` is really meant for python typecheckers, and none of them supports Cython code as far as I know. I think that for public functions (meant to be used from Python modules and/or downstream code), the annotations should really live in stub files so they remain discoverable for typechekers, but don't have any effect at runtime, which I think answers your other question.

ajelenak: Thanks. I've added annotations to the three things this PR introduces: `get_opaque_tag` (both the `dt` parameter and the return), the new `tag` keyword on `opaque_dtype` along with its return type, and the `Datatype.opaque_tag` property. I left the pre-existing `np_dtype` parameter alone. Annotating it would require `npt.DTypeLike`, and I think pulling `numpy.typing` into `h5t.templ.pyx` for one pre-existing parameter is too much for this PR.

I validated, at least on my laptop, that Cython's default `annotation_typing` value does not turn on runtime type checks.

PR #2874, opened 2026-05-08
WHL: add abi3 nightly wheels

neutrinoceros:
This should allow downstream to test cp315 earlier. Will also continue to serve unreleased versions of Python at no extra cost in the future.


neutrinoceros:
I think there's actually an argument to be made to do this even for PyPI releases. If we just treat abi3 as another target, we could make all versions of h5py future compatible with unreleased Pythons going forward, without having to renounce specialized wheels (yet).

neutrinoceros:
I don't see this being controversial, but there's also no real gain from merging this until numpy has cp315 nightlies, so I'm happy to wait a couple days. I'll merge then if no objection is raised.

Issue #239, opened 2012-12-27
Dict interface to File class inappropriately converts to unicode


andrewcollette:
_Original author: barry.wa...@gmail.com (March 06, 2012 18:03:17)_

The dict interface to the high-level File class converts dataset names to unicode independently of whether the original dataset name was unicode or not. As unicode support was only introduced in HDF5 1.8, this has the unexpected result that the following code produces two files (test1.h5 and test3.h5) which are readable with HDF5 1.6 and two files (test2.h5 and test4.h5) which are not. I have tested this using HDF5 1.8.8 along with the latest mercurial version of h5py.

import h5py
import numpy

f1=h5py.File('test1.h5', 'w')
f2=h5py.File('test2.h5', 'w')
f3=h5py.File('test3.h5', 'w')
f4=h5py.File('test4.h5', 'w')

f1.create_dataset('test1', data=numpy.array([1,2,3]))
f1.close()

f1=h5py.File('test1.h5', 'r')

for (name, ds) in f1.items():
  f2.create_dataset(name, data=ds)

f3['test3'] = numpy.array([1,2,3])

f4['test4'] = f1['test1']

f1.close()
f2.close()
f3.close()
f4.close()

_Original issue: http://code.google.com/p/h5py/issues/detail?id=239_

PR #2878, opened 2026-05-20
Decode non-UTF-8 names using surrogateescape mechanism

takluyver:
Non-UTF-8 names are currently returned from the high-level API as bytes, which will be surprising for any code that assumes names are always str (an assumption I'm sure I could have made). This uses surrogateescape so we can consistently return str.

While doing this, I spotted that the low-level `_path_valid` method converted bytes paths to str and then back to bytes; this splits them up as bytes instead.

As discussed in #2756. This is a breaking change, intended for h5py 4.0.

### DIFF
--- a/h5py/_hl/base.py
+++ b/h5py/_hl/base.py
@@ -173,57 +173,54 @@ class CommonStateObject:
         """
         return dlcpl
 
-    def _e(self, name, lcpl=None):
-        """ Encode a name according to the current file settings.
-
-        Returns name, or 2-tuple (name, lcpl) if lcpl is True
-
-        - Binary strings are always passed as-is, h5t.CSET_ASCII
-        - Unicode strings are encoded utf8, h5t.CSET_UTF8
-
-        If name is None, returns either None or (None, None) appropriately.
-        """
-        def get_lcpl(coding):
-            """ Create an appropriate link creation property list """
-            lcpl = self._lcpl.copy()
-            lcpl.set_char_encoding(coding)
-            return lcpl
-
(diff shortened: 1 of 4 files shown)


h5py/_hl/base.py
```diff
@@ -173,57 +173,54 @@ def _lcpl(self):
...
-        Returns name, or 2-tuple (name, lcpl) if lcpl is True
+    def _e_lcpl(self, name):
+        """ Encode a name and prepare a link creation property list.
 
-        - Binary strings are always passed as-is, h5t.CSET_ASCII
-        - Unicode strings are encoded utf8, h5t.CSET_UTF8
+        Returns a 2-tuple (name, lcpl)
 
-        If name is None, returns either None or (None, None) appropriately.
+        - str with only ASCII characters are marked as ASCII, otherwise UTF-8
```
takluyver, line 193: I've elected to keep this behaviour for creating names with ASCII strings. We could probably get rid of it and mark any name coming from `str` as UTF-8, since all (strict) ASCII is also valid as UTF-8. But the cost of doing it is pretty low, and it seems preferable to avoid changes to how h5py creates files without a decent reason, even in major versions.

There is a test checking this behaviour, and it refers to the very old issue #239.

takluyver:
Close/reopen to rerun tests

neutrinoceros:
(looks like you forgot to re-open, did it for you)

takluyver:
:laughing: Thanks, no idea how I managed to lose track half way through two button clicks.

PR #2879, opened 2026-05-20
Require specifying either data or dtype when creating dataset

takluyver:
As discussed in #2756. The warning was added in #2775, for 3.16.


takluyver:
(I'm fine leaving this and #2878 around for a while if we want to leave a bit more time before 4.0. I just thought I'd make a start on implementing some of the ideas)

Issue #1912, opened 2021-06-24
Feature request: Support mypy type checking?


rggjan:
Would be great if h5py could be statically type checked with:
https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-type-hints-for-third-party-library

takluyver:
I suspect this would be a bit messy, because some of the most commonly used operations can give you several possible types - e.g. `file['name']` can give you a Dataset, Group or Datatype object depending on the contents of the file. And the less said about the return type of `file.get()`, the better.

I quite like type hints where the types involved are simple, concrete things. I quite dislike them when you need things like `Union` and `TypeVar` and `Callable[]` to  describe the things involved. I suspect fully annotating h5py would involve quite a bit of the latter sort of type hints.

hhoeflin:
I also love using h5py, but now that I use type-hints extensively I get a lot of complaints from pyright due to it.

Of course, you are right that many common operations can give several types which makes type hinting not very useful in practice. I wonder if a possible solution would be to add methods that guarantee to return a specific datatype or raise an exception otherwise. This is of course also easily done manually by wrapping the file['name'] into an 'isinstance' call, however that would be less readable. 

So for example, a 'get_dataset' method added to the 'Group' class that returns the requested object if it is a dataset and throws an exception otherwise (so in a sense the missing companion to 'create_dataset', and 'require_dataset'). These methods are then very clean and easy to type hint.

From my perspective this would also make code cleaner and at least error messages more expressive. I suspect that currently many people use h5py, accessing objects in files under implicit assumptions of their datastype without ever really testing it and only in subsequent steps errors occur (e.g. if the object retrieved really wasn't an expected group but a dataset).

takluyver:
I can see the attraction of having a set of specific methods like that, but I'm -0.1 on the idea (i.e. slightly against it) because it duplicates existing functionality, expanding the API, and it wouldn't be easy to explain to people when it's best to use one way or the other.

> I suspect that currently many people use h5py, accessing objects in files under implicit assumptions of their datastype without ever really testing it and only in subsequent steps errors occur

I'm confident people do, not least because I do. :wink: But I strongly suspect that would remain true even if we offered alternative strictly-typed APIs. I'm assuming here that we're never going to get rid of the flexibly-typed APIs, because they're very central to h5py.

In some cases, we can make the error messages clearer - e.g. #1856 improved the case where you have a group but expected a dataset - though obviously that has its limitations.

[...]

PR #2880, opened 2026-05-21
Pyi type stubs

jon-edward:
Hello! I am the creator of [stubgen-pyx](https://github.com/jon-edward/stubgen-pyx), and this applies the tool over `./h5py` to produce more complete type stubs. The following commands were ran after building h5py (I removed generated `*.temp.pyi` files):
```bash
pip install stubgen-pyx=0.2.10
stubgen-pyx ./h5py --include-private
```
Stubs can be regenerated automatically with the same command in a workflow to perpetually keep the type stubs in sync, or regenerated manually as needed. This addresses #1912 

(Note: I did manually remove some identity TypeAlias assignments. This should be fixed at next release)

CC: @neutrinoceros


neutrinoceros (commented):
Hi, thanks for opening this.
Before I run an in-depth review, I would like to address 4 high level points:
- I don't think we should duplicate docstrings (is there any reason for doing so that I'm missing ?)
- if we're going to add generated stubs, I'd rather all signatures be typed *completely* (including return types, even when it's a simple `-> None`)
- we should be running a type checker in CI
- I would really like some comparison with `h5py-stubs`. Is there an overlap with this and  are there any incompatible annotations between the two sets ?

jon-edward:
That's a valid point concerning docstrings, I added an `--exclude-docstrings` CLI arg to `stubgen-pyx` that skips docstring inclusion and regenerated the stubs. I will work towards completing type hints in Cython source and afterwords provide a comparison between the stubs. Marking this as a draft for the time being, thank you!

jon-edward:
I added typing for the signatures of public methods and attributes. Additionally, I tried to narrow types for some cdef functions to help with type checking.

I checked out [h5py-stubs](https://github.com/Vidium/h5py-stubs) to draw a comparison, and it seems like the output of stubgen-pyx is largely a superset of h5py-stubs typings. h5py-stubs provides the names of some exposed classes and some function signatures of four public modules (`h5d`, `h5f`, `h5g`, `h5t`). However, the majority of module members are missing and all included classes are incomplete.

This is the current status of the h5py-stubs typings:

```python
# h5py-stubs/h5d.pyi

import numpy as np

from . import _objects as __h5py__objects

class DatasetID(__h5py__objects.ObjectID):
    @property
    def dtype(self) -> np.generic: ...
```

_As compared to [h5d suggested stubs](https://github.com/jon-edward/h5py-pyi-experiment/blob/982db719a0d4a11991a5817616d9f55094868a0c/h5py/h5d.pyi)_

```python
# h5py-stubs/h5f.pyi

from . import h5g as __h5py_h5g

class FileID(__h5py_h5g.GroupID): ...
```

_As compared to [h5f suggested stubs](https://github.com/jon-edward/h5py-pyi-experiment/blob/982db719a0d4a11991a5817616d9f55094868a0c/h5py/h5f.pyi)_

```python
# h5py-stubs/h5g.pyi

from . import _objects as __h5py__objects

class GroupID(__h5py__objects.ObjectID): ...
```

_As compared to [h5g suggested stubs](https://github.com/jon-edward/h5py-pyi-experiment/blob/982db719a0d4a11991a5817616d9f55094868a0c/h5py/h5g.pyi)_

```python
# h5py-stubs/h5t.pyi

import numpy as np

from typing import Any
from typing import Literal

class string_info(tuple[str, int]): ...

def string_dtype(
    encoding: Literal["utf-8", "ascii"] = "utf-8", length: int | None = None
) -> np.dtype[np.bytes_] | np.dtype[np.object_]: ...
def check_string_dtype(*args: Any, **kwargs: Any) -> string_info | None: ...
```

_As compared to [h5t suggested stubs](https://github.com/jon-edward/h5py-pyi-experiment/blob/982db719a0d4a11991a5817616d9f55094868a0c/h5py/h5t.pyi)_

neutrinoceros:
Sorry about the delay in my response. I really appreciate the work here and want it to get the attention that it needs. Unfortunately +4k LoC is a bit too much for me to review in one seating, even if it's just stubs. I think we need to spread the work over several PRs for this to be manageable. Ideally, I would love to get a first one with *just* infrastructure:  enabling *some* type checking in CI. It doesn't really matter to me which type checker you want to setup first, as long as it can be configured to not emit diagnostics about missing annotations. After that, maybe we can go for one stub file per PR ? It's still useful to have one giant PR such as this one to serve as a reference, so we can also see and validate the end result in CI.
Does this seem practical for you ?

neutrinoceros:
About the errors we're seeing in CI: it'll be crucial that we don't break runtime compatibility with numpy 2.0.0. If we *have* to drop 1.x that's negotiable, I think, though I would actually not expect stub files to have *any* impact at runtime, so there's clearly something fishy going on that we need to understand. Other than that, I think it should be okay that type checking h5py requires a slightly more up-to-date version of numpy than runtime, given how much better the support has gotten in that area along the 2.x series.

h5py/_objects.pyi
```diff
@@ -0,0 +1,99 @@
...
+from .defs import *
+
+
+class BogoLock:
+
+    def __enter__(self) -> None:
+        ...
+
+    def __exit__(self, *args) -> None:
+        ...
```
neutrinoceros, line 15: I'd suggest running `ruff format` on these, which should inline function bodies that are exactly `...`, as well as reduce the number of empty lines in stub files. This alone could make the change *look* much more reasonable in size (and I think it'll also make it easier for me to review)

PR #2884, opened 2026-05-31
docs: warn about external storage paths in untrusted files

dfgvaetyj3456356-hash:
## Summary

Add security guidance for handling untrusted HDF5 files that contain external dataset storage or external links.

HDF5 resolves these paths when dataset data or external links are accessed. For applications accepting files from other users, this means the file can reference filesystem paths outside the main HDF5 file. The new notes point users toward inspecting `Dataset.external` and approving link/storage targets before reading or writing data from untrusted files.

## Verification

- `git diff --check`
- `python -m py_compile h5py\_hl\group.py h5py\_hl\dataset.py`
- Local benign reproduction with temporary files confirmed external storage reads and writable opens affect the referenced external path, matching the warning text.

### DIFF
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -718,6 +718,11 @@ Reference
        of 3-tuples, like the ``external=`` parameter to
        :meth:`Group.create_dataset`. Otherwise, it is ``None``.
 
+       For untrusted HDF5 files, inspect this before reading or writing dataset
+       data.  External storage entries can name files outside the HDF5 file, so
+       HDF5 may read from or write to those paths when dataset data is
+       accessed.
+
     .. attribute:: is_virtual
 
        True if this dataset is a :doc:`virtual dataset </vds>`, otherwise False.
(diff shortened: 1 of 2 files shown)


aragilar:
I have no objection to adding this, but I'm not sure saying "be careful" is that helpful? Arguably if you are concerned about untrusted files, compression plugins might be of more interest?

dfgvaetyj3456356-hash:
Thanks, that makes sense. I tightened the wording so it gives concrete checks instead of just saying to be careful, and added a short note in the custom filter section about only enabling native filter plugins from trusted locations. The external storage/link guidance is still scoped to path resolution behavior, but hopefully the untrusted-file advice is now more specific and useful.

takluyver:
I'm also fine with adding these specific warnings, but I'll repeat my opinion that if your threat model includes potentially malicious HDF5 files, you probably want to read them entirely within a secure sandbox. HDF5 is a massively complex file format, libhdf5 is written in C, and a big chunk of use is in institutional environments where we largely don't need to worry about malicious files. This is not a slight on HDFGroup, who have resolved a good number of CVEs.

Also, virtual datasets give a third way for an HDF5 file to point to another file. Although these and external links should only be able to point to another HDF5 file; external dataset storage is scarier, because it can potentially read any file.

PR #2885, opened 2026-05-31
REL: publish trailing abi3 wheels in addition to version-specific ones

### LINKED ISSUE #2881: WHL: publishing `cp<latest>-abi3` wheels in addition to version specific ones ?
As a follow up to #2874, I would like to propose we do more of the same for PyPI releases. This would completely eliminate the need for cutting releases around October when the most pressing motivation is to avoid breaking downstream projects who adopt new Python versions early.
These additional wheels would single-handedly make each release future-proof for much longer.

neutrinoceros:
Close #2881
Still needs to be discussed, but so far feedback is positive.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -130,6 +130,10 @@ jobs:
                 github.event_name == 'push'
               ) &&
               contains( env.TIP_COMMIT_MSG , '[pip-pre]' )
+            ) || (
+              github.repository_owner == 'h5py' &&
+              github.event_name == 'push' &&
+              github.ref_type == 'tag'
             )
         env:
           H5PY_LIMITED_API: 1


neutrinoceros:
Adding this to the 3.17 milestone mostly to make sure it is considered before the release is cut, but this shouldn't be blocking.

neutrinoceros:
blocked by #2959

neutrinoceros:
Nevermind. I was planning to also upgrade to `cp315-abi3` here but #2959 is blocked itself, so let's just undraft this again.

neutrinoceros:
no (known) blocker left

takluyver:
Have we done any kind of performance measurements with abi3 builds yet? I hope that any differences will be minor, but if not people are still going to push for version-specific wheels as well.

We don't need an extensive suite of benchmarks, but I'd like to see some numbers for a couple of things where we might be doing a lot of Python operations, such as reading variable-length strings into Python string objects, and walking a file with a lot of nested groups.

neutrinoceros:
Agreed that this should be a requirement for dropping version-specific wheels. Are you saying  should block this approach too ?

neutrinoceros:
(deep down, the reason I'm not doing this kind of benchmarking myself is that I was never really a user of h5py, and I don't have massive or typical datasets at hand, so I think it'd take me an innordiate amount of time to just assemble the benchmark. I'd be more than happy to run comparisons given a bit more pointers and resources)

takluyver:
I don't think it needs to take an inordinate amount of time - construct a couple of suitable files to test the cases I outlined, the timing can just be a `%%timeit` cell magic in IPython/Jupyter for now, run it with a regular cp3xy build and an abi3 build, and see how the numbers compare.

I'd like to have at least a basic sanity check that performance hasn't obviously tanked before we promote abi3 wheels to general users via a PyPI release.

neutrinoceros:
Alright I'll give it a try today

takluyver:
Thanks :-)

neutrinoceros:
Here's my attempt to implement your first proposal. The numbers look promising to me (essentially no measurable impact), but I'm *really* not certain that I'm being faithful to what you had in mind, and wouldn't be surprised if I was doing something obviously wrong  😅

generator script
```py
import random
from pathlib import Path

import h5py

random.seed(0)

NELEMS = 1_000_000
NMAX = 1_000
FILE = Path(__file__).parent / "dataset.h5"

data = ["a"] * NELEMS
for i in range(NELEMS):
    data[i] = "b" * random.randint(1, NMAX)

dt = h5py.vlen_dtype(str)


with h5py.File(FILE, mode="w") as f:
    f.create_dataset("vlen-str", NELEMS, dtype=dt, data=data)
```

benchmark script
```py
from pathlib import Path
from time import monotonic_ns
import h5py

FILE = Path(__file__).parent / "dataset.h5"

f = h5py.File(FILE)
tstart = monotonic_ns()
bytes_: bytes = list(f["vlen-str"])
tstop = monotonic_ns()
f.close()

print(f"read {len(bytes_):_} vlen Python bytes objects in {(tstop - tstart)/1e6:_.0f} ms")
```

Results

| Python | abi    | best-of-five (ms) |
| ------- | ----- | -----------------| 
| 3.11 | `cp311` | 11762 |
| 3.11 | `abi3` | 11701 |
| 3.15 | `cp315` | 12926 |
| 3.15 | `abi3` | 12994 |

(I suspect the ~1s difference between 3.11 and 3.15 is just a sign that python-build-standalone currently doesn't make a lot of effort optimizing 3.15 builds)

takluyver:
Thanks, that's the sort of thing I was after, although I was actually thinking of reading like this (about 100× faster for me):

```python
f["vlen-str"][:]
```

The `list()` call iterates over the dataset, so does 1M [separate read calls](https://github.com/h5py/h5py/blob/3abbc3e5fd94e74a9dffc3a3bbb661840d11bcc7/h5py/_hl/dataset.py#L850-L851), whereas `[:]` reads the whole lot to an array. Both are worth testing, but doing one big read focuses more on the [conversion to Python bytes objects](https://github.com/h5py/h5py/blob/3abbc3e5fd94e74a9dffc3a3bbb661840d11bcc7/h5py/_conv.templ.pyx#L200-L224) where I know we're doing Python operations inside a C-level loop.

neutrinoceros:
Great. Here are results for the same benchmark, except what's measured is `f["vlen-str"][:]`

| Python | abi    | best-of-five (ms) |
| ------- | ----- | -----------------| 
| 3.11 | `cp311` | 518 |
| 3.11 | `abi3` | 503 |
| 3.15 | `cp315` | 507 |
| 3.15 | `abi3` | 529 |

neutrinoceros:
About this one
> walking a file with a lot of nested groups.

should the data itself be "big" arrays ? Does it make a difference ?
Should I prioritize depth (many nesting levels), width (many top-level groups, some of which containing nesting up to a couple levels), everything at once, all of the above ?

takluyver:
For what I'm thinking of, the size of the datasets shouldn't make any difference. I would start with something like this:

```python
for tpl in itertools.product(range(10), repeat=3):
    f.create_dataset('a{}/b{}/c{}'.format(*tpl), shape=5, dtype=np.int32)
```

And then test wrapping all the groups and datasets in Python objects with something like this:

```python
l = []
f.visititems(l.append)
assert len(l) == 1110  # Sanity check - I think this is the right number ;-)
```

Then increase the length and/or depth until the measured time on each run is over about 1 second. I don't think nesting depth should actually matter for this test, but it doesn't hurt to have some groups in there too.

(I just looked at a random EuXFEL data file for comparison: 5217 objects (groups and datasets), up to 8 levels deep :slightly_smiling_face: )

neutrinoceros:
Length first:
```py
import random
from pathlib import Path
from itertools import product
import h5py
import numpy as np
from time import monotonic_ns

FILE = Path(__file__).parent / "dataset.h5"


f = h5py.File(FILE, mode="w")

tstart = monotonic_ns()
for tpl in product(range(34), repeat=3):
    f.create_dataset('a{}/b{}/c{}'.format(*tpl), shape=5, dtype=np.int32)

tmid = monotonic_ns()

L = []
f.visititems(lambda x, y: L.append((x, y)))
tstop = monotonic_ns()

f.close()

print(f"Creation took {(tmid - tstart) / 1e6:_.0f} ms")
print(f"Visit took {(tstop - tmid) / 1e6:_.0f} ms")
print(f"Total runtime {(tstop - tstart) / 1e6:_.0f} ms")
```

Creation is about 60% of the entire runtime

| Python | abi    | best-of-five (ms) |
| ------- | ----- | -----------------| 
| 3.11 | `cp311` | 2671 |
| 3.11 | `abi3` | 2670 |
| 3.15 | `cp315` | 2880 |
| 3.15 | `abi3` | 2884 |

neutrinoceros:
Now depth.
The script is identical except for the loop line, changed to
```py
for tpl in product(range(8), repeat=5):
```

This time creation is about 25% to 30% of the entire runtime

| Python | abi    | best-of-five (ms) |
| ------- | ----- | -----------------| 
| 3.11 | `cp311` | 2828 |
| 3.11 | `abi3` | 2785 |
| 3.15 | `cp315` | 3033 |
| 3.15 | `abi3` | 3039 |

takluyver:
Thank-you, this looks like a good initial indication that abi3 builds should be reasonably performant :smile:

neutrinoceros:
super. Can we get this one in then ? 😇

PR #2887, opened 2026-06-04
fix: reject short fileobj reads

dfgvaetyj3456356-hash:
## Summary

Fixes #2835.

The `fileobj` VFD read callback now treats short reads as failures for both file-like read paths:

- checks the byte count returned by `readinto()` before returning success
- returns `-1` for short `read()` results, matching `herr_t` failure conventions

Previously, `readinto()` results were ignored, so a file-like object that returned fewer bytes than requested could leave part of the destination buffer untouched while the callback still returned success. The fallback `read()` path also returned `1` for a short read, which is non-negative and therefore not a conventional `herr_t` failure.

## Tests

Added `TestFileObj.test_short_readinto`, which builds a valid in-memory HDF5 file and then serves it through a file-like object whose `readinto()` intentionally returns one byte short.

Local checks run:

- `python -m py_compile h5py\tests\test_file2.py`
- `git diff --check`

I could not run the new runtime test locally in this checkout because Cython is not installed here, and this change touches a Cython template that needs the normal h5py extension build.

### DIFF
--- a/h5py/h5fd.templ.pyx
+++ b/h5py/h5fd.templ.pyx
@@ -156,16 +156,19 @@ cdef haddr_t H5FD_fileobj_get_eof(const H5FD_fileobj_t *f, H5FD_mem_t type) exce
 
 cdef herr_t H5FD_fileobj_read(H5FD_fileobj_t *f, H5FD_mem_t type, hid_t dxpl, haddr_t addr, size_t size, void *buf) except -1 with gil:
     cdef unsigned char[:] mview
+    cdef Py_ssize_t bytes_read
     (<object>f.fileobj).seek(addr)
     if hasattr(<object>f.fileobj, 'readinto'):
         mview = <unsigned char[:size]>(buf)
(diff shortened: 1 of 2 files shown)


tacaswell:
@dfgvaetyj3456356-hash Please test this locally.

dfgvaetyj3456356-hash:
Tested locally on Windows with an editable build of this branch against HDF5 1.12.2.\n\nEnvironment:\n- Python 3.11.9\n- h5py 3.16.0 from this checkout\n- HDF5 1.12.2 built locally with Visual Studio 2022 Build Tools\n- NumPy 2.4.6\n- Cython 3.2.5\n\nCommands/results:\n- \\pytest h5py/tests/test_file2.py::TestFileObj::test_short_readinto -q\\ -> \\1 passed\\\n- \\pytest h5py/tests/test_file2.py::TestFileObj -q\\ -> \\9 passed, 1 skipped\\\n\nAlso confirmed the PR CI/Codecov are green from the GitHub checks.

PR #2889, opened 2026-06-09
REL: use a pre-release version number for nightlies

neutrinoceros:
Gain: allow downstream projects to configure additional registry indexes at the project level (rather than on a per-job basis) while preserving the ability to exclude anaconda.org nightly channels by not accepting pre-releases
Cost: add 2 steps to the release process (is it documented somewhere ? I couldn't find where): dropping and re-adding the `dev0` prefix.

For reference, this would unblock https://github.com/astropy/astropy/pull/19886

### DIFF
--- a/h5py/version.py
+++ b/h5py/version.py
@@ -28,7 +28,7 @@ class _H5PY_VERSION_CLS(NamedTuple):
     def __str__(self) -> str:
         s = f"{self.major}.{self.minor}.{self.bugfix}"
         if self.pre is not None:
-            s += version_tuple.pre
+            s += self.pre
         if self.post is not None:
             s += f".post{self.post}"
         if version_tuple.dev is not None:
--- a/h5py/version.py
+++ b/h5py/version.py
@@ -10,28 +10,36 @@
 """
     Versioning module for h5py.
 """
-
-from collections import namedtuple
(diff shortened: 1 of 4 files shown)


takluyver:
The release process is indeed documented: https://docs.h5py.org/en/stable/release_guide.html

I think it's just 1 extra step, as we have to change the version number once anyway. :slightly_smiling_face: If you want to add it to that guide, I'm fine with this.

neutrinoceros:
Cool. Done. I also added inline comments and type annotations to `_H5PY_VERSION_CLS`

h5py/version.py
```diff
@@ -10,28 +10,36 @@
...
+    bugfix: int
+    pre: Literal["a", "b", "rc"] | None = None
+    post: int | None = None
+    dev: int | None = None
 
-hdf5_built_version_tuple = _h5.HDF5_VERSION_COMPILED_AGAINST
+    def __str__(self) -> str:
+        s = f"{self.major}.{self.minor}.{self.bugfix}"
+        if self.pre is not None:
+            s += version_tuple.pre
```
neutrinoceros:
```suggestion
            s += self.pre
```

takluyver:
Thanks!

PR #2890, opened 2026-06-10
DEP: drop support for CPython 3.10

neutrinoceros:
Tis the season.
On the surface the gain isn't really apparent but here's what this unblocks:
- `typing` features that could be used in the on-going effort to add stub files
- reduce complexity around abi selection in the build process which is about to go up again when abi3t (PEP 803) becomes usable.
- if we were to drop version-specific wheels in favor of abi3 ones, this removes the one python version for which we cannot make abi3 wheels, thus halving the number of artifacts (excluding sdists)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,16 +8,16 @@ environment:
 
     ### USING HDF5 1.10 ###
 
-    - PYTHON: "C:\\Python310"
-      TOXENV: "py310-test-deps"
+    - PYTHON: "C:\\Python311"
+      TOXENV: "py311-test-deps"
       HDF5_VSVERSION: "14"
       HDF5_DIR: "C:\\hdf5\\%HDF5_VERSION%\\%HDF5_VSVERSION%"
       HDF5_VERSION: "1.10.7"
 
     ### USING HDF5 1.12 ###
 
-    - PYTHON: "C:\\Python310"
-      TOXENV: "py310-test-deps"
+    - PYTHON: "C:\\Python311"
+      TOXENV: "py311-test-deps"
       HDF5_VSVERSION: "14"
       HDF5_DIR: "C:\\hdf5\\%HDF5_VERSION%\\%HDF5_VSVERSION%"
       HDF5_VERSION: "1.12.1"
@@ -31,10 +31,10 @@ environment:
 install:
   # We need wheel installed to build wheels
   - "%PYTHON%\\python.exe -m pip install --upgrade wheel pip setuptools"
-  - "py -3.10 -m pip install --upgrade wheel pip setuptools virtualenv"
-  - "py -3.10 -m pip install requests"
-  - "py -3.10 ci\\get_hdf5_win.py"
-  - "py -3.10 -m pip install tox codecov"
+  - "py -3.11 -m pip install --upgrade wheel pip setuptools virtualenv"
+  - "py -3.11 -m pip install requests"
+  - "py -3.11 ci\\get_hdf5_win.py"
+  - "py -3.11 -m pip install tox codecov"
 
 build: off
 
@@ -43,7 +43,7 @@ test_script:
   # Note that you must use the environment variable %PYTHON% to refer to
   # the interpreter you're using - Appveyor does not do anything special
   # to put the Python evrsion you want to use on PATH.
-  - "py -3.10 -m tox run --discover %PYTHON%\\python.exe"
+  - "py -3.11 -m tox run --discover %PYTHON%\\python.exe"
 
 # This is commented out as there's no easy way to deal with numpy dropping
 # older python versions without a recent pip/setuptools.
@@ -60,4 +60,4 @@ cache:
   - "C:\\hdf5"
 
 on_success:
-  - "py -3.10 ci\\upload_coverage.py"
+  - "py -3.11 ci\\upload_coverage.py"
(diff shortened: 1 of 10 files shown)


neutrinoceros:
Something is off on azure's macOS build:
```
2026-06-10T08:38:14.4031330Z       /private/var/folders/ss/yttmxf7j3k584mnld5ldxg100000gn/T/pip-req-build-kr0lhvil/h5py/h5t.c:25000:20: error: use of undeclared identifier 'h5py_size_n256'
2026-06-10T08:38:14.4031830Z        25000 |     __pyx_v_size = h5py_size_n256;
2026-06-10T08:38:14.4032140Z              |                    ^
2026-06-10T08:38:14.4032610Z       /private/var/folders/ss/yttmxf7j3k584mnld5ldxg100000gn/T/pip-req-build-kr0lhvil/h5py/h5t.c:25009:21: error: use of undeclared identifier 'h5py_offset_n256_real'
2026-06-10T08:38:14.4033130Z        25009 |     __pyx_v_off_r = h5py_offset_n256_real;
2026-06-10T08:38:14.4033450Z              |                     ^
2026-06-10T08:38:14.4033930Z       /private/var/folders/ss/yttmxf7j3k584mnld5ldxg100000gn/T/pip-req-build-kr0lhvil/h5py/h5t.c:25018:21: error: use of undeclared identifier 'h5py_offset_n256_imag'
```

these symbols are 
- defined if `np.complex256` is available
- used if the pre-processor var `NPY_COMPLEX256` is defined

normally these two conditions are equivalent, but here it seems the former is False and the latter is True.

neutrinoceros:
I'm hesitant to even investigate this problem; we already exercise both archs available on macOS through GHA, so I don't know what more azure is supposed to bring to the table.

aragilar:
If azure pipelines is running the same tests as GHA, maybe it's something to do with the MacOS versions (given it appears to be x86 that is failing)? I don't know of a reason to not drop the failing azure pipelines.

neutrinoceros:
Ok let's give it a try then.

h5py/_hl/dataset.py
```diff
@@ -402,9 +403,8 @@ def __init__(self, dset, source_sel=None):
                         start = s.start or 0
                         stop = s.stop or self._shape[dim]
                         step = s.step
-                    case _:
-                        # TODO: use typing.assert_never when Python 3.10 is dropped
-                        raise AssertionError(f'{s}: Selection object must be a slice or integer')
+                    case _ as unreachable:
+                        assert_never(unreachable)
```
tacaswell: Can we get the nice message back?
```suggestion
                        assert_never(f'{unreachable}: Selection object must be a slice or integer')
```

appveyor.yml
```diff
@@ -8,16 +8,16 @@ environment:
...
+      TOXENV: "py311-test-deps"
       HDF5_VSVERSION: "14"
       HDF5_DIR: "C:\\hdf5\\%HDF5_VERSION%\\%HDF5_VSVERSION%"
       HDF5_VERSION: "1.10.7"
 
     ### USING HDF5 1.12 ###
 
-    - PYTHON: "C:\\Python310"
-      TOXENV: "py310-test-deps"
+    - PYTHON: "C:\\Python311"
```
tacaswell: I think we can drop this one as the next test down is with hdf5 1.12.2 and the same everything else.

h5py/_hl/dataset.py
neutrinoceros: this could arguably produce a more confusing error message, and wouldn't typecheck properly: the expected input type is Never.

neutrinoceros: Actually I think it's possible to get the best of both worlds:
```py
raise AssertionError(...)
assert_never(unreachable)
```
Here, `assert_never` is *purely* a type-check time assertion. I was afraid type checkers would (somewhat ironically) report `raise` statement as unreachable, but apparently they don't. I tried ty, pyrefly and mypy, and none of them complained.

appveyor.yml
neutrinoceros: good call

h5py/_hl/dataset.py
tacaswell: I am a bit confused by the typing as

```python

In [4]: a = 1
   ...: match a:
   ...:     case 'a':
   ...:         print('nope')
   ...:     case _ as f:
   ...:         print(f, type(f))
   ...: 
1 <class 'int'>
```

I'm also less convinced that this is actually an `assert_never` as nothing in this function ensures we never reach here and if someone were to sub-class us and then write a non-compliant iter this could _would_ be reachable (but maybe this is where my view of reality and the type checkers views diverge 🤣 ).

I think either we just leave the exception or we make the string that goes into `assert_never` meaningful.

neutrinoceros: > nothing in this function ensures we never reach here

that's correct at runtime. The `assert_never` is really only there for type checkers, and has the advantage of actually blocking execution at runtime if it's ever reached.

neutrinoceros: if this is the only blocker towards merging this PR I'm happy to split it out

neutrinoceros: I reverted this bit for now, and intend to merge this once CI passes

neutrinoceros: moved to #2969

PR #2893, opened 2026-06-17
TST: start testing nightly cp315 builds

neutrinoceros:

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -135,6 +135,8 @@ jobs:
           H5PY_LIMITED_API: 1
           CIBW_BUILD: 'cp315*'
           CIBW_ENABLE: cpython-prerelease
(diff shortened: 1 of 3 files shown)


neutrinoceros:
our manylinux images do not have cp315 yet, which is the current bottleneck.
I just triggered a rebuild+push at https://github.com/h5py/hdf5-manylinux/actions/runs/27741828771

neutrinoceros:
Finally got tests running but now hitting what looks like a bug in tox. (`py315-test-deps-pre: internal error`)
I'll split out the working bits into separate PRs so this one gets simpler.

neutrinoceros:
helper PRs at:
- #2900
- #2901
- #2902

neutrinoceros:
the end result is correct but build times are inflated from compiling numpy 2.4 from source (before it's immediately discarded for 2.6dev). I'd like to address this but it's not a blocker.

neutrinoceros:
fixed. I plan to merge this tomorrow (about 24h from now) if no objection is raised.

PR #2895, opened 2026-06-18
Fix page_buf_size=0 bug

### LINKED ISSUE #2894: the h5py.File argument `page_buf_size=0` is effectively ignored
I wanted to disable page buffering for some performance comparisons, and was surprised at the following:

```python
with h5py.File(path, driver="ros3", page_buf_size=0) as f:
    p = f.id.get_access_plist()
    print(p.get_page_buffer_size())
```
```
(67108864, 0, 0)
```

The page buffer is not zero, as requested.

I can work around this with the argument `page_buf_size="0"`, since this value does not evaluate to `False` at [_hl/files#144](https://github.com/h5py/h5py/blob/f11419e2d8c603bca8ee133651ff7f34846368f2/h5py/_hl/files.py#L144).

ajelenak:
`File(..., page_buf_size=0)` now accepts an integer `0` whereas previously an integer `0` was silently ignored (the same as the `None` default).

Fixes #2894.

### DIFF
--- a/h5py/tests/test_file.py
+++ b/h5py/tests/test_file.py
@@ -26,7 +26,7 @@ from hashlib import sha256
 import pytest
 
 from .common import ut, TestCase, UNICODE_FILENAMES, closed_tempfile, make_name
-from h5py._hl.files import direct_vfd
+from h5py._hl.files import direct_vfd, make_fapl
 from h5py import File
 import h5py
 import pathlib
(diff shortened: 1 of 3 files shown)


news/page-buf-size-zero.rst
```diff
@@ -0,0 +1,9 @@
+Bug fixes
+---------
+
+* ``File(..., page_buf_size=0)`` now accepts an integer ``0`` and treats it the
+  same as the string ``"0"``. Previously an integer ``0`` was silently ignored
+  (the same as the ``None`` default), which also dropped any ``min_meta_keep``
+  and ``min_raw_keep`` values passed alongside it. As a consequence, a
+  nonsensical empty string ``page_buf_size=""`` now raises ``ValueError``
+  instead of being silently ignored as before.
```
neutrinoceros, line 9: if that's important enough to make it to changelog I think we ought to test it too

h5py/tests/test_file.py
```diff
@@ -277,6 +277,16 @@ def test_actual_pbs(self):
             fapl = f.id.get_access_plist()
             self.assertEqual(fapl.get_page_buffer_size()[0], fsp)
 
+    def test_page_buf_size_zero(self):
```
neutrinoceros, line 280: I'd prefer this to be a parametrized test with a single `assertEqual` call

news/page-buf-size-zero.rst
ajelenak: It’s not significant at all. I thought it was worth mentioning only because I’ve noticed that people often resort to strange behavior simply because the library didn’t complain and then demanded endless support invoking backward compatibility. Should I remove that text to have this PR accepted?

neutrinoceros: To be clear, I agree it's worth mentioning here. I just also think it's worth testing.

h5py/tests/test_file.py
ajelenak: Done in [fea0038](https://github.com/h5py/h5py/pull/2895/commits/fea003847c36deefd33a9bd211a225b2a54f955b).

news/page-buf-size-zero.rst
ajelenak: Done in [fea0038](https://github.com/h5py/h5py/pull/2895/commits/fea003847c36deefd33a9bd211a225b2a54f955b).

neutrinoceros (approved):

PR #2900, opened 2026-06-18
MNT: move uv.index settings to pyproject.toml

neutrinoceros:
This makes sp-nightlies permanently configured as an extra index for any uv-based workflow, but artifacts won't be pulled from there unless pre-releases are allowed in resolution.


neutrinoceros:
I intend to merge this in a couple days if no objection is raised

PR #2901, opened 2026-06-18
TST: define py315(t) tox factors

neutrinoceros:


tox.ini
```diff
@@ -55,6 +55,8 @@ uv_resolution =
     !mindeps: highest
 pip_pre =
     pre: True
+    py315: True
+    py315t: True
```
neutrinoceros: these are meant to be temporary and removed when 3.15 becomes testable with stable versions of numpy

neutrinoceros: I removed this part because tox is buggy and couldn't handle it this way. I'll move the change to #2893 where it'll be immediately exercised

neutrinoceros:
there's nothing remotely controversial left here so I'll merge when CI passes

PR #2903, opened 2026-06-18
BUG: check realloc() return value in conv_vlen2ndarray

### LINKED ISSUE #2896: [BUG] realloc NULL unchecked in conv_vlen2ndarray - segfault on OOM
## Bug Description

In `h5py/_conv.pyx` (template file `_conv.templ.pyx`), `realloc()` is called without checking for NULL. If allocation fails, the NULL pointer is passed to `H5Tconvert`, causing a segfault.

## Location

`h5py/_conv.pyx` (generated from `_conv.templ.pyx:712-713`)

## Code

```cython
if itemsize > H5Tget_size(intype.id):
    data = realloc(data, itemsize * size)
    # No NULL check! If realloc fails:
    # - data is now NULL
    # - original pointer is leaked (realloc doesn't free on failure)
    # - H5Tconvert(intype.id, outtype.id, size, data, ...) dereferences NULL → crash
```

## Impact

Under memory pressure, `realloc()` returns NULL and the original buffer is leaked. The NULL `data` pointer is then passed to HDF5's `H5Tconvert` which dereferences it, causing a segfault. This affects reading variable-length datasets when the output type is larger than the input type.

## Suggested Fix

```cython
if itemsize > H5Tget_size(intype.id):
    new_data = realloc(data, itemsize * size)
    if new_data == NULL:
        free(data)
        raise MemoryError("Failed to allocate conversion buffer")
    data = new_data
```

## Environment

- Affected: reading variable-length datasets with type conversion

Fuyugithub:
## What does this PR do?

Fixes #2896 — `realloc()` return value not checked in `conv_vlen2ndarray`.

## Root cause

In `h5py/_conv.templ.pyx`, when the output itemsize is larger than the input, the buffer is reallocated:

```cython
data = realloc(data, itemsize * size)
```

`realloc()` returns `NULL` on allocation failure. Overwriting `data` with `NULL` causes:

1. **Memory leak** — the original HDF5-allocated buffer at `in_vlen[0].ptr` is lost.
2. **Segfault** — `NULL` is passed to `H5Tconvert()` at the next line.

## Fix

Store the result in a temporary variable, raise `MemoryError` if `NULL`, then update `data`:

```cython
new_data = realloc(data, itemsize * size)
if new_data == NULL:
    efree(data)
    raise MemoryError("Failed to reallocate vlen conversion buffer")
data = new_data
```

This follows the standard safe-realloc pattern required by the C standard (§7.22.3.5).

## Testing

The bug triggers when reading a variable-length dataset whose element type requires upcasting (output `itemsize > input itemsize`) under low-memory conditions.  No existing test covers the OOM path; the fix is a mechanical correction of a well-known C pattern.

---

### DIFF
--- a/h5py/_conv.templ.pyx
+++ b/h5py/_conv.templ.pyx
@@ -697,6 +697,7 @@ cdef int conv_vlen2ndarray(void* ipt,
         int flags = NPY_ARRAY_WRITEABLE | NPY_ARRAY_C_CONTIGUOUS | NPY_ARRAY_OWNDATA
         npy_intp dims[1]
         void* data
+        void* new_data
         char[:] buf
         void* back_buf = NULL
         cnp.ndarray ndarray
@@ -710,7 +711,11 @@ cdef int conv_vlen2ndarray(void* ipt,
     dims[0] = size
     itemsize = H5Tget_size(outtype.id)
     if itemsize > H5Tget_size(intype.id):
-        data = realloc(data, itemsize * size)
+        new_data = realloc(data, itemsize * size)
+        if new_data == NULL:
+            efree(data)
+            raise MemoryError("Failed to reallocate vlen conversion buffer")
+        data = new_data
 
     if needs_bkg_buffer(intype.id, outtype.id):
         back_buf = emalloc(H5Tget_size(outtype.id)*size)


neutrinoceros:
> Bug found via static FFI boundary analysis of h5py source.

Any chance you could share the tool(s) used to accomplish this ?

I also noticed that the PR and the commit attached to it were authored by 2 different accounts, which is unusual. Given the form of the text in both the issue and the PR is very evocative of LLMs outputs, I'd also like to request some disclosure for any use of AI in your workflow. Thank you !

Fuyugithub:
Hi @neutrinoceros, thanks for asking!

We're working on a research project about detecting FFI boundary bugs in Python-C and JNI code. We built a static analysis prototype that scans C-API call sites and checks them against common safety patterns. Claude (Sonnet 4.6) is used in the pipeline to help verify uncertain cases. The fixes are written and reviewed by us humans. Sorry I didn't disclose this upfront.

About the two accounts — that was just me messing up the git config in my local clone. Everything will come from this account going forward.

Thanks again for the review!

neutrinoceros (approved):
Thank you !

PR #2904, opened 2026-06-18
BUG: copy vlen data into NumPy-owned buffer, avoid allocator mismatch

### LINKED ISSUE #2897: [BUG] Allocator mismatch: NumPy free() on HDF5-allocated memory in conv_vlen2ndarray
## Bug Description

In `h5py/_conv.pyx`, the `frombuffer` code path wraps HDF5-allocated memory into a NumPy array and sets `NPY_ARRAY_OWNDATA`, telling NumPy to free the buffer on deallocation. However, NumPy uses `PyDataMem_FREE`/`free()` while the buffer was allocated by HDF5's internal allocator.

## Location

`h5py/_conv.pyx` (generated from `_conv.templ.pyx:731-754`)

## Code

```cython
# frombuffer path:
buf = <char[:itemsize * size]> data        # wraps HDF5-allocated data
ndarray = np.frombuffer(buf, dtype=elem_dtype)

PyArray_ENABLEFLAGS(ndarray, flags)        # flags includes NPY_ARRAY_OWNDATA
# Now NumPy thinks it owns `data` and will call free() on it
# But `data` was allocated by HDF5 (via vlen read), not by malloc/PyMem
```

## Impact

If HDF5 uses a different allocator than the system `malloc` (which can happen with custom HDF5 builds, debug builds, or Windows CRT configurations), calling `free()` on HDF5-allocated memory is undefined behavior that causes:
- Heap corruption
- Double-free if HDF5 also tries to free it
- Crash in the memory allocator

## Suggested Fix

Copy the data into a NumPy-owned buffer instead of wrapping with OWNDATA:

```cython
ndarray = np.empty(size, dtype=elem_dtype)
memcpy(PyArray_DATA(ndarray), data, itemsize * size)
efree(data)  # free using h5py's allocator wrapper
```

## Environment

- Affected: reading variable-length datasets (compound types with frombuffer path)

Fuyugithub:
## What does this PR do?

Fixes #2897 — `conv_vlen2ndarray` sets `NPY_ARRAY_OWNDATA` on HDF5-allocated memory, causing heap corruption.

## Root cause

`conv_vlen2ndarray` had three branches for building the output ndarray:

| Branch | dtype | What it did |
|---|---|---|
| `SimpleNewFromData` | `biufcmMO` (numeric/datetime/object) | Wrapped HDF5 pointer directly |
| `frombuffer` | strings / simple compounds | Wrapped via Cython memoryview |
| `np.empty + memcpy` | compound with object fields | Copied; correct |

Both the first two branches then hit:
```cython
PyArray_ENABLEFLAGS(ndarray, NPY_ARRAY_WRITEABLE | NPY_ARRAY_C_CONTIGUOUS | NPY_ARRAY_OWNDATA)
```

Setting `NPY_ARRAY_OWNDATA` tells NumPy to free the array's data pointer via `PyDataMem_FREE` (i.e. `free()` or `PyMem_RawFree()`). But the data came from HDF5's internal vlen allocator. On systems where HDF5 uses a custom allocator, this is **heap corruption or a double-free**.

The third branch already used the correct pattern: allocate a fresh NumPy buffer, `memcpy` the data, `efree` the HDF5 buffer.

## Fix

Consolidate all three branches into the single correct pattern:

```cython
# Always copy HDF5-allocated vlen data into a fresh NumPy-owned buffer and
# release the HDF5 buffer with the matching deallocator.
ndarray = np.empty(size, dtype=elem_dtype)
if size != 0:
    memcpy(PyArray_DATA(ndarray), data, itemsize * size)
efree(data)
```

`np.empty` creates a NumPy-owned buffer; NumPy will free it correctly with its own allocator. `efree` releases the HDF5 buffer with `free()`, matching how h5py allocates vlen memory.

## Trade-offs

- **One extra copy** for the `biufcmMO` path (numeric dtypes). This is the only way to avoid the UB without registering a custom base-object capsule with a HDF5-aware deallocator, which is significantly more complex.
- String / compound dtypes were already copying (the `frombuffer` shim copies on platforms where the buffer is not directly usable), so no observable difference there.

---

### DIFF
--- a/h5py/_conv.templ.pyx
+++ b/h5py/_conv.templ.pyx
@@ -720,36 +720,15 @@ cdef int conv_vlen2ndarray(void* ipt,
     finally:
         free(back_buf)
 
-    # We need to use different approaches to creating the ndarray with the converted
-    # data depending on the destination dtype.
-    # For simple dtypes, we can use SimpleNewFromData, but types like
-    # string & void need a size specified, so this function can't be used.
-    # Additionally, Cython doesn't expose NumPy C-API functions like NewFromDescr,
-    # so we fall back on copying directly to the underlying buffer
-    # of a new ndarray for other types.
-
-    if elem_dtype.kind in b"biufcmMO":
-        # type_num is enough to create an array for these dtypes
-        ndarray = cnp.PyArray_SimpleNewFromData(1, dims, elem_dtype.type_num, data)
-    elif not elem_dtype.hasobject:
-        # This covers things like string dtypes and simple compound dtypes,
-        # which can't be used with SimpleNewFromData.
-        # Cython doesn't expose NumPy C-API functions
-        # like NewFromDescr, so we'll construct this with a Python function.
-        if size != 0:
-            buf = <char[:itemsize * size]> data
-            ndarray = np.frombuffer(buf, dtype=elem_dtype)
-        else:
-            ndarray = np.empty(0, dtype=elem_dtype)
-    else:
-        # Compound dtypes containing object fields: frombuffer() refuses these,
-        # so we'll fall back to allocating a new array and copying the data in.
-        ndarray = np.empty(size, dtype=elem_dtype)
+    # Always copy HDF5-allocated vlen data into a fresh NumPy-owned buffer and
+    # release the HDF5 buffer with the matching deallocator.  Wrapping the
+    # HDF5 pointer directly (SimpleNewFromData / frombuffer) and then setting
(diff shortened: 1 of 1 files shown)


neutrinoceros:
Sorry it's been so long.
> Both the first two branches then hit

I'm confused, it looks to me that *all* three branches hit this call ?

other than that, LGTM

neutrinoceros (approved):

PR #2905, opened 2026-06-18
BUG: guard PyBuffer_Release on uninitialized view in write_direct_chunk

### LINKED ISSUE #2898: [BUG] Uninitialized Py_buffer released in write_direct_chunk on GetBuffer failure
## Bug Description

In `h5py/h5d.pyx`, if `PyObject_GetBuffer` fails, the `finally` block calls `PyBuffer_Release(&view)` on an uninitialized `Py_buffer` struct.

## Location

`h5py/h5d.pyx` (generated from `h5d.templ.pyx:519-529`)

## Code

```cython
try:
    offset = <hsize_t*>emalloc(sizeof(hsize_t)*rank)
    convert_tuple(offsets, offset, rank)
    PyObject_GetBuffer(data, &view, PyBUF_ANY_CONTIGUOUS)  # can fail!
    H5Dwrite_chunk(dset_id, dxpl_id, filter_mask, offset, view.len, view.buf)
finally:
    efree(offset)
    PyBuffer_Release(&view)  # BUG: view may be uninitialized if GetBuffer failed!
```

## Impact

If `PyObject_GetBuffer` fails (e.g., `data` doesn't support buffer protocol), `view` on the stack contains garbage. `PyBuffer_Release` reads from this garbage struct, causing:
- Crash (reading invalid pointer from `view.obj`)
- Memory corruption (calling `Py_XDECREF` on a garbage pointer)
- Potential code execution via controlled stack data

## Suggested Fix

Initialize `view` before the try block:

```cython
cdef Py_buffer view
memset(&view, 0, sizeof(view))  # zero-initialize
try:
    ...
```

Or guard the release:

```cython
cdef bint buffer_acquired = False
try:
    PyObject_GetBuffer(data, &view, PyBUF_ANY_CONTIGUOUS)
    buffer_acquired = True
    ...
finally:
    efree(offset)
    if buffer_acquired:
        PyBuffer_Release(&view)
```

## Environment

- Affected: `Dataset.id.write_direct_chunk()` with non-buffer objects

Fuyugithub:
## What does this PR do?

Fixes #2898 — `PyBuffer_Release` is called on an uninitialized `Py_buffer` when `PyObject_GetBuffer` fails in `write_direct_chunk`.

## Root cause

```cython
cdef Py_buffer view   # declared on stack, never initialized

try:
    ...
    PyObject_GetBuffer(data, &view, PyBUF_ANY_CONTIGUOUS)  # can raise
    H5Dwrite_chunk(...)
finally:
    efree(offset)
    PyBuffer_Release(&view)   # BUG: called even if GetBuffer raised
```

If `PyObject_GetBuffer` raises (e.g. `data` does not support the buffer protocol), `view` was never populated — its fields contain whatever happened to be on the stack.  `PyBuffer_Release` unconditionally calls `Py_XDECREF(view.obj)`, which dereferences a garbage pointer.  This is undefined behaviour and typically crashes the interpreter.

## Fix

Add a `view_acquired` flag.  `PyBuffer_Release` is only called when `PyObject_GetBuffer` returned successfully, which matches the CPython documentation contract:

> The exporter of the buffer is responsible for deciding whether `bf_releasebuffer` is necessary.  Only call `PyBuffer_Release` if it was successfully acquired via `PyObject_GetBuffer`.

```cython
cdef bint view_acquired = False

try:
    ...
    PyObject_GetBuffer(data, &view, PyBUF_ANY_CONTIGUOUS)
    view_acquired = True
    H5Dwrite_chunk(...)
finally:
    efree(offset)
    if view_acquired:
        PyBuffer_Release(&view)
```

---

### DIFF
--- a/h5py/h5d.templ.pyx
+++ b/h5py/h5d.templ.pyx
@@ -510,6 +510,7 @@ cdef class DatasetID(ObjectID):
         cdef size_t data_size
         cdef int rank
         cdef Py_buffer view
+        cdef bint view_acquired = False
 
         dset_id = self.id
         dxpl_id = pdefault(dxpl)
@@ -523,10 +524,12 @@ cdef class DatasetID(ObjectID):
             offset = <hsize_t*>emalloc(sizeof(hsize_t)*rank)
             convert_tuple(offsets, offset, rank)
             PyObject_GetBuffer(data, &view, PyBUF_ANY_CONTIGUOUS)
+            view_acquired = True
             H5Dwrite_chunk(dset_id, dxpl_id, filter_mask, offset, view.len, view.buf)
         finally:
             efree(offset)
-            PyBuffer_Release(&view)
+            if view_acquired:
+                PyBuffer_Release(&view)
             if space_id:
                 H5Sclose(space_id)
 


neutrinoceros (approved):
Thanks !

PR #2907, opened 2026-06-19
BLD: fix abi3 builds

### LINKED ISSUE #2906: BLD: non-abi3 symbols detected by `abi3audit` on LIMITED_API builds
[example logs](https://github.com/h5py/h5py/actions/runs/27761551203/job/82136971369?pr=2902)

Seen on macOS and Windows. Linux is clear.
The reason this only showed up now is that I upgraded cibuildwheel from <4 to 4.1.0, and 4.0.0 introduced auditing for `abi3` wheels.

neutrinoceros:
Tentatively address #2906

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -120,10 +120,8 @@ jobs:
       # Triage build
       - run: bash ./ci/triage_build.sh "${{ matrix.arch }}" "${{ github.event.pull_request.head.sha || github.sha }}"
 
-      # Now actually build the wheels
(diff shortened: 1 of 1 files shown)


neutrinoceros:
This solves the issue I was chasing. However, I just now noticed that Linux nightlies do *not* ship as abi3, contrary to my intention, so I'll try to also address this here.

neutrinoceros:
Interestingly, the error I'm after comes back the minute I switch the first cibw job back on, suggesting the problem is really a leak of some sort from one job to the other.
Maybe I should just re-order them.

PR #2929, opened 2026-07-07
`H5dont_atexit()` Control (for WASM)

### LINKED ISSUE #2928: WASM Builds: `H5dont_atexit()`
Hi,

I am doing a WASM32 build with multiple (private, static) HDF5 libraries in [openPMD-api](https://github.com/openPMD/openPMD-api/), [ImpactX](https://github.com/ECP-WarpX/impactx) and `h5py` for pyodide PyPI packages.

[Since the WASM build process and export of symbols is a bit special and involved compared to native builds](https://emscripten.org/docs/compiling/Dynamic-Linking.html), a problem is that WASM does a symbol interposition that leads to multiple libraries depending on HDF5 to call its teardown multiple times.
I am working around resulting [teardown issues](https://github.com/openPMD/openPMD-api/pull/1900) setting [H5dont_atexit](https://support.hdfgroup.org/documentation/hdf5/latest/group___h5.html#title7) for WASM-only builds

Would you be open to a PR that calls `H5dont_atexit()` guarded by `#if defined(__EMSCRIPTEN__)` or `if _sys.platform == "emscripten"` at the appropriate location in `h5py`? The result is a small memory leak at shutdown, which will be reclaimed by the OS anyway on process end, fixing a non-zero exit status otherwise that raises on process shutdown from HDF5.

ax3l:
Close #2928: On Emscripten/WebAssembly (e.g. Pyodide), h5py now calls HDF5's ``H5dont_atexit()`` before its first HDF5 call, preventing crashes at interpreter exit when several wheels bundle their own statically linked copies of HDF5.  The new environment variable ``H5PY_DONT_ATEXIT`` (``1``/``0``) forces this behavior on or off on any platform. For all existing platforms besides Emscripten, the existing cleanup is used as the default.

### DIFF
--- a/h5py/_errors.templ.pxd
+++ b/h5py/_errors.templ.pxd
@@ -411,6 +411,18 @@ cdef extern from "hdf5.h":
     ctypedef herr_t (*H5E_walk_t)(unsigned int n, const H5E_error_t *err_desc, void* client_data)
     herr_t    H5Ewalk(hid_t estack_id, H5E_direction_t direction, H5E_walk_t func, void* client_data)
 
+    # --- Not error-related ---------------------------------------------------
+    # H5dont_atexit() is declared here rather than in api_functions.txt
+    # because the auto-generated wrappers in defs.pyx call
+    # set_default_error_handler() (-> H5Eset_auto) before the real HDF5
+    # function, which would initialise the HDF5 library and turn the
+    # H5dont_atexit() call into a no-op.  It is called at the top of
+    # _errors.pyx: _errors is the first HDF5-linked module h5py imports,
+    # and its own module init is also the first thing to initialise the
+    # HDF5 library (building the error tables expands each H5E_* code to
+    # `H5open(), H5E_..._g` in C).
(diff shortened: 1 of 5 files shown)


h5py/_errors.templ.pxd
```diff
@@ -411,6 +411,18 @@ cdef extern from "hdf5.h":
     ctypedef herr_t (*H5E_walk_t)(unsigned int n, const H5E_error_t *err_desc, void* client_data)
     herr_t    H5Ewalk(hid_t estack_id, H5E_direction_t direction, H5E_walk_t func, void* client_data)
 
+    # --- Not error-related ---------------------------------------------------
```
tacaswell, line 414: Can you add a note on the `_errors` import in `__init__.py` as well?  We don't want to accidentally re-arrange things and break this!

tacaswell:
Is it worth making a new `_configure_atexit` modules to have be the first import?  Putting this in `_errors` is practical but maybe a bit brittle/surprising.

PR #2931, opened 2026-07-20
Release the GIL in H5Ocopy()

takluyver:
`f.copy()` can copy an arbitrary amount of data, including from one file to another. It would be handy to release the GIL while doing this so other threads can make progress.

### DIFF
--- a/h5py/api_functions.txt
+++ b/h5py/api_functions.txt
@@ -249,7 +249,7 @@ hdf5:
   herr_t    H5Oget_info_by_idx(hid_t loc_id, char *group_name,  H5_index_t idx_type, H5_iter_order_t order, hsize_t n, H5O_info_t *oinfo, hid_t lapl_id)
 
   herr_t    H5Olink(hid_t obj_id, hid_t new_loc_id, char *new_name, hid_t lcpl_id, hid_t lapl_id)
-  herr_t    H5Ocopy(hid_t src_loc_id, char *src_name, hid_t dst_loc_id,  char *dst_name, hid_t ocpypl_id, hid_t lcpl_id)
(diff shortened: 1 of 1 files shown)


h5py/api_functions.txt
```diff
@@ -249,7 +249,7 @@ hdf5:
   herr_t    H5Oget_info_by_idx(hid_t loc_id, char *group_name,  H5_index_t idx_type, H5_iter_order_t order, hsize_t n, H5O_info_t *oinfo, hid_t lapl_id)
 
   herr_t    H5Olink(hid_t obj_id, hid_t new_loc_id, char *new_name, hid_t lcpl_id, hid_t lapl_id)
-  herr_t    H5Ocopy(hid_t src_loc_id, char *src_name, hid_t dst_loc_id,  char *dst_name, hid_t ocpypl_id, hid_t lcpl_id)
+  herr_t    H5Ocopy(hid_t src_loc_id, char *src_name, hid_t dst_loc_id,  char *dst_name, hid_t ocpypl_id, hid_t lcpl_id) nogil
```
neutrinoceros, line 252: As I understand it this doesn't release the GIL but merely *allows* it to be released. Is this what you meant ?

https://cython.readthedocs.io/en/latest/src/userguide/nogil.html#marking-functions-as-able-to-run-without-the-gil

takluyver: You're right if this was just the signature of a Cython function, but our custom wrapper generator uses this as a marker to say that it should make a `with nogil:` block inside the function ~~as well~~ instead.

https://github.com/h5py/h5py/blob/5069ff2705e0b992e83486e07bf410f3663fbdd5/api_gen.py#L268-L272

neutrinoceros: wow, I completely missed that. Carry on then !

PR #2932, opened 2026-07-20
Let cmake select which Visual Studio version to build with

### LINKED ISSUE #2913: WHL: Windows AMD64 jobs fail to build zlib
[Example logs](https://github.com/h5py/h5py/actions/runs/27834731361/job/82379910088?pr=2893)

```
Building zlib into D:\a\h5py\h5py/zlib-win-x64
CMake Error at CMakeLists.txt:3 (project):
  Generator

    Visual Studio 17 2022

  could not find any instance of Visual Studio.



-- Configuring incomplete, errors occurred!
```

I've been ignoring this problem for a couple days because I thought it was #2816, but I now realized it's an other target.

takluyver:
Fixes #2913 (hopefully)

I'm roughly hoping that specifying the VS version is no longer necessary. I think it used to be necessary to match up the version building extensions with the version used to build Python itself, but we haven't paid much attention to doing so for years, and it seems to still work.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -135,11 +135,14 @@ jobs:
           H5PY_LIMITED_API: 1
           CIBW_BUILD: 'cp314-* cp315t-*'
           CIBW_ENABLE: cpython-prerelease
+          CIBW_REPAIR_WHEEL_COMMAND_WINDOWS: ''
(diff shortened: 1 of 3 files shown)


takluyver:
As this appears to be working, I'll merge it so that the CI works again.

neutrinoceros:
Thank you !

PR #2933, opened 2026-07-21
Don't set HDF5_VSVERSION in Azure pipelines

takluyver:
Follow up to #2932, to see if we can skip setting this on Azure as well.

I've left the setting on Appveyor in place for now, because I think it's how we tell it to do a 32-bit build (we've kept Appveyor specifically for testing 32-bit builds).


takluyver:
That seems to be OK. CMake says:

```
-- Building for: Visual Studio 17 2022
-- Selecting Windows SDK version 10.0.26100.0 to target Windows 10.0.20348.
-- The C compiler identification is MSVC 19.44.35228.0
```

These are the same versions as when we were specifying HDF5_VSVERSION.

neutrinoceros:
Cool. Thanks !

neutrinoceros (approved):

PR #2935, opened 2026-07-28
Fix NULL pointer dereference in h5g.get_objname_by_idx and h5o.get_comment

lilu5458:
## Description

Both `h5g.get_objname_by_idx()` and `h5o.get_comment()` call HDF5 C functions that return a `ssize_t` value representing the required buffer size, which is negative (-1) on error. However, neither function checked the return value before passing it to `emalloc(size+1)`.

When the underlying HDF5 C function returns -1 (e.g., when the object name does not exist or the index is out of range), `emalloc(size+1)` evaluates to `emalloc(0)`. Depending on the platform:

- **NULL pointer dereference (CWE-476)**: `emalloc(0)` may return NULL, and the subsequent `H5Gget_objname_by_idx` / `H5Oget_comment_by_name` call writes to this NULL pointer, causing a crash.
- **Uninitialized memory read (CWE-908)**: On some allocators, `emalloc(0)` returns a small non-NULL buffer, and the code reads uninitialized heap memory, potentially exposing sensitive data.

## Fix

Add `assert size >= 0` after the first HDF5 API call (the sizing call) in both functions, matching the existing pattern already used in:

- `h5f.get_name()` (h5f.pyx)
- `h5g.get_comment()` (h5g.pyx)

This ensures that if the HDF5 C function returns an error, an `AssertionError` is raised immediately rather than proceeding with an invalid buffer size.

## Files changed

- `h5py/h5g.pyx`: Added `assert size >= 0` after `H5Gget_objname_by_idx()` sizing call (1 line)
- `h5py/h5o.pyx`: Added `assert size >= 0` after `H5Oget_comment_by_name()` sizing call (1 line)
- `h5py/tests/test_null_deref_fix.py`: Unit tests verifying error handling for both functions

## Testing

The fix was verified locally:

1. **PoC verification**: Before the fix, calling `h5o.get_comment(group.id, b"", obj_name=b"nonexistent_object")` returned uninitialized heap memory. After the fix, it raises `AssertionError`.

2. **Unit tests**: `python -m pytest h5py/tests/test_null_deref_fix.py` — all tests pass.

3. **Existing functionality**: Verified that normal use cases (valid objects, valid indices) still work correctly.

## Risk assessment

This is a minimal, low-risk change (2 lines of production code) that follows an existing pattern in the codebase. It only adds an assertion check that triggers on error paths — normal operation is unaffected.

### DIFF
--- a/h5py/h5g.pyx
+++ b/h5py/h5g.pyx
@@ -353,6 +353,7 @@ cdef class GroupID(ObjectID):
         buf = NULL
 
         size = H5Gget_objname_by_idx(self.id, idx, NULL, 0)
+        assert size >= 0
 
         buf = <char*>emalloc(sizeof(char)*(size+1))
         try:
--- a/h5py/h5o.pyx
+++ b/h5py/h5o.pyx
@@ -316,6 +316,7 @@ def get_comment(ObjectID loc not None, char* comment, *, char* obj_name=".",
     cdef char* buf
 
     size = H5Oget_comment_by_name(loc.id, obj_name, NULL, 0, pdefault(lapl))
+    assert size >= 0
     buf = <char*>emalloc(size+1)
     try:
         H5Oget_comment_by_name(loc.id, obj_name, buf, size+1, pdefault(lapl))
--- a/h5py/tests/test_null_deref_fix.py
+++ b/h5py/tests/test_null_deref_fix.py
@@ -1,0 +1,106 @@
+#!/usr/bin/env python3
+"""
+Unit tests for h5py NULL-pointer-dereference / uninitialized-memory-read fix.
+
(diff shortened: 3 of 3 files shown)


neutrinoceros:
While the problem is real, there are a couple issues with this PR as it stands:
- CI failed (logs are too old and unavailable, so I can't tell what happened)
- the tests are not following existing style and hardly maintainable (nested try/finally blocks)

The second point makes me suspect this patch was written entirely by, or with the assistance of LLMs; if there's a human reading this, please let us know if you are still interested in fixing this, and please disclose any use of LLMs in your work. I reserve the right to close this without merge if I don't hear back or if this is all LLM-backed.

PR #2936, opened 2026-07-30
DEP: upgrade cibuilwheel to 4.1.1

neutrinoceros:
This should fix recent failures with Python 3.15, where I think the root problem is simply some ABI incompatibility between beta 2 (which we're currently on) and beta 4 (which latest numpy nightlies were built against)


neutrinoceros:
Merging now to confirm the fix

neutrinoceros:
This isn't sufficient to fix Linux builds because we also need to update our docker images. Doing it now.

PR #2938, opened 2026-07-30
MNT: disable pre-commit.ci autofixes to PRs

neutrinoceros:
While genuine, this PR is mostly a pretext to exercise cp315 wheels and check the stability of the build process.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,3 +14,4 @@ repos:
 
 ci:
     autoupdate_schedule: monthly
+    autofix_prs: false


neutrinoceros:
macOS and Linux are stable. Windows-AMD64 isn't:
```
Loading library to get build settings and version: D:\a\h5py\h5py\cache\hdf5\2.0.0\lib\hdf5.dll
      on sys.platform='win32' with platform.machine()='AMD64'
      Library dirs checked: ['D:\\a\\h5py\\h5py/cache/hdf5/2.0.0\\lib', 'D:\\a\\h5py\\h5py/cache/hdf5/2.0.0\\bin']
      error: [WinError 193] %1 is not a valid Win32 application
      error: subprocess-exited-with-error
      
      Building wheel for h5py (pyproject.toml) did not run successfully.
      exit code: 1
```

neutrinoceros:
If there's the slightest hint of controversy here I'll simply drop it. This isn't a hill worth dying on.

tacaswell:
did we give pre-commit.io the ability to push back?  If so we should revoke that at the GH permissions level not at its config level.

neutrinoceros:
I don't know that it is supported. Last time I checked there was no other way to fine tune permissions for the pre-commit.ci app

tacaswell:
Are we even still using pre-commit.ci?  I tried to check but do not want to authorize it against my GH account (I do not trust them) but I do not see it in the status pushed back here.

I think we now run pre-commit via the `static_checks` job on azure and we can purge this config file.

neutrinoceros:
Oh right, we don't. How did I miss this 😅
Thanks !

Issue #2802, opened 2026-02-23
`.require_group()` missing `track_order`, Conflicting doc.


desplenterkarel:
The `.require_group()` function is missing the argument `track_order`.

the docs say: 
_**Parameters as in [Group.create_group()](https://docs.h5py.org/en/3.12.1/high/group.html#h5py.Group.create_group).**_

But the `track_order `is argument is missing.
Thus given a goup doesn't exist you can create it this way an get `tracked_order`.
Forcing the use of manually checking and creating for going the use of this function.
It can also be fixt by setting `h5py.get_config().track_order = True` globally (which is not always ideal).

```python

if name not in f:
        f.create_group(name, track_order=True)
else
    group = f[name]
```

Doc for [reference](https://docs.h5py.org/en/latest/high/group.html#h5py.Group.require_group):
```
create_group(name, track_order=None) [*](https://docs.h5py.org/en/3.12.1/high/group.html#h5py.Group.create_group)
Create and return a new group in the file.

Parameters:
name (String or None) – Name of group to create. May be an absolute or relative path.
 Provide None to create an anonymous group, to be linked into the file later.

track_order – Track dataset/group/attribute creation order under this group if True.
 Default is h5.get_config().track_order.

Returns:
The new [Group](https://docs.h5py.org/en/3.12.1/high/group.html#h5py.Group) object.

require_group(name) [*](https://docs.h5py.org/en/3.12.1/high/group.html#h5py.Group.require_group)
Open a group in the file, creating it if it doesn’t exist. TypeError is raised if a conflicting object already exists.
 Parameters as in [Group.create_group()](https://docs.h5py.org/en/3.12.1/high/group.html#h5py.Group.create_group).
```

takluyver:
Parameters in 'require' methods are always a little bit ambiguous: do they mean 'use these settings if creating a group' or 'require that a group exists with these properties'? For `require_dataset()`, it's a mixture. @tacaswell probably has an opinion on this.

I'd be OK with fixing it if we can agree what fixing it should mean. Or I'd be fine with changing the docs and telling people to use an `if` if they need something more. I don't think this is too ugly, especially if you use the walrus operator:

```python
if (group := f.get(name)) is None:
    group = f.create_group(name, ...)
```

(This doesn't consider the case where `name` exists as e.g. a dataset instead of a group. But I think in a lot of use cases, you're confident that this will be a group if it is present, so this could be good enough.)

PR #2939, opened 2026-07-30
ENH allow require_group to set track_order

Whning0513:
Addresses h5py/h5py#2802.

Add the missing `track_order` argument to `Group.require_group`, pass it through when creating a new group, and document the behavior. A regression test verifies that creation order is preserved.

### DIFF
--- a/docs/high/group.rst
+++ b/docs/high/group.rst
@@ -378,11 +378,15 @@ Reference
         :return:        The new :class:`Group` object.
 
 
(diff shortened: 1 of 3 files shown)


neutrinoceros:
As per https://github.com/h5py/h5py/issues/2802#issuecomment-3950587821

> I'd be OK with fixing it if we can agree what fixing it should mean.

No further discussion happened since so consensus isn't there yet to support this approach.

PR #2943, opened 2026-07-31
WHL: stop building and shipping win32 wheels

neutrinoceros:
Following this arch being dropped by numpy.
Builds are currently failing so I'm hoping this could get us back to a stable state.


tacaswell (approved):

neutrinoceros (commented):
Whoops. I accidentally included workflow upgrades too.

PR #2946, opened 2026-08-01
Update version info in version.py

### LINKED ISSUE #2945: h5py version info out of sync
The two places with h5py version value are out of sync:

1. https://github.com/h5py/h5py/blob/597fb5f3e498ef0ebbf0592467df64a449dc398b/pyproject.toml#L17
2. https://github.com/h5py/h5py/blob/597fb5f3e498ef0ebbf0592467df64a449dc398b/h5py/version.py#L40-L41

ajelenak:
Brings it in sync with the `pyproject.toml` value.

Fixes #2945.


neutrinoceros (approved):
Whoops, good catch

Issue #2930, opened 2026-07-17
Regression in h5py 3.16 / HDF5 2.0.0: valid chunked datasets fail to open


Abbo93:
## Title
Regression in h5py 3.16 / HDF5 2.0.0: valid chunked datasets fail to open with "stored datatype size in chunk layout does not match datatype description" / "stored chunk dimension encoding length does not match value calculated from chunk dimensions"

## Summary

A production HDF5 file that opens correctly in h5py 3.15.1 (HDF5 1.14.6) fails to open three chunked, gzip+shuffle-compressed datasets after upgrading to h5py 3.16.0 (HDF5 2.0.0). The file is written by a third-party C# library (PureHDF) and has not changed — the same file opens cleanly with h5dump from HDF5 1.10.10 and with h5py 3.15.1, but is rejected by h5py 3.16.0.

This looks like a decoding/validation regression introduced alongside the HDF5 2.0.0 chunk-layout rework (release notes for 2.0.0 mention a new, larger chunk-size limit and a new file format version 4.0, which changed how the chunk layout object header message is encoded/decoded), rather than an actual defect in the file.

## Environment

**Failing:**
- h5py 3.16.0
- HDF5 2.0.0
- Python 3.12.3
- numpy 2.4.4
- OS: Ubuntu 24.04 (also occurs on the reporter's original machine after upgrading from h5py 3.15.1 → 3.16)

**Working (same file):**
- h5py 3.15.1
- HDF5 1.14.6
- Python 3.12.3
- numpy 2.5.1

Also confirmed readable with `h5dump` from HDF5 1.10.10 (system Ubuntu package), so the file is not affected by two independent older library versions.

## Steps to reproduce

```python
import h5py

f = h5py.File("test_file.hdf5", "r")
f["Profile"]["Luminance"]   # raises KeyError
f["Profile"]["X_mm"]        # raises KeyError
f["Profile"]["Z_mm"]        # raises KeyError
```

## Observed errors

```
--- Luminance ---
KeyError: 'Unable to synchronously open object (stored chunk dimension encoding
length does not match value calculated from chunk dimensions)'

--- X_mm ---
KeyError: 'Unable to synchronously open object (stored datatype size in chunk
layout does not match datatype description)'

--- Z_mm ---
KeyError: 'Unable to synchronously open object (stored datatype size in chunk
layout does not match datatype description)'
```

Full traceback (representative, same shape for all three):

```
Traceback (most recent call last):
[...]

PR #2947, opened 2026-08-02
Bundle HDF5 2.2.0 in wheels

jrideout:
The current wheels bundle HDF5 2.0.0, which is affected by a number of published CVEs that were fixed upstream in HDF5 2.1.0 and 2.2.0 (there are no 2.0.x patch releases, so bumping is the only way to pick up the fixes). This PR moves the bundled HDF5 for the Mac and Windows wheels to 2.2.0 (released 2026-07-29).

## CVEs present in HDF5 2.0.0

Fixed in 2.1.0:

* **CVE-2025-44904** — heap buffer overflow in `H5VM_memcpyvv` via an unvalidated stored chunk size, reachable by opening a crafted file (NVD rates this 8.8 HIGH)
* **CVE-2025-2308** — heap overflow in the scale-offset filter decompression (`H5Z__scaleoffset_decompress_one_byte`)
* **CVE-2025-2309** — heap overflow in type conversion (`H5T__bit_copy`)
* **CVE-2026-26197** ([GHSA-gh44-7wpq-622f](https://github.com/HDFGroup/hdf5/security/advisories/GHSA-gh44-7wpq-622f)) — out-of-bounds read decoding a malformed array datatype; the advisory explicitly lists 2.0.0 as the affected version
* **CVE-2026-26199** ([GHSA-5c6x-jmgf-f5vc](https://github.com/HDFGroup/hdf5/security/advisories/GHSA-5c6x-jmgf-f5vc)) — buffer underflow in `H5Iget_name`/`H5G_get_name` when called with size 0

Fixed in 2.2.0:

* **CVE-2026-17572** — heap buffer overflow deserializing a crafted SOHM list index (HDF Group CNA lists all versions <= 2.1.1 as affected)
* **CVE-2026-17573** — double free in the chunk-copy code via a crafted chunk-index size field
* **CVE-2026-17574** — NULL pointer dereference decoding an invalid variable-length datatype
* **CVE-2025-9274** — NULL callback dereference via `H5Aiterate2` (root cause fixed in the library in 2.2.0; not reachable through h5py itself, which always passes a real callback)

The HDF Group's own CVE tracking is at [HDFGroup/cve_hdf5](https://github.com/HDFGroup/cve_hdf5).

## Changes

* `HDF5_VERSION` 2.0.0 → 2.2.0 in `ci/cibw_before_all_macos.sh` and `ci/cibw_before_all_windows.sh`
* HDF5 releases after 2.1.0 are tagged with the plain version only (e.g. `2.2.0` — the `hdf5-2_2_0`/`hdf5_2.2.0` conventions stopped after 2.1.0), so `get_hdf5_if_needed.sh` and `get_hdf5_win.py` now try the plain tag first and fall back to the older conventions
* Added an HDF5 2.2.0 entry to the Azure test matrix, mirroring the 2.0.0 entry added in #2750

Linux wheels take their HDF5 from the `h5py/hdf5-manylinux` images; companion PR: h5py/hdf5-manylinux#33

## Testing

Built HDF5 2.2.0 locally on macOS arm64 with the same CMake invocation `get_hdf5_if_needed.sh` uses, built h5py against it (Python 3.14), and ran the full test suite:

```
874 passed, 28 skipped, 15 warnings, 3 subtests passed in 17.14s
```

All skips are environment-specific (MPI, ros3, szip, the Linux-only DIRECT driver, complex256 on arm64, and one test marked "Requires HDF5 <= 1.14.3").

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -71,6 +71,13 @@ jobs:
         HDF5_VERSION: 2.0.0
         HDF5_DIR: $(HDF5_CACHE_DIR)/$(HDF5_VERSION)
         H5PY_ENFORCE_COVERAGE: yes
+      py314-deps-hdf5-2_2:
+        python.version: '3.14'
+        python.architecture: 'x64'
+        TOXENV: py314-test-deps
+        HDF5_VERSION: 2.2.0
+        HDF5_DIR: $(HDF5_CACHE_DIR)/$(HDF5_VERSION)
+        H5PY_ENFORCE_COVERAGE: yes
     # do mpi tests
       py310-deps-hdf51122-mpi:
         python.version: '3.10'
--- a/ci/get_hdf5_if_needed.sh
+++ b/ci/get_hdf5_if_needed.sh
@@ -34,7 +34,9 @@ else
         echo "building HDF5"
 
         IFS='.-' read MAJOR_V MINOR_V REL_V PATCH_V <<< "$HDF5_VERSION"
-        # the assets in GitHub (currently) have two naming conventions
+        # the assets in GitHub (currently) have three naming conventions;
+        # releases after 2.1.0 are tagged with the plain version only
+        ASSET_FMT3="${HDF5_VERSION}"
         if [[ -n "${PATCH_V}" ]]; then
             ASSET_FMT1="hdf5-${MAJOR_V}_${MINOR_V}_${REL_V}-${PATCH_V}"
             ASSET_FMT2="hdf5_${MAJOR_V}.${MINOR_V}.${REL_V}.${PATCH_V}"
@@ -79,7 +81,7 @@ else
 
         pushd /tmp
         url_base="https://github.com/HDFGroup/hdf5/archive/refs/tags/"
-        curl -fsSL -o "hdf5-$HDF5_VERSION.tar.gz" "${url_base}${ASSET_FMT1}.tar.gz" || curl -fsSL -o "hdf5-$HDF5_VERSION.tar.gz" "${url_base}${ASSET_FMT2}.tar.gz"
+        curl -fsSL -o "hdf5-$HDF5_VERSION.tar.gz" "${url_base}${ASSET_FMT3}.tar.gz" || curl -fsSL -o "hdf5-$HDF5_VERSION.tar.gz" "${url_base}${ASSET_FMT1}.tar.gz" || curl -fsSL -o "hdf5-$HDF5_VERSION.tar.gz" "${url_base}${ASSET_FMT2}.tar.gz"
 
         mkdir -p hdf5-$HDF5_VERSION && tar -xzvf hdf5-$HDF5_VERSION.tar.gz --strip-components=1 -C hdf5-$HDF5_VERSION
 
--- a/ci/get_hdf5_if_needed.sh
+++ b/ci/get_hdf5_if_needed.sh
@@ -81,7 +81,23 @@ else
 
         pushd /tmp
         url_base="https://github.com/HDFGroup/hdf5/archive/refs/tags/"
-        curl -fsSL -o "hdf5-$HDF5_VERSION.tar.gz" "${url_base}${ASSET_FMT3}.tar.gz" || curl -fsSL -o "hdf5-$HDF5_VERSION.tar.gz" "${url_base}${ASSET_FMT1}.tar.gz" || curl -fsSL -o "hdf5-$HDF5_VERSION.tar.gz" "${url_base}${ASSET_FMT2}.tar.gz"
+        urls=(
+            "${url_base}${ASSET_FMT3}.tar.gz"
+            "${url_base}${ASSET_FMT1}.tar.gz"
+            "${url_base}${ASSET_FMT2}.tar.gz"
+        )
+        set +e
+        for url in "${urls[@]}"; do
+            echo "downloading from $url"
+            curl --location "$url" --output "hdf5-$HDF5_VERSION.tar.gz" --fail --silent --show-error
+            if [[ "$?" == 0 ]]; then
+                echo "download succeeded"
+                break
+            else
+                echo "download failed"
+            fi
+        done
+        set -e
 
         mkdir -p hdf5-$HDF5_VERSION && tar -xzvf hdf5-$HDF5_VERSION.tar.gz --strip-components=1 -C hdf5-$HDF5_VERSION
 
--- a/ci/get_zlib_windows.sh
+++ b/ci/get_zlib_windows.sh
@@ -12,11 +12,25 @@ ZLIB_VERSION="1.3.2"
 ZLIB_DIR="zlib-$ZLIB_VERSION"
 
 if [ ! -d "$ZLIB_DIR" ]; then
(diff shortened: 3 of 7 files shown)


neutrinoceros:
As I said on the companion thread I'm favorable to the change but not to listing Claude (or any LLM) as a contributor.

azure-pipelines.yml
```diff
@@ -71,6 +71,13 @@ jobs:
         HDF5_VERSION: 2.0.0
         HDF5_DIR: $(HDF5_CACHE_DIR)/$(HDF5_VERSION)
         H5PY_ENFORCE_COVERAGE: yes
+      py313-deps-hdf5-2_2:
+        python.version: '3.13'
```
neutrinoceros:
```suggestion
      py314-deps-hdf5-2_2:
        python.version: '3.14'
```

neutrinoceros (changes requested):

jrideout:
Thanks for the review! Applied the suggestion — the new matrix entry is now py314/'3.14' (with matching TOXENV) — and removed the AI co-author trailer from the commit and the PR description.

neutrinoceros:
Thank you. Do you know of any backward compatibility issue this upgrade might cause ?

jrideout:
Nothing breaking for h5py itself. The C API delta is purely additive, the soname is unchanged, and write-side format defaults are the same as 2.0.0, so files written by these wheels stay readable by older libraries. Test suite passes unchanged.

One thing worth knowing about: 2.2.0 added filter-pipeline validation that rejects a write when a filter reports a data size larger than its reported buffer size. I audited the filters in hdf5plugin main plus lzf and deflate — only Blosc2 trips it, on poorly-compressible chunks. A fix is proposed upstream at Blosc/HDF5-Blosc2#5 but isn't in any released hdf5plugin, and it doesn't look reported on their tracker; I can open an issue there.

Two smaller behavior changes: `can_apply`/`set_local` now run for vlen datatypes (so creating a filtered vlen dataset can fail, or silently drop an optional filter), and `h5py.version.hdf5_version` becomes `"2.2.0"`, which breaks exact-version assertions downstream.

On the plus side, the relaxed chunk dimension-size check (HDFGroup/hdf5#6432) means the PureHDF- and JLD2.jl-written files rejected by the current wheels (#2930, JuliaIO/HDF5.jl#1224) are readable again.

The CVE fixes only change behavior on malformed or truncated files. One case a real workflow might hit: new superblock EOF/address checks mean recovery reads of truncated files that previously opened partially may now fail at open.

jrideout:
The Windows wheel failure was zlib.net serving an error page that curl (without `-f`) saved as the tarball. Pushed a small hardening of that download: fail on HTTP errors, retry, and fall back to the zlib GitHub release mirror.

ajelenak:
> so files written by these wheels stay readable by older libraries.

Depends what _older_ means here. If older means prior to 2.0.0 then the statement is not always correct.

As for backward compatibility and soname... Since 2.0.0 when the SemVer convention was first adopted, any release where the major number is unchanged is ABI compatible.

jrideout:
Windows wheels failed again with the same signature even with `curl -f`: zlib.net is intermittently answering HTTP 200 with an error page, so there is no HTTP failure to detect. Switched the download to the zlib GitHub release mirror first with a `gzip -t` integrity check, keeping zlib.net as fallback. Apologies for the extra workflow-approval round-trip this push causes.

ci/get_hdf5_if_needed.sh
```diff
@@ -79,7 +81,7 @@ else
 
         pushd /tmp
         url_base="https://github.com/HDFGroup/hdf5/archive/refs/tags/"
-        curl -fsSL -o "hdf5-$HDF5_VERSION.tar.gz" "${url_base}${ASSET_FMT1}.tar.gz" || curl -fsSL -o "hdf5-$HDF5_VERSION.tar.gz" "${url_base}${ASSET_FMT2}.tar.gz"
+        curl -fsSL -o "hdf5-$HDF5_VERSION.tar.gz" "${url_base}${ASSET_FMT3}.tar.gz" || curl -fsSL -o "hdf5-$HDF5_VERSION.tar.gz" "${url_base}${ASSET_FMT1}.tar.gz" || curl -fsSL -o "hdf5-$HDF5_VERSION.tar.gz" "${url_base}${ASSET_FMT2}.tar.gz"
```
neutrinoceros: I suggest re-using the refactor I wrote on your companion PR here

ci/get_zlib_windows.sh
```diff
@@ -12,7 +12,12 @@ ZLIB_VERSION="1.3.2"
 ZLIB_DIR="zlib-$ZLIB_VERSION"
 
 if [ ! -d "$ZLIB_DIR" ]; then
-  curl -sLO https://zlib.net/fossils/$ZLIB_DIR.tar.gz
+  # zlib.net intermittently answers HTTP 200 with an error page, which curl -f
+  # cannot detect (it broke this job twice), so fetch from the zlib GitHub
+  # release mirror first and verify the archive, falling back to zlib.net.
+  curl -fsSLO --retry 5 https://github.com/madler/zlib/releases/download/v$ZLIB_VERSION/$ZLIB_DIR.tar.gz && gzip -t $ZLIB_DIR.tar.gz 2>/dev/null \
```
neutrinoceros: and here

jrideout:
Applied your url-loop refactor to both scripts — the HDF5 download now iterates the three tag conventions, and the Windows zlib download iterates GitHub-mirror-then-zlib.net with a `gzip -t` check inside the loop (zlib.net's failure mode here was HTTP 200 with an error page, so verifying content is what actually catches it).

neutrinoceros (changes requested):
I think all we need now is a changelog entry for 3.17

tacaswell:
Just to confirm: there is no format change between 2.0.0 and 2.2.0?  When swmr came in (iirc) there was a change to the format where files written with the new scheme could not be read by older version of the library which has made be knee-jerk cautious about updating the bundled version of libhdf5.

ajelenak:
No format change. When format changes the major number will increase, e.g., from 2.x to 3.0.

jrideout:
Confirmed from the source diff as well: default libver bounds, the `H5F_LIBVER_LATEST` mapping, and the maximum superblock version are identical between the 2.0.0 and 2.2.0 tags (`H5Fsuper.c` is byte-identical). The only format-adjacent additions are opt-in FP6/FP4 predefined datatypes (2.1.0), which produce datatype messages older readers can't decode only if explicitly used — and h5py doesn't expose them. Nothing comparable to the 1.10 SWMR/superblock change.

Changelog entry for 3.17 added in the latest push.

ajelenak:
The ABI report for the release is available at https://github.com/HDFGroup/hdf5/releases/download/2.2.0/hdf5-2.2.0.html.abi.reports.tar.gz.

tacaswell (approved):

neutrinoceros (approved):

neutrinoceros:
I think failures are not directly caused by this change but may be due to some defect from the hdf5-manylinux companion PR, which I'd like to address before merging this one.

See #2948

neutrinoceros:
thank you !

PR #2949, opened 2026-08-05
WHL: explicitly set HDF5_LIBDIR in Linux builds

### LINKED ISSUE #2948: CI: Linux builds are failing (cannot find `libhdf5.so`)
[example logs](https://github.com/h5py/h5py/actions/runs/30975373533/job/92208153465)

I suspect https://github.com/h5py/hdf5-manylinux/pull/33 to be the cause, as the only significant change recently merged, though I don't get off hand how it could result in this error.

@ajelenak , any clue ?

neutrinoceros:
tentatively fix #2948

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -153,6 +153,7 @@ environment-pass = [
 
 [tool.cibuildwheel.linux.environment]
 CFLAGS = "-g1"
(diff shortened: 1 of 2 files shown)


neutrinoceros:
clearly not correct. But I think I'll just keep the debug info I added. Opening a separate PR for posterity.

PR #2953, opened 2026-08-07
Add ROS3 block caching feature

ajelenak:
ROS3 driver can now have improved performance with typical, non-cloud optimized, HDF5 files in S3.

Available since HDF5 2.2.0.


ajelenak:
I plan to merge this on September 7 unless there is any discussion or review beforehand.

tacaswell:
I missed this coming in, but after the fact of it being merged looks 👍🏻 .

PR #2954, opened 2026-08-08
Fix writing a single vlen field of a compound dataset

### LINKED ISSUE #1857: setting dataset string field results in TypeError
I'm using Ubuntu 18.04.5 LTS with python 3.8.0 from apt.  I created a fresh virtualenv and pip installed h5py for testing. 

h5py version info:
```python
>>> print(h5py.version.info)

h5py    3.2.1
HDF5    1.12.0
Python  3.8.0 (default, Oct 28 2019, 16:14:01) 
[GCC 8.3.0]
sys.platform    linux
numpy   1.20.1
cython (built with) 0.29.22
numpy (built against) 1.17.5
HDF5 (built against) 1.12.0
```
I have a simple script where I'm trying to change the individual fields of a dataset.

```python
import h5py
f = h5py.File('test.hdf5','w')
ds = f.create_dataset('test_dataset',1,dtype=[('a_float','<u8'),('a_string',h5py.string_dtype(encoding='utf-8'))])
ds[0] = (1, "some string")

test_float = 2
ds['a_float'] = test_float
assert ds[0][0] == test_float

test_string = "test string"
ds['a_string'] = test_string
```
In this script setting the float works as expected, but `ds['a_string'] = test_string` errors with:

```python
  File "/softdev/akuurstr/python/h5py_test.py", line 11, in <module>
    ds['a_string'] = test_string
  File "/softdev/akuurstr/python/virtualenvs/h5py_test/lib/python3.8/site-packages/h5py/_hl/dataset.py", line 845, in __setitem__
    val = val.view(numpy.dtype([(names[0], dtype)]))
  File "/softdev/akuurstr/python/virtualenvs/h5py_test/lib/python3.8/site-packages/numpy/core/_internal.py", line 459, in _view_is_safe
    raise TypeError("Cannot change data-type for object array.")
TypeError: Cannot change data-type for object array.
```
The only way I can set the dataset's string is to replace the entire record. However it seems like I should be able to set the string field individually.

teddytennant:
Fixes #1857.

## The bug

Assigning to a single variable-length-string field of a compound dataset raises `TypeError`:

```python
dt = np.dtype([('num', '<u8'), ('a_string', h5py.string_dtype())])
ds = f.create_dataset('x', (3,), dtype=dt)
ds['a_string'] = 'test string'
```

```
TypeError: Cannot change data-type for array of references.
```

Numeric fields work; vlen fields do not. (The NumPy message has changed wording since the issue was
filed — it used to say "object array" — but it is the same failure.)

## Cause

`Dataset.__setitem__` builds a one-field structured array by reinterpreting the value with
`ndarray.view()`:

```python
val = val.view(numpy.dtype([(names[0], dtype)]))
```

NumPy refuses `.view()` on object-dtype arrays, and object dtype is exactly how h5py represents vlen
data. Fixed-width and numeric fields take the same path successfully because `.view()` is legal for
them.

## The fix

When the target field's dtype is object (`dtype.base.kind == 'O'`), allocate the one-field array and
copy the value in instead of reinterpreting it. Every other dtype keeps the existing zero-copy
`view()` path, so there is no behaviour or performance change for numeric or fixed-length string
fields.

## Verification

Built from source against HDF5 1.14.6 (Python 3.13, NumPy 2.5.1, Cython 3.2.9).

Fail-before (fix reverted, new test present):

```
>           raise TypeError("Cannot change data-type for array of references.")
E           TypeError: Cannot change data-type for array of references.
FAILED h5py/tests/test_dataset.py::TestCompound::test_assign_vlen_field
====================== 1 failed, 240 deselected in 0.58s =======================
```

Pass-after:

```
h5py/tests/test_dataset.py::TestCompound::test_assign_vlen_field PASSED  [100%]
====================== 1 passed, 240 deselected in 0.17s =======================
```

Suites: `test_dataset.py` 240 passed / 1 skipped. Full `h5py/tests` 849 passed / 54 skipped in 10.4s.
No pre-existing failures.

I also checked by hand that numeric scalar and array field assignment, numeric subarray fields
(`('a', 'f4', (3,))`), fixed-length `S5` fields, and object-reference fields all round-trip
identically before and after.

## Notes

- The `kind == 'O'` guard is deliberately narrow. The copy path also works for numeric fields, so an
  unconditional copy would pass the suite too — but `view()` is zero-copy, and keeping the branch
  narrow limits the change to the broken case. Happy to simplify it if you'd prefer.
- The two `reshape` calls were collapsed into one shared local so the branches stay symmetric. That
  is the only incidental change and it is behaviour-preserving.
- Thanks to @takluyver, whose workaround on the issue pointed at where this goes wrong.

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -1010,8 +1010,16 @@ class Dataset(HLObject):
 
             val = numpy.asarray(val, dtype=dtype.base, order='C')
             if cast_compound:
-                val = val.view(numpy.dtype([(names[0], dtype)]))
-                val = val.reshape(val.shape[:len(val.shape) - len(dtype.shape)])
+                cast_dtype = numpy.dtype([(names[0], dtype)])
+                cast_shape = val.shape[:len(val.shape) - len(dtype.shape)]
+                if dtype.base.kind == 'O':
+                    # Vlen data is held in object arrays, which numpy refuses to
+                    # reinterpret as another dtype, so copy it into place instead.
+                    tmp = numpy.empty(cast_shape, dtype=cast_dtype)
+                    tmp[names[0]] = val
+                    val = tmp
+                else:
+                    val = val.view(cast_dtype).reshape(cast_shape)
         elif (self.dtype.kind == 'S'
               and (h5t.check_string_dtype(self.dtype).encoding == 'utf-8')
               and (find_item_type(val) is str)
--- a/h5py/tests/test_dataset.py
+++ b/h5py/tests/test_dataset.py
@@ -1572,6 +1572,32 @@ class TestCompound(BaseDataset):
         self.assertTrue(np.all(outdata == testdata))
         self.assertEqual(outdata.dtype, testdata.dtype)
 
+    def test_assign_vlen_field(self):
+        """ A single vlen field can be written on its own (issue 1857)"""
+        dt = np.dtype([('a_float', np.float64),
+                       ('a_string', h5py.string_dtype(encoding='utf-8')),
+                       ('a_seq', h5py.vlen_dtype(np.int32))])
+
+        ds = self.f.create_dataset(make_name(), (3,), dtype=dt)
+        ds['a_float'] = [1.0, 2.0, 3.0]
+        ds['a_string'] = ['one', 'two', 'three']
+        ds['a_seq'] = np.array([np.arange(n, dtype=np.int32) for n in (1, 2, 3)],
+                               dtype=object)
+
+        outdata = ds[...]
+        self.assertArrayEqual(outdata['a_float'], np.array([1.0, 2.0, 3.0]))
+        self.assertArrayEqual(outdata['a_string'],
+                              np.array([b'one', b'two', b'three'], dtype=object))
+        for got, expected in zip(outdata['a_seq'], (1, 2, 3)):
+            self.assertArrayEqual(got, np.arange(expected, dtype=np.int32))
+
+        # A scalar is broadcast over the field, leaving other fields alone
+        ds['a_string'] = 'same'
+        outdata = ds[...]
+        self.assertArrayEqual(outdata['a_string'],
+                              np.array([b'same'] * 3, dtype=object))
+        self.assertArrayEqual(outdata['a_float'], np.array([1.0, 2.0, 3.0]))
+
     def test_fields(self):
         dt = np.dtype([
             ('x', np.float64),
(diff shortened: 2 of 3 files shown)


teddytennant:
Anything I can do to help move this along? CI is green and it still merges cleanly.

ajelenak:
Hi @teddytennant,

I have an open PR #2964 that reworks the fill value code path, and because both changes touch `_hl/dataset.py` I wanted to make sure they would not step on each other before either one is merged. They turn out to be completely independent, but while I was checking I ended up exercising your fix across a fairly wide range of compound field dtypes, and a few of the results seemed worth passing along. I tested on macOS arm64, Python 3.12.11, HDF5 2.2.0, NumPy 2.3.0, against your branch merged onto mine. Thanks for picking this one up. Issue #1857 has been open a long while, and the copy-instead-of-view approach looks like the right fix to me.

The two PRs compose cleanly. Your patch applies on top of my branch without any conflict, and with both applied the full `test_dataset.py` suite passes, including your new `test_assign_vlen_field`. Creating a dataset with a compound fill value and then writing an individual field afterwards also behaves correctly. So there is no ordering dependency between the two PRs from my side, and nothing here that should hold yours up, as far as I can tell. Whichever merges first, the other should only need a trivial rebase in the test file, since we both add tests to `test_dataset.py`.

Your fix covers more ground than the only test currently asserts. I ran a matrix of compound field shapes through `__setitem__`, and in addition to the plain vlen cases your test covers, the fix also repairs all three of the array-valued object shapes: an array of vlen strings, an array of vlen sequences, and an array of object references, e.g., `np.dtype([('num', '<u8'), ('b', h5py.string_dtype(), (2,))])`. Every one of those raises the same `TypeError` on current master and works with your change applied. This is because `dtype.base` strips the subarray shape before `.kind` is checked, so the guard sees through to the element type, which I think is exactly right. I did check the values themselves rather than just the absence of an exception, and the contents round-trip correctly, the neighbouring fields are left undisturbed, and the references still resolve to their target. You may want to consider adding one array-valued case to `test_assign_vlen_field`, since it exercises a genuinely distinct path and, I think, nothing in the suite currently covers it.

teddytennant:
Added the array-valued case to `test_assign_vlen_field`, a `(string_dtype(), (2,))` field. I ran that write on its own against master first to be sure it fails with the same TypeError, since folded into the existing test it would otherwise never get reached.

I left out the vlen sequence and reference shapes because all three land on the same `kind == 'O'` branch, so they would not cover anything new. Happy to add the reference one if you would rather have it pinned.

Thanks for checking this against #2964. I'll rebase whichever way that lands.

ajelenak:
I plan to merge this on September 7, unless there is any discussion before then.

ajelenak:
Thanks @teddytennant, merged.

PR #2956, opened 2026-08-12
Fix file images with libver on older HDF5

### LINKED ISSUE #2558: Passing libver to in_memory() constructor produces file which can't be reopened
From #2517: For some reason passing `libver=` when constructing the in-memory file (tried with `'v108'` and `'v110'`) then leads to an error attempting to reopen that file from the bytes 'file image':

```
OSError: Unable to open file (incorrect metadata checksum after all read attempts)
```

This still occurs if I pass the same libver when reopening it. And the test already calls `f.flush()` before getting the file image.

CAOShurong:
## Summary

- repair the version 2/3 superblock checksum in file images returned by HDF5 versions before 2.0
- compile the compatibility path out for HDF5 2.0 and later
- add a regression covering reopening a `File.in_memory(libver="v108")` image before and after closing the original file

Before HDF5 2.0, `H5Fget_file_image()` clears the file-open status flags without updating the affected superblock checksum. HDFGroup/hdf5#5489 fixed that upstream; this PR mirrors the repair for older supported HDF5 versions.

Closes #2558.

## Validation

- HDF5 1.14.6: focused file-image tests passed (4); full suite passed (847 passed, 56 skipped, 3 subtests passed); a clean wheel install reopened all six available libver bounds
- HDF5 2.0.0: clean wheel install and focused file-image tests passed (4)
- `pre-commit run --all-files` and `git diff --check` passed
- follow-up `4d268f01`: rebuilt against HDF5 1.14.6 and reran the focused file-image tests (4 passed)

## AI assistance

OpenAI Codex was used to investigate the issue, implement and test the change, and draft this pull request. The checksum implementation was checked against the cited HDF5 1.14.6 source, and the base failure and repaired behavior were reproduced locally.

### DIFF
--- a/h5py/h5f.templ.pyx
+++ b/h5py/h5f.templ.pyx
@@ -15,6 +15,8 @@
 from cpython.buffer cimport PyObject_CheckBuffer, \
                             PyObject_GetBuffer, PyBuffer_Release, \
                             PyBUF_SIMPLE
+from libc.stdint cimport uint8_t, uint32_t
+from libc.string cimport memcmp
 from ._objects cimport pdefault
 from .h5p cimport propwrap, PropFAID, PropFCID
 from .h5i cimport wrap_identifier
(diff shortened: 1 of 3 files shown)


h5py/h5f.templ.pyx
```diff
@@ -451,11 +573,21 @@ cdef class FileID(GroupID):
...
+        cdef char *image_buf
 
         size = H5Fget_file_image(self.id, NULL, 0)
         image = PyBytes_FromStringAndSize(NULL, size)
+        image_buf = PyBytes_AsString(image)
+
+        H5Fget_file_image(self.id, image_buf, size)
 
-        H5Fget_file_image(self.id, PyBytes_AsString(image), size)
+        # HDF5 < 2.0 clears the file-open status flags in the returned image
```
neutrinoceros, line 584: The comment should be part of the templated section

neutrinoceros (changes requested):
the provenance of `_lookup3_checksum` is insufficiently documented
furthermore, the PR description looks suspiciously verbose (in particular the validation section looks way more detailed than necessary), so I would like to require disclosure on any use of LLMs in this work. I reserve the right to close without merge depending on the answer.

CAOShurong:
Thanks for the review. OpenAI Codex was used to investigate, implement, test, and draft this change.

Commit `4d268f01` now documents the lookup3 implementation's exact HDF5 1.14.6 source and the existing `licenses/hdf5.txt` attribution, and moves the `get_file_image()` explanation inside the HDF5 `< 2.0` template section. I also shortened the PR description and added the explicit disclosure there.

I rebuilt the updated head against HDF5 1.14.6 and reran `h5py/tests/test_file_image.py`: 4 passed.

PR #2957, opened 2026-08-12
Avoid stale caches for growing virtual datasets

### LINKED ISSUE #1693: Don't cache shape for virtual datasets
From discussion on #1689, it turns out that virtual datasets can magically resize themselves if the source data is resized. This is true even if they're in a read-only file, in which case we currently cache the shape. We should probably avoid caching the shape for virtual datasets.

This illustrates the issue:

```python
import h5py
import numpy as np
import os.path as osp
import tempfile

tmpdir = tempfile.mkdtemp()
path = osp.join(tmpdir, "resize.h5")
vds_path = osp.join(tmpdir, "resize_vds.h5")
with h5py.File(path, "w") as f:
        "source",
        data=np.arange(20),
        shape=(10, 2),
        maxshape=(None, 2),
        chunks=(10, 1),
        fillvalue=-1
    )
    layout = h5py.VirtualLayout((10, 1), np.int, maxshape=(None, 1))
    layout[:h5py.UNLIMITED, 0] = layout_source[:h5py.UNLIMITED, 1]

with h5py.File(vds_path, "w") as f:
    f.create_virtual_dataset("virtual", layout)


with h5py.File(path, "a") as f, h5py.File(vds_path, 'r') as fv:
    source_dset = f['source']
    virtual_dset = fv['virtual']

    print("Before", virtual_dset.shape)  # (10, 1)
    source_dset.resize(20, axis=0)
    print("After resize", virtual_dset.shape)  # (10, 1)
    virtual_dset.refresh()
    print("After refresh", virtual_dset.shape)  # (20, 1)
```

CAOShurong:
Closes #1693.

## Summary

- Do not cache dataspace-derived shape, size, selector, or fast-reader objects
  for virtual datasets opened from read-only files.
- Preserve the existing cache optimization for ordinary read-only datasets.
- Add a regression test which reads an unlimited VDS, grows its source dataset,
  and observes the new shape and data without calling `refresh()`.

## Root cause

The file containing a virtual dataset can be read-only while the source dataset
continues to grow. h5py cached the VDS shape and reader because the VDS file was
read-only, so both metadata and later reads could remain limited to the old
extent. The new cache eligibility check excludes virtual datasets, whose extent
can change through their sources.

## Validation

- `pytest h5py/tests/test_vds/test_highlevel_vds.py -q`: 13 passed.
- Clean CPython 3.13 wheel built against HDF5 1.14.6: full suite 847 passed,
  56 skipped, 3 subtests passed.
- Clean CPython 3.13 wheel built from a fresh sdist against HDF5 2.0.0: full
  suite 874 passed, 29 skipped, 3 subtests passed.
- Independent installed-wheel workflow on both HDF5 versions: the read-only VDS
  grew from `(10, 1)` to `(20, 1)` and exposed the new values without a refresh.
- Ordinary read-only datasets still populated the shape, size, and fast-reader
  caches.
- `pytest-run-parallel --parallel-threads 4`: the unmarked file-mutating test
  reproduced a shared-file race; with the repository's `thread_unsafe` marker
  it ran serially and passed.
- `pre-commit run --all-files`: passed.

An initial HDF5 2.0.0 clean-wheel full-suite run omitted the repository's
declared `pytest-mpi` and `pytest-cov` dependencies and failed only on missing
plugin/fixture setup. The same unchanged wheel passed the full suite after those
declared test dependencies and the repository pytest configuration were loaded.

Not run locally: Linux, macOS, MPI-enabled builds, free-threaded Python, and the
complete upstream HDF5/Python matrix. Those remain for project CI.

## AI assistance

AI tools were used in preparing this contribution. This disclosure and the follow-up reply were prepared with OpenAI Codex; the original model/version has not been verified.

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -549,9 +549,9 @@ class Dataset(HLObject):
         with phil:
             shape = self.id.shape
 
-        # If the file is read-only, cache the shape to speed-up future uses.
-        # This cache is invalidated by .refresh() when using SWMR.
-        if self._readonly:
+        # Cache the shape for read-only, non-VDS datasets. This cache is
+        # invalidated by .refresh() when using SWMR.
+        if self._should_cache_props:
             self._cache_props['shape'] = shape
         return shape
 
@@ -572,11 +572,16 @@ class Dataset(HLObject):
         else:
             size = product(self.shape)
 
-        # If the file is read-only, cache the size to speed-up future uses.
-        # This cache is invalidated by .refresh() when using SWMR.
-        if self._readonly:
+        # Cache the size for read-only, non-VDS datasets. This cache is
+        # invalidated by .refresh() when using SWMR.
+        if self._should_cache_props:
             self._cache_props['size'] = size
         return size
+
+    @property
+    def _should_cache_props(self):
+        """Whether dataspace-derived properties can be cached safely."""
+        return self._readonly and (not vds_support or not self.is_virtual)
 
     @property
     def nbytes(self):
@@ -594,9 +599,9 @@ class Dataset(HLObject):
 
         slr = _selector.Selector(self.id.get_space())
 
-        # If the file is read-only, cache the reader to speed up future uses.
-        # This cache is invalidated by .refresh() when using SWMR.
-        if self._readonly:
+        # Cache the selector for read-only, non-VDS datasets. This cache is
+        # invalidated by .refresh() when using SWMR.
+        if self._should_cache_props:
             self._cache_props['_selector'] = slr
         return slr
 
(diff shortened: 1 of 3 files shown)


neutrinoceros:
Quoting myself from #2956
> the PR description looks suspiciously verbose (in particular the validation section looks way more detailed than necessary), so I would like to require disclosure on any use of LLMs in this work. I reserve the right to close without merge depending on the answer.

CAOShurong:
AI-assisted reply prepared with OpenAI Codex on behalf of CAOShurong.

In response to the disclosure request: AI tools were used in preparing this contribution. The original model/version has not been verified. This disclosure has now been added to the PR description.

PR #2959, opened 2026-08-14
WHL: upgrade abi3 nightlies to cp315

neutrinoceros:
Now possible with #2958 + numpy 2.5.2 which has cp315* wheels


neutrinoceros:
Hum, the abi3 violation is real. I assume that it's a Cython issue.
More specifically, `PyObject_CallFinalizerFromDealloc` is being used, and while it's been added in Python 3.4, it's never been part of the Limited API.

neutrinoceros:
reported upstream at https://github.com/cython/cython/issues/7955

PR #2960, opened 2026-08-18
Fix memory leak reading vlen string fill values

ajelenak:
`H5Pget_fill_value()` converts the stored fill value into the type the caller
asks for. When that type is a variable-length string, the conversion allocates
the string and hands ownership to the caller. `PropDCID.get_fill_value()`
copied the bytes into Python but never released the buffer, so every read of
`Dataset.fillvalue` leaked it, and since it is an uncached property, this compounds
with each access:

```python
import h5py
import resource

with h5py.File('t.h5', 'w') as f:
    d = f.create_dataset('s', (4,), dtype=h5py.string_dtype(),
                         fillvalue=b'x' * 4096)
    for _ in range(50_000):
        d.fillvalue
    # ru_maxrss is bytes on macOS, kilobytes on Linux
    print(resource.getrusage(resource.RUSAGE_SELF).ru_maxrss)
```

Peak RSS growth over 50,000 reads (macOS/arm64, HDF5 2.2.0):

| fill value | before | after |
| --- | --- | --- |
| `f8` scalar (control) | 0.0 MB | 0.0 MB |
| vlen string, 64 B | 4.1 MB | 0.0 MB |
| vlen string, 4 KiB | 231.3 MB | 0.0 MB |

Growth tracks the payload size, and the fixed-size control is flat.

## Fix

Reclaim the buffer with `H5Dvlen_reclaim()` after copying the value out. The
copy is wrapped in `try`/`finally` so a failure while building the Python bytes
object cannot lose the buffer.

I used `H5Dvlen_reclaim()` rather than `H5Treclaim()` because it is the one
h5py already wraps and what `_proxy.pyx` uses throughout. Happy to add an
`H5Treclaim()` binding instead if that is preferred.

The undefined-fill path returns early on a `NULL` pointer and is unaffected,
there is nothing to reclaim there.

## Tests

Added `test_vlen_fillvalue_not_leaked` in `h5py/tests/test_dataset.py`. I
confirmed it fails without the fix and passes with it:

| build | RSS growth, 4000 reads of a 64 KiB fill | result |
| --- | --- | --- |
| unpatched | 327 MB | fail (threshold 32 MB) |
| patched | 0 MB | pass |

This is the only memory test in the suite and an
RSS-based test may not be something desirable in CI. It skips on Windows (no
`resource`), normalizes the macOS/Linux `ru_maxrss` unit difference, and leaves
a 10x margin against the threshold. I am willing to reduce it to just a correctness-only test.

### DIFF
--- a/h5py/h5p.templ.pyx
+++ b/h5py/h5p.templ.pyx
@@ -532,7 +532,9 @@ cdef class PropDCID(PropOCID):
         rank, only the first element will contain the value.
         """
         from .h5t import check_string_dtype
+        from .h5s import create as create_space, SCALAR
         cdef TypeID tid
+        cdef SpaceID sid
         cdef char * c_ptr = NULL
 
         check_numpy_write(value, -1)
@@ -542,7 +544,7 @@ cdef class PropDCID(PropOCID):
         string_info = check_string_dtype(value.dtype)
         if string_info is not None and string_info.length is None:
             tid = py_create(value.dtype, logical=1)
-            ret = H5Pget_fill_value(self.id, tid.id, &c_ptr)
+            H5Pget_fill_value(self.id, tid.id, &c_ptr)
             if c_ptr == NULL:
                 # If the pointer is NULL (either the value did not get changed,
                 # or maybe the 0 length string, it's unclear currently), if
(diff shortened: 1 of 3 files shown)


ajelenak:
#2964 supersedes this PR. Closing.

PR #2961, opened 2026-08-18
Support fill values for array fixed-length datatypes

ajelenak:
Note that this branch is stacked on PR #2960.

Datasets with an array fixed-size datatype (`H5T_ARRAY`) now accept a fill value:

```python
dt = np.dtype(('f4', (3,)))
d = f.create_dataset('x', (2,), dtype=dt, fillvalue=[1.5, 2.5, 3.5])
d.fillvalue        # array([1.5, 2.5, 3.5], dtype=float32)
d[1]               # array([1.5, 2.5, 3.5], dtype=float32)
```

The value is broadcast to the shape of one element, so `fillvalue=7` becomes
`[7., 7., 7.]`. For an array datatype `Dataset.fillvalue` returns an array of
the base type. Before this, any fill value failed at dataset creation:

```
ValueError: Unable to synchronously create dataset
  (no appropriate function for conversion path)
```

and reading an array datatype fill value from a file raised
`RuntimeError: Can't get fill value`, which this also fixes.

`PropDCID.set_fill_value()` inferred the HDF5 datatype from `value.dtype`.
NumPy collapses an array dtype into the array's own shape, so `ndarray.dtype`
can never itself be an array dtype: for `('f4', (3,))`, `np.array([1, 2, 3])`
has dtype `f4` and shape `(3,)`. HDF5 was therefore asked to convert `f4` to
`3 x f4`, and there is no such conversion path. The array type was unreachable
through the signature, whatever the caller passed.

Now `PropDCID.set_fill_value()` and `.get_fill_value()` take an optional `mtype`
giving the HDF5 datatype of the buffer for datatypes NumPy cannot express on an
array itself. Omitting it keeps the current inference, so the change is backward
compatible.

The high-level callers pass it when the dtype has a `subdtype`:
`make_new_dset()`, the `Dataset.fillvalue` property, and
`VirtualLayout.make_dataset()` for virtual datasets. The buffer is
`numpy.zeros((), dtype=dtype)`, which is exactly one element of the array
datatype, filled by broadcast assignment and no manual allocation involved.

Supported are any array datatypes with a fixed-size base type, for example:
multidimensional bases (`('i2', (2, 2))`), non-native byte order (`('>f8',
(2,))`), or fixed-length ASCII or UTF-8 strings.

Array datatypes whose base has variable-length components raise
`NotImplementedError` for now. Reading `.fillvalue` on such a dataset is left
unchanged.

### Buffer safety

`H5Pset_fill_value()`/`H5Pget_fill_value()` read and write exactly
`H5Tget_size(mtype)` bytes at the buffer with no checking of their own, and
`check_numpy_read()`/`check_numpy_write()` only validate contiguity and
writability. Since `mtype` is public API, the new `_check_fill_buffer()` rejects two
things before HDF5 is called:

* a buffer smaller than the datatype, which would otherwise overrun the heap;
* an object array, whose elements are `PyObject*` rather than the
  variable-length data HDF5 expects.

The second cannot be folded into the first: those are the same width, so an
object array paired with an array-of-vlen type passes the size check exactly.
There's also a new test for this.

### Tests

Twenty-three new tests. The main ones are parametrized across seven dtype/fill-value
combinations and run twice, once in-memory after creation and once after a
round trip through a closed file.

### DIFF
--- a/docs/high/dataset.rst
+++ b/docs/high/dataset.rst
@@ -712,6 +712,15 @@ Reference
         type-appropriate default value.  Can't be changed after the dataset is
         created.
 
+        For a dataset with an array datatype, such as ``('f4', (3,))``, this is
+        an array of the base type.  The base type may be anything of a fixed
+        size, including a compound type. Array datatypes built on
+        variable-length strings, variable-length sequences or references do not
+        support fill values yet.
+
+        .. versionchanged:: 3.17
+           Array fixed-size datatypes are supported.
+
     .. attribute:: external
 
        If this dataset is stored in one or more external files, this is a list
--- a/docs/high/group.rst
+++ b/docs/high/group.rst
@@ -420,7 +420,9 @@ Reference
         :keyword fletcher32: Enable Fletcher32 checksum (T/**F**).  See :ref:`dataset_fletcher32`.
 
         :keyword fillvalue: This value will be used when reading
-                            uninitialized parts of the dataset.
+                            uninitialized parts of the dataset.  For an array
+                            datatype it is broadcast to the shape of one
+                            element, so both a scalar and a sequence work.
 
         :keyword fill_time: Control when to write the fill value. One of the
             following choices: `alloc`, write fill value before writing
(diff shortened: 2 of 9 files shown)


ajelenak:
#2964 supersedes this PR. Closing.

PR #2963, opened 2026-08-20
Add alloc_time parameter to create_dataset()

takluyver:
I wanted the option to allocate dataset space up front from the high-level API. At the moment, I'm working around it by writing a single number into the dataset to force allocation.

I also moved the handling of `fill_time` out of `filters.py`, because I don't think it's related to filtering.


ajelenak:
Is this ready for merge? It looks so to me.

takluyver:
Thanks, yes, this is indeed ready.

PR #2964, opened 2026-08-20
Fix fill values for array and variable-length datatypes

ajelenak:
This PR supersedes #2960 and #2961 since fixes for both are folded in here.

### Summary

Fill values were broken for several datatypes, in four different ways:

| datatype | before |
| --- | --- |
| compound with a variable-length member | wrote a file no HDF5 tool could read back, **no error reported** |
| variable-length sequence | reading a stored fill value **aborted the process** |
| variable-length string | leaked its buffer on every read of `Dataset.fillvalue` |
| array (`H5T_ARRAY`) | rejected outright at dataset creation |

All four now work, including array datatypes over a variable-length base.

### The fix

Convert fill values the same way dataset data is converted: a buffer sized for the file datatype, `H5Tconvert` between `py_create(dtype, logical=1)` and `py_create(dtype, logical=0)`, a background buffer where the conversion needs one, and `H5Dvlen_reclaim` for the staging copy on write. On read the conversion hands ownership of its allocations to the objects it builds, so nothing is left to reclaim (why the leak goes away without an explicit free).

Because `H5Tconvert` recurses, one mechanism covers plain vlen, compound members, and array elements. Both string special cases are deleted rather than extended.

`set_fill_value()` and `get_fill_value()` take an optional `dtype` naming the NumPy datatype the buffer describes, defaulting to `value.dtype`. That is what array datatypes need, and it lets both HDF5 types be derived from one argument. Both methods now also check the buffer against the datatype: HDF5 transfers a
fixed number of bytes, and `check_numpy_read`/`check_numpy_write` validate only contiguity and writability, never size.

### Behaviour changes

* `Dataset.fillvalue` on a variable-length dataset with no fill value returns an empty value instead of `None`. That matches HDF5's default fill for a variable-length type, a zero-length sequence. The previous `None` came from reading 8 bytes of a 16-byte value.
* Setting a fill value on a compound with a variable-length member used to "succeed" and produced an unreadable file. It now works properly.

### Tests

46 new tests. Array datatypes are parametrized over ten dtype/value combinations and run twice, in memory and after a round trip through a closed file. Variable-length datatypes over five base types. Compound-with-vlen-member in both flavours. Array-over-vlen for sequences and strings; plus fill status, unset values, UTF-8 character set preservation, native `H5T_COMPLEX` base (skipped below HDF5 2.0), a virtual dataset, and the buffer size guard.

Three cases run in a child process, because two of them previously killed the interpreter which would take the whole pytest run down rather than fail one test. `h5py/tests/data_files/vlen_seq_fillvalue.h5` (6 KB) is a new test file. h5py could not write such a file before, and it is also the realistic case built with the HDF5 C API.

I confirmed the new tests fail without the change by reverting the source files and rebuilding.

Two tests carry `thread_unsafe`: one measures process-wide `maxRSS` which will be inflated by multiple threads, and the other one occasionaly segfaulted with 8 or more threads.


ajelenak:
Added more tests for object and region references in various advanced configurations. They are all parameterized so not much new code. 😁

ajelenak:
By removing the failing tests caused by the bug reported in PR #2966, this PR is all green now. Ready for review.

PR #2965, opened 2026-08-21
Enable Windows ARM64 builds on CI

MugundanMCW:
PR Description:

* As discussed in [[h5py#2816](https://github.com/h5py/h5py/issues/2816#issuecomment-5311606178)](https://github.com/h5py/h5py/issues/2816#issuecomment-5311606178), Windows ARM64 CI builds currently fail while running the test suite.
* This is caused by the UV Python manager selecting an x86_64 Python interpreter instead of the ARM64 interpreter configured by cibuildwheel (CIBW).
* The CI passes for Python versions that are preinstalled on the Windows ARM64 runner. Therefore, this PR skips the test suite for Python versions that are not preinstalled on the runner.
* Several downstream projects that depend on h5py for Windows ARM64 are currently blocked by this issue. This PR addresses the problem and helps unblock Windows ARM64 support for these projects.

cc: @neutrinoceros & @khmyznikov

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -70,9 +70,8 @@ jobs:
             arch: aarch64
           - os: windows-latest
             arch: AMD64
-          # See https://github.com/h5py/h5py/issues/2816
-          #- os: windows-11-arm
-          #  arch: ARM64
+          - os: windows-11-arm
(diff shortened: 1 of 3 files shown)


neutrinoceros:
Thank you !

PR #2966, opened 2026-08-24
Fix crash reading variable-length region reference data

ajelenak:
Reading a dataset whose datatype is a variable-length sequence of region references segfaults:

```python
dt = h5py.vlen_dtype(h5py.regionref_dtype)
d = f.create_dataset('x', (2,), dtype=dt)
d[0] = np.array([ref, ref], dtype=object)   # write: fine
d[0]                                        # read: SIGSEGV
```

This bug was revealed by failing CI workflows in my other PR #2964.

| platform | |
| --- | --- |
| Linux x86_64 | segfaults |
| Linux aarch64 | segfaults |
| Windows AMD64 / ARM64 | segfaults |
| macOS | passes |

`conv_vlen2ndarray` allocates the background buffer for the element conversion with `emalloc` and never initializes it. `conv_regref2pyref` treats that buffer as holding the existing destination value and releases it:

```cython
bkg_obj0 = bkg_obj[0]
...
Py_XDECREF(bkg_obj0)
```

so it decrefs whatever pointer happened to be on the heap. The gdb backtrace lands in `_Py_DECREF` with a garbage `op`, three frames below `conv_vlen2ndarray`.

Object references never hit this because they register with `H5T_BKG_NO` and never read the buffer, while region references register `H5T_BKG_YES`. That asymmetry is exactly why `vlen_dtype(ref_dtype)` is fine and `vlen_dtype(regionref_dtype)` is not. MacOS escapes because its allocator hands back zeroed pages.

### Fix

Zero the buffer after allocating it, in both `conv_vlen2ndarray` and `conv_ndarray2vlen`.

The second function is not causing any crash but it is the same unzeroed allocation, and garbage there would reach unwritten compound fields. Fixing one and not the other seemed worse than fixing both.

### DIFF
--- a/h5py/_conv.templ.pyx
+++ b/h5py/_conv.templ.pyx
@@ -18,7 +18,7 @@ from .h5 import get_config
 from .h5r cimport Reference, RegionReference, hobj_ref_t, hdset_reg_ref_t
 from .h5t cimport H5PY_OBJ, typewrap, py_create, TypeID, H5PY_PYTHON_OPAQUE_TAG
 from libc.stdlib cimport realloc
-from libc.string cimport strcmp
+from libc.string cimport strcmp, memset
 from .utils cimport emalloc, efree
 from ._proxy cimport needs_bkg_buffer
 cfg = get_config()
@@ -718,7 +718,11 @@ cdef int conv_vlen2ndarray(void* ipt,
         data = new_data
 
     if needs_bkg_buffer(intype.id, outtype.id):
+        # The background buffer holds the existing destination value, which a
+        # converter may release -- conv_regref2pyref decrefs it.  It has to be
+        # zeroed rather than left holding whatever was on the heap.
         back_buf = emalloc(H5Tget_size(outtype.id)*size)
+        memset(back_buf, 0, H5Tget_size(outtype.id)*size)
 
     try:
         H5Tconvert(intype.id, outtype.id, size, data, back_buf, H5P_DEFAULT)
@@ -869,7 +873,9 @@ cdef int conv_ndarray2vlen(void* ipt,
         PyBuffer_Release(&view)
 
         if needs_bkg_buffer(intype.id, outtype.id):
+            # Zeroed for the same reason as in conv_vlen2ndarray above
             back_buf = emalloc(H5Tget_size(outtype.id)*len)
+            memset(back_buf, 0, H5Tget_size(outtype.id)*len)
 
         H5Tconvert(intype.id, outtype.id, len, data, back_buf, H5P_DEFAULT)
 
--- a/h5py/tests/test_dataset.py
+++ b/h5py/tests/test_dataset.py
@@ -2363,3 +2363,28 @@ def test_store_regionrefs(writable_file):
     refs_ds = writable_file.create_dataset(make_name("refs"), shape=(1,), dtype=h5py.ref_dtype)
     with pytest.raises(TypeError, match="convert"):
         refs_ds[0] = ds1.regionref[:6]
+
+
+@pytest.mark.parametrize('ref_dtype,make_ref', [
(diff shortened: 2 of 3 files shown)


ajelenak:
Forgot to mention, I used `memset()` because there is no currently `ecalloc()` in h5py. Willing to add `ecalloc()` to replace the `emalloc()`/`memset()` combo.

ajelenak:
I plan to merge this on September 7, unless changes are requested before then.

neutrinoceros:
It's September 8, so feel free to merge now if you want but just as a note, I'm finally going back to my backlog and I could review this after lunch (a couple hours from now) if that's still on the table.

h5py/_conv.templ.pyx
```diff
@@ -869,7 +873,9 @@ cdef int conv_ndarray2vlen(void* ipt,
         PyBuffer_Release(&view)
 
         if needs_bkg_buffer(intype.id, outtype.id):
+            # Zeroed for the same reason as in conv_vlen2ndarray above
```
neutrinoceros: I'd prefer the comment be copied verbatim, in case the other one ever changes or goes away and this becomes a dead ref

neutrinoceros:
> so it decrefs whatever pointer happened to be on the heap.

Let me get this straight: `back_buf` is only ever included in `bkg_obj` if it's initialized first ? If so, where is that behavior defined ? Otherwise, clearly I must be misunderstanding something.

ajelenak:
> It's September 8

I forgot yesterday was a holiday in the US. 😁

> Let me get this straight: back_buf is only ever included in bkg_obj if it's initialized first ? If so, where is that behavior defined ? Otherwise, clearly I must be misunderstanding something.

You are not misunderstanding it, there is no such condition. `back_buf` is passed through to `bkg_obj` unconditionally whenever a background buffer is requested at all, initialized or not. Below is the chain of calls I followed for this bug, all in `_conv.templ.pyx` on this branch:

1. `conv_vlen2ndarray` allocates it but does not zero the buffer at line 724, `back_buf = emalloc(H5Tget_size(outtype.id)*size)`.
2. The buffer goes straight into `H5Tconvert(intype.id, outtype.id, size, data, back_buf, H5P_DEFAULT)` at line 728, as the `background` argument.
3. HDF5 hands that pointer to the registered converter as `bkg_i`. For this conversion that is `conv_regref2pyref`, registered at line 474. Note the `H5T_BKG_YES` there.
4. `generic_converter` (line 58) does `char* bkg = <char*>bkg_i` and then, per element, `op(buf + i*src_size, buf + i*dst_size, bkg + i*bkg_stride, priv)` around lines 93-97.
5. `conv_regref2pyref` (line 386) casts that straight back: `bkg_obj = <PyObject**>bkg`, reads `bkg_obj0 = bkg_obj[0]`, and calls `Py_XDECREF(bkg_obj0)` at line 400.

So `bkg_obj[0]` is literally the first eight bytes of the `emalloc`'d block, and those bytes get treated as a `PyObject*` and decref'd. `Py_XDECREF` only prevents `NULL`, which is why zeroing is a sufficient fix.

As for where the behavior is defined: the `H5T_BKG_YES` at line 474 is the h5py side, and the HDF5 side for the `background` argument of `H5Tconvert` is documented as holding the *existing* destination value. HDF5 neither allocates nor initializes it. Supplying meaningful contents is the caller's job. For a conversion whose destination is `H5PY_OBJ`, "existing destination value" means a `PyObject*` so the buffer has to hold either real object pointers or `NULL`. `emalloc` gives neither.

This is also why object references never tripped on this bug: `conv_objref2pyref` is registered with `H5T_BKG_NO` at line 462. Region references are the only reference converter registered with `H5T_BKG_YES`.

h5py/_conv.templ.pyx
ajelenak: Done.

PR #2969, opened 2026-09-02
RFC: use typing.assert_never for a statically unreachable branch

neutrinoceros:
Follow up to, and extracted from #2890
The branch is still reachable at runtime, which is why the original `AssertionError` message is preserved.

### DIFF
--- a/h5py/_hl/dataset.py
+++ b/h5py/_hl/dataset.py
@@ -14,6 +14,8 @@
 import posixpath as pp
 import sys
 from abc import ABC, abstractmethod
+from collections.abc import Iterable
+from typing import assert_never, TypeAlias
 from warnings import warn
 
 import numpy
@@ -364,11 +366,13 @@ if MPI:
             self._dset._dxpl.set_dxpl_mpio(h5fd.MPIO_INDEPENDENT)
 
 
+MaybeInt: TypeAlias = int | None
+Selector: TypeAlias = "int | slice[MaybeInt, MaybeInt, MaybeInt]"
 class ChunkIterator:
     """
     Class to iterate through list of chunks of a given dataset
     """
-    def __init__(self, dset, source_sel=None):
+    def __init__(self, dset, source_sel: "Selector | Iterable[Selector] | None" = None):
         self._shape = dset.shape
         rank = len(dset.shape)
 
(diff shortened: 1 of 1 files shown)


tacaswell:
I remain confused by the utility of both asserts.  Can the type checkers see past the exception?

Even if it did, we don't have enough type information for them to conclude that  from a static typing point of view we should never reach here (`'abc'` would be kosher to pass in from a typing point of view and would run up until here as `source_sel` is implicitly `Any` if I understand correctly).  I think the typechecker is only happy because it agrees that after an exception is raised we will never reach the next line of code!

I did have a concern that in `python -O` would make this a no-op, but fortunatly upstream thought of that.

```
✔ 16:29:36 [belanna] @ python -O
Python 3.14.7 (main, Aug 10 2026, 07:46:56) [GCC 16.1.1 20260728] on linux
>>> from typing import assert_never
>>> assert_never('bob')
  File "<python-input-2>", line 1, in <module>
    assert_never('bob')
    ~~~~~~~~~~~~^^^^^^^
  File "/usr/lib/python3.14/typing.py", line 2594, in assert_never
    raise AssertionError(f"Expected code to be unreachable, but got: {value}")
AssertionError: Expected code to be unreachable, but got: 'bob'
>>> assert False
>>> 
```

The more I think about it, the more I think the initial comment was wrong unless we have full type hint coverage.  Even then I am less that sold as `assert_never` makes the type checkers marginally better it degrades run-time error message.   I also think that this should have been a `ValueError` rather than an assertion error, but it is not worth the effort to change that.

neutrinoceros:
🤦🏻‍♂️ 
Of course, you're right that without type hints, this `assert_never` is useless to typecheckers. I think I originally thought of that but considered it so obvious I didn't mention it on the comment.

I think you're also right that an `assert_never` immediately following a `raise` statement can never be invalidated by type checkers and thus is, again, useless. I changed `raise AssertionError` to a normal `assert` statement, which I think should solve this problem.

> I also think that this should have been a ValueError rather than an assertion error, but it is not worth the effort to change that.

The error type *was* `ValueError` originally. It was changed in #2657 following my suggestion, which was made explicitly it in anticipation to having an `assert_never` there at some point in the future.

PR #2970, opened 2026-09-02
Test File context-manager cleanup on exceptions

wangyusheng1985:
## Summary
- add regression coverage for exceptional exits from an `h5py.File` context manager
- verify the raised exception propagates and the file is closed afterward

## Testing
- `/tmp/h5py-test-env/bin/python -m pytest h5py/tests/test_file.py::TestContextManager::test_context_manager_closes_on_exception`
- `VIRTUAL_ENV=/tmp/h5py-test-env /tmp/h5py-test-env/bin/pre-commit run --files h5py/tests/test_file.py`

### DIFF
--- a/h5py/tests/test_file.py
+++ b/h5py/tests/test_file.py
@@ -693,6 +693,16 @@ class TestContextManager(TestCase):
             self.assertTrue(fid)
         self.assertTrue(not fid)
 
+    def test_context_manager_closes_on_exception(self):
+        """File objects close when an exception is raised in a with block."""
+        exception = RuntimeError("test exception")
+
+        with pytest.raises(RuntimeError, match="test exception"):
+            with File(self.mktemp(), 'w') as fid:
+                raise exception
+
+        self.assertFalse(fid.id.valid)
+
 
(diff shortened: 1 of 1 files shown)


neutrinoceros (changes requested):
I was surprised I couldn't find an existing test case for this, but I agree this should be covered. I just note that there's a precedent in `test_h5o.py` for making the exception raised a custom type, so we're sure the test will never have false positives.

wangyusheng1985:
Thanks for the review. I’ll address this in the existing pull request and rerun the focused validation.

tacaswell:
I am curious the method for determining that this was a path that needed to be tested.  Looking at `File` we explicitly define `__enter__` and `__exit__` so this amounts to testing that context managers behave as documented as much or more so than we implemented the context manager correctly.

The test is not wrong, but I am very curious for the motivation of adding it.

wangyusheng1985:
The motivation was that existing context-manager tests only exercised normal block exits. There was no explicit regression test confirming both that `File.__exit__` closes the file and that exceptions raised inside the `with` block propagate outward without suppression.

PR #2976, opened 2026-09-21
Install RTD deps from pyproject.toml

ulgens:
Relates to https://github.com/h5py/h5py/pull/2789#issuecomment-5762724799

Supersedes #2789


ulgens:
@neutrinoceros 👋🏻 Ready for review.

ulgens:
This can be handled without the noop `python.install.command` entry but that requires updating the install block to

```
    install:
    - asdf plugin add uv
    - asdf install uv latest
    - asdf global uv latest
    - UV_PROJECT_ENVIRONMENT=$READTHEDOCS_VIRTUALENV_PATH uv sync --no-install-project --group docs
```

which is a bit more explicit, but not clean. I can switch to these version if you think that's better.

docs/requirements-rtd.txt
neutrinoceros, line 1: we should also clean up the part of `.github/dependabot.yml` that's specifically dedicated to this file

neutrinoceros (changes requested):
the problem with this approach right now is that we're loosing the lock file without a replacement. Now `uv sync` just re-resolves the whole environment from scratch every time. I'm in favor of having a `uv.lock` file, but that has implications beyond locking docs builds, and it's not portable. Maybe a `pylock.docs.toml` would make sense here, but then I'm not sure dependabot supports auto-upgrades to it.

neutrinoceros:
In the long run I like `uv sync` better, but right now it'd be much easier moving to something like

```
    - method: uv
      command: pip
      requirements: docs/requirements.txt
```

tacaswell:
It is my understanding that currently we are using pip on RTD?  If that is the case I would either leave it as-is or move to pixi.

